PR #1, opened 2021-09-22
Fix background of attribution list for Mac.

antonbauhofer:
When using dark mode on Mac, the headers of the attribution list were hard to read because of a dark background.


maxhbr (approved):
Code looks good

sebathomas (approved):
Tested it on Linux without dark mode and it doesn't break anything.

PR #6, opened 2021-09-22
Fix zip step in ship-mac script

nicarl:
There was an error in the zip command required to compress the `.app` artifact


sebathomas (approved):
Works on my laptop. :+1:

PR #7, opened 2021-09-22
Show more icons in package cards

sebathomas:
To make signals and attributions that are marked as follow-up or
as excluded more notable, we add the icons that are already used
in the report view table also in the other two views.

- Add a grid of icons to package cards with two icons per column
- Remove blue text color for excluded attributions

Screenshot_20210922_134539

### DIFF
--- a/src/Frontend/Components/ListCard/ListCard.tsx
+++ b/src/Frontend/Components/ListCard/ListCard.tsx
@@ -10,12 +10,14 @@ import React, { ReactElement } from 'react';
 import { OpossumColors } from '../../shared-styles';
 import { ListCardConfig } from '../../types/types';
 
+const defaultCardHeight = 40;
+
 const useStyles = makeStyles({
   root: {
     flex: 1,
     display: 'flex',
     alignItems: 'center',
-    height: 40,
+    height: defaultCardHeight,
     '&:hover': {
       cursor: 'pointer',
     },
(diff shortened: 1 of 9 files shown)


src/Frontend/Components/ListCard/ListCard.tsx
```diff
@@ -88,12 +90,17 @@ const useStyles = makeStyles({
...
   },
   iconColumn: {
-    flexDirection: 'column',
     display: 'flex',
-    justifyContent: 'center',
+    justifyContent: 'start',
     alignItems: 'flex-end',
+    flexWrap: 'wrap-reverse',
+    // fix for width of column flexbox container not growing after wrap
+    // -> use row flexbox with vertical writing mode
```
sebathomas, line 101: More details:
https://stackoverflow.com/questions/33891709/when-flexbox-items-wrap-in-column-mode-container-does-not-grow-its-width

nicarl (approved):

src/Frontend/Components/PackageCard/PackageCard.tsx
```diff
@@ -25,6 +37,16 @@ interface PackageCardProps {
...
 
+function getKey(
+  prefix: string,
+  cardContent: ListCardContent,
+  packageLabels: Array<string>
+): string {
+  return `${prefix}-${cardContent.name}-${cardContent.packageVersion}-${
+    packageLabels[0] || ''
+  }`;
+}
```
sebathomas, line 48: I'm seeing duplicate key warnings here in some cases. (This may already have been a problem before.) Will fix before merging.

sebathomas: Will do it in the follow-up diff that adds "pre-selected" icons instead.

JakobSchubert (approved):
Nice feature, looks very good!

PR #8, opened 2021-09-22
Fix horizontal scrolling in resources list

antonbauhofer:
Before this fix, the background color and the highlighting color of the resources list were not coherent for long paths.

### DIFF
--- a/src/Frontend/Components/List/List.tsx
+++ b/src/Frontend/Components/List/List.tsx
@@ -16,6 +16,7 @@ interface ListProps {
   cardVerticalDistance?: number;
   alwaysShowHorizontalScrollBar?: boolean;
   addPaddingBottom?: boolean;
+  allowHorizontalScrolling?: boolean;
 }
 
 const useStyles = makeStyles({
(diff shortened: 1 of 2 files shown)


antonbauhofer:
When scrolling is active, the highlighting for short paths stops at the path ending. It would be nice to have highlighting for the full width, but I could not think of an inexpensive way to implement this, yet.
Screenshot 2021-09-22 at 15 03 41

jonas-tai (approved):

src/Frontend/Components/List/List.tsx
```diff
@@ -26,10 +27,7 @@ const useStyles = makeStyles({
 function maxHeightWasGiven(
   max: NumberOfDisplayedItems | Height
 ): max is Height {
-  if ((max as Height).height) {
-    return true;
-  }
-  return false;
+  return !!(max as Height).height;
```
leslielazzarino, line 30: better Boolean(...) than !!

leslielazzarino (approved):

PR #9, opened 2021-09-22
Fix tests in Windows with manual fix for temporary folders.

leslielazzarino:


package.json
```diff
@@ -82,7 +82,7 @@
     "compile-all": "yarn update-commit-hash && tsc -p ./ && tsc --noEmit -p src/ElectronBackend",
     "test": "run-script-os",
     "test:darwin:linux": "react-scripts test --watchAll=false",
-    "test:win32": "EXIT 1",
+    "test:win32": "react-scripts test --watchAll=false",
```
nicocarltng, line 85: You can just remove the two sub-commands if the test command is the same for all operating systems :-)

src/ElectronBackend/test-helpers.ts
```diff
@@ -0,0 +1,9 @@
+import fs from 'fs';
+
+export function createTempFolder(): string {
+  return fs.mkdtempSync(`temp-folder-${Date.now()}`);
```
nicocarltng, line 4: It is better to use a UUID as jest makes massive use of parallelization and the chances of collision are not that small.

src/ElectronBackend/test-helpers.ts
```diff
@@ -0,0 +1,9 @@
...
+}
+
+export function deleteFolder(folderName: string): void {
+  fs.rmdirSync(folderName, { recursive: true });
```
nicarl, line 8: `DeprecationWarning: In future versions of Node.js, fs.rmdir(path, { recursive: true }) will be removed. Use fs.rm(path, { recursive: true }) instead`

src/ElectronBackend/test-helpers.ts
```diff
@@ -0,0 +1,9 @@
+import fs from 'fs';
+
+export function createTempFolder(): string {
+  return fs.mkdtempSync(`temp-folder-${Date.now()}`);
+}
+
+export function deleteFolder(folderName: string): void {
```
antonbauhofer, line 7: To avoid confusion, we could call the variable `folderPath` instead of `folderName`

antonbauhofer (commented):
See comments

antonbauhofer (commented):
I tested it on Mac and the app, as well as the tests are running fine.

package.json
leslielazzarino: True, I forgot to check :)

src/ElectronBackend/output/__tests__/writeSpdxFile.test.ts
```diff
@@ -27,10 +23,11 @@ describe('writeSpdxFile', () => {
     expect(fs.existsSync(yamlPath)).toBe(true);
     const fileContent = fs.readFileSync(yamlPath, 'utf-8');
     expect(fileContent).not.toBeNull();
+    deleteFolder(temporaryPath);
```
nicarl, line 26: Is there some mechanism like in python that you can have fixtures which are executed before and after a test? This wold be helpful, because I fear that one forgets to use `deleteFolder`.

nicarl (approved):

antonbauhofer (approved):

src/ElectronBackend/output/__tests__/writeSpdxFile.test.ts
leslielazzarino: Never heard of them. But you will not forget this, it will pollute your git repo with files and git will complain.

src/ElectronBackend/test-helpers.ts
```diff
@@ -0,0 +1,20 @@
+// SPDX-FileCopyrightText: Facebook, Inc. and its affiliates
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+
+import fs from 'fs';
+import { v4 as uuid4 } from 'uuid';
+
+export function createTempFolder(): string {
```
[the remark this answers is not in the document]
leslielazzarino: Usually, as a pattern, test helpers are not tested, and this is a one liner.

PR #12, opened 2021-09-23
Add icon for pre-selected attributions, remove gradients

sebathomas:
The gradients were hard to understand and their visibility was
not ideal. This adds an icon with hover text instead.

Additionally, we add a unique ID to the package card to avoid
duplicate key errors, which were probably already a problem
before but have become more likely since we now have many more
icons.

Screenshot_20210923_102546

### DIFF
--- a/src/Frontend/Components/AttributionList/AttributionList.tsx
+++ b/src/Frontend/Components/AttributionList/AttributionList.tsx
@@ -62,6 +62,7 @@ export function AttributionList(props: AttributionListProps): ReactElement {
         cardConfig={cardConfig}
         key={`AttributionCard-${attribution.packageName}-${attributionId}`}
         cardContent={{
+          id: `attribution-list-${attributionId}`,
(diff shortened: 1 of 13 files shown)


benedikt-richter (approved):

src/Frontend/Components/AttributionList/AttributionList.tsx
```diff
@@ -62,6 +62,7 @@ export function AttributionList(props: AttributionListProps): ReactElement {
         cardConfig={cardConfig}
         key={`AttributionCard-${attribution.packageName}-${attributionId}`}
         cardContent={{
+          id: `attribution-${attributionId}`,
```
antonbauhofer, line 65: nitpick: To be consistent with the id above, we could call this `attribution-list-${attributionId}`

antonbauhofer (approved):
Just one nitpick to consider

PR #14, opened 2021-09-23
Show more indicators for pre-selected attributions

sebathomas:
1. Show the new icon in report table view.
(This makes it necessary to increase the row height.)
Screenshot_20210923_104424

2. Show a hint in the attribution details column.
Screenshot_20210923_112002

### DIFF
--- a/src/Frontend/Components/AttributionColumn/ButtonRow.tsx
+++ b/src/Frontend/Components/AttributionColumn/ButtonRow.tsx
@@ -8,12 +8,19 @@ import React, { ReactElement } from 'react';
 import { PackageInfo } from '../../../shared/shared-types';
 import { ToggleButton } from '../ToggleButton/ToggleButton';
 import { ButtonGroup } from '../ButtonGroup/ButtonGroup';
+import MuiTypography from '@material-ui/core/Typography';
 import {
   getContextMenuButtonConfigs,
   getMainButtonConfigs,
 } from './attribution-column-helpers';
 
+const preSelectedLabel = 'Attribution was pre-selected';
+
 const useStyles = makeStyles({
+  preSelectedLabel: {
+    marginLeft: 10,
+    marginTop: 12,
+  },
   buttonRow: {
     position: 'absolute',
     bottom: 0,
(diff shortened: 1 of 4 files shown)


antonbauhofer:
When running the app on Mac, the text "Attribution was pre-selected" is not left-aligned with the content above.
Screenshot 2021-09-23 at 14 14 42

antonbauhofer:
Why do we have different blue tones for the icons? Should not the blue icons have the same color?
Screenshot 2021-09-23 at 14 17 25

src/Frontend/Components/AttributionColumn/ButtonRow.tsx
```diff
@@ -50,35 +57,45 @@ export function ButtonRow(props: ButtonRowProps): ReactElement {
...
-        <ToggleButton
-          buttonText={'hide'}
-          className={classes.resolveButton}
-          selected={props.selectedPackageIsResolved}
-          handleChange={props.resolvedToggleHandler}
-          ariaLabel={'resolve attribution'}
-        />
-      )}
-    </div>
+    <>
```
antonbauhofer, line 60:
```suggestion
    <div>
```

This fixes the left-alignment (see previous comment) with the content above on Mac.

src/Frontend/Components/AttributionColumn/ButtonRow.tsx
```diff
@@ -50,35 +57,45 @@ export function ButtonRow(props: ButtonRowProps): ReactElement {
...
+          <ToggleButton
+            buttonText={'hide'}
+            className={classes.resolveButton}
+            selected={props.selectedPackageIsResolved}
+            handleChange={props.resolvedToggleHandler}
+            ariaLabel={'resolve attribution'}
+          />
+        )}
+      </div>
+    </>
```
antonbauhofer, line 99:
```suggestion
    </div>
```

antonbauhofer (changes requested):
See comments

sebathomas:
> When running the app on Mac, the text "Attribution was pre-selected" is not left-aligned with the content above.
> Screenshot 2021-09-23 at 14 14 42

That one is really strange, I cannot reproduce it on different sizes and zoom levels. Seems to be OS dependent. (See screenshot in description.)

sebathomas:
> Why do we have different blue tones for the icons? Should not the blue icons have the same color?
> Screenshot 2021-09-23 at 14 17 25

I thought the idea was to have different colors for every icon. Matter of taste though.

sebathomas:
As discussed in person, we'll make the icon orange.

src/Frontend/Components/ReportTableItem/ReportTableItem.tsx
```diff
@@ -24,19 +25,22 @@ import { PathPredicate } from '../../types/types';
...
+const padding = 10;
+
 const useStyles = makeStyles({
   tableData: {
     overflow: 'auto',
     whiteSpace: 'pre-line',
-    padding: 10,
-    height: 140,
+    padding: padding,
+    height: reportTableRowHeight - 2 * padding,
```
nicarl, line 36: :+1: removing magic numbers is always good :mage_man:

nicarl (approved):

nicarl:
As discussed we agreed to keep the colors as they are

sebathomas:
> When running the app on Mac, the text "Attribution was pre-selected" is not left-aligned with the content above.
> Screenshot 2021-09-23 at 14 14 42

Looks like it was a difference between the auto-update version and the built version of the app (weird...). I think I fixed it. Can you try again?

antonbauhofer:
> > When running the app on Mac, the text "Attribution was pre-selected" is not left-aligned with the content above.
> > Screenshot 2021-09-23 at 14 14 42
> 
> Looks like it was a difference between the auto-update version and the built version of the app (weird...). I think I fixed it. Can you try again?

Great, it works now as expected.

antonbauhofer (approved):

PR #17, opened 2021-09-23
Fix app name

antonbauhofer:
Before this fix, the name of the built app was shown as 'opossum-ui'. With the fix, it is shown as 'OpossumUI'. This fixes the name of the executables and the name shown when the app is running.

### DIFF
--- a/.circleci/build_opossum_frontend.sh
+++ b/.circleci/build_opossum_frontend.sh
@@ -17,11 +17,11 @@ yarn ship-linux
 yarn ship-mac
 yarn ship-win
 cd release/macOS/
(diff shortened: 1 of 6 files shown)


antonbauhofer:
I tested this on Mac. Please check if the building and naming also works correct in Linux and Windows.

benedikt-richter (approved):

sebathomas (approved):
I can confirm that I can still build releases for all platforms on Linux.

There will be a merge conflict with the adapted documentation though.

antonbauhofer:
> I can confirm that I can still build releases for all platforms on Linux.
> 
> There will be a merge conflict with the adapted documentation though.

The conflict should be easy to fix.

PR #23, opened 2021-09-27
Show PURL if no package type is given in input file

antonbauhofer:
Before this fix, informations like namespace that are present in the PURL were not shown if an attribution did not have a package type. To make these information visible, we now generate a PURL with the type 'generic' that is shown when no type is present.


maxhbr (approved):
Code looks good

benedikt-richter (approved):
It's kind of a work around. However, fine for now.

antonbauhofer:
> It's kind of a work around. However, fine for now.

Yes, I have the same feeling.

PR #24, opened 2021-09-27
Fix a bug in the github action triggered upon release

nicarl:
The action triggered 3 runs with 3 build and publish steps (9 jobs in total) due to a problem with matrix definition.
Furthermore, the timeout for yarn install was increased (similar as in build.yml).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,9 +11,8 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     strategy:
-      fail-fast: true
+      fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
         include:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -11,9 +11,8 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     strategy:
-      fail-fast: true
+      fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
```
nicarl, line 16: see discussion here: https://github.community/t/create-matrix-with-multiple-os-and-env-for-each-one/16895/7

antonbauhofer (approved):
Let's cross fingers and see in the next release if this works.

sebathomas (approved):

PR #26, opened 2021-09-28
catch frontend errors that are not caught at error boundary

benedikt-richter:
In the past only frontend errors that were thrown during render were caught. Now we catch all errors in order to signal the user that something went wrong.
Note: in dev mode both, the error boundary and the event listener, are called in event of errors. In prod, however, this is not the case.

To test: -> test with build app
- create one error in render
-  create one error not in render

### DIFF
--- a/src/Frontend/Components/ErrorBoundary/ErrorBoundary.tsx
+++ b/src/Frontend/Components/ErrorBoundary/ErrorBoundary.tsx
@@ -21,6 +21,14 @@ const styles = createStyles({
   },
 });
 
+// catches errors that are not thrown during render
+// it's known to fire twice in dev mode: https://github.com/facebook/react/issues/19613
+window.addEventListener('error', (event): void => {
+  sendErrorInfo(event.error, {
+    componentStack: event.error.stack || '',
+  });
+});
+
 interface ErrorBoundaryState {
   hasError: boolean;
 }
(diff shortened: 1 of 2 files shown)


nicarl (commented):
It works but the stack trace is super long and I can not even see the close button for the popup :-( 
Is there a way to fix this in an easy way?

src/Frontend/Components/ErrorBoundary/ErrorBoundary.tsx
```diff
@@ -21,6 +21,14 @@ const styles = createStyles({
   },
 });
 
+// catches errors that are not thrown during render
+// it's known to fire twice in dev mode: https://github.com/facebook/react/issues/19613
+window.addEventListener('error', (event): void => {
+  sendErrorInfo(event.error, {
+    componentStack: event.error.stack || '',
```
leslielazzarino, line 28: I would prefer to have the fallback in the function itself, it would be cleaner

benedikt-richter:
> It works but the stack trace is super long and I can not even see the close button for the popup :-(
> Is there a way to fix this in an easy way?

I'll look into it

leslielazzarino (approved):

benedikt-richter:
> It works but the stack trace is super long and I can not even see the close button for the popup :-( Is there a way to fix this in an easy way?

On Mac that's not an issue. Should be due to the default settings for the system popups... A potential solution is to create a new window with custom html to be able to properly style it. However, not sure it's worth it currently.

Screenshot 2021-09-28 at 12 43 11

nicarl:
> > It works but the stack trace is super long and I can not even see the close button for the popup :-( Is there a way to fix this in an easy way?
> 
> On Mac that's not an issue. Should be due to the default settings for the system popups... A potential solution is to create a new window with custom html to be able to properly style it. However, not sure it's worth it currently.
> 
> Screenshot 2021-09-28 at 12 43 11

This is fine for me too.

nicarl (approved):

PR #27, opened 2021-09-28
Add resourceIdMarkedForReplacment to redux state

nicarl:
For adding a feature to merge/replace an attribution by another one it is necessary to remember the attribution to be replaced. This attributionId is written to the redux store. For now the feature is only available in the attributionView and therefore placed in the corresponding sub-store.

### DIFF
--- a/src/Frontend/state/reducers/resource-reducer.ts
+++ b/src/Frontend/state/reducers/resource-reducer.ts
@@ -29,6 +29,7 @@ import {
   ACTION_REPLACE_ATTRIBUTION_WITH_MATCHING,
   ACTION_RESET_RESOURCE_STATE,
   ACTION_SET_ATTRIBUTION_BREAKPOINTS,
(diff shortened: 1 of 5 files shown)


src/Frontend/state/reducers/resource-reducer.ts
```diff
@@ -92,6 +93,7 @@ export const initialResourceState: ResourceState = {
   attributionView: {
     selectedAttributionId: '',
     targetSelectedAttributionId: '',
+    attributionIdMarkedForReplacement: '',
```
nicarl, line 96: I have no strong opinion on making this value nullable instead of an empty string. What is your opinion here? Seems to be a pattern we partly have.

sebathomas: I think it's fine to stick to the pattern, probably doesn't make a big difference in usage.

sebathomas (approved):

benedikt-richter (approved):

PR #31, opened 2021-09-28
Add entries in contextMenu to allow merging of attributions

nicarl:
- add a helper to calculate which button is shown and activated
- add new entries for buttons

### DIFF
--- a/src/Frontend/Components/AttributionColumn/ButtonRow.tsx
+++ b/src/Frontend/Components/AttributionColumn/ButtonRow.tsx
@@ -45,12 +45,19 @@ interface ButtonRowProps {
   initialPackageInfo: PackageInfo;
   hideDeleteButtons?: boolean;
   selectedPackageIsResolved: boolean;
+  hideMarkForReplacementButton: boolean;
+  hideUnmarkForReplacementButton: boolean;
+  hideOnReplaceMarkedByButton: boolean;
+  deactivateReplaceMarkedByButton: boolean;
   onSaveButtonClick(): void;
   onSaveForAllButtonClick(): void;
   onDeleteButtonClick(): void;
   onDeleteForAllButtonClick(): void;
   onUndoButtonClick(): void;
   resolvedToggleHandler(): void;
+  onMarkForReplacementButtonClick(): void;
+  onUnmarkForReplacementButtonClick(): void;
+  onReplaceMarkedByButtonClick(): void;
 }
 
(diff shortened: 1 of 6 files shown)


src/Frontend/Components/AttributionColumn/ButtonRow.tsx
```diff
@@ -45,12 +45,19 @@ interface ButtonRowProps {
...
+  hideOnReplaceMarkedByButton: boolean;
+  deactivateReplaceMarkedByButton: boolean;
   onSaveButtonClick(): void;
   onSaveForAllButtonClick(): void;
   onDeleteButtonClick(): void;
   onDeleteForAllButtonClick(): void;
   onUndoButtonClick(): void;
   resolvedToggleHandler(): void;
+  onMarkForReplacementButtonClick(): void;
+  onUnmarkForReplacementButtonClick(): void;
```
benedikt-richter, line 59: The bottonsRows interface could be more generic to accept an array containing the config of the individual buttons. However, as there are currently other diffs open touching the same code. I'd rather not refactor it here.

benedikt-richter (approved):
nice!

src/Frontend/Components/AttributionColumn/ButtonRow.tsx
nicarl: I already hesitated to refactor this component... lets do it at the end of the feature.
I think that we still need a small timeout when clicking the button (it switches right away to `Unmark for replacement`). As a user I would prefer to not see this change, only when I open the context menu again.

src/Frontend/Components/AttributionColumn/attribution-column-helpers.ts
```diff
@@ -178,5 +185,57 @@ export function getContextMenuButtonConfigs(
...
+      !isAttributionView ||
+      !anyAttributionMarkedForReplacement ||
+      !selectedAttributionIsMarkedForReplacement,
+    hideOnReplaceMarkedByButton:
+      !isAttributionView ||
+      !anyAttributionMarkedForReplacement ||
+      selectedAttributionIsMarkedForReplacement,
+    deactivateReplaceMarkedByButton:
+      packageInfoWereModified || attributionIsPreSelected,
+  };
```
leslielazzarino, line 240: Yes, this stuff is growing out of control, it should be refactored at the end of the story.

leslielazzarino (approved):

PR #33, opened 2021-09-29
Integrate popup

benedikt-richter:
The ReplaceAttributionPopup is integrated to enable the full workflow.
Furthermore, a integration test is added and the definition for two attributions to be equal is improved (we now ignore the confidence for that decision).


src/Frontend/Components/ReplaceAttributionPopup/__tests__/ReplaceAttributionPopup.test.tsx
```diff
@@ -31,8 +38,9 @@ describe('ReplaceAttributionPopup and do not change view', () => {
...
     setupTestState(store);
 
-    expect(screen.queryByText('Warning')).toBeTruthy();
-    expect(screen.queryAllByText('React').length).toBe(2);
+    expect(screen.queryByText('Replacing an attribution')).toBeTruthy();
+    expect(screen.queryByText('React')).toBeTruthy();
+    expect(screen.queryByText('Vue')).toBeTruthy();
 
     fireEvent.click(screen.queryByText(ButtonTitle.Cancel) as Element);
     expect(getOpenPopup(store.getState())).toBe(null);
```
nicarl, line 46: Could we also add a test here for clicking okay?

src/Frontend/state/helpers/save-action-helpers.ts
```diff
@@ -457,7 +457,17 @@ export function attributionForTemporaryPackageInfoExists(
 ): boolean {
   const manualAttributions = getManualAttributions(state);
   return Object.values(manualAttributions).some((packageInfo) =>
-    isEqual(packageInfo, packageInfoToMatch)
+    attributionsAreEqual(packageInfo, packageInfoToMatch)
+  );
+}
+
+export function attributionsAreEqual(
```
nicarl, line 464: For next time: here a commit message with additional information would be super valuable for the future

nicarl (approved):
Great integration test :+1: Works nicely

src/Frontend/Components/ReplaceAttributionPopup/__tests__/ReplaceAttributionPopup.test.tsx
nicarl: Somehow optional as tested by integration test

leslielazzarino (approved):

PR #35, opened 2021-09-29
add the following lint plugin: plugin:testing-library/react

benedikt-richter:
add the following lint plugin: plugin:testing-library/react

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -10,18 +10,6 @@ module.exports = {
     node: true,
     jest: true,
   },
-  parser: '@typescript-eslint/parser',
-  extends: [
-    'plugin:react/recommended',
-    'plugin:@typescript-eslint/recommended',
-    'plugin:prettier/recommended',
-    'prettier',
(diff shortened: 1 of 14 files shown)


sebathomas (approved):

.eslintrc.js
```diff
@@ -46,6 +47,7 @@ module.exports = {
     '@typescript-eslint/await-thenable': 2,
     '@typescript-eslint/require-await': 2,
     'react/prop-types': 'off',
+    'testing-library/no-node-access': 'off', // TODO: remove this line and fix warnings
```
leslielazzarino, line 50: If they are just warnings I would remove this line, so that we keep seeing them and fix them at some point. This todo will be forgotten.

leslielazzarino (changes requested):

.eslintrc.js
benedikt-richter: These are errors. Therefore, we have tp turn these off

leslielazzarino (approved):

PR #36, opened 2021-09-29
Fix bug when selecting frequent licenses

nicarl:
There was some hidden magic that converted the license name the user wrote to uppercase letters if there was a match with a frequent license. This logic is removed and the user has to actively choose the frequent license now.
Only in the case of a perfect match (user input: MIT, frequent license: MIT) the frequent license is automatically chosen.
Also removed some formatting, which was confusing for the user and added an integration test.

### DIFF
--- a/src/Frontend/Components/InputElements/AutoComplete.tsx
+++ b/src/Frontend/Components/InputElements/AutoComplete.tsx
@@ -46,8 +46,8 @@ export function AutoComplete(props: AutoCompleteProps): ReactElement {
 
   const inputValue = props.text || '';
   const inputValueIndexInOptions: number = props.options
-    .map((name) => name.toLowerCase())
-    .indexOf(inputValue.toLowerCase());
+    .map((name) => name)
+    .indexOf(inputValue);
   const isInputValueInOptions: boolean = inputValueIndexInOptions > -1;
 
   return (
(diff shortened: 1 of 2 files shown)


src/Frontend/integration-tests/__tests__/navigate-around.test.tsx
```diff
@@ -229,4 +229,48 @@ describe('The App integration', () => {
...
+    screen.getByText('gpl-2.0');
+    expect(screen.queryByText('frequent license')).toBeFalsy();
+
+    goToView(screen, View.Audit);
+    insertValueIntoTextBox(screen, 'License Name', 'GPL-2.0');
+
+    clickOnButton(screen, ButtonTitle.Save);
+    goToView(screen, View.Report);
+    screen.getByText('GPL-2.0');
+    screen.getByText('frequent license');
```
nicarl, line 274: Maybe also test the select feature of the dropdown?

benedikt-richter (approved):
👍

src/Frontend/integration-tests/__tests__/navigate-around.test.tsx
nicarl: see part below

JakobSchubert (approved):

src/Frontend/Components/InputElements/AutoComplete.tsx
```diff
@@ -46,8 +46,8 @@ export function AutoComplete(props: AutoCompleteProps): ReactElement {
 
   const inputValue = props.text || '';
   const inputValueIndexInOptions: number = props.options
-    .map((name) => name.toLowerCase())
-    .indexOf(inputValue.toLowerCase());
+    .map((name) => name)
```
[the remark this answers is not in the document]
nicarl: Good catch

PR #38, opened 2021-09-30
Refactor button row

benedikt-richter:
The botton row is refactored such that the config is created outside and provided via props.
Two custom hooks are extracted from the attribution coulmn.

### DIFF
--- a/src/Frontend/Components/AttributionColumn/AttributionColumn.tsx
+++ b/src/Frontend/Components/AttributionColumn/AttributionColumn.tsx
@@ -5,7 +5,7 @@
 
 import { makeStyles } from '@material-ui/core/styles';
 import clsx from 'clsx';
-import React, { ChangeEvent, ReactElement, useEffect, useState } from 'react';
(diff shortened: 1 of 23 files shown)


nicarl (approved):

src/Frontend/Components/AttributionColumn/AttributionColumn.tsx
```diff
@@ -112,43 +119,74 @@ export function AttributionColumn(props: AttributionColumnProps): ReactElement {
...
+        : ButtonTitle.SaveForAll,
+      disabled: isSavingDisabled,
+      onClick: props.onSaveForAllButtonClick,
+      hidden: !Boolean(props.showSaveForAllButton),
+    },
+  ];
+
+  const contextMenuButtonConfigs = [
+    {
+      buttonTitle: ButtonTitle.Undo,
```
leslielazzarino, line 143: I would either call both text or both title in the two configs

leslielazzarino (approved):

PR #39, opened 2021-09-30
Activate CI for spdx-tools-ts

nicarl:
I had issues also setting up eslint (it got confused with a second config file in subfolder). It should be set up at some point

### DIFF
--- a/spdx-tools-ts/package.json
+++ b/spdx-tools-ts/package.json
@@ -6,22 +6,24 @@
   "private": true,
   "homepage": "./",
   "dependencies": {
-    "js-yaml": "^4.1.0",
-    "packageurl-js": "^0.0.5",
-    "uuid": "^8.3.2",
-    "typescript": "^4.3.5",
     "@types/jest": "^27.0.1",
     "@types/js-yaml": "^4.0.3",
-    "@types/uuid": "^8.3.1"
+    "@types/uuid": "^8.3.1",
+    "js-yaml": "^4.1.0",
+    "packageurl-js": "^0.0.5",
+    "typescript": "^4.3.5",
+    "uuid": "^8.3.2"
   },
   "devDependencies": {
     "@babel/preset-typescript": "^7.15.0",
     "jest": "^27.2.0",
+    "prettier": "^2.4.1",
     "ts-jest": "^27.0.5"
   },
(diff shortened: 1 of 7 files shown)


spdx-tools-ts/package.json
```diff
@@ -6,22 +6,24 @@
   "private": true,
   "homepage": "./",
   "dependencies": {
+    "@types/jest": "^27.0.1",
+    "@types/js-yaml": "^4.0.3",
+    "@types/uuid": "^8.3.1",
```
benedikt-richter, line 11: these are dev dependencies

spdx-tools-ts/package.json
```diff
@@ -6,22 +6,24 @@
...
+    "uuid": "^8.3.2"
   },
   "devDependencies": {
     "@babel/preset-typescript": "^7.15.0",
     "jest": "^27.2.0",
+    "prettier": "^2.4.1",
     "ts-jest": "^27.0.5"
   },
   "scripts": {
     "postinstall": "tsc",
```
benedikt-richter, line 24: if we have "compile" it should be used in "postinstall"

spdx-tools-ts/package.json
nicarl: No this will not work. As `tsc` is part of the postinstall it needs the corresponding types when we install it in the main package json

spdx-tools-ts/package.json
```diff
@@ -6,22 +6,24 @@
...
     "jest": "^27.2.0",
+    "prettier": "^2.4.1",
     "ts-jest": "^27.0.5"
   },
   "scripts": {
-    "postinstall": "tsc",
-    "test": "jest"
+    "test": "jest",
+    "compile": "tsc",
+    "postinstall": "compile"
```
sebathomas, line 26:
```suggestion
    "postinstall": "yarn compile"
```

spdx-tools-ts/package.json
nicarl: This does not work... https://github.com/opossum-tool/OpossumUI/pull/39/checks?check_run_id=3753216416

sebathomas (approved):

benedikt-richter (approved):

PR #41, opened 2021-09-30
Fix bug of deleting an marked attribution

benedikt-richter:
The following bug is fixed: If an attribution is marked and then deleted, subsequently replacing by marked throws.

### DIFF
--- a/src/Frontend/state/reducers/resource-reducer.ts
+++ b/src/Frontend/state/reducers/resource-reducer.ts
@@ -366,27 +366,29 @@ export const resourceState = (
         },
       };
     case ACTION_DELETE_ATTRIBUTION:
+      const attributionToDeleteId = action.payload;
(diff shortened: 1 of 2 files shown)


nicarl (approved):

src/Frontend/state/reducers/resource-reducer.ts
```diff
@@ -384,6 +384,12 @@ export const resourceState = (
           ? ''
           : state.attributionView.selectedAttributionId;
 
+      const newAttributionIdMarkedForReplacement: string =
+        state.attributionView.attributionIdMarkedForReplacement ===
+        action.payload
```
sebathomas, line 389: The action is getting long enough that it might be helpful to extract `action.payload` to a variable with a more readable name.

sebathomas (approved):

PR #44, opened 2021-10-01
Do not show 'marked for replacement' label in Audit View

sebathomas:
The check was broken because it is based on variables of the
Attribution View that are constant while the user is using the
Audit View.

Signed-off-by: Sebastian Thomas <sebastian.thomas@tngtech.com>


nicarl (approved):

maxhbr (approved):
Quickfix, but will fail once we have the "mark for replacement" feature in the AUDIT view

PR #51, opened 2021-10-04
Split ResourcePathPopup resource list in current and other folders

jonas-tai:
If resources in other folders exist, the resource list shows the resources in the current folder at the top and the resources in other folders underneath, separated by a new header.

Signed-off-by: Jonas Tai <jonas.tai@tngtech.com>

### DIFF
--- a/src/Frontend/Components/ResourcePathPopup/__tests__/resource-path-popup-helpers.test.ts
+++ b/src/Frontend/Components/ResourcePathPopup/__tests__/resource-path-popup-helpers.test.ts
@@ -3,44 +3,7 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
-import {
-  insertHeaderForOtherFolderResourceIds,
-  splitResourceItToCurrentAndOtherFolder,
-} from '../resource-path-popup-helpers';
-
-describe('insertHeaderForOtherFolderResourceIds', () => {
-  test('inserts headers if resources in other folders present', () => {
-    const currentFolderResourceIds = ['resource_1'];
-    const otherFolderResourceIds = ['resource_2', 'resource_3'];
-
-    const headerIndices = insertHeaderForOtherFolderResourceIds(
-      currentFolderResourceIds,
-      otherFolderResourceIds
-    );
-
-    expect(headerIndices).toStrictEqual([1, 2]);
-    expect(currentFolderResourceIds).toStrictEqual([
-      'resource_1',
-      '',
-      'Resources in Other Folders',
-    ]);
-    expect(otherFolderResourceIds).toStrictEqual(['resource_2', 'resource_3']);
-  });
-
-  test('inserts no headers if no resources in other folders present', () => {
-    const currentFolderResourceIds = ['resource_1'];
-    const otherFolderResourceIds: string[] = [];
-
-    const headerIndices = insertHeaderForOtherFolderResourceIds(
-      currentFolderResourceIds,
-      otherFolderResourceIds
-    );
-
-    expect(headerIndices).toStrictEqual([]);
-    expect(currentFolderResourceIds).toStrictEqual(['resource_1']);
-    expect(otherFolderResourceIds).toStrictEqual([]);
-  });
-});
+import { splitResourceItToCurrentAndOtherFolder } from '../resource-path-popup-helpers';
 
 describe('splitResourceItToCurrentAndOtherFolder', () => {
   test('splits resource ids correctly', () => {
@@ -51,7 +14,7 @@ describe('splitResourceItToCurrentAndOtherFolder', () => {
     ];
     const folderPath = '/folder_1/';
 
-    const [currentFolderResourceIds, otherFolderResourceIds] =
+    const { currentFolderResourceIds, otherFolderResourceIds } =
       splitResourceItToCurrentAndOtherFolder(allResourceIds, folderPath);
 
     expect(currentFolderResourceIds).toStrictEqual([
@@ -65,7 +28,7 @@ describe('splitResourceItToCurrentAndOtherFolder', () => {
     const allResourceIds = undefined;
     const folderPath = '/folder_1/';
 
-    const [currentFolderResourceIds, otherFolderResourceIds] =
+    const { currentFolderResourceIds, otherFolderResourceIds } =
       // @ts-ignore
       splitResourceItToCurrentAndOtherFolder(allResourceIds, folderPath);
 
--- a/src/Frontend/Components/ResourcePathPopup/__tests__/resource-path-popup-helpers.test.ts
+++ b/src/Frontend/Components/ResourcePathPopup/__tests__/resource-path-popup-helpers.test.ts
@@ -1,0 +1,75 @@
+// SPDX-FileCopyrightText: Facebook, Inc. and its affiliates
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+
+import {
+  insertHeaderForOtherFolderResourceIds,
+  splitResourceItToCurrentAndOtherFolder,
+} from '../resource-path-popup-helpers';
+
+describe('insertHeaderForOtherFolderResourceIds', () => {
+  test('inserts headers if resources in other folders present', () => {
+    const currentFolderResourceIds = ['resource_1'];
+    const otherFolderResourceIds = ['resource_2', 'resource_3'];
+
+    const headerIndices = insertHeaderForOtherFolderResourceIds(
+      currentFolderResourceIds,
+      otherFolderResourceIds
+    );
+
+    expect(headerIndices).toStrictEqual([1, 2]);
+    expect(currentFolderResourceIds).toStrictEqual([
+      'resource_1',
+      '',
+      'Resources in Other Folders',
+    ]);
+    expect(otherFolderResourceIds).toStrictEqual(['resource_2', 'resource_3']);
+  });
+
+  test('inserts no headers if no resources in other folders present', () => {
+    const currentFolderResourceIds = ['resource_1'];
+    const otherFolderResourceIds: string[] = [];
+
+    const headerIndices = insertHeaderForOtherFolderResourceIds(
+      currentFolderResourceIds,
+      otherFolderResourceIds
+    );
+
+    expect(headerIndices).toStrictEqual([]);
+    expect(currentFolderResourceIds).toStrictEqual(['resource_1']);
+    expect(otherFolderResourceIds).toStrictEqual([]);
+  });
+});
+
+describe('splitResourceItToCurrentAndOtherFolder', () => {
+  test('splits resource ids correctly', () => {
+    const allResourceIds = [
+      '/folder_1/',
+      '/folder_1/resource_1',
+      '/folder_2/resource_3',
+    ];
+    const folderPath = '/folder_1/';
+
+    const [currentFolderResourceIds, otherFolderResourceIds] =
+      splitResourceItToCurrentAndOtherFolder(allResourceIds, folderPath);
+
+    expect(currentFolderResourceIds).toStrictEqual([
+      '/folder_1/',
+      '/folder_1/resource_1',
+    ]);
+    expect(otherFolderResourceIds).toStrictEqual(['/folder_2/resource_3']);
+  });
+
+  test('handles undefined allResourceIds correctly', () => {
+    const allResourceIds = undefined;
+    const folderPath = '/folder_1/';
+
+    const [currentFolderResourceIds, otherFolderResourceIds] =
+      // @ts-ignore
+      splitResourceItToCurrentAndOtherFolder(allResourceIds, folderPath);
+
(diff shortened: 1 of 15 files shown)


src/Frontend/Components/ResourcePathPopup/resource-path-popup-helpers.ts
```diff
@@ -0,0 +1,52 @@
+// SPDX-FileCopyrightText: Facebook, Inc. and its affiliates
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+
+export function insertHeaderForOtherFolderResourceIds(
+  currentFolderResourceIds: string[],
```
benedikt-richter, line 7: Arrays should use `Array<string>`

benedikt-richter: also other places

src/Frontend/Components/ResourcePathPopup/resource-path-popup-helpers.ts
```diff
@@ -0,0 +1,52 @@
...
+    currentFolderResourceIds.push('', otherFolderResourceIdsHeader);
+    otherFolderResourceIds.sort();
+  }
+  return headerIndices;
+}
+
+export function splitResourceItToCurrentAndOtherFolder(
+  allResourceIds: string[],
+  folderPath: string
+): [string[], string[]] {
```
benedikt-richter, line 28: here, returning an object improves clarity, as keys can have speaking names.

src/Frontend/Components/ResourcesList/ResourcesList.tsx
```diff
@@ -44,6 +47,16 @@ export function ResourcesList(props: ResourcesListProps): ReactElement {
...
+    if (props.headerIndices && props.headerIndices.includes(index)) {
+      return (
+        <ListCard
+          text={resourcePath}
+          onClick={doNothing}
+          cardConfig={{ isHeader: true }}
+        />
+      );
+    }
+
```
benedikt-richter, line 59: would be nice to have a single return.
Instead of using 'doNothing' the prop `onClick' could be optional

src/Frontend/Components/ResourcePathPopup/resource-path-popup-helpers.ts
```diff
@@ -0,0 +1,52 @@
+// SPDX-FileCopyrightText: Facebook, Inc. and its affiliates
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+
+export function insertHeaderForOtherFolderResourceIds(
```
benedikt-richter, line 6: functions should not have dide effects. And if they have they should not return a value.

src/Frontend/Components/ResourcesList/ResourcesList.tsx
```diff
@@ -24,6 +24,7 @@ const useStyles = makeStyles({
 
 interface ResourcesListProps {
   resourceIds: Array<string>;
+  headerIndices?: Array<number>;
```
benedikt-richter, line 27: instead of indices, we could provide the actual paths

src/Frontend/Components/ResourcesList/ResourcesList.tsx
leslielazzarino: I agree on combining the two returns int oone, makes it clear what the difference is. But I am fine with doNothing: it makes it more prominent that in this very special case the card is not clickable.

leslielazzarino (changes requested):
I agree on all comments from Benedikt, where I haven't commented.

leslielazzarino (approved):

src/Frontend/Components/ResourcePathPopup/resource-path-popup-helpers.ts
```diff
@@ -0,0 +1,26 @@
...
+  (allResourceIds ?? []).forEach((resourceId) => {
+    resourceId.startsWith(folderPath)
+      ? currentFolderResourceIds.push(resourceId)
+      : otherFolderResourceIds.push(resourceId);
+  });
+  currentFolderResourceIds.sort();
+  return {
+    currentFolderResourceIds: currentFolderResourceIds,
+    otherFolderResourceIds: otherFolderResourceIds,
+  };
```
benedikt-richter, line 25: could be 

```
  return {
    currentFolderResourceIds: currentFolderResourceIds.sort(),
    otherFolderResourceIds
  };


```

src/Frontend/Components/ResourcePathPopup/resource-path-popup-helpers.ts
```diff
@@ -0,0 +1,26 @@
...
+} {
+  const currentFolderResourceIds: Array<string> = [];
+  const otherFolderResourceIds: Array<string> = [];
+
+  (allResourceIds ?? []).forEach((resourceId) => {
+    resourceId.startsWith(folderPath)
+      ? currentFolderResourceIds.push(resourceId)
+      : otherFolderResourceIds.push(resourceId);
+  });
+  currentFolderResourceIds.sort();
```
benedikt-richter, line 21: why are `otherFolderResourceIds` not sorted?

src/Frontend/Components/ResourcePathPopup/resource-path-popup-helpers.ts
```diff
@@ -0,0 +1,26 @@
...
+  allResourceIds: Array<string>,
+  folderPath: string
+): {
+  currentFolderResourceIds: Array<string>;
+  otherFolderResourceIds: Array<string>;
+} {
+  const currentFolderResourceIds: Array<string> = [];
+  const otherFolderResourceIds: Array<string> = [];
+
+  (allResourceIds ?? []).forEach((resourceId) => {
```
benedikt-richter, line 16: could be `allResourceIds?.forEach((resourceId) => {`

src/Frontend/Components/ResourcesList/ResourcesList.tsx
```diff
@@ -32,25 +34,39 @@ export function ResourcesList(props: ResourcesListProps): ReactElement {
 
   const isFileWithChildren = useSelector(getIsFileWithChildren);
   const dispatch = useDispatch();
-  const sortedResourcePaths = props.resourceIds.sort();
   const onClickCallback = props.onClickCallback ?? doNothing;
 
+  props.resourcesListBatches.forEach((resourcesListBatch) => {
+    resourcesListBatch.resourceIds.sort();
+  });
```
benedikt-richter, line 41: shouzld probably be moved into : `convertResourcesListBatchesToResourcesListItems`

src/Frontend/Components/ResourcePathPopup/resource-path-popup-helpers.ts
```diff
@@ -0,0 +1,26 @@
+// SPDX-FileCopyrightText: Facebook, Inc. and its affiliates
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+
+export function splitResourceItToCurrentAndOtherFolder(
```
benedikt-richter, line 6: `splitResourceItToCurrentAndOtherFolder` -> `splitResourceIdsToCurrentAndOtherFolder`

src/Frontend/Components/ResourcesList/ResourcesList.tsx
```diff
@@ -13,6 +13,8 @@ import { doNothing } from '../../util/do-nothing';
 import { removeTrailingSlashIfFileWithChildren } from '../../util/remove-trailing-slash-if-file-with-children';
 import { getIsFileWithChildren } from '../../state/selectors/all-views-resource-selectors';
 import { OpossumColors } from '../../shared-styles';
+import { ResourcesListBatch, ResourcesListItem } from '../../types/types';
```
benedikt-richter, line 16: would be nice to define `ResourcesListItem` in the component and export from there as it's quite specific

src/Frontend/Components/ResourcesList/ResourcesList.tsx
```diff
@@ -32,25 +34,39 @@ export function ResourcesList(props: ResourcesListProps): ReactElement {
...
-        onClick={onPathClick}
-        cardConfig={{ isResource: true }}
+        text={
+          resourcesListItems[index].isHeader
+            ? cardText
+            : removeTrailingSlashIfFileWithChildren(
+                cardText,
+                isFileWithChildren
+              )
+        }
```
benedikt-richter, line 63: could be extracted into a constant as there is quite some logic involved

src/Frontend/Components/ResourcesList/ResourcesList.tsx
```diff
@@ -32,25 +34,39 @@ export function ResourcesList(props: ResourcesListProps): ReactElement {
...
+        onClick={resourcesListItems[index].isHeader ? doNothing : onPathClick}
```
benedikt-richter, line 64: `resourcesListItems[index].isHeader` could be a constant as it's used repeatedly

src/Frontend/Components/ResourcesList/__tests__/resource-list-helpers.test.ts
```diff
@@ -0,0 +1,43 @@
+// SPDX-FileCopyrightText: Facebook, Inc. and its affiliates
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+
+import { convertResourcesListBatchesToResourcesListItems } from '../resource-list-helpers';
+
+describe('convertResourceListBatchesToResourceListItems', () => {
+  test('inserts headers correctly', () => {
+    const resourcesListBatches = [
```
benedikt-richter, line 10: in tests it is very helpful to declare the type for such test data `const resourcesListBatches: $TYPE`

src/Frontend/Components/ResourcesList/__tests__/resource-list-helpers.test.ts
```diff
@@ -0,0 +1,43 @@
...
+      convertResourcesListBatchesToResourcesListItems(resourcesListBatches);
+
+    expect(resourcesListItems).toStrictEqual([
+      { text: 'resource_1' },
+      { text: '', isHeader: true },
+      { text: 'Resources in Other Folders', isHeader: true },
+      { text: 'resource_2' },
+      { text: 'resource_3' },
+    ]);
+  });
```
benedikt-richter, line 28: should be extracted to constant. Similar in some other tests

src/Frontend/state/helpers/action-and-reducer-helpers.ts
```diff
@@ -105,7 +105,7 @@ export function addUnresolvedAttributionsToResourcesWithAttributedChildren(
 
 export function removeResolvedAttributionsFromResourcesWithAttributedChildren(
   resourcesWithAttributedChildren: ResourcesWithAttributedChildren,
-  resourceIds: string[]
+  resourceIds: Array<string>
```
benedikt-richter, line 108: 👍

benedikt-richter (changes requested):
Looks good and works nicely 👍 

Just some code style comments

src/Frontend/Components/ResourcePathPopup/resource-path-popup-helpers.ts
jonas-tai: Deleted the sorting, because the ResourcesList component already does that

src/Frontend/Components/ResourcePathPopup/resource-path-popup-helpers.ts
jonas-tai: No longer required because of the ResourcesList component

src/Frontend/Components/ResourcesList/ResourcesList.tsx
jonas-tai: I decided to separate them, in case we do not want to sort for other use cases of the component.

src/Frontend/Components/ResourcePathPopup/__tests__/resource-path-popup-helpers.test.ts
```diff
@@ -0,0 +1,45 @@
...
+    const allResourceIds = undefined;
+    const folderPath = '/folder_1/';
+
+    const { currentFolderResourceIds, otherFolderResourceIds } =
+      // @ts-ignore
+      splitResourceIdsToCurrentAndOtherFolder(allResourceIds, folderPath);
+
+    expect(currentFolderResourceIds).toStrictEqual([]);
+    expect(otherFolderResourceIds).toStrictEqual([]);
+  });
```
benedikt-richter, line 44: What is the purpose of this test?
The function signature suggests that `allResourceIds` is an array of strings. If it is not the function signature should be adapted. However, I guess, we can just remove this test.

benedikt-richter (approved):
nice 👍 

just a comment regarding a test

PR #70, opened 2021-10-05
Persist follow-up filter flag when changing views

antonbauhofer:
Before this fix, the follow-up filter flag was cleaned when the view was changed.


src/Frontend/util/use-follow-up-filter.ts
```diff
@@ -18,9 +21,11 @@ export function useFollowUpFilter(): {
     attributions: AttributionsWithResources | Attributions
   ): AttributionsWithResources | Attributions;
 } {
-  const [filterForFollowUp, setFilterForFollowUp] = useState(false);
+  const dispatch = useDispatch();
```
benedikt-richter, line 24: since the local state is removed, this should not be a cusom hook

src/Frontend/util/use-follow-up-filter.ts
```diff
@@ -18,9 +21,11 @@ export function useFollowUpFilter(): {
...
+
+  const filterForFollowUp = useSelector(areOnlyFollowUpAttributionsShown);
```
benedikt-richter, line 26: `filterForFollowUp` could be an input that is provided by the component that calls it

src/Frontend/util/use-follow-up-filter.ts
benedikt-richter, line 24: also dispatch could be an input. 
Then this would not be  acustom hook anymore, but a "normal" util function.

benedikt-richter (changes requested):
works 👍 
would be nice to get rid of the custom hook

src/Frontend/util/__tests__/use-follow-up-filter.test.tsx
```diff
@@ -3,17 +3,24 @@
 //
```
benedikt-richter, line 3: nsme of file has to be adapted

src/Frontend/util/provide-follow-up-filter.ts
```diff
@@ -9,21 +9,19 @@ import {
...
 } from '../../shared/shared-types';
-import { areOnlyFollowUpAttributionsShown } from '../state/selectors/view-selector';
-import { useSelector } from 'react-redux';
 import { setFollowUpFilter } from '../state/actions/view-actions/view-actions';
-import { useDispatch } from 'react-redux';
+import { SimpleThunkDispatch } from '../state/actions/types';
 
-export function useFollowUpFilter(): {
+export function provideFollowUpFilter(
+  filterForFollowUp: boolean,
```
benedikt-richter, line 16: `filterForFollowUp` should not be returned. It is given as input and un-modified returned again

PR #100, opened 2021-10-14
add devcontainer configuration

maxhbr:
This adds [devcontainer](https://code.visualstudio.com/docs/remote/containers) support for developing containerized within vscode.

This should also allow to use GitHub Codespaces.


maxhbr:
Yes, open with VSCode, and then it should ask you whether to reopen in container

PR #114, opened 2021-10-20
Add context menu for package cards show on right click

jonas-tai:
Add ContextMenu for package cards that shows on right click

This commit adds a context menu for external and manual attributions PackageCards to allow for faster execution of operations on attributions (delete, confirm, hide, show resources). This commit only adds the correctly displayed options and not the operations themself. There are different options, depending on external / manual attribution and whether the attributions belongs to multiple resources. The context menu is implemented as a separate WithContextMenu component, which can be used to open a context menu on right or left click for any component, by using it as the wrapping component and providing the menu items as a parameter.

To test the changes, open a file in Opossum and right click on attributions / signals and open the button context menu next to the save button.


benedikt-richter:
should be rebased as it's 50 commits behind master

PR #146, opened 2021-10-29
Remove unused devDependencies

benedikt-richter:
### Summary of changes

* Unused dependencies are removed
* the package is renamed to only consist of valid characters

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "OpossumUI",
+  "name": "opossum-ui",
   "description": "The OpossumUI enables the editing of attribution information that is assigned to a resource tree.",
   "license": "Apache-2.0",
   "version": "0.1.0",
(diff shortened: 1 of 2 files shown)


nicarl (approved):

nicarl (changes requested):
Can you also modify the release github action?

benedikt-richter:
The github action does not have to be modified as we rename the files when shipping. These internals are encapsulated in the ship scripts :)

nicarl (approved):

package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "OpossumUI",
+  "name": "opossum-ui",
```
antonbauhofer, line 2: This changes the shown name of the app (i.e. for Mac, the name shown in the top left of the screen and in the applications bar at the bottom). I am not sure about the other systems, but in Mac it is common, to have the name shown including capitalization.
I think I would rather have the warning than changing the name of the app. Or is there another fix for this?

antonbauhofer (changes requested):

package.json
benedikt-richter: Good catch! I agree that we want to show the name "OpossumUI". The config for the builder is adapted to make sure that's the case even though we use a different name here.

antonbauhofer (approved):
Looks good now.

PR #151, opened 2021-11-02
Add e2e test to CI

nicarl:
### Summary of changes

Adds a step to the CI which executes the CI for linux and mac. The trick is to tell concurrently to stop when jest exits and persist jests exit code upwards. Also force jest to exit when tests are done.
Increase file watchers for linux. The end2end tests currently do not work under windows yet.

We recently added e2e-tests, but the tests were not running in the CI yet.

`yarn e2e-test`

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -31,6 +31,9 @@ jobs:
         uses: actions/setup-node@v1
         with:
           node-version: 16
+      - name: Increase watchers
+        run: echo fs.inotify.max_user_watches=524288 | sudo tee -a /etc/sysctl.conf && sudo sysctl -p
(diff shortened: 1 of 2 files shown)


benedikt-richter:
Nice! 

There is a minor issue (at least for mac. Is it the same on Linux?). After the e2e tests run there is a exit code 1 (related to https://github.com/yarnpkg/yarn/issues/4667).

Screenshot 2021-11-03 at 07 47 14

benedikt-richter (approved):
It's great to have the e2e tests in the CI!

The other comment does in my opinion not block this diff.

nicarl:
I do not see the same problem with exit code 1 with yarn version 1.22.15.

jonas-tai (approved):
Nice👍

PR #153, opened 2021-11-03
Disable multiline for some fields

mayayunx:

### Summary of changes

- Add new property multiline for Textbox
- Disable multiline for name, version, purl and url
- Add new test helper functions for confidence fields

Name, version, purl and url should only be one line

Run `yarn test` and test in the UI

### DIFF
--- a/src/Frontend/Components/InputElements/TextBox.tsx
+++ b/src/Frontend/Components/InputElements/TextBox.tsx
@@ -14,6 +14,7 @@ interface TextProps extends InputElementProps {
   minRows?: number;
   maxRows?: number;
   endIcon?: ReactElement;
+  multiline?: boolean;
 }
 
 export function TextBox(props: TextProps): ReactElement {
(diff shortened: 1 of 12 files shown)


src/Frontend/Components/InputElements/TextBox.tsx
```diff
@@ -14,6 +14,7 @@ interface TextProps extends InputElementProps {
   minRows?: number;
   maxRows?: number;
   endIcon?: ReactElement;
+  multiline?: boolean;
```
benedikt-richter, line 17: this property is set everytime the component is called. However, it would be better to have a default -> multiline
Then it must only be set when we want to have the component as singleLine

src/Frontend/test-helpers/test-helpers.ts
```diff
@@ -395,14 +395,32 @@ export function insertValueIntoTextBox(
...
+
 export function expectValueInTextBox(
   screen: Screen,
   textBoxLabel: string,
   value: string
 ): void {
   const textBox = screen.getByLabelText(textBoxLabel);
   // eslint-disable-next-line testing-library/prefer-screen-queries
-  getByText(textBox, value);
+  expect(textBox).toHaveValue(value);
```
nicarl, line 423: Nice :+1: 
Short background: `getByText` does not work for `<form>`. MUI converts the text to a `<textarea>` element in case of multiline. `toHaveValue` works for `<form>` and `<textarea>`

nicarl (approved):

benedikt-richter (approved):

PR #160, opened 2021-11-04
enforce object shorthand notation

benedikt-richter:
### Summary of changes

Introduce new lint rule to enforce use of object shorthand notation

We had this rule implicitly. However, automation is better than manual curation.


sebathomas (approved):
:+1:

nicarl (approved):

PR #165, opened 2021-11-05
Introduce playwright

nicarl:
### Summary of changes

Replaces spectron with playwright as framework for e2e tests.
Unfortunately, spectron will be deprecated in Feb 2022. Playwright seems to be a well-maintained alternative for e2e supporting electron.  

- replace spectron with playwright
- update github action (using xvfb-run action for linux does not work, calling Xvfb manually does though)
- activate e2e test for windows :tada: 
- remove check for e2e env, which avoids installation of devtools in e2e (was necessary for spectron, tests were flaky and timing out randomly before)

`yarn test:e2e` + CI feedback


benedikt-richter (approved):

src/e2e-tests/test-helpers/test-helpers.ts
```diff
@@ -3,21 +3,20 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
-import { Application } from 'spectron';
-import path from 'path';
+import { _electron, ElectronApplication } from 'playwright';
 
-export const INTEGRATION_TEST_TIMEOUT = 60000;
+export const INTEGRATION_TEST_TIMEOUT = 30000;
```
leslielazzarino, line 8: Runs into timeout when there is a lot to rebuild. Keep it at 60s.

leslielazzarino (changes requested):

src/e2e-tests/test-helpers/test-helpers.ts
leslielazzarino: And why is it called "INTEGRATION" when it is in the e2e folder?

nicarl: The timeout should have nothing to do with the building step.We build the app and wait until it is reachable before we start the test. 
The timeout is the actual timeout of the test. Did you see any timeouts in your local test runs?

leslielazzarino: Ok, probably something else was happening on my laptop.

leslielazzarino (approved):

PR #168, opened 2021-11-08
Suppress lint errors in dev mode

benedikt-richter:
### Summary of changes

* yarn start now does work even if there are lint errors
* there is yarn build:dev and yarn buid:prod not respecting and respecting linting, respectively

So far the app died in dev mode, when run via "yarn start" if there was a lint error (e.g. unused variable). This was annoying. Now there is a second build script "yarn build:dev" that has linting disabled but still logs warnings.


nicarl (approved):

nicarl:
Windows ci seems to still fail :-( @leslielazzarino can  you try to debug/reproduce locally?
Previously we had a timeout because we were installing devtools every time we open the app. We could try to give this a shot.

benedikt-richter:
> Windows ci seems to still fail :-( @leslielazzarino can you try to debug/reproduce locally? Previously we had a timeout because we were installing devtools every time we open the app. We could try to give this a shot.

Maybe, https://github.com/opossum-tool/OpossumUI/pull/168#issuecomment-963259829 already reduces the amount of failing ones...

benedikt-richter:
The e2e test on windows is fixed now. A `cross-env` was missing ...

leslielazzarino (approved):

PR #172, opened 2021-11-09
Use latest electron version

benedikt-richter:
### Summary of changes

The electron version is modified to use the latest version.

We used an outdated version.


benedikt-richter:
There is still a bug on Linux when opening files.

PR #173, opened 2021-11-09
Add hide button to context menu

jonas-tai:
### Summary of changes

Adds the implementation for using the hide / unhide function to the context menu for external attributions.

The hide functionality was missing in the context menu for external attributions in the previous version of the context menu. 

Run the integration tests in package-card-context-menu.test.tsx and / or try it out in the app with example file.

### DIFF
--- a/src/Frontend/Components/AttributionColumn/AttributionColumn.tsx
+++ b/src/Frontend/Components/AttributionColumn/AttributionColumn.tsx
@@ -100,6 +100,9 @@ export function AttributionColumn(props: AttributionColumnProps): ReactElement {
   );
   const view = useAppSelector(getSelectedView);
 
+  const selectedPackageId = selectedPackage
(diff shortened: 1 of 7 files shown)


benedikt-richter:
Works nicely!

We could think of icons for hide and unhide :)
E.g. 
import VisibilityOffIcon from '@mui/icons-material/VisibilityOff';
import VisibilityIcon from '@mui/icons-material/Visibility';

leslielazzarino (changes requested):
Attributions in folder content created from signals get hidden too: that should not happen.

leslielazzarino (approved):

jonas-tai:
> 
> 
> Attributions in folder content created from signals get hidden too: that should not happen.

This seems to be the normal hide logic at the moment, no conjunction with this branch.

PR #174, opened 2021-11-09
Change state for filtering

mayayunx:
### Summary of changes
- create 3 filters as enum: only 1st party , hide 1st party, only follow-up 
- save active filters in state
- update the follow-up action and selector
- add logic enforces only 1st party or hide 1st party filter can be active at once
- fix follow up filtering logic, to use the new state

- new filters for show only first party and hide first party are required

run `yarn test:unit`

### DIFF
--- a/src/Frontend/Components/AttributionView/AttributionView.tsx
+++ b/src/Frontend/Components/AttributionView/AttributionView.tsx
@@ -7,7 +7,7 @@ import { makeStyles } from '@material-ui/core/styles';
 import React, { ReactElement } from 'react';
 import { useAppDispatch, useAppSelector } from '../../state/hooks';
 import { Attributions } from '../../../shared/shared-types';
-import { PackagePanelTitle } from '../../enums/enums';
+import { FilterType, PackagePanelTitle } from '../../enums/enums';
 import { changeSelectedAttributionIdOrOpenUnsavedPopup } from '../../state/actions/popup-actions/popup-actions';
 import { getManualAttributions } from '../../state/selectors/all-views-resource-selectors';
 import {
@@ -21,7 +21,7 @@ import { AttributionList } from '../AttributionList/AttributionList';
 import { Checkbox } from '../Checkbox/Checkbox';
 import { OpossumColors } from '../../shared-styles';
 import { topBarHeight } from '../TopBar/TopBar';
-import { areOnlyFollowUpAttributionsShown } from '../../state/selectors/view-selector';
+import { getActiveFilters } from '../../state/selectors/view-selector';
 
 const countAndSearchOffset = 119;
 
@@ -52,7 +52,8 @@ export function AttributionView(): ReactElement {
   const attributionIdMarkedForReplacement: string = useAppSelector(
     getAttributionIdMarkedForReplacement
   );
-  const filterForFollowUp = useAppSelector(areOnlyFollowUpAttributionsShown);
+  const activeFilters = useAppSelector(getActiveFilters);
+  const filterForFollowUp = activeFilters.has(FilterType.OnlyFollowUp);
 
   const { handleFilterChange, getFilteredAttributions } = provideFollowUpFilter(
     filterForFollowUp,
--- a/src/Frontend/enums/enums.ts
+++ b/src/Frontend/enums/enums.ts
@@ -58,3 +58,9 @@ export enum ButtonText {
   UnmarkForReplacement = 'Unmark for replacement',
   ShowResources = 'Show resources',
 }
+
+export enum FilterType {
+  OnlyFirstParty = 'Only First Party',
+  HideFirstParty = 'Hide First Party',
+  OnlyFollowUp = 'Only Follow Up',
+}
--- a/src/Frontend/state/actions/view-actions/__tests__/view-actions.test.ts
+++ b/src/Frontend/state/actions/view-actions/__tests__/view-actions.test.ts
@@ -3,9 +3,10 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
-import { PopupType, View } from '../../../../enums/enums';
+import { FilterType, PopupType, View } from '../../../../enums/enums';
 import { createTestAppStore } from '../../../../test-helpers/render-component-with-store';
 import {
+  getActiveFilters,
   getOpenPopup,
   getSelectedView,
   getTargetAttributionId,
@@ -20,6 +21,7 @@ import {
   openPopup,
   openPopupWithTargetAttributionId,
   resetViewState,
+  updateActiveFilters,
   setTargetView,
 } from '../view-actions';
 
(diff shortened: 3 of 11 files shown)


benedikt-richter:
The commits should be squashed before merging

src/Frontend/Components/AttributionView/AttributionView.tsx
```diff
@@ -52,7 +52,8 @@ export function AttributionView(): ReactElement {
   const attributionIdMarkedForReplacement: string = useAppSelector(
     getAttributionIdMarkedForReplacement
   );
-  const filterForFollowUp = useAppSelector(areOnlyFollowUpAttributionsShown);
+  const activeFilters = useAppSelector(getActiveFilters);
+  const filterForFollowUp = activeFilters.has(FilterEnum.OnlyFollowUp);
```
benedikt-richter, line 56: The name `FilterEnum` should not contain `Enum`

src/Frontend/enums/enums.ts
```diff
@@ -58,3 +58,9 @@ export enum ButtonText {
   UnmarkForReplacement = 'Unmark for replacement',
   ShowResources = 'Show resources',
 }
+
+export enum FilterEnum {
```
benedikt-richter, line 62: couzld be `FilterType`

src/Frontend/state/actions/view-actions/__tests__/view-actions.test.ts
```diff
@@ -98,16 +100,44 @@ describe('view actions', () => {
...
+    testStore.dispatch(setFilter(FilterEnum.OnlyFirstParty));
+    expect(
+      getActiveFilters(testStore.getState()).has(FilterEnum.OnlyFirstParty)
+    ).toBe(false);
+
+    testStore.dispatch(setFilter(FilterEnum.OnlyFirstParty));
+    testStore.dispatch(setFilter(FilterEnum.HideFirstParty));
+    expect(
+      getActiveFilters(testStore.getState()).has(FilterEnum.OnlyFirstParty)
+    ).toBe(false);
```
benedikt-richter, line 137: we could check the case of having two filters set at the same time

src/Frontend/state/reducers/view-reducer.ts
```diff
@@ -58,10 +59,10 @@ export function viewState(
         ...state,
         openPopup: action.payload,
       };
-    case ACTION_SET_FOLLOW_UP_FILTER:
+    case ACTION_SET_FILTER:
```
benedikt-richter, line 62: actually it is more a `ACTION_TOGGLE_FILTER`, as the action might lead to a removed filter. In this case the name would be missleading

src/Frontend/state/helpers/set-filters.ts
```diff
@@ -0,0 +1,47 @@
...
+  newFilter: FilterEnum
+): Set<FilterEnum> {
+  const currentFilters = new Set(activeFilters);
+  if (newFilter == FilterEnum.HideFirstParty) {
+    currentFilters.delete(FilterEnum.OnlyFirstParty);
+  } else if (newFilter == FilterEnum.OnlyFirstParty) {
+    currentFilters.delete(FilterEnum.HideFirstParty);
+  }
+  return currentFilters;
+}
```
benedikt-richter, line 47: It would be nice to have a configuration like:
{mutuallyExclusiveFilters: [filterA, filterB]} or similar.

In this case adding a new filter would be trivial.

benedikt-richter (changes requested):
see comments

src/Frontend/state/actions/view-actions/types.ts
```diff
@@ -51,9 +51,9 @@ export interface OpenPopupAction {
   payload: OpenPopupActionPopupType;
 }
 
-export interface SetFollowUpFilter {
-  type: typeof ACTION_SET_FOLLOW_UP_FILTER;
-  payload: boolean;
+export interface SetFilter {
```
benedikt-richter, line 54: The Interface should also reflect that it is not a simple set (also the name of the action and all other occurences).

I just realized that it's also not a simple toggle...

Maybe it could be: `ACTION_ADD_ACTIVE_FILTER`, `AddActiveFilter`, ...

src/Frontend/state/helpers/set-filters.ts
```diff
@@ -0,0 +1,71 @@
+// SPDX-FileCopyrightText: Facebook, Inc. and its affiliates
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+
+import { FilterType } from '../../enums/enums';
+const mutuallyExclusiveFilters = new Set([
+  new Set<FilterType>([FilterType.OnlyFirstParty, FilterType.HideFirstParty]),
+]);
```
benedikt-richter, line 9: This implementation seems to be more complex than necessary. Could be

```

import { FilterType } from '../../enums/enums';

const mutuallyExclusiveFilters = [
  [FilterType.OnlyFirstParty, FilterType.HideFirstParty],
];

export function getUpdatedFilters(
  activeFilters: Set<FilterType>,
  newFilter: FilterType
): Set<FilterType> {
  const currentFilters = new Set(activeFilters);

  if (currentFilters.has(newFilter)) {
    currentFilters.delete(newFilter);
  } else {
    const filtersToRemove = getFiltersToRemove(currentFilters, newFilter);
    if (filtersToRemove.size !== 0) {
      filtersToRemove.forEach((element) => currentFilters.delete(element));
    }
    currentFilters.add(newFilter);
  }

  return currentFilters;
}

export function getFiltersToRemove(
  activeFilters: Set<FilterType>,
  newFilter: FilterType
): Set<FilterType> {
  return new Set(
    mutuallyExclusiveFilters
      .filter((filterPair) => filterPair.includes(newFilter))
      .map((filterPair) =>
        filterPair.indexOf(newFilter) ? filterPair[0] : filterPair[1]
      )
  );
}


```

benedikt-richter (changes requested):

benedikt-richter (approved):
nice 👍 

just squash commits before merging

src/Frontend/state/actions/view-actions/types.ts
```diff
@@ -3,25 +3,25 @@
...
 
 export const ACTION_SET_TARGET_VIEW = 'ACTION_SET_TARGET_VIEW';
 export const ACTION_SET_VIEW = 'ACTION_SET_VIEW';
 export const ACTION_OPEN_POPUP = 'ACTION_OPEN_POPUP';
 export const ACTION_CLOSE_POPUP = 'ACTION_CLOSE_POPUP';
 export const ACTION_RESET_VIEW_STATE = 'ACTION_RESET_VIEW_STATE';
-export const ACTION_SET_FOLLOW_UP_FILTER = 'ACTION_SET_FOLLOW_UP_FILTER';
 export const ACTION_OPEN_POPUP_WITH_TARGET_ATTRIBUTION_ID =
   'ACTION_OPEN_POPUP_WITH_TARGET_ATTRIBUTION_ID';
+export const ACTION_ADD_ACTIVE_FILTER = 'ACTION_ADD_ACTIVE_FILTER';
```
antonbauhofer, line 15: I think the action and the types should be called "...SET..." instead of "...ADD..." since they not only add, but also remove the respective filter.

antonbauhofer (approved):
Just one comment regarding renaming. The rest looks good.

src/Frontend/state/actions/view-actions/types.ts
benedikt-richter: I understand your point. `Add` is not perfect. However, if I'd use an action called `set...`, I'd expect the value to be the value I've set. This is, however, not the case, as it could also remove other values...

antonbauhofer: Maybe `add_or_delete` would fit

benedikt-richter: Maybe more generic `update`

PR #185, opened 2021-11-11
improve test tooling

benedikt-richter:
### Summary of changes

The integration tests are split into one part that is only run in CI. For this purpose there is now `yarn test:local`.

The goal is to decrease the total time tests running locally by not always running all tests. For local development there is now `yarn test:local` that runs all unit tests and only a subset of the integration tests.

`yarn test:local`
`yarn test:integration-ci`


CONTRIBUTING.md
```diff
@@ -75,9 +75,13 @@ yarn start
 Unit tests are provided for all features, aided by integration tests when helpful. The testing framework is jest + react
 testing library. They can be run locally by executing:
 
+```
+yarn test:local
```
JakobSchubert, line 79: This will run the unit tests + the local integration tests, right? Is there a way to only run the local integration tests?

JakobSchubert (approved):

CONTRIBUTING.md
benedikt-richter: There is no way (in the form of `yarn ...`) to run the local integration tests. There is imo no usecase for that. The idea was to have a single command `yarn test:local` that provides a good confidence whether tests are broken. If you want to be sure `test:integration-ci` has to be used.

antonbauhofer (approved):

PR #188, opened 2021-11-12
keep comments when promoting signal

benedikt-richter:
### Summary of changes

keep comments when promoting signal

In the past we removed the comment of a signal when it was promoted to an attribution. This behaviour is modified to keep the comment in this case.


src/Frontend/Components/PackageCard/PackageCard.tsx
```diff
@@ -105,7 +104,7 @@ export function PackageCard(props: PackageCardProps): ReactElement | null {
     <IconButton
       tooltipTitle="add"
       placement="left"
-      onClick={props.onIconClick ? props.onIconClick : doNothing}
+      onClick={props.onIconClick}
```
benedikt-richter, line 107: unrelated change

benedikt-richter:
Waits for decision regarding versioning of the input schema.

antonbauhofer (approved):

leslielazzarino (approved):

PR #189, opened 2021-11-12
improve table styling

benedikt-richter:
### Summary of changes

Do not break on white spaces in table cells.

Do not break on white spaces in table cells.


antonbauhofer (changes requested):
Why do we want this?
I personally don't like the horizontal scrolling in the table cells and would prefer to leave it as is.

nicarl (commented):
Let's discuss together, I am not sure if it looks better

antonbauhofer (approved):
Looks good. I looked at it in the app and it works.

nicarl (approved):

PR #201, opened 2021-11-17
Add replace attribution to context menu and audit view

jonas-tai:
### Summary of changes

This commit adds the replace attribution function to the context menu and to the hamburger menu in audit view.

Replacing attributions was only possible in attribution view so far and not available in the context menu. This adds these functionalities and allows the user to replace attributions with the context of the file tree and to switch between views while replacing attributions.

Run the context menu integration tests (part of the frontend ci integration tests) and / or test replace attributions with example file with context menu and in audit view.

### DIFF
--- a/src/Frontend/Components/AttributionColumn/AttributionColumn.tsx
+++ b/src/Frontend/Components/AttributionColumn/AttributionColumn.tsx
@@ -123,15 +123,14 @@ export function AttributionColumn(props: AttributionColumnProps): ReactElement {
       ? selectedPackageId
       : '';
 
-  const mergeButtonDisplayState = getMergeButtonsDisplayState(
-    view,
+  const mergeButtonDisplayState = getMergeButtonsDisplayState({
     attributionIdMarkedForReplacement,
-    currentViewSelectedAttributionId,
-    currentViewSelectedAttributionId,
+    targetAttributionId: currentViewSelectedAttributionId,
+    selectedAttributionId: currentViewSelectedAttributionId,
     packageInfoWereModified,
-    Boolean(temporaryPackageInfo.preSelected),
-    false
-  );
+    targetAttributionIsPreSelected: Boolean(temporaryPackageInfo.preSelected),
+    targetAttributionIsExternalAttribution: false,
+  });
 
   const mainButtonConfigs: Array<MainButtonConfig> = [
     {
@@ -199,7 +198,7 @@ export function AttributionColumn(props: AttributionColumnProps): ReactElement {
           )
         );
       },
-      hidden: mergeButtonDisplayState.hideOnReplaceMarkedByButton,
+      hidden: mergeButtonDisplayState.hideReplaceMarkedByButton,
     },
   ];
 
--- a/src/Frontend/Components/AttributionColumn/AttributionColumn.tsx
+++ b/src/Frontend/Components/AttributionColumn/AttributionColumn.tsx
@@ -10,6 +10,7 @@ import React, { ChangeEvent, ReactElement } from 'react';
 import { PackageInfo } from '../../../shared/shared-types';
 import { setTemporaryPackageInfo } from '../../state/actions/resource-actions/all-views-simple-actions';
 import {
+  getAttributionIdMarkedForReplacement,
   getIsSavingDisabled,
   getPackageInfoOfSelected,
   getTemporaryPackageInfo,
@@ -41,12 +42,9 @@ import { LicenseSubPanel } from './LicenseSubPanel';
 import { AuditingSubPanel } from './AuditingSubPanel';
 import { ButtonRow } from './ButtonRow';
 import { setAttributionIdMarkedForReplacement } from '../../state/actions/resource-actions/attribution-view-simple-actions';
-import {
-  getAttributionIdMarkedForReplacement,
-  getSelectedAttributionId,
-} from '../../state/selectors/attribution-view-resource-selectors';
-import { openPopup } from '../../state/actions/view-actions/view-actions';
-import { ButtonText, PopupType } from '../../enums/enums';
+import { getSelectedAttributionId } from '../../state/selectors/attribution-view-resource-selectors';
+import { openPopupWithTargetAttributionId } from '../../state/actions/view-actions/view-actions';
+import { ButtonText, PopupType, View } from '../../enums/enums';
 import { MainButtonConfig } from '../ButtonGroup/ButtonGroup';
 import { ContextMenuItem } from '../ContextMenu/ContextMenu';
 import { useAppDispatch, useAppSelector } from '../../state/hooks';
@@ -118,13 +116,21 @@ export function AttributionColumn(props: AttributionColumnProps): ReactElement {
     selectedPackage
   );
   const nameAndVersionAreEditable = props.isEditable && temporaryPurl === '';
+  const currentViewSelectedAttributionId =
+    view === View.Attribution
+      ? selectedAttributionId
+      : view === View.Audit
+      ? selectedPackageId
+      : '';
 
   const mergeButtonDisplayState = getMergeButtonsDisplayState(
     view,
     attributionIdMarkedForReplacement,
-    selectedAttributionId,
+    currentViewSelectedAttributionId,
+    currentViewSelectedAttributionId,
     packageInfoWereModified,
-    Boolean(temporaryPackageInfo.preSelected)
+    Boolean(temporaryPackageInfo.preSelected),
+    false
   );
 
   const mainButtonConfigs: Array<MainButtonConfig> = [
@@ -169,7 +175,9 @@ export function AttributionColumn(props: AttributionColumnProps): ReactElement {
     {
       buttonText: ButtonText.MarkForReplacement,
       onClick: (): void => {
-        dispatch(setAttributionIdMarkedForReplacement(selectedAttributionId));
+        dispatch(
+          setAttributionIdMarkedForReplacement(currentViewSelectedAttributionId)
+        );
       },
       hidden: mergeButtonDisplayState.hideMarkForReplacementButton,
     },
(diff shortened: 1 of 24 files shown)


src/Frontend/Components/AttributionColumn/AttributionColumn.tsx
```diff
@@ -118,13 +116,21 @@ export function AttributionColumn(props: AttributionColumnProps): ReactElement {
...
   const mergeButtonDisplayState = getMergeButtonsDisplayState(
     view,
     attributionIdMarkedForReplacement,
-    selectedAttributionId,
+    currentViewSelectedAttributionId,
+    currentViewSelectedAttributionId,
     packageInfoWereModified,
-    Boolean(temporaryPackageInfo.preSelected)
+    Boolean(temporaryPackageInfo.preSelected),
+    false
```
leslielazzarino, line 133: This function has a huge number of parameters, and I do not see a meaningful way of breaking the function into parts. To make the code more readable, I would suggest the following js/ts pattern: the function takes a single object as input, where the keys are the names of the variables. Therefore when you are using the function you can see which input is what.

leslielazzarino (changes requested):

antonbauhofer:
When selecting replace attribution from the context menu, the popup says `...to the current attribution`. This might be misleading, since the `current attribution` might not be the one selected (see screenshot). So, I suggest to remove `current` from the popup tex
Screenshot 2021-11-18 at 09 39 00
t.

src/Frontend/Components/ResourceDetailsTabs/ResourceDetailsTabs.tsx
```diff
@@ -73,12 +73,17 @@ export function ResourceDetailsTabs(
...
-    // manualData is excluded from dependencies on purpose to avoid recalculation when
-    // it changes. Usually this is not an issue as the displayed data remains correct.
-    // In consequence, the panelData is eventually consistent.
+    /*
+      manualData is excluded from dependencies on purpose to avoid recalculation when
+      it changes. Usually this is not an issue as the displayed data remains correct.
+      In consequence, the panelData is eventually consistent.
+      Only manualData.attributionsToResources in dependencies to update panelData, when replaceAttributionPopup
+      was called in AuditView for manual attributions in attributions in folder content panel, but not update them
+      after every save click.
```
antonbauhofer, line 82: This sentence is very long and hard to read. Can you break it down into simpler sentences?

src/Frontend/Components/AttributionColumn/attribution-column-helpers.ts
```diff
@@ -160,29 +160,39 @@ interface MergeButtonDisplayState {
...
   attributionIdMarkedForReplacement: string,
+  targetAttributionId: string,
   selectedAttributionId: string,
   packageInfoWereModified: boolean,
-  attributionIsPreSelected: boolean
+  targetAttributionIsPreSelected: boolean,
+  isExternalAttribution: boolean
 ): MergeButtonDisplayState {
-  const isAttributionView = currentView === View.Attribution;
+  const isTableView = currentView === View.Report;
```
antonbauhofer, line 169: Let's call it `isReportView` to keep it consistent

antonbauhofer (commented):
Looks good.
Just a few minor comments.

src/Frontend/Components/AttributionColumn/attribution-column-helpers.ts
leslielazzarino, line 169: Is it at all possible to have the context menu on ReportView? I do not think so, as we have no cards there.

src/Frontend/Components/PackageCard/PackageCard.tsx
```diff
@@ -259,6 +283,33 @@ export function PackageCard(props: PackageCardProps): ReactElement | null {
...
+          disabled: mergeButtonDisplayState.deactivateReplaceMarkedByButton,
+          onClick: (): void => {
+            dispatch(
+              openPopupWithTargetAttributionId(
+                PopupType.ReplaceAttributionPopup,
+                attributionId
+              )
+            );
+          },
+          hidden: mergeButtonDisplayState.hideOnReplaceMarkedByButton,
```
leslielazzarino, line 311: Why "On" in the eneum name? I would remove it.

src/Frontend/Components/AttributionColumn/attribution-column-helpers.ts
```diff
@@ -160,29 +160,39 @@ interface MergeButtonDisplayState {
...
+      isTableView ||
+      targetAttributionIsMarkedForReplacement,
     hideUnmarkForReplacementButton:
-      !isAttributionView ||
+      isExternalAttribution ||
+      isTableView ||
       !anyAttributionMarkedForReplacement ||
-      !selectedAttributionIsMarkedForReplacement,
+      !targetAttributionIsMarkedForReplacement,
     hideOnReplaceMarkedByButton:
```
leslielazzarino, line 188: Why "On" in the name? I would remove it.

src/Frontend/integration-tests/all-views-tests/__tests-ci__/context-menu-all-views.test.tsx
```diff
@@ -431,4 +434,71 @@ describe('The ContextMenu', () => {
     clickOnPathInPopupWithResources(screen, '/secondResource.js');
     expectValueInTextBox(screen, 'Name', 'Vue');
   });
+
+  test('replace attributions buttons are synced with hamburger menu buttons', () => {
```
leslielazzarino, line 438: I do not think we need such an integration test. It is enough to test with unit tests that all buttons read/write correctly in the state. We do not need to test that Redux works with React.

src/Frontend/integration-tests/attribution-view-tests/__tests-ci__/context-menu-attribution-view.test.tsx
```diff
@@ -139,4 +146,129 @@ describe('The ContextMenu', () => {
     expectValueNotInConfidenceField(screen, '10');
     expectValueInConfidenceField(screen, `High (${DiscreteConfidence.High})`);
   });
+
+  describe('replace attribution buttons', () => {
```
leslielazzarino, line 150: "replaces", the subject is the context menu. Same in the following row.

src/Frontend/integration-tests/audit-view-tests/__tests-ci__/context-menu-audit-view.test.tsx
```diff
@@ -303,4 +294,195 @@ describe('The ContextMenu', () => {
     );
     expectAddIconInAddToAttributionCardIsNotHidden(screen, 'Vue, 1.2.0');
   });
+
+  describe('replace attribution buttons', () => {
```
leslielazzarino, line 298: "replaces"

src/Frontend/integration-tests/audit-view-tests/__tests-ci__/context-menu-audit-view.test.tsx
```diff
@@ -303,4 +294,195 @@ describe('The ContextMenu', () => {
...
+        comment: 'ManualPackage',
+        preSelected: true,
+      },
+    };
+    const testResourcesToManualAttributions: ResourcesToAttributions = {
+      '/root/src/file_1': ['uuid_1'],
+      '/root/src/file_2': ['uuid_2', 'uuid_3'],
+    };
+
+    test('replace attributions in Package panel', () => {
```
leslielazzarino, line 346: "replaces"

leslielazzarino (changes requested):
Too many integration tests. One should not test every bit of the feature through integration tests, as they are long to write/read and slow. My suggestion would be:
- a single integration test, that tests a complex enough case, e.g. that you can mark for replacement an attribution on attribution view and replace it with another on audit view
- everything else you can test through unit tests, where you just test that each component interacts correctly with the state, writing what it should or being updated correctly when something changes.

src/Frontend/Components/AttributionColumn/AttributionColumn.tsx
jonas-tai: I changed the paramters of the function to a single object👍

src/Frontend/Components/AttributionColumn/attribution-column-helpers.ts
jonas-tai: Removed the parameter, because it is currently not necessary.

jonas-tai: Removed the parameter, because it is currently not necessary👍

src/Frontend/integration-tests/attribution-view-tests/__tests-ci__/context-menu-attribution-view.test.tsx
jonas-tai: I think the whole naming was confusing to read, so I simplified the naming of these tests.

src/Frontend/integration-tests/all-views-tests/__tests-ci__/context-menu-all-views.test.tsx
jonas-tai: Removed this test as part of the test refactoring.

jonas-tai:
> Too many integration tests. One should not test every bit of the feature through integration tests, as they are long to write/read and slow. My suggestion would be:
> 
>     * a single integration test, that tests a complex enough case, e.g. that you can mark for replacement an attribution on attribution view and replace it with another on audit view
> 
>     * everything else you can test through unit tests, where you just test that each component interacts correctly with the state, writing what it should or being updated correctly when something changes.

Testing the replace attribution feature requires opening the corresponding replace attribution pop-up. The pop-up is opened at the top-level of the app, so it is not possible to test the feature in unit tests for each component. To make the integration tests easier to read, I removed the correct context menu configuration check from the integration tests. For that, I added separate unit tests for each component that verify the correct context menu configuration for the replace attribution buttons.

antonbauhofer (approved):

leslielazzarino (approved):

PR #215, opened 2021-11-24
Fix hidden signals in Signals in Folder Content tab

jonas-tai:
### Summary of changes

Add filter for resolved attributions to 'Signals in Folder Content' tab.

Hiding a signal should remove it from the 'Signals in Folder Content' tab and make it only visible through the 'Signals' tab of the resources, the signal is attached to. So far, the signal stayed greyed out in the 'Signals in Folder Content' tab, if at least one resource, to which the signal was attached to, had one or more other non-hidden signals. This commit fixes this behavior by adding a filter for resolved attributions to the 'Signals in Folder Content' tab.

Run tests and / or use example file to hide a signal that is attached to a resource with at least one other non-hidden signal. Verify that the signal is not visible in the 'Signals in Folder Content' tab after hiding it

### DIFF
--- a/src/Frontend/Components/ResourceDetailsTabs/ResourceDetailsTabs.tsx
+++ b/src/Frontend/Components/ResourceDetailsTabs/ResourceDetailsTabs.tsx
@@ -19,6 +19,7 @@ import { getPanelData, PanelData } from './resource-details-tabs-helpers';
 import {
   getAttributionIdsOfSelectedResource,
   getDisplayedPackage,
+  getResolvedExternalAttributions,
   getSelectedResourceId,
 } from '../../state/selectors/audit-view-resource-selectors';
 import { OpossumColors } from '../../shared-styles';
@@ -66,6 +67,9 @@ export function ResourceDetailsTabs(
     getAttributionIdsOfSelectedResource,
     isEqual
   );
+  const resolvedExternalAttributions: Set<string> = useAppSelector(
+    getResolvedExternalAttributions
+  );
 
   enum Tabs {
     SignalsAndContent = 0,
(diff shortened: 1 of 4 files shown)


src/Frontend/Components/ResourceDetailsTabs/resource-details-tabs-helpers.ts
```diff
@@ -103,13 +107,16 @@ function getContainedManualPackages(
 export function computeAggregatedAttributionsFromChildren(
   attributions: Attributions,
   resourcesToAttributions: ResourcesToAttributions,
-  attributedChildren: Set<string>
+  attributedChildren: Set<string>,
+  resolvedExternalAttributions = new Set<string>()
```
nicarl, line 111: I prefer to make `resolvedExternalAttributions` nullable with default null instead of creating a new set every time the function gets called.

nicarl (changes requested):

benedikt-richter (approved):
works smoothly 👍

nicarl:
One test is still failing, let's have a look together.

src/Frontend/Components/ResourceDetailsTabs/ResourceDetailsTabs.tsx
```diff
@@ -88,7 +98,7 @@ export function ResourceDetailsTabs(
     */
 
     // eslint-disable-next-line react-hooks/exhaustive-deps
-    [selectedResourceId, externalData, manualData.attributionsToResources]
+    [selectedResourceId, externalData, resolvedExternalAttributions]
```
nicarl, line 101: @benedikt-richter

PR #243, opened 2021-12-03
POC, DO NOT MERGE: Web workers with comlink

leslielazzarino:
### Summary of changes

Introduce web workers and compute accordion data in a worker (async, not in the main process). That sadly requires to eject from react-scripts, as they do not support web workers in the webpack config.

This PR is not supposed to be merged: if found worth it, the ejection will be spit in a separate PR, and the web worker code will be properly refactored.

The app gets stuck for a while when clicking on a folder with a lot of signals inside. Web workers allow to calculate the accordion faster, meanwhile leaving the main process free to process user input.

Open keycloak input and navigate away from the root folder (somewhere where there is an attribution). Then click on the root folder. On the right and above the accordion the UI will change instantly, meanwhile the old accordion will be shown until the new one is ready.


package.json
```diff
@@ -68,28 +127,29 @@
     "nunjucks": "^3.2.3",
     "playwright": "^1.17.1",
     "prettier": "^2.5.0",
-    "react-scripts": "4.0.3",
+    "react-app-rewired": "^2.1.8",
```
benedikt-richter, line 130: ?

package.json
```diff
@@ -153,6 +213,53 @@
...
+    "setupFiles": [
+      "react-app-polyfill/jsdom"
+    ],
+    "setupFilesAfterEnv": [],
+    "testMatch": [
+      "<rootDir>/src/**/__tests__/**/*.{js,jsx,ts,tsx}",
+      "<rootDir>/src/**/*.{spec,test}.{js,jsx,ts,tsx}"
+    ],
+    "testEnvironment": "jsdom",
+    "testRunner": "C:\\Users\\lesli\\IdeaProjects\\OpossumUI\\node_modules\\jest-circus\\runner.js",
```
benedikt-richter, line 232: ?

config-overrides.js
```diff
@@ -0,0 +1,10 @@
+// This tutorial uses webpack directly https://www.youtube.com/watch?v=1QFTfbDXJCI
+// This configuration file is probably inaccurate
+
+module.exports = function override(config, env) {
```
leslielazzarino, line 4: Remove this file

PR #251, opened 2021-12-08
Fix display of invalid package urls

antonbauhofer:
### Summary of changes

Invalid purls are supposed to be displayed in red.

Before this fix, the text color was black.

Enter an invalid text in the PURL field in the attribution column of OpossumUI


benedikt-richter (approved):
👍

leslielazzarino (approved):

PR #258, opened 2021-12-09
Add icon to hide multiselect in attribution view

mayayunx:

### Summary of changes
- Add an icon in attribution view to hide or show the multiselect filter

- The multiselect component is too prominent and is not used so frequently

- The icon can be tested in the UI

### DIFF
--- a/src/Frontend/Components/AttributionView/AttributionView.tsx
+++ b/src/Frontend/Components/AttributionView/AttributionView.tsx
@@ -4,7 +4,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import makeStyles from '@mui/styles/makeStyles';
-import React, { ReactElement } from 'react';
+import React, { ReactElement, useState } from 'react';
 import { useAppDispatch, useAppSelector } from '../../state/hooks';
 import { Attributions } from '../../../shared/shared-types';
 import { PackagePanelTitle } from '../../enums/enums';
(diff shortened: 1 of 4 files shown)


src/Frontend/Components/Filter/FilterMultiSelect.tsx
```diff
@@ -70,6 +70,10 @@ export function FilterMultiSelect(props: FilterMultiSelectProps): ReactElement {
   const activeFilters = Array.from(useAppSelector(getActiveFilters));
 
   const updateFilters = (filter: FilterType): void => {
+    const filterFormControl = document.getElementById('filter-form-control');
```
nicarl, line 73: I think the more "react"-like way is to use a boolean state to indicate if the element is shown. I think the state should be declared in `AttributionView`

``` typescript
const [showMultiSelect, setShowMultiselect] = useState<boolean>(false);
``` 
The onClick function of the icon will then just use 
``` typescript
onClick = () => {setShowMultiselect(!showMultiSelect)}

**Review by nicarl — CHANGES_REQUESTED** [review]

File: src/Frontend/Components/Filter/FilterMultiSelect.tsx (hunk 1 above)
Comment (leslielazzarino) [reply src/Frontend/Components/Filter/FilterMultiSelect.tsx:73]: It would also be cool if the multiselect menu would open when the icon is clicked. As the filters multiselect is not shown without selected filters, if I click on it it is to select some.

**Review by leslielazzarino — CHANGES_REQUESTED** [review]

**Review by leslielazzarino — APPROVED** [review]

File: src/Frontend/Components/Filter/FilterMultiSelect.tsx [hunk 2]
```diff
@@ -91,6 +91,7 @@ export function FilterMultiSelect(props: FilterMultiSelectProps): ReactElement {
 
   return (
     <MuiFormControl
+      id="filter-form-control"
```
nicarl, line 94: Do we need this id ?

src/Frontend/Components/AttributionView/AttributionView.tsx
```diff
@@ -56,6 +68,14 @@ export function AttributionView(): ReactElement {
     Object.keys(attributions).length
   })`;
 
+  const activeFilters = Array.from(useAppSelector(getActiveFilters));
+
+  const [showMultiSelect, setShowMultiselect] = useState<boolean>(false);
+
+  if (activeFilters.length != 0 && !showMultiSelect) {
+    setShowMultiselect(!showMultiSelect);
+  }
```
nicarl, line 77: I think this can be deleted

nicarl: It is actually needed as `showMultiSelect` is not in the store and we might have a conflict when changing views

nicarl (approved):

PR #279, opened 2021-12-16
Add checkbox for multi-select to package cards

jonas-tai:
### Summary of changes

Adds a checkbox to the package card while in multi-select mode for multi-select deletion feature. The checkbox has the same color as the package card. The selection gets reset on view change.

Preparation for adding multi-select deletion.

Run tests or test checkboxes in attribution view.

### DIFF
--- a/src/Frontend/Components/PackageCard/PackageCard.tsx
+++ b/src/Frontend/Components/PackageCard/PackageCard.tsx
@@ -89,7 +89,7 @@ interface PackageCardProps {
   cardConfig: ListCardConfig;
   onClick(): void;
   onIconClick?(): void;
-  hideContextMenu?: boolean;
+  hideContextMenuAndMultiSelect?: boolean;
   hideResourceSpecificButtons?: boolean;
 }
 
@@ -185,7 +185,6 @@ export function PackageCard(props: PackageCardProps): ReactElement | null {
       : selectedAttributionIdAuditView;
 
   const isMultiSelected =
-    multiSelectMode &&
     multiSelectSelectedAttributionIds.includes(attributionId);
 
   function openConfirmDeletionPopup(): void {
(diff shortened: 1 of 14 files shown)


src/Frontend/Components/PackageCard/PackageCard.tsx
```diff
@@ -319,8 +343,35 @@ export function PackageCard(props: PackageCardProps): ReactElement | null {
     setIsContextMenuOpen(!isContextMenuOpen);
   }
 
+  function handleMultiSelectAttributionSelected(
+    event: React.ChangeEvent<HTMLInputElement>
+  ): void {
+    if (event.target.checked) {
+      multiSelectSelectedAttributionIds.push(attributionId);
```
nicarl, line 350: I think this is a bit of an anti-pattern:
What you are doing here is mutating `multiSelectSelectedAttributionIds` and then calling `dispatch` to "persist" this mutation.
What I would expect here is that you have an action which does the mutation and persistence.
Let's wait for Benedikt's feedback

src/Frontend/Components/PackageCard/PackageCard.tsx
```diff
@@ -319,8 +343,35 @@ export function PackageCard(props: PackageCardProps): ReactElement | null {
...
+      dispatch(
+        setMultiSelectSelectedAttributionIds(multiSelectSelectedAttributionIds)
+      );
+    } else {
+      dispatch(
+        setMultiSelectSelectedAttributionIds(
+          multiSelectSelectedAttributionIds.filter((id) => id !== attributionId)
```
nicarl, line 357: Filter is fine because it creates a new copy (basically an anonymous copy of your filtered array)

src/Frontend/state/actions/resource-actions/save-actions.ts
```diff
@@ -221,6 +233,16 @@ export function deleteAttributionAndSave(
   return (dispatch: AppThunkDispatch, getState: () => State): void => {
     const attributionsToResources: AttributionsToResources =
       getManualAttributionsToResources(getState());
+    const multiSelectSelectedAttributionIds =
```
nicarl, line 236: this is duplicated. From above. Maybe one wants to have 1 thunk action for the duplicated code, but I have no strong opinion on this.

src/Frontend/Components/PackageCard/PackageCard.tsx
```diff
@@ -164,6 +184,10 @@ export function PackageCard(props: PackageCardProps): ReactElement | null {
       ? selectedAttributionIdAttributionView
       : selectedAttributionIdAuditView;
 
+  const isMultiSelected =
+    multiSelectMode &&
```
benedikt-richter, line 188: this line should not be needed

src/Frontend/state/actions/resource-actions/save-actions.ts
```diff
@@ -209,7 +211,17 @@ export function addToSelectedResource(
...
+  return (dispatch: AppThunkDispatch, getState: () => State): void => {
+    const multiSelectSelectedAttributionIds =
+      getMultiSelectSelectedAttributionIds(getState());
+    if (multiSelectSelectedAttributionIds.includes(attributionId)) {
+      dispatch(
+        setMultiSelectSelectedAttributionIds(
+          multiSelectSelectedAttributionIds.filter((id) => id !== attributionId)
+        )
+      );
+    }
```
benedikt-richter, line 223: This logic could probably go into `savePackageInfo`. At the end of this function there is already some clean up implemented that aims at keeping the state consistent

src/Frontend/state/actions/resource-actions/save-actions.ts
benedikt-richter: I'd avoid this deduplication. Maybe (as noted above) `savePackageInfo` could be a good place

src/Frontend/Components/PackageCard/PackageCard.tsx
```diff
@@ -319,8 +343,35 @@ export function PackageCard(props: PackageCardProps): ReactElement | null {
...
+        )
+      );
+    }
+  }
+
+  const leftElement = multiSelectMode ? (
+    <Checkbox
+      checked={multiSelectSelectedAttributionIds.includes(attributionId)}
+      onChange={handleMultiSelectAttributionSelected}
+      className={clsx(classes.multiSelectCheckbox)}
```
benedikt-richter, line 367: here `clsx` is not needed

src/Frontend/Components/PackageCard/PackageCard.tsx
```diff
@@ -319,8 +343,35 @@ export function PackageCard(props: PackageCardProps): ReactElement | null {
...
+    />
+  ) : undefined;
+
   return (
-    <>
+    <div
+      className={clsx(multiSelectMode ? classes.multiSelectPackageCard : '')}
```
benedikt-richter, line 373: could be 
`className={clsx(multiSelectMode && classes.multiSelectPackageCard )}`

src/Frontend/Components/PackageCard/PackageCard.tsx
benedikt-richter: I agree that the mutation is not great. In particular, for one case there is a mutation and for the other one there is none. What we want to do is to calculate the new value and then to set it in state. The following solution would avoid the asymmetry between the cases as well as avoid the issue Nico pointed out:

```

  function handleMultiSelectAttributionSelected(
    event: React.ChangeEvent<HTMLInputElement>
  ): void {
    const newMultiSelectSelectedAttributionIds = event.target.checked
      ? multiSelectSelectedAttributionIds.concat([attributionId])
      : multiSelectSelectedAttributionIds.filter((id) => id !== attributionId);

    dispatch(
      setMultiSelectSelectedAttributionIds(newMultiSelectSelectedAttributionIds)
    );
  }

```

benedikt-richter:
The component looks nice!

One point that might be worth discussing is the behavior when clicking a checkbox. Currently, clicking a checkbox to either multi-select or de-multi-select an attribution selects the attribution (it is shown on the right hand side). While for multi-selecting an attribution this might be a neat feature (not sure), for de-multi-selecting it might be anoying. At the moment I feel like preferring to add a `event.stopPropagation()` might be a good idea. 
Are there any other opinions?

src/Frontend/Components/PackageCard/PackageCard.tsx
```diff
@@ -104,6 +120,10 @@ export function PackageCard(props: PackageCardProps): ReactElement | null {
   const attributionIdMarkedForReplacement = useAppSelector(
     getAttributionIdMarkedForReplacement
   );
+  const multiSelectMode = useAppSelector(getMultiSelectMode);
```
benedikt-richter, line 123: (not required in this PR) This component got huge. Feels like refactoring might be a good idea.

src/Frontend/Components/PackageCard/__tests__/PackageCard.test.tsx
```diff
@@ -156,4 +161,50 @@ describe('The PackageCard', () => {
       preSelected: undefined,
     });
   });
+
+  test('has working multi-select box in multi-select mode', () => {
```
benedikt-richter, line 165: 👍

src/Frontend/Components/PackageCard/PackageCard.tsx
jonas-tai: Yes, I think we should definitely try to split it up / refactor it.

jonas-tai:
> The component looks nice!
> 
> One point that might be worth discussing is the behavior when clicking a checkbox. Currently, clicking a checkbox to either multi-select or de-multi-select an attribution selects the attribution (it is shown on the right hand side). While for multi-selecting an attribution this might be a neat feature (not sure), for de-multi-selecting it might be anoying. At the moment I feel like preferring to add a `event.stopPropagation()` might be a good idea. Are there any other opinions?

I think this is something we probably want. We cannot use event.stopPropagation and other implementations are costly, because of css problems. I will leave it out of the current version and we can look at it again at another iteration.

The rest of the comments is implemented now and I also hid the checkboxes in the replace attributions popup.

benedikt-richter (approved):
works nicely 👍

nicarl (approved):

PR #282, opened 2021-12-17
NOT NEEDED ANY LONGER FOR WEB WORKERS Eject from react-scripts

leslielazzarino:
# Probably we just want to avoid all the hassle of this change
and just don't do it, as this PR would only be a start and there isn't a strong reason for it any longer.

### Summary of changes

Eject from react-scripts

Required to use web workers

Check that all yarn scripts still work


config/env.js
```diff
@@ -0,0 +1,109 @@
+// SPDX-FileCopyrightText: Facebook, Inc. and its affiliates
```
benedikt-richter, line 1: I'm not sure we have the copyright for these files

config/env.js
```diff
@@ -0,0 +1,109 @@
...
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
```
benedikt-richter, line 4: Not sure this license applies here

config/env.js
leslielazzarino: I think we need a different header, with both create-react-app compl. info and ours..

config/env.js
nicarl: We should be careful with such files generated by the eject command. @maxhbr can you give us guidance on this?

package.json
```diff
@@ -32,10 +32,10 @@
     "stream-json": "^1.7.2",
     "upath": "^2.0.1",
     "uuid": "^8.3.0",
-    "write-file-atomic": "^3.0.3",
-    "react-scrollbar-size": "4.0.0"
+    "write-file-atomic": "^3.0.3"
```
nicarl, line 35: I think you need to rebase to latest master

src/react-app-env.d.ts
```diff
@@ -3,4 +3,77 @@
...
+}
+
+declare module '*.module.scss' {
+  const classes: { readonly [key: string]: string };
+  export default classes;
+}
+
+declare module '*.module.sass' {
+  const classes: { readonly [key: string]: string };
+  export default classes;
```
benedikt-richter, line 78: this is not needed, I guess

src/react-app-env.d.ts
```diff
@@ -3,4 +3,77 @@
...
+
+declare module '*.avif' {
+  const src: string;
+  export default src;
+}
+
+declare module '*.bmp' {
+  const src: string;
+  export default src;
+}
```
benedikt-richter, line 28: this is not needed, I guess

jest.config.js
```diff
@@ -2,3 +2,49 @@
...
+  setupFiles: ['react-app-polyfill/jsdom'],
+  setupFilesAfterEnv: [],
+  testMatch: [
+    '<rootDir>/src/**/__tests__/**/*.{js,jsx,ts,tsx}',
+    '<rootDir>/src/**/*.{spec,test}.{js,jsx,ts,tsx}',
+  ],
+  testEnvironment: 'jsdom',
+  transform: {
+    '^.+\\.(js|jsx|mjs|cjs|ts|tsx)$': '<rootDir>/config/jest/babelTransform.js',
+    '^.+\\.css$': '<rootDir>/config/jest/cssTransform.js',
```
benedikt-richter, line 18: this is not needed, I guess

config/jest/cssTransform.js
```diff
@@ -0,0 +1,20 @@
+// SPDX-FileCopyrightText: Facebook, Inc. and its affiliates
```
benedikt-richter, line 1: this file is not needed, I guess

LICENSES/MIT.txt
```diff
@@ -0,0 +1,21 @@
+MIT License
```
benedikt-richter, line 1: why do we add this one?

benedikt-richter:
Some of the following dependencies might be unused:

Unused dependencies
* @mui/lab
Unused devDependencies
* babel-plugin-named-asset-import
* browserslist
* identity-obj-proxy
* jest-resolve
* postcss-flexbugs-fixes
* postcss-normalize
* postcss-preset-env
* prompts
* sass-loader
* tailwindcss
* wait-on

PR #284, opened 2021-12-17
Add web workers without ejecting & extra packages

leslielazzarino:
### Summary of changes

Introduce web workers and compute accordion data in a worker (async, not in the main process). Since Webpack 5 web workers are supported by default, not requiring any external package or change in the Webpack configuration. Further changes:
- Web worker can run out of memory with big files, such error is caught and the computation is then performed as fallback on main,
- jest does not support web workers yet, therefore they need to be mocked in the test. The mock just throws an error in the worker and triggers the fallback computation on main.

The app gets stuck for a while when clicking on a folder with a lot of signals inside. Web workers allow to calculate the accordion faster, meanwhile leaving the main process free to process user input.

Open keycloak input and navigate away from the root folder (somewhere where there is an attribution). Then click on the root folder. On the right and above the accordion the UI will change instantly, meanwhile the old accordion will be shown until the new one is ready.
The fallback path can be tested using the ORT input file.

### DIFF
--- a/src/Frontend/Components/AppContainer/__tests__/AppContainer.test.tsx
+++ b/src/Frontend/Components/AppContainer/__tests__/AppContainer.test.tsx
@@ -10,6 +10,8 @@ import { IpcRenderer } from 'electron';
 
 let originalIpcRenderer: IpcRenderer;
 
+jest.mock('../../ResourceDetailsTabs/get-new-accordion-worker');
+
 describe('The AppContainer', () => {
   beforeAll(() => {
     originalIpcRenderer = global.window.ipcRenderer;
--- a/src/Frontend/Components/ResourceDetailsTabs/ResourceDetailsTabs.tsx
+++ b/src/Frontend/Components/ResourceDetailsTabs/ResourceDetailsTabs.tsx
@@ -24,6 +24,7 @@ import {
 } from '../../state/selectors/audit-view-resource-selectors';
 import { OpossumColors } from '../../shared-styles';
 import { useAppSelector } from '../../state/hooks';
+import { getNewAccordionWorker } from './get-new-accordion-worker';
 
 const useStyles = makeStyles({
   tabsRoot: {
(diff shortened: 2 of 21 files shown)


src/Frontend/Components/AppContainer/__tests__/AppContainer.test.tsx
```diff
@@ -10,6 +10,8 @@ import { IpcRenderer } from 'electron';
 
 let originalIpcRenderer: IpcRenderer;
 
+jest.mock('../../ResourceDetailsTabs/get-new-accordion-worker');
```
benedikt-richter, line 13: As discussed: move to `setupTests.js`

src/Frontend/Components/ResourceDetailsTabs/__mocks__/get-new-accordion-worker.ts
```diff
@@ -0,0 +1,8 @@
+// SPDX-FileCopyrightText: Facebook, Inc. and its affiliates
```
benedikt-richter, line 1: This should probably not be placed here, next to ResourceDetailsTabs

src/Frontend/Components/ResourceDetailsTabs/accordion-worker.ts
```diff
@@ -0,0 +1,26 @@
...
+
+import { getPanelData } from './resource-details-tabs-helpers';
+
+self.onmessage = ({
+  data: {
+    selectedResourceId,
+    manualData,
+    externalData,
+    resolvedExternalAttributions,
+  },
```
benedikt-richter, line 14: Might be useful to extract an interface for this type

src/Frontend/Components/ResourceDetailsTabs/ResourceDetailsTabs.tsx
```diff
@@ -80,14 +81,55 @@ export function ResourceDetailsTabs(
...
+            return;
+          }
+
+          worker.onmessage = ({ data: { output } }): void => {
+            setPanelData(output);
+          };
+        } catch (error) {
+          console.log('Error in ResourceDetailsTab worker: ', error);
+
+          const output = getPanelData(
```
nicarl, line 118: A comment for the `try .. catch` block would be nice

benedikt-richter:
Just did a quick performance test with the build version. Opening the root folder of a huge app:
Before: 800ms (total) = approx 1s
Now: 260ms (to display the root folders attributions) + 1100ms (load package panels) + approx 700ms (idle) = approx 2s

src/Frontend/Components/ResourceDetailsTabs/ResourceDetailsTabs.tsx
```diff
@@ -80,14 +81,55 @@ export function ResourceDetailsTabs(
...
+  const [panelData, setPanelData] = useState<Array<PanelData>>([]);
+
+  useMemo(
+    () => {
+      let active = true;
+      setPanelData([]);
+
+      loadPanelData();
+      // @ts-ignore
+      return (): never => {
```
benedikt-richter, line 93: never -> void

benedikt-richter (changes requested):
performance issues shave to be fixed

PR #295, opened 2021-12-21
Remove empty comment and change logic of checking same attributions between parent and child

mayayunx:

### Summary of changes
- Remove empty comment by import.
- Change logic for checking same attributions between parent and child to check all attributions and not only the ids.

- Empty string are imported as a comment

- Changes can be test via `yarn test:unit`, `yarn test:integration-ci` and `yarn test:e2e`

### DIFF
--- a/src/Frontend/state/helpers/save-action-helpers.ts
+++ b/src/Frontend/state/helpers/save-action-helpers.ts
@@ -392,9 +392,10 @@ function removeManualAttributionFromChildIfInferable(
   }
 
   if (
-    isEqual(
-      manualData.resourcesToAttributions[closestParentWithAttribution]?.sort(),
-      manualData.resourcesToAttributions[childId]?.sort()
+    resourcesHaveTheSameAttributions(
+      closestParentWithAttribution,
+      childId,
+      manualData
     )
   ) {
     const childAttributions: Array<string> =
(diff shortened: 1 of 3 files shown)


benedikt-richter (approved):

src/Frontend/state/helpers/save-action-helpers.ts
```diff
@@ -421,6 +418,36 @@ function removeManualAttributionFromChildIfInferable(
...
+    attributionsOfChild.length !=
+    attributionsOfClosestParentWithAttribution.length
+  ) {
+    return false;
+  }
+  return attributionsOfClosestParentWithAttribution.every((attribution) =>
+    attributionsOfChild.some((childAttribution) =>
+      attributionsAreEqual(attribution, childAttribution)
+    )
+  );
```
antonbauhofer, line 448: It seems all correct, but I find this part a bit hard to read.
I would prefer to have both conditions connected with && and return true and else false.
Maybe it could help to extract a function for this with the name `allAttributionsAreEqual` and give the containing function another name, since it does more than only comparing the attributions.

antonbauhofer (commented):
Seems all correct, but I made a suggestion for refactoring because I found one part a bit hard to read.

src/Frontend/state/helpers/save-action-helpers.ts
```diff
@@ -419,6 +417,20 @@ function removeManualAttributionFromChildIfInferable(
...
+  const allAttributionsAreInOtherAttributions = attributions.every(
+    (attribution) =>
+      otherAttributions.some((otherAttribution) =>
+        attributionsAreEqual(attribution, otherAttribution)
+      )
+  );
+  return hasSameLength && allAttributionsAreInOtherAttributions;
+}
+
+function hasSameAttributions(
```
antonbauhofer, line 433: I think it's better now.
- To be more specific, this function could be called `resourcesHaveTheSameAttributions`.
- The arguments and the constants defined in the function could be more general, since the function is not limited to parent and child, but would work for any two resources. So, the arguments could just be `first_resource` and `second_resource` or something else that is generic.

src/Frontend/state/helpers/save-action-helpers.ts
```diff
@@ -419,6 +417,20 @@ function removeManualAttributionFromChildIfInferable(
...
+  otherAttributions: PackageInfo[]
+): boolean {
+  const hasSameLength = attributions.length === otherAttributions.length;
+  const allAttributionsAreInOtherAttributions = attributions.every(
+    (attribution) =>
+      otherAttributions.some((otherAttribution) =>
+        attributionsAreEqual(attribution, otherAttribution)
+      )
+  );
+  return hasSameLength && allAttributionsAreInOtherAttributions;
```
antonbauhofer, line 430: This is much easier to read now, I think 👍

antonbauhofer (commented):
Much better now. I left a suggestion for some renamings.

src/Frontend/state/helpers/save-action-helpers.ts
```diff
@@ -421,6 +416,43 @@ function removeManualAttributionFromChildIfInferable(
...
+}
+
+function hasSameAttributions(
+  closestParentWithAttribution: string,
+  childId: string,
+  manualData: AttributionData
+): boolean {
+  const attributionIdsOfClosestParentWithAttribution =
+    manualData.resourcesToAttributions[closestParentWithAttribution]?.sort() ||
```
benedikt-richter, line 439: is sorting necessary here?

mayayunx: Actually I think not.

mayayunx: Actually it is needed in the comparison of two `AttributionData `in the tests. Without sorting, comparing the `resourcesToAttributions` fails if one resource has attributions ['uuid1', 'uuid2'] and the other one has ['uuid2', 'uuid1'].

antonbauhofer (approved):
Looks good now.

PR #302, opened 2021-12-23
Add multiselect deletion to attribution view

benedikt-richter:
### Summary of changes

A mode to multi-select attributions for deletion is added in attribution view.

So far attributions had to be deleted one by one. This diff adds the possibility for multi-deletion.


nicarl (approved):
:rocket:

nicarl (approved):

PR #307, opened 2021-12-23
Fix bug for inference

nicarl:
### Summary of changes

- when deleting one attribution for a resource with several attributions one has to check if the remaining attributions are might be covered by a parent

There was a bug:

- add two attributions to some resource
- then add the same two attributions to some other resource
 - then add one of these to the parent of one of these
 - then remove the other one from its child using "Delete" in context menu
 - observe that attribution is not removed from child

The described bug should be fixed


benedikt-richter (commented):
works 👍

benedikt-richter (approved):

PR #321, opened 2022-01-03
Fix CORS issues when fetching from pypi

nicarl:
### Summary of changes

This PR conditionally adds a Access-Control-Allow-Origin header to requests going to npmjs urls. 

Adding the Access-Control-Allow-Origin header is only necessary for npmjs urls. Using a filter allows to conditionally attach this header to requests going to certain urls. For pypi urls (and github, WIP) this is not necessary and actually will complain.

Test license fetching with
`https://pypi.org/project/numpy`
`https://npmjs.com/package/react`


antonbauhofer (approved):

FlorianSchepers (approved):
Looks good to me.

PR #353, opened 2022-01-18
ProgressBar: make every file with signal reachable through clicking

leslielazzarino:

### Summary of changes

Folders with signals only can also be reached by means of clicking on the progress bar.

When a folder has signal and no attribution his files are shown in the progress bar as having only signal. Anyway, those files and their parent folders with signals where unreachable by means of clicking on the progress bar.
 
By changing an input file to recreate such configuration.

### DIFF
--- a/src/Frontend/Components/ProgressBar/ProgressBar.tsx
+++ b/src/Frontend/Components/ProgressBar/ProgressBar.tsx
@@ -45,7 +45,7 @@ export function ProgressBar(): ReactElement {
     useAppSelector(getProgressBarData);
 
   const onProgressBarClick = useOnProgressBarClick(
-    progressBarData?.filesWithNonInheritedSignalOnly || []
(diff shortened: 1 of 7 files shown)


benedikt-richter:
Wouldn't it be easier to also jump to folders with signal? 
In the current implementation, if there is a folder with signal high up in the tree, there are many files without signal one visits when clicking on the progress bar.

leslielazzarino:
> Wouldn't it be easier to also jump to folders with signal? In the current implementation, if there is a folder with signal high up in the tree, there are many files without signal one visits when clicking on the progress bar.

Would require a bigger change in the logic, with increased complexity. Not sure if it is worth it.

antonbauhofer (approved):
Looks good.
We tested it manually and it works like expected.

benedikt-richter (approved):

PR #363, opened 2022-01-20
Remove hint if resource or its parent has signal

mayayunx:

### Summary of changes
- While inputting data, concatenate attributions with children's attributions, and check whether hint and other signals exists
- If it is the case, remove the hint signal 

User does not want to see the hint signal if the resource has other signals or its parent have other signals.

The changes can be tested by running `yarn test:unit` or add a  signal (external attribution) in the `oppsum_input.json` file for either `/package.json` or its parent (`/`) and test in the UI.

### DIFF
--- a/src/ElectronBackend/input/cleanInputData.ts
+++ b/src/ElectronBackend/input/cleanInputData.ts
@@ -19,6 +19,7 @@ import {
   RawFrequentLicense,
 } from '../types/types';
 import { canResourceHaveChildren } from '../../Frontend/util/can-resource-have-children';
+import { getParents } from '../../Frontend/state/helpers/get-parents';
 
 function addTrailingSlashIfAbsent(resourcePath: string): string {
   return resourcePath.endsWith('/') ? resourcePath : resourcePath.concat('/');
@@ -166,4 +167,51 @@ export function sanitizeRawBaseUrlsForSources(
         })
       )
     : {};
+}
+
+function hasNonHintAttribution(
+  attributionIds: string[],
+  externalAttributions: Attributions
+): boolean {
+  return (
+    attributionIds &&
+    attributionIds.some(
+      (attributionId) =>
+        externalAttributions[attributionId]?.source?.name !== 'HINT'
+    )
+  );
+}
+
+export function removeHintIfSignalsExist(
+  resourcesToExternalAttributions: ResourcesToAttributions,
+  externalAttributions: Attributions
+): void {
+  const hintAttributionIds = Object.keys(externalAttributions).filter(
+    (attributionId) =>
+      externalAttributions[attributionId].source?.name === 'HINT'
+  );
+  hintAttributionIds.forEach((attributionId) => {
+    Object.keys(resourcesToExternalAttributions).forEach((resourceId) => {
+      const selfAndParents = getParents(resourceId);
+      selfAndParents.push(resourceId);
+      if (
+        resourcesToExternalAttributions[resourceId].includes(attributionId) &&
+        selfAndParents.some((resourcePath) =>
+          hasNonHintAttribution(
+            resourcesToExternalAttributions[resourcePath],
+            externalAttributions
+          )
+        )
+      ) {
+        delete externalAttributions[attributionId];
+      }
+      resourcesToExternalAttributions[resourceId] =
+        resourcesToExternalAttributions[resourceId].filter((id) =>
+          Object.keys(externalAttributions).includes(id)
+        );
+      if (resourcesToExternalAttributions[resourceId].length === 0) {
+        delete resourcesToExternalAttributions[resourceId];
+      }
+    });
+  });
 }
--- a/src/ElectronBackend/input/importFromFile.ts
+++ b/src/ElectronBackend/input/importFromFile.ts
@@ -17,6 +17,7 @@ import {
   cleanNonExistentAttributions,
   cleanNonExistentResolvedExternalSignals,
   parseFrequentLicenses,
+  removeHintIfSignalsExist,
   sanitizeRawAttributions,
   sanitizeRawBaseUrlsForSources,
   sanitizeResourcesToAttributions,
@@ -101,6 +102,12 @@ export async function loadJsonFromFilePath(
   );
 
   log.info('Converting and cleaning data');
+  // Hints are now a special kind of external attribution,
+  // and will be supported in a less hacky way in the near future.
+  removeHintIfSignalsExist(
+    resourcesToExternalAttributions,
+    externalAttributions
+  );
   const parsedFileContent: ParsedFileContent = {
     metadata: parsingResult.metadata,
     resources: parsingResult.resources,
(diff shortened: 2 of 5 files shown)


src/ElectronBackend/input/cleanInputData.ts
```diff
@@ -167,3 +167,71 @@ export function sanitizeRawBaseUrlsForSources(
       )
     : {};
 }
+
+export function removeHintIfSignalsExist(
```
benedikt-richter, line 171: this seems not to be tested in `cleanInputData.test.ts`

src/ElectronBackend/input/cleanInputData.ts
```diff
@@ -167,3 +167,71 @@ export function sanitizeRawBaseUrlsForSources(
...
+  resourcesToExternalAttributions: ResourcesToAttributions,
+  externalAttributions: Attributions
+): void {
+  const resourcesToExternalAttributionsIncludeChildren = Object.assign(
```
nicarl, line 175: I think we can skip the `Object.assign` here as we already create a new object in line 223 -237.

src/ElectronBackend/input/cleanInputData.ts
```diff
@@ -167,3 +167,71 @@ export function sanitizeRawBaseUrlsForSources(
...
+      resourcesToExternalAttributionsIncludeChildren
+    )) {
+      if (resource !== otherResource && otherResource.startsWith(resource)) {
+        resourcesToExternalAttributionsIncludeChildren[resource] =
+          resourcesToExternalAttributionsIncludeChildren[resource].concat(
+            resourcesToExternalAttributionsIncludeChildren[otherResource]
+          );
+      }
+    }
+    const externalNonHintAttribtion =
```
nicarl, line 192: typo: `Attribtion` instead of `Attribution`

src/ElectronBackend/input/cleanInputData.ts
```diff
@@ -167,3 +167,71 @@ export function sanitizeRawBaseUrlsForSources(
...
+      resourcesToExternalAttributionsIncludeChildren[resource].filter(
+        (attributionId) =>
+          externalAttributions[attributionId]?.source?.name !== 'HINT'
+      );
+
+    const externalHintAttribtion =
```
nicarl, line 198: same typo here

src/ElectronBackend/input/cleanInputData.ts
nicarl: I am not sure if the variable name makes much sense here. It `resourcesToExternalAttributions` already contains the children?

src/ElectronBackend/input/cleanInputData.ts
```diff
@@ -167,3 +167,71 @@ export function sanitizeRawBaseUrlsForSources(
...
+
+export function removeHintIfSignalsExist(
+  resourcesToExternalAttributions: ResourcesToAttributions,
+  externalAttributions: Attributions
+): void {
+  const resourcesToExternalAttributionsIncludeChildren = Object.assign(
+    {},
+    sortResources(resourcesToExternalAttributions)
+  );
+  for (const resource of Object.keys(
```
nicarl, line 179: This function is very long and quite complicated. Just to help me understanding what is going on I summarize the steps:

1. Sort `resourcesToExternalAttributions` by resource (we should double check if this is necessary)
2. Loop over `resourcesToExternalAttributions` (`resource`)
2.1 Loop over `resourcesToExternalAttributions` (`otherResource`) (watch out this gives `O(N**2)` run time behavior)
2.1.1 If `resource` and `otherResource` are not identical and `otherResource` is a child of `resource` merge the attributions to the parent
2.1.2 Get all attributions with and without hint for `resource`
2.1.3 If `externalNonHintAttribution` > 0 and `externalHintAttribution` delete `externalHintAttribtion` entry for `resource`
2.1.4 Remove attributions for `resourceId` which are not in `externalAttributions`
2.1.5 If there are not attributions for `resourceId` left, delete the entry

nicarl: An alternative could be to:
1. Find all attributions/`attributionId` which are `HINTS`
2. For each `attributionId` get all `resourceIds`
2.1 For each `resourceId` look if the attribution itself or a parent has a non hint attribution
2.1.1 If yes remove it
2.2.2 if not continue

We can discuss together with @benedikt-richter

src/ElectronBackend/input/cleanInputData.ts
```diff
@@ -167,3 +167,71 @@ export function sanitizeRawBaseUrlsForSources(
       )
     : {};
 }
+
```
leslielazzarino, line 170: This looks like a feature that does not belong in the app.

leslielazzarino (changes requested):

src/ElectronBackend/input/cleanInputData.ts
benedikt-richter: I think the approach Nico is suggesting sounds good.

src/ElectronBackend/input/cleanInputData.ts
```diff
@@ -167,3 +167,48 @@ export function sanitizeRawBaseUrlsForSources(
...
+    for (const resource of Object.keys(resourcesToExternalAttributions)) {
+      if (
+        resourcesToExternalAttributions[resource].includes(attribution) &&
+        getSelfAndParents(resource).some((resourcePath) =>
+          hasNonHintAttribution(resourcesToExternalAttributions[resourcePath])
+        )
+      ) {
+        delete externalAttributions[attribution];
+      }
+      resourcesToExternalAttributions[resource] =
```
nicarl, line 205: Can't we delete the reference in `resourcesToExternalAttributions` after line 203 ?

nicarl (changes requested):

src/ElectronBackend/input/cleanInputData.ts
mayayunx: I thought there might be the case where other resource has the same hint signal, then it should also be deleted. Therefore, the deletion of the reference in `resourcesToExternalAttributions` shouldn't only be in the if-loop.

leslielazzarino (changes requested):

src/ElectronBackend/input/importFromFile.ts
```diff
@@ -100,6 +101,11 @@ export async function loadJsonFromFilePath(
     parsingResult.resourcesToAttributions
   );
 
+  removeHintIfSignalsExist(
```
leslielazzarino, line 104: We should have a comment here explaining that hints are now a special kind of external attribution, and that we plan to support them in a less hacky way in the near future.

leslielazzarino: It also should be documented in the docs.

nicarl (approved):

leslielazzarino (approved):

PR #377, opened 2022-01-25
Remove cache in CI

benedikt-richter:
### Summary of changes

Caching is removed for the CI.

Caching actually did not save time. Without caching the CI is about 10% faster


nicarl (approved):
Crazy

leslielazzarino (approved):

PR #398, opened 2022-02-03
Revert hint removal

benedikt-richter:
### Summary of changes

Revert hint removal

We introduced the removal of unnecessary hints in #363.
This caused a significant decrease in loading performance
and is removed here.

Check opening huge file with built version.


nicarl (approved):
Works :) Tested with a large file (10mb compressed)

leslielazzarino (approved):

PR #403, opened 2022-02-04
Clean up dependencies for VirtualizedTree

benedikt-richter:
### Summary of changes

Clean up dependencies for VirtualizedTree.

In order to extract a package there should be no dependencies between the VirtualizedTree and the other code.

### DIFF
--- a/src/Frontend/Components/VirtualisedTree/List.tsx
+++ b/src/Frontend/Components/VirtualisedTree/List.tsx
@@ -1,0 +1,77 @@
+// SPDX-FileCopyrightText: Facebook, Inc. and its affiliates
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+
+import React, { CSSProperties, ReactElement } from 'react';
+import { FixedSizeList as VirtualizedList } from 'react-window';
+import makeStyles from '@mui/styles/makeStyles';
+
+const useStyles = makeStyles({
+  scrollChild: {
+    direction: 'ltr',
+  },
+});
+
+interface Height {
+  height: number;
+}
+
+interface NumberOfDisplayedItems {
+  numberOfDisplayedItems: number;
+}
+
+interface ListProps {
+  length: number;
+  max: NumberOfDisplayedItems | Height;
+  getListItem(index: number): ReactElement | null;
+  cardVerticalDistance?: number;
+  alwaysShowHorizontalScrollBar?: boolean;
+  className?: string;
+}
+
+function maxHeightWasGiven(
+  max: NumberOfDisplayedItems | Height
+): max is Height {
+  return Boolean((max as Height).height);
+}
+
+export function List(props: ListProps): ReactElement {
+  const classes = useStyles();
+  const cardHeight = props.cardVerticalDistance || 24;
+  const maxHeight = maxHeightWasGiven(props.max)
+    ? props.max.height
+    : props.max.numberOfDisplayedItems * cardHeight;
+  const currentHeight = props.length * cardHeight;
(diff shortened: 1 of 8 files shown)


src/Frontend/Components/VirtualisedTree/VirtualizedTree.tsx
```diff
@@ -38,22 +29,23 @@ const useStyles = makeStyles({
...
-    resource: Resources | 1,
+    itemName: string,
+    item: ItemsForTree | 1,
     nodeId: string
   ) => ReactElement;
   expandedIds: Array<string>;
-  selectedResourceId: string;
-  isFileWithChildren: PathPredicate;
+  selectedItemId: string;
+  isFileWithChildren: PathPredicateForTree;
```
nicarl, line 40: Should be `isNodeWithChildren` ?

src/Frontend/Components/VirtualisedTree/VirtualizedTree.tsx
```diff
@@ -38,22 +29,23 @@ const useStyles = makeStyles({
 const DEFAULT_MAX_TREE_DISPLAYED_ITEMS = 5;
 
 interface VirtualizedTreeProps {
-  resources: Resources;
+  items: ItemsForTree;
```
nicarl, line 32: alternative: `nodes: NodesForTree`

nicarl: I think we should use either `node` or `item` term consequently over all files we will extract as package. I have a preference for `node`

src/Frontend/Components/VirtualisedTree/List.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+function maxHeightWasGiven(
+  max: NumberOfDisplayedItemsForTree | HeightForTree
+): max is HeightForTree {
+  return Boolean((max as HeightForTree).height);
+}
+
+export function List(props: ListProps): ReactElement {
+  const classes = useStyles();
+  const cardHeight = props.cardVerticalDistance || 24;
```
nicarl, line 34: 24 could be moved to a `const` in the beginning of the file

nicarl (commented):
I have some general comments concerning naming. The rest looks fine

src/Frontend/Components/VirtualisedTree/VirtualizedTree.tsx
benedikt-richter: Agree. Node sounds good to me!

benedikt-richter: I'll do so in follow up diff

src/Frontend/Components/VirtualisedTree/VirtualizedTree.tsx
leslielazzarino: It is only the "resource" folders that we label as files. As now the icons are controlled from the outside we should extract this logic from the tree in a follow up pr.

src/Frontend/Components/VirtualisedTree/List.tsx
```diff
@@ -0,0 +1,72 @@
...
+    direction: 'ltr',
+  },
+});
+
+interface ListProps {
+  length: number;
+  max: NumberOfDisplayedItemsForTree | HeightForTree;
+  getListItem(index: number): ReactElement | null;
+  cardVerticalDistance?: number;
+  alwaysShowHorizontalScrollBar?: boolean;
```
leslielazzarino, line 24: I guess we should either surface this property in the tree, or hard-code the option and simplify the logic.

leslielazzarino (changes requested):

nicarl (approved):
I will check the renaming in the follow up ;-)

PR #405, opened 2022-02-04
Extract virtualized tree

benedikt-richter:
### Summary of changes

Extract virtualized tree into own folder

Preparation for publishing in own package


nicarl (approved):
Nit: `get-tree-node-props.test.ts` has one variable left which is called `item`

nicarl (approved):

src/Frontend/Components/ResourceBrowser/ResourceBrowser.tsx
```diff
@@ -43,6 +43,35 @@ const useStyles = makeStyles({
     background: OpossumColors.white,
     height: '100%',
   },
+  treeItemLabel: {
+    height: 19,
```
leslielazzarino, line 47: This should be controlled through cardHeight, not through styling

src/Frontend/extracted/VirtualisedTree/VirtualizedTree.tsx
```diff
@@ -46,6 +48,7 @@ interface VirtualizedTreeProps {
   expandedNodeIcon?: ReactElement;
   nonExpandedNodeIcon?: ReactElement;
   className?: string;
+  treeNodeStyle?: TreeNodeStyle;
```
leslielazzarino, line 51: I would like to have our style as default here, for quick fist use by the user. The user-defined classes should be added to the default ones with clsx, so that the user can even just overwrite a few.

leslielazzarino (changes requested):

src/Frontend/extracted/VirtualisedTree/VirtualizedTree.tsx
benedikt-richter: Our style depends on the opossum colors. Not sure how this could be the default without duplicating the colors. However, having some reasonable default style is a good idea.

benedikt-richter: I'd suggest a separate diff to come up with a solid default styling.

leslielazzarino (approved):

PR #406, opened 2022-02-04
Add eslint plugin to enforce consistent naming of files

nicarl:
### Summary of changes

Adds a linter rule to only allows files with upper or lower CamelCase naming as well as dashes (-).

File naming was not enforced previously and there were some inconsistencies. Furthermore, the naming for tests might be important for test discovery.

Create a file with an invalid file name, e.g. `my_invalid_filename.ts` and run `yarn lint`


benedikt-richter (approved):
nice

leslielazzarino (approved):

PR #407, opened 2022-02-07
Fix styling

benedikt-richter:
### Summary of changes

Fix styling

After:
Screenshot 2022-02-07 at 12 45 17

In the attribution column there was no vertical spacing between individual
text boxes. Furthermore, sometimes the buttons did overlap the comment
text box.

Before:
Screenshot 2022-02-07 at 12 44 21

Check appearance with several zoom levels.


nicarl (approved):
Screenshots would be nice and a better branch name :-) Otherwise it is fine

Issue #424, opened 2022-02-09
Green color should propagate upwards in tree


benedikt-richter:
If all resources in a folder have an attribution the user shall get visible feedback about that.

Currently, there is no visual feedback for the user (ElectronBackend should be green):
![Screenshot 2022-02-07 at 11 34 35](https://user-images.githubusercontent.com/46576389/153176433-59a3476c-6d90-44a0-acc4-1f68d82d6d7c.png)

Important: check performance afterwards.

PR #425, opened 2022-02-09
Propagate green if all children have attributions

mayayunx:

### Summary of changes
A folder should be light green if:
- some contained resources have signals
- all contained resources with signals have attribution or inferred attribution
- the folder itself has no attribution inferred attribution

- "resolved" signals should not have impact on coloring
- Fix: #424

- In the UI and with `yarn test:unit`

### DIFF
--- a/src/Frontend/Components/ResourceBrowser/get-tree-item-label.tsx
+++ b/src/Frontend/Components/ResourceBrowser/get-tree-item-label.tsx
@@ -62,6 +62,15 @@ export function getTreeItemLabel(
       )}
       isAttributionBreakpoint={isAttributionBreakpoint(nodeId)}
       showFolderIcon={canHaveChildren && !isFileWithChildren(nodeId)}
+      containsResourcesWithOnlyExternalAttribution={
+        canHaveChildren &&
+        containsResourcesWithOnlyExternalAttribution(
+          nodeId,
+          resourcesToManualAttributions,
+          resourcesToExternalAttributions,
+          resource
+        )
+      }
     />
   );
 }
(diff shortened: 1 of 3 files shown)


benedikt-richter:
Looks good!

I'll still test performance.

src/Frontend/Components/ResourceBrowser/get-tree-item-label.tsx
```diff
@@ -152,3 +162,25 @@ function hasParentWithManualAttributionAndNoOwnAttribution(
...
+
+function allChildrenWithExternalAttributionHaveManualAttribution(
+  nodeId: string,
+  resourcesWithExternalAttributedChildren: ResourcesWithAttributedChildren,
+  manualAttributions: Attributions,
+  resourcesToManualAttributions: ResourcesToAttributions,
+  isAttributionBreakpoint: PathPredicate
+): boolean {
+  return (
+    resourcesWithExternalAttributedChildren[nodeId] &&
```
leslielazzarino, line 174: The logic in this function is extremely inefficient for folders with a lot of signals. We look for (mostly parent) attributions for each and every resource with signals. We should navigate the node sub-tree top-down recursively, ignoring branches as soon as a folder has an attribution. That would be much more efficient. 
If one wanted to be thorough, the children with external/manual attributions was implemented a while ago, when we wanted to show arrows for each resource. Now that we want to show other information it would be best to rebuild it, optimized for the new visuals. But it would be more expensive and should be packed in a technical debt ticket.

leslielazzarino (changes requested):

benedikt-richter (approved):

leslielazzarino (approved):

PR #427, opened 2022-02-10
Improve buttons in edit popup

benedikt-richter:
### Summary of changes

Exchange left and right button in edit popup. Refacator NotificationPopup interface.

The left button should be the "save" button, and it should have the more prominent color (blue). This was not the case.
Further, instead of providing button configuration for all buttons as a flat object, a new type - ButtonConfig - is introduced to improve the interface.

Before:
Screenshot 2022-02-10 at 10 38 48

After:
Screenshot 2022-02-10 at 10 39 47

### DIFF
--- a/src/Frontend/Components/ConfirmationPopup/ConfirmationPopup.tsx
+++ b/src/Frontend/Components/ConfirmationPopup/ConfirmationPopup.tsx
@@ -31,10 +31,14 @@ export function ConfirmationPopup(props: ConfirmationPopupProps): ReactElement {
     <NotificationPopup
       content={props.content}
       header={props.header}
-      leftButtonText={ButtonText.Confirm}
-      onLeftButtonClick={handleDeletionClick}
-      rightButtonText={ButtonText.Cancel}
-      onRightButtonClick={handleCancelClick}
+      leftButtonConfig={{
+        onClick: handleDeletionClick,
+        buttonText: ButtonText.Confirm,
+      }}
+      rightButtonConfig={{
+        onClick: handleCancelClick,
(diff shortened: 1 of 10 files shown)


src/Frontend/Components/ConfirmationPopup/ConfirmationPopup.tsx
```diff
@@ -31,10 +31,14 @@ export function ConfirmationPopup(props: ConfirmationPopupProps): ReactElement {
     <NotificationPopup
       content={props.content}
       header={props.header}
-      leftButtonText={ButtonText.Confirm}
-      onLeftButtonClick={handleDeletionClick}
-      rightButtonText={ButtonText.Cancel}
-      onRightButtonClick={handleCancelClick}
+      leftButtonConfig={{
+        onClick: handleDeletionClick,
```
nicarl, line 35: Should we rename the function from `handleDeletionClick` to `handleLeftButtonClick` ?

src/Frontend/Components/ConfirmationPopup/ConfirmationPopup.tsx
```diff
@@ -31,10 +31,14 @@ export function ConfirmationPopup(props: ConfirmationPopupProps): ReactElement {
...
+        buttonText: ButtonText.Confirm,
+      }}
+      rightButtonConfig={{
+        onClick: handleCancelClick,
```
nicarl, line 39: Same here

nicarl (approved):

leslielazzarino (approved):

Issue #428, opened 2022-02-10
Log errors in frontend in backend log file


benedikt-richter:
There is a log file for the electron backend. Include there error messages from the frontend.

Also see whether logging in backend in general could be improved. E.g. log file exports

PR #429, opened 2022-02-10
Improvements on backend logging

mayayunx:

### Summary of changes
- Log errors in frontend in log file
- Change create output file log
- Add logs for export files
- Add 'help' tab with option to open folder for log files

- Errors in frontend should be also save in log files
- Enable users to find location of log files
- Fix: #428

In the app, click on 'Help' and 'Open log files folder' and check logs in the file

### DIFF
--- a/src/ElectronBackend/input/importFromFile.ts
+++ b/src/ElectronBackend/input/importFromFile.ts
@@ -61,7 +61,7 @@ export async function loadJsonFromFilePath(
     return;
   }
 
-  log.info('Successfully parsed input file.');
+  log.info('... Successfully parsed input file.');
   const externalAttributions = sanitizeRawAttributions(
     parsingResult.externalAttributions
   );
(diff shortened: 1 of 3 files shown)


src/ElectronBackend/input/importFromFile.ts
```diff
@@ -72,15 +72,16 @@ export async function loadJsonFromFilePath(
...
-  log.info(
-    `Creating output file if it does not exist, project ID is ${projectId}`
-  );
-  createOutputFileIfItDoesNotExist(
-    manualAttributionFilePath,
-    externalAttributions,
-    parsingResult.resourcesToAttributions,
-    projectId
-  );
+  if (!fs.existsSync(manualAttributionFilePath)) {
```
nicarl, line 75: What is your motivation to move this logic from `createOutputFile` up?

nicarl (commented):
Works nicely, I just added a `throw Error("my custom error")` to any frontend component which is rendered by default and checked that the stack trace appears in the log file.

src/ElectronBackend/input/importFromFile.ts
mayayunx: My motivation is to be able to directly log it if a file is created. But I can also move the log inside the function if it is better.

src/ElectronBackend/main/listeners.ts
```diff
@@ -243,14 +244,19 @@ export function _exportFileAndOpenFolder(mainWindow: BrowserWindow) {
...
         await fileExporter(exportedFilePath, exportArgs);
       } else {
+        log.error(`Failed to create ${exportArgs.type} export.`);
         throw new Error(`Failed to create ${exportArgs.type} export.`);
       }
     } finally {
       loadingWindow.close();
 
       if (exportedFilePath) {
+        log.info(`... Successfully created ${exportArgs.type} export`);
```
benedikt-richter, line 259: Sometimes it is `... Successfully` and sometimes `Successfully`. Should be consistent

benedikt-richter (approved):
works 👍 

just one small comment

nicarl (approved):

Issue #451, opened 2022-02-17
Introduce jest test environment for integration tests


benedikt-richter:
Currently, we pretty much always use the same setup (e.g. ipcRenderer mocking) in integration tests. However, this is duplicated code.

A nice way to extract this would be to use something like jest testEnvironment:

https://jestjs.io/docs/configuration#testenvironment-string

PR #464, opened 2022-02-24
Extract test setting to reduce duplicates

mayayunx:

### Summary of changes
- Extract duplicate test mock function to helper 
- Extract duplicate test setting to `setupTest.js` file

- Lot of duplicated codes in the tests
- Fix: #451

The changes can be tested by running `yarn test:unit` and `yarn test:integration-ci`. (Also test the speed of the tests against main branch. On my computer it is roughly the same.)


setupTests.js
```diff
@@ -3,4 +3,29 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
-require('@testing-library/jest-dom/extend-expect');
+import '@testing-library/jest-dom/extend-expect';
```
nicarl, line 6: Can we make this file typed (`.ts` instead of `.js`)?
You will need to change something in the `package.json`.
This should be a first hint:
https://stackoverflow.com/questions/57442151/i-cannot-configure-jest-properly-to-import-modules-setupfilesafterenv

benedikt-richter: Simply renaming to `setupTests.ts` and adapting the scripts in `package.json` accordingly should do the job

benedikt-richter (approved):

nicarl (approved):

PR #487, opened 2022-03-03
Highlight incomplete attributions attribution view

mayayunx:

### Summary of changes
- Highlight missing fields in attribution view with the same logic as in report view
- Highlight package card on the left side with triangle if there are missing fields

- As a user I want to be able to quickly see which attributions are missing some important fields. 

The changes can be tested in the UI.


src/Frontend/Components/AttributionColumn/AuditingSubPanel.tsx
```diff
@@ -102,6 +103,11 @@ export function AuditingSubPanel(props: AuditingSubPanelProps): ReactElement {
                 name: `Low (${DiscreteConfidence.Low})`,
               },
             ]}
+            isHighlighted={
+              props.showHighlight &&
+              (!props.displayPackageInfo.attributionConfidence ||
+                props.displayPackageInfo.attributionConfidence < 50)
```
benedikt-richter, line 109: this is duplicated logic

src/Frontend/Components/ListCard/ListCard.tsx
```diff
@@ -39,6 +39,23 @@ const useStyles = makeStyles({
...
+      'linear-gradient(225deg, ' +
+      OpossumColors.lightOrange +
+      ' 44.5px, ' +
+      defaultBackgroundColor +
+      ' 0) 0 0/100% 40px no-repeat',
+    '&:hover': {
+      background:
+        'linear-gradient(225deg, ' +
+        OpossumColors.lightOrange +
+        ' 44.5px, ' +
```
benedikt-richter, line 54: this is duplicated multiple times

benedikt-richter (changes requested):
In the context of the report view the logic whether a field is highlighted is encoded in the function `isMarkedTableCell` this logic should not be duplicated for report view and for attribution view but only be implemented once.

src/Frontend/util/is-information-missing.ts
```diff
@@ -0,0 +1,30 @@
+// SPDX-FileCopyrightText: Meta Platforms, Inc. and its affiliates
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+
+import { PackageInfo } from '../../shared/shared-types';
+
+export function isInformationMissing(
+  attributionProperty: string,
```
nicarl, line 9: I think we could benefit from stricter typing here. Either use an enum or a string literal

src/Frontend/Components/ListCard/ListCard.tsx
```diff
@@ -16,6 +16,16 @@ const hoveredBackgroundColor = OpossumColors.lightestBlueOnHover;
 const defaultBackgroundColor = OpossumColors.lightestBlue;
 const packageBorder = `1px ${OpossumColors.white} solid`;
 
+const getBackground = (backgroundColor: string): string => {
```
benedikt-richter, line 19: this name is very generic. Would be good to find a more specific one.

src/Frontend/Components/ListCard/ListCard.tsx
```diff
@@ -184,10 +208,13 @@ export function ListCard(props: ListCardProps): ReactElement | null {
         props.cardConfig.isSelected
           ? props.cardConfig.isContextMenuOpen
             ? classes.hoveredSelected
+            : props.highlightedCard
+            ? classes.highLightedSelected
```
benedikt-richter, line 212: `highLightedSelected` -> `highlightedSelected`

src/Frontend/Components/ListCard/ListCard.tsx
```diff
@@ -184,10 +208,13 @@ export function ListCard(props: ListCardProps): ReactElement | null {
...
             : classes.selected
           : null,
         props.cardConfig.isMarkedForReplacement && classes.markedForReplacement,
-        props.cardConfig.isResolved && classes.resolved
+        props.cardConfig.isResolved && classes.resolved,
+        props.highlightedCard && classes.highLightedPackage
```
benedikt-richter, line 217: general comment:

consolidate "high**l**ighted" vs "high**L**ighted"

src/Frontend/util/is-information-missing.ts
```diff
@@ -0,0 +1,31 @@
+// SPDX-FileCopyrightText: Meta Platforms, Inc. and its affiliates
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+
+import { PackageInfo } from '../../shared/shared-types';
+import { AttributionInfo } from '../Components/Table/Table';
+
+export function isInformationMissing(
```
benedikt-richter, line 9: this name is very generic and be made more specific

src/Frontend/util/is-information-missing.ts
```diff
@@ -0,0 +1,31 @@
+// SPDX-FileCopyrightText: Meta Platforms, Inc. and its affiliates
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+
+import { PackageInfo } from '../../shared/shared-types';
+import { AttributionInfo } from '../Components/Table/Table';
```
benedikt-richter, line 7: we should not import from components to util

src/Frontend/util/is-information-missing.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export function isInformationMissing(
+  attributionProperty: keyof PackageInfo | keyof AttributionInfo | 'icons',
```
benedikt-richter, line 10: if the variable `attributionInfo` is of tpye `PackageInfo` how can there be the property `icons`? Also why would we need `keyof AttributionInfo`?

benedikt-richter (changes requested):

src/Frontend/util/is-information-missing.ts
nicarl: We should have an eslint rule for this

src/Frontend/util/is-information-missing.ts
mayayunx: I added `keyof AttributionInfo` and `icons` since otherwise there are errors regarding to the types when using it in the report view. But I changed it now and check the type before calling the function

benedikt-richter:
When hovering the package card, the red part of the card should also change its color.

benedikt-richter (approved):

nicarl (approved):

PR #498, opened 2022-03-08
Add field "criticality" to attributions

benedikt-richter:
### Summary of changes

There is a new field introduced in PackageInfo: "criticality". This field has values in an enum and is optional.

The new field will be used to highlight signals that are critical. Criticality is set in the input file used for OpossumUI.

### DIFF
--- a/src/ElectronBackend/input/OpossumInputFileSchema.json
+++ b/src/ElectronBackend/input/OpossumInputFileSchema.json
@@ -118,6 +118,10 @@
             "type": "string",
             "description": "Additional human-readable comments about the attribution that don't fit into other fields"
           },
+          "criticality": {
+            "type": "string",
+            "description": "Indication on how critical a signal is. Possible values are \"high\" and \"medium\"."
+          },
           "copyright": {
             "type": "string",
             "description": "Copyright of the package"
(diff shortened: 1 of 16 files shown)


antonbauhofer (approved):
Looks good 👍

src/ElectronBackend/input/OpossumInputFileSchema.json
```diff
@@ -118,6 +118,11 @@
             "type": "string",
             "description": "Additional human-readable comments about the attribution that don't fit into other fields"
           },
+          "criticality": {
+            "type": "string",
+            "enum": ["high", "medium"],
```
nicarl, line 123: As discussed this breaks backward compatibility once we add new values for criticality. What is our strategy here? Converting to the enum in code would be possible and use `null` as fallback.

nicarl (approved):
Just one comment concerning the backwards compatibility

src/ElectronBackend/input/OpossumInputFileSchema.json
benedikt-richter: So far we did not support this kind of backward compatibility. However, I still weakened the schema and now expect any string.

nicarl (approved):

PR #510, opened 2022-03-10
Highlight critical signals

antonbauhofer:
### Summary of changes

Critical signals and pre-selected attributions are highlighted in the audit and attribution view. If a critical pre-selected attribution is confirmed, it is not highlighted anymore.

In input documents, signals can now be marked as critical, if they are specifically important to consider or marked as red flags. To make the review process more effective, these signals are highlighted in OpossumUI.

Open the test input file `opossum_input.json` and view the signals contained in the root folder or navigate to the attribution view.


example-files/opossum_input.json
```diff
@@ -137,17 +137,34 @@
...
       "originId": "1eb75f3b-fc84-e781-ad9d-d46d6d61abf2"
     },
+    "bb61b094-b5ba-4bd0-b218-8ec96b7c5ae1": {
```
benedikt-richter, line 151: Now we have typescript three times as signal. All very similar (with same url...). Therefore, it's not easy to distinguish these on e.g. package cards. Maybe this is an opportunity to improve on this

src/Frontend/Components/AttributionList/AttributionList.tsx
```diff
@@ -66,6 +66,9 @@ export function AttributionList(props: AttributionListProps): ReactElement {
       firstParty: attribution.firstParty,
       excludeFromNotice: attribution.excludeFromNotice,
       followUp: Boolean(attribution.followUp),
+      criticality: Boolean(attribution.preSelected)
+        ? attribution.criticality
+        : undefined,
```
benedikt-richter, line 71: could be `attribution.preSelected && attribution.criticality`

src/Frontend/Components/PackageList/PackageList.tsx
```diff
@@ -83,6 +84,7 @@ export function PackageList(props: PackageListProps): ReactElement {
       firstParty: packageInfo.firstParty,
       excludeFromNotice: packageInfo.excludeFromNotice,
       followUp: Boolean(packageInfo.followUp),
+      criticality: isPreselected ? packageInfo.criticality : undefined,
```
benedikt-richter, line 87: as above

src/Frontend/Components/AttributionList/AttributionList.tsx
antonbauhofer: This gives a typing error. But we can leave out the cast to Boolean.

example-files/opossum_input.json
antonbauhofer: I replaced two of the occurrences.

benedikt-richter (approved):
please squash commits

nicarl (approved):
Changes and styling looks good. I just wonder one thing: for now `criticality` is a somewhat hidden property of an attribution. In particular in the `Attribution View` it is hard to distinguish between incomplete and critical attributions.
Maybe we can at least add an on hover text to display the reason why an attribution is marked?

benedikt-richter (changes requested):
the package card looks broken on hover on attribution view

src/Frontend/Components/ListCard/ListCard.tsx
```diff
@@ -205,7 +260,21 @@ export function ListCard(props: ListCardProps): ReactElement | null {
...
+    return props.cardConfig.criticality ? (
+      <MuiTooltip
+        classes={{ tooltip: classes.tooltip }}
+        title={`has criticality ${props.cardConfig.criticality}`}
+        placement="left"
+      >
+        {children}
+      </MuiTooltip>
+    ) : (
+      <span>{children}</span>
```
nicarl, line 273: I do not like the ternary operator inside `wrapInTooltip`. If I pass something into `wrapInTooltip` I want to have a tool-tip for sure and not depend on some additional logic. Otherwise the function should be named `eventuallyWrapInTooltip` :wink:

benedikt-richter:
The fact that on attribution view critical attributions are displayed the same as incomplete ones confused me and, in my opinion, should not be the case.

benedikt-richter:
I just realised that there is a tooltip. However, the placement is a bit odd. First I did not see it and then I did not realize it belongs to the card:
Screenshot 2022-03-16 at 10 00 44

src/Frontend/Components/ListCard/ListCard.tsx
```diff
@@ -205,7 +260,21 @@ export function ListCard(props: ListCardProps): ReactElement | null {
     }
   }
 
-  return (
+  function wrapInTooltip(children: ReactElement): ReactElement {
```
benedikt-richter, line 263: I realized the same pattern in `IconButton`. It should not be duplicated.

src/Frontend/Components/ListCard/ListCard.tsx
benedikt-richter: That's true. It feels a bit odd. It would probably better to always add a tooltip and disabling or hiding it under certain conditions.

benedikt-richter (changes requested):
see comments

src/Frontend/Components/ListCard/ListCard.tsx
antonbauhofer: The logic was taken from the icons that also (sometimes) have a tooltip.
I agree, and we should fix it in the icons, too.

antonbauhofer:
> The fact that on attribution view critical attributions are displayed the same as incomplete ones confused me and, in my opinion, should not be the case.

Yes, I agree. Originally, I chose a different highlighting for that reason, but it was decided in a team meeting to have the same highlighting and possibly fix it later.

antonbauhofer:
> I just realised that there is a tooltip. However, the placement is a bit odd. First I did not see it and then I did not realize it belongs to the card: Screenshot 2022-03-16 at 10 00 44
> 
> It would be great to anchor it at the current mouse position or otherwise much closer to the card

Agree. I'll fix the placement.

src/Frontend/Components/ListCard/ListCard.tsx
antonbauhofer: I extracted the tooltip wrapper into a separate component that is used for both occurrences.

antonbauhofer:
After some discussion, we decided to change the styling so that it is not the same as for incomplete attributions. Consequently, we also decided to drop the tooltip.

src/Frontend/Components/ListCard/ListCard.tsx
```diff
@@ -176,6 +242,7 @@ const useStyles = makeStyles({
       background: OpossumColors.white,
     },
   },
+  tooltip: tooltipStyle,
```
benedikt-richter, line 245: unused

benedikt-richter (approved):

src/Frontend/Components/PackageList/PackageList.tsx
```diff
@@ -83,6 +84,11 @@ export function PackageList(props: PackageListProps): ReactElement {
       firstParty: packageInfo.firstParty,
       excludeFromNotice: packageInfo.excludeFromNotice,
       followUp: Boolean(packageInfo.followUp),
+      criticality: props.isExternalAttribution
```
nicarl, line 87:
```typescript
props.isExternalAttribution || packageInfo.preSelect ?? packageInfo.criticality

nicarl (approved):

PR #530, opened 2022-03-17
Add progress bar for folders

mayayunx:

### Summary of changes
- Render progress bar for folders with workers

- User wants to see a small progress bar for ther folder he/she is currently in

The changes can be tested with `yarn test:unit` and in the UI.


src/Frontend/Components/ResourceDetailsViewer/__tests__/ResourceDetailsViewer.test.tsx
```diff
@@ -164,7 +165,7 @@ describe('The ResourceDetailsViewer', () => {
   });
 
   test('renders Contained External Packages', () => {
-    const { store } = renderComponentWithStore(<ResourceDetailsViewer />);
+    const testStore = createTestAppStore();
```
nicarl, line 168: What is the motivation for this change?

src/Frontend/state/helpers/progress-bar-data-helpers.ts
```diff
@@ -161,3 +163,58 @@ export function resourceHasOnlyPreSelectedAttributions(
...
+    filesWithOnlyPreSelectedAttributionCount: 0,
+    filesWithOnlyExternalAttributionCount: 0,
+    resourcesWithNonInheritedSignalOnly: [],
+  };
+
+  const parentResources = args.resourceId
+    .split('/')
+    .filter((resource: string) => resource !== '');
+
+  let currentResources: Resources | 1 = args.resources;
```
nicarl, line 193: Just to make sure that I understood the code correctly:
At the end (after line 201) currentResources are the resources inside `resourceId`?

src/Frontend/state/helpers/progress-bar-data-helpers.ts
```diff
@@ -161,3 +163,58 @@ export function resourceHasOnlyPreSelectedAttributions(
...
+    for (const resource of parentResources) {
+      if (currentResources && currentResources != 1) {
+        currentResources = currentResources[resource];
+      }
+    }
+  }
+
+  if (currentResources !== 1) {
+    updateProgressBarDataForResources(
+      progressBarData,
```
nicarl, line 205: never liked the mutation taking place here, but I guess it is okay :no_mouth:

src/Frontend/Components/ProgressBar/ProgressBar.tsx
```diff
@@ -37,29 +54,103 @@ const useStyles = makeStyles({
...
+    getResourcesToManualAttributions
+  );
+  const resourcesToExternalAttributions = useAppSelector(
+    getResourcesToExternalAttributions
+  );
+  const resolvedExternalAttributions = useAppSelector(
+    getResolvedExternalAttributions
+  );
+  const attributionBreakpoints = useAppSelector(getAttributionBreakpoints);
+  const filesWithChildren = useAppSelector(getFilesWithChildren);
```
nicarl, line 90: Can you make empty lines in between the different groups of hooks we are using here?

src/Frontend/Components/ResourceDetailsViewer/__tests__/ResourceDetailsViewer.test.tsx
mayayunx: It was changed while I was trying to fix the test and forgot to change it back. It is now reverted,

src/Frontend/state/helpers/progress-bar-data-helpers.ts
mayayunx: Yes, that's right

src/Frontend/state/helpers/progress-bar-data-helpers.ts
```diff
@@ -161,3 +163,58 @@ export function resourceHasOnlyPreSelectedAttributions(
...
+  const isFileWithChildren = getFileWithChildrenCheck(args.filesWithChildren);
+  const progressBarData: ProgressBarData = {
+    fileCount: 0,
+    filesWithManualAttributionCount: 0,
+    filesWithOnlyPreSelectedAttributionCount: 0,
+    filesWithOnlyExternalAttributionCount: 0,
+    resourcesWithNonInheritedSignalOnly: [],
+  };
+
+  const parentResources = args.resourceId
```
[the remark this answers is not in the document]
mayayunx: I think the `getParents` function works a little different. It returns the complete path for the parents but I only want the name of parent and current resource. I renamed the variables to make it clearer.

src/Frontend/state/helpers/progress-bar-data-helpers.ts
```diff
@@ -161,3 +163,56 @@ export function resourceHasOnlyPreSelectedAttributions(
...
+  };
+
+  const parentAndCurrentResources = args.resourceId
+    .split('/')
+    .filter((resource: string) => resource !== '');
+
+  let childResources: Resources | 1 = args.resources;
+
+  for (const resource of parentAndCurrentResources) {
+    if (childResources && childResources != 1) {
```
leslielazzarino, line 196: I do not like this if, it is just there because the typing isn't smart enough to understand what is going on. I would do as follows:

1. childResources is Resources, this function is called getFolderProgressBarData, the id should be that of a folder, therefore it is ok if the function throws when called on a file
2. remove the ifs in rows 196 and 201, they are not really needed
3. cast the next line to Resources

Ps: never use != instead of !==, unless you have a really good reason for it. We should probably introduce a lint rule for it.

src/Frontend/state/helpers/progress-bar-data-helpers.ts
```diff
@@ -161,3 +163,56 @@ export function resourceHasOnlyPreSelectedAttributions(
...
+  resourcesToExternalAttributions: ResourcesToAttributions;
+  resolvedExternalAttributions: Set<string>;
+  attributionBreakpoints: Set<string>;
+  filesWithChildren: Set<string>;
+}): ProgressBarData {
+  const isAttributionBreakpoint = getAttributionBreakpointCheck(
+    args.attributionBreakpoints
+  );
+  const isFileWithChildren = getFileWithChildrenCheck(args.filesWithChildren);
+  const progressBarData: ProgressBarData = {
```
leslielazzarino, line 181: I would extract the right side to a getEmptyProgressBarData() return value, and use it here and for the other progress bar.

src/Frontend/state/helpers/progress-bar-data-helpers.ts
```diff
@@ -161,3 +163,56 @@ export function resourceHasOnlyPreSelectedAttributions(
...
+  let childResources: Resources | 1 = args.resources;
+
+  for (const resource of parentAndCurrentResources) {
+    if (childResources && childResources != 1) {
+      childResources = childResources[resource];
+    }
+  }
+
+  if (childResources !== 1) {
+    updateProgressBarDataForResources(
```
leslielazzarino, line 202: This function has currently a bug. If the folder you are in has one or more attributions, they are ignored for the computations, therefore the sub-progress bar isn't all green. One solution could be to give {[$RESOURCE_NAME]: childResources} as resources to the function.

leslielazzarino (changes requested):

src/Frontend/state/helpers/progress-bar-data-helpers.ts
```diff
@@ -161,3 +163,56 @@ export function resourceHasOnlyPreSelectedAttributions(
...
+  const isFileWithChildren = getFileWithChildrenCheck(args.filesWithChildren);
+  const progressBarData: ProgressBarData = {
+    fileCount: 0,
+    filesWithManualAttributionCount: 0,
+    filesWithOnlyPreSelectedAttributionCount: 0,
+    filesWithOnlyExternalAttributionCount: 0,
+    resourcesWithNonInheritedSignalOnly: [],
+  };
+
+  const parentAndCurrentResources = args.resourceId
```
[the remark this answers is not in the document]
mayayunx: substring() does not work with negative index, I changed it to slice()

src/Frontend/state/helpers/progress-bar-data-helpers.ts
```diff
@@ -161,3 +157,64 @@ export function resourceHasOnlyPreSelectedAttributions(
...
+    resources,
+    args.manualAttributions,
+    args.resourcesToManualAttributions,
+    filterResourcesToAttributions(
+      args.resourcesToExternalAttributions,
+      args.resolvedExternalAttributions
+    ),
+    isAttributionBreakpoint,
+    isFileWithChildren,
+    parentPath
```
leslielazzarino, line 206: If you change the tree than you do not need a parent path any longer.

src/Frontend/Components/ProgressBar/__tests__/ProgressBar.test.tsx
```diff
@@ -94,4 +105,180 @@ describe('ProgressBar', () => {
     renderComponentWithStore(<ProgressBar />);
     expect(screen.queryByLabelText('ProgressBar')).not.toBeInTheDocument();
   });
+
+  test('ProgressBar for folders renders correctly', () => {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    (global as any).document.createRange = (): unknown => ({
```
nicarl, line 111: What is this code doing exactly? It is used in used several times in this file, can you refactor it ?

nicarl (approved):

src/Frontend/Components/ProgressBar/FolderProgressBar.tsx
```diff
@@ -0,0 +1,214 @@
...
+  } else {
+    displayedProgressBarData = null;
+  }
+
+  const onProgressBarClick = useOnProgressBarClick(
+    displayedProgressBarData?.resourcesWithNonInheritedSignalOnly || []
+  );
+
+  return (
+    <div className={classes.root}>
```
leslielazzarino, line 141: I would extract this visual part to a component, that is shared between the old progress bar and the folder one.

leslielazzarino (approved):

PR #540, opened 2022-03-23
Fix bug that changes disappear after context menu

mayayunx:

### Summary of changes
Fixed bug that changes for one attribution disappear after delete or confirm another attribution via context menu in audit and attribution view.

Changes should not be discarded 

The changes can be tested in the UI

### DIFF
--- a/src/Frontend/Components/PackageCard/PackageCard.tsx
+++ b/src/Frontend/Components/PackageCard/PackageCard.tsx
@@ -165,7 +165,13 @@ export function PackageCard(props: PackageCardProps): ReactElement | null {
   function getContextMenuItems(): Array<ContextMenuItem> {
     function openConfirmDeletionPopup(): void {
       if (isPreselected) {
-        dispatch(deleteAttributionAndSave(selectedResourceId, attributionId));
+        dispatch(
+          deleteAttributionAndSave(
+            selectedResourceId,
+            attributionId,
+            selectedAttributionId
+          )
+        );
       } else {
         dispatch(openPopup(PopupType.ConfirmDeletionPopup, attributionId));
       }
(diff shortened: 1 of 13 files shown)


src/Frontend/Components/PackageCard/PackageCard.tsx
```diff
@@ -165,15 +165,28 @@ export function PackageCard(props: PackageCardProps): ReactElement | null {
   function getContextMenuItems(): Array<ContextMenuItem> {
```
leslielazzarino, line 165: It wasn't in the task description, but probably it makes sense to fix the behavior also for replacing an attribution.

leslielazzarino (changes requested):

JakobSchubert:
We also need to fix the same bug for the case that we delete multiple selected attributions via the context menu "Delete selected globally".

JakobSchubert (changes requested):
See my comment above (the multi-select case is missing).

src/Frontend/state/actions/resource-actions/__tests__/save-actions.test.ts
```diff
@@ -697,6 +702,200 @@ describe('The savePackageInfo action', () => {
...
+    const testResources: Resources = {
+      root: { src: { 'something.js': 1 }, 'somethingElse.js': 1 },
+    };
+    const testManualAttributions: Attributions = {
+      [testUuidA]: {
+        packageVersion: '1.0',
+        packageName: 'test Package',
+        licenseText: ' test License text',
+      },
+      [testUuidB]: {},
```
leslielazzarino, line 718: Please do not use empty attributions, they are not realistic.

src/Frontend/state/actions/resource-actions/__tests__/save-actions.test.ts
```diff
@@ -697,6 +702,200 @@ describe('The savePackageInfo action', () => {
...
+      attributionConfidence: DiscreteConfidence.High,
+    };
+    const testResources: Resources = {
+      'something.js': 1,
+      'somethingElse.js': 1,
+    };
+    const testInitialManualAttributions: Attributions = {
+      uuid1: testPackageInfo,
+      toReplaceUuid: { packageName: 'Vue' },
+      uuid2: {},
```
leslielazzarino, line 768: here too

leslielazzarino (changes requested):

JakobSchubert (approved):

JakobSchubert (approved):
Nice!

leslielazzarino (approved):

PR #563, opened 2022-03-29
Update electron builder

benedikt-richter:
### Summary of changes

Upgrade version of `electron-builder`

The build failed on macOS for v. 12.3

`yarn ship`


nicarl (approved):
Tried on Linux (Manjaro) and works well

leslielazzarino (approved):

Issue #581, opened 2022-04-05
Signals in Folder content not shown with ORT test file


leslielazzarino:
### What is wrong
When opening a large file the `Signals in Folder Contents` tab does not show any content even though there are signals further down the tree.
Moreover, an error message is logged:
```
Web worker error in workers context provider:  DOMException: Failed to execute 'postMessage' on 'Worker': Data cannot be cloned, out of memory.
```

### What is expected
The `Signals in Folder Contents` tab should show some content.
How it looks:
![bug_opossum](https://user-images.githubusercontent.com/26271333/161760556-0ceb642f-a5a0-4c95-ae8f-3b6fc4fdf5e0.png)
How it should look like:
![opossum_without_bug](https://user-images.githubusercontent.com/26271333/161761213-a108f687-f20d-403f-a712-62aeffa36b82.png)

### Steps to reproduce
Open https://github.com/opossum-tool/OpossumUI/blob/main/example-files/input_from_ort.json.gz and navigate to the root folder.

PR #582, opened 2022-04-05
Fix signals in folder content when loading ORT input file

leslielazzarino:

When ExternalData is too big posting it to the contained-external-attributions-accordion-worker at file loading fails. If the cachedExternalData in the worker was null, a null output will be returned posting to the worker for a resourceId. That will trigger the fallback synchronous calculation of the accordion content. But this process can fail if old data was cached in the worker when the data loading fails (including empty externalData when the app is started). 

### Summary of changes

Now whenever the externalData in the store change (i.e. when a file is opened) first the cached data in the worker are set to null, then the loading of the actual data is attempted.

Open the ORT test file, click in the root folder. The Signals in Folder Content accordion should be populated.

### DIFF
--- a/src/Frontend/Components/WorkersContextProvider/WorkersContextProvider.tsx
+++ b/src/Frontend/Components/WorkersContextProvider/WorkersContextProvider.tsx
@@ -25,6 +25,10 @@ export const AccordionWorkersContextProvider: FC = ({ children }) => {
   const externalData = useAppSelector(getExternalData);
   useMemo(() => {
     try {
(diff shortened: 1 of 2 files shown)


benedikt-richter:
the commit message should reference the respective issue number

nicarl:
> the commit message should reference the respective issue number

It should actually contain
`Fix: #581 `

benedikt-richter (approved):
works

nicarl (approved):
Works

PR #595, opened 2022-04-07
Experiment for replacing makeStyles in MUI 5

mayayunx:

### Summary of changes
- Experiment for replacing makeStyles with styled() and sx prop
- For Checkbox, ContextMenu and FilterMultiSelect

- makeStyles etc. are deprecated in MUI v5

- In the UI

### DIFF
--- a/src/Frontend/Components/AttributionColumn/AuditingSubPanel.tsx
+++ b/src/Frontend/Components/AttributionColumn/AuditingSubPanel.tsx
@@ -64,21 +64,18 @@ export function AuditingSubPanel(props: AuditingSubPanelProps): ReactElement {
     <MuiPaper className={classes.panel} elevation={0} square={true}>
       <div className={classes.displayRow}>
         <Checkbox
-          className={classes.checkBox}
           label={CheckboxLabel.FirstParty}
           disabled={!props.isEditable}
           checked={Boolean(props.displayPackageInfo.firstParty)}
           onChange={props.firstPartyChangeHandler}
         />
         <Checkbox
-          className={classes.checkBox}
           label={CheckboxLabel.FollowUp}
           disabled={!props.isEditable}
           checked={Boolean(props.displayPackageInfo.followUp)}
           onChange={props.followUpChangeHandler}
         />
         <Checkbox
-          className={classes.checkBox}
           label={CheckboxLabel.ExcludeFromNotice}
           disabled={!props.isEditable}
           checked={Boolean(props.displayPackageInfo.excludeFromNotice)}
--- a/src/Frontend/Components/Checkbox/Checkbox.tsx
+++ b/src/Frontend/Components/Checkbox/Checkbox.tsx
@@ -8,13 +8,15 @@ import React, { ReactElement } from 'react';
 import MuiTypography from '@mui/material/Typography';
 import { OpossumColors } from '../../shared-styles';
 import { styled } from '@mui/material/styles';
+
 interface CheckboxProps {
   label?: string;
   disabled?: boolean;
   checked: boolean;
   onChange(event: React.ChangeEvent<HTMLInputElement>): void;
   white?: boolean;
-  isMultiSelect?: boolean;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  component?: any;
 }
 
 const CheckboxComponent = styled('div')({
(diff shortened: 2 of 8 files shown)


src/Frontend/Components/ReportView/ReportView.tsx
```diff
@@ -64,7 +61,7 @@ export function ReportView(): ReactElement {
         attributionsWithResources={useFilters(attributionsWithResources)}
         isFileWithChildren={isFileWithChildren}
         onIconClick={getOnIconClick()}
-        topElement={<FilterMultiSelect className={classes.multiselect} />}
+        topElement={<FilterMultiSelect sx={{ maxWidth: '300px' }} />}
```
benedikt-richter, line 64: why is this changed here? Probably it's best to migrate file by file and not do partial migrations (except there are good reasons to do so).

src/Frontend/Components/AttributionColumn/AuditingSubPanel.tsx
```diff
@@ -64,21 +64,18 @@ export function AuditingSubPanel(props: AuditingSubPanelProps): ReactElement {
     <MuiPaper className={classes.panel} elevation={0} square={true}>
       <div className={classes.displayRow}>
         <Checkbox
-          className={classes.checkBox}
```
benedikt-richter, line 67: this removes the following styling
```

  checkBox: {
    height: 40,
    display: 'flex',
    alignItems: 'center',
    marginRight: 12,
    marginLeft: -2,
  },

```
Doesn't this have consequences?

src/Frontend/Components/Checkbox/Checkbox.tsx
```diff
@@ -5,34 +5,39 @@
...
+  alignItems: 'center',
+  marginRight: 12,
+  marginLeft: -2,
+});
+
+const MultiSelectCheckboxComponent = styled('div')({
+  height: 40,
+  marginTop: 1,
+});
+
```
benedikt-richter, line 32: not sure I'm happy about having this hardcoded here instead of providing as props. This makes the component less generic...

src/Frontend/Components/ReportView/ReportView.tsx
mayayunx: since we removed className from FilterMultiSelectProps

src/Frontend/Components/Checkbox/Checkbox.tsx
```diff
@@ -5,34 +5,39 @@
...
 export function Checkbox(props: CheckboxProps): ReactElement {
-  const classes = useStyles();
-  const whiteMode = clsx(props.white && classes.white);
+  const whiteMode = props.white ? { color: OpossumColors.white } : {};
+  const Component = props.isMultiSelect
```
benedikt-richter, line 35: This is an interesting case. We have to find a good pattern to attach styling from the outside (as props) to the root element. One solution is to use e.g. `MuiBox` and the `sx` property. There however the performance is not so good. Are you aware of another one?

mayayunx: https://mui.com/system/styled/#how-can-i-use-the-sx-syntax-with-the-styled-utility
This could be a solution, we can pass sx prop from parent to Checkbox and use the sx prop inside the styled component.

src/Frontend/Components/Checkbox/Checkbox.tsx
```diff
@@ -5,34 +5,34 @@
...
 
 interface CheckboxProps {
   label?: string;
   disabled?: boolean;
   checked: boolean;
   onChange(event: React.ChangeEvent<HTMLInputElement>): void;
-  className?: string;
   white?: boolean;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  component?: any;
```
leslielazzarino, line 19: This way a parent styled div can be passed. But the typing is a nightmare, I could not fix it quickly. If we decide to go with this solution it has to be fixed though.

src/Frontend/Components/PackageCard/PackageCard.tsx
```diff
@@ -361,7 +367,7 @@ export function PackageCard(props: PackageCardProps): ReactElement | null {
       <Checkbox
         checked={multiSelectSelectedAttributionIds.includes(attributionId)}
         onChange={handleMultiSelectAttributionSelected}
-        className={classes.multiSelectCheckbox}
+        component={MultiSelectCheckboxComponent}
```
leslielazzarino, line 370: Example of styled root component being passed on. Still not as flexible as passing a classname, that potentially just overwrites parts of the style.

src/Frontend/util/get-sx-from-props-and-classes.ts
```diff
@@ -0,0 +1,27 @@
...
+import { SystemStyleObject } from '@mui/system/styleFunctionSx';
+
+export type MuiSx = (
+  | boolean
+  | SystemStyleObject
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  | ((theme: {}) => SystemStyleObject)
+)[];
+
+export function getSxFromPropsAndClasses({
```
leslielazzarino, line 16: If we decide to use this function it needs to be properly tested, also in order to document what it does.

leslielazzarino:
Proper implementation will be done in other PRs, this one was just to try out different solutions.

PR #612, opened 2022-04-13
Update react-query to version 4

nicarl:
### Summary of changes
Update `react-query` to version 4. 

We recently updated to `react` version 18. When updating the `react` types we saw some errors associated with `react-query` and it turns out that version 3 does not support `react` 18. 

Existing unit tests should cover the basics, I tried the "package search" feature manually.

### DIFF
--- a/src/Frontend/Components/PackageSearchPopup/PackageSearchPopup.tsx
+++ b/src/Frontend/Components/PackageSearchPopup/PackageSearchPopup.tsx
@@ -109,7 +109,7 @@ export function PackageSearchPopup(): ReactElement {
         search={currentSearchTerm}
         autoFocus={false}
       />
-      {isLoading ? (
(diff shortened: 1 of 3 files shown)


src/Frontend/Components/PackageSearchPopup/PackageSearchPopup.tsx
```diff
@@ -109,7 +110,7 @@ export function PackageSearchPopup(): ReactElement {
         search={currentSearchTerm}
         autoFocus={false}
       />
-      {isLoading ? (
+      {isLoading && Boolean(currentSearchTerm) ? (
```
nicarl, line 113: Not sure if the behavior changed or this is a bug. Will try to come up with a minimal example and eventually open a bug ticket.

leslielazzarino (approved):

PR #773, opened 2022-06-23
New command that runs only tests related to changes in the current repo

MarkusObendrauf:
### Summary of changes

Added the command `yarn test:changed`. This command uses the `--onlyChanged` flag to run only tests related to changes in the current repo.

The current test commands run all tests within the repo. This can be quite time-consuming (40 seconds for me). The new command speeds up the process significantly (around 2 seconds) by only running necessary tests.

Make some changes and run `yarn test:changed`.


JakobSchubert (approved):
Great idea to include that command. That will be very helpful for developers!

FlorianSchepers (approved):

PR #803, opened 2022-07-06
Use node v18

benedikt-richter:
### Summary of changes

Use node v18 instead of node v17.

v18 is the latest lts version of node. v17 is no longer maintained.

Build the app:
`yarn ship`
Run the app: 
`yarn start`

### DIFF
--- a/.github/workflows/build-and-e2e-test.yml
+++ b/.github/workflows/build-and-e2e-test.yml
@@ -33,7 +33,7 @@ jobs:
       - name: Install Node.js, NPM and Yarn
         uses: actions/setup-node@v1
         with:
-          node-version: 17
+          node-version: 18
       - name: Increase watches
         run: echo fs.inotify.max_user_watches=524288 | sudo tee -a /etc/sysctl.conf && sudo sysctl -p
         if: matrix.os == 'ubuntu-latest'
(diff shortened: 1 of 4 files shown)


JakobSchubert (approved):
Looks good. I was only able to test the building the Linux version on my Ubuntu (as I do not have wine installed for building the windows version and when trying to build the mac version, I get the error `Cannot find module 'dmg-license'` independent for both node v17 and v18).
`yarn start` worked.

nicarl (commented):
You could add the following to `package.json`:
```
 "engines" : { 
    "node" : ">=18.0.0"
  }
```
I think yarn will then enforce this version, so whenever you are running a different version yarn will complain and actually fail.

nicarl:
I had a quick look at the `tsconfig` and you could also update the `target`to `es2022`. I think [this one](https://github.com/tsconfig/bases/blob/main/bases/node18-strictest-esm.combined.json) example is a good reference.

benedikt-richter:
> You could add the following to `package.json`:
> 
> ```
>  "engines" : { 
>     "node" : ">=18.0.0"
>   }
> ```
> 
> I think yarn will then enforce this version, so whenever you are running a different version yarn will complain and actually fail.

This would force developers to use node 18 and therefore enforce consistency, therefore I like it. It on the other hand does not allow to use node 17 which would actually still work. From my perspective this would be OK.

leslielazzarino (approved):

PR #808, opened 2022-07-07
Changed '!=' to '!=='

FlorianSchepers:
Signed-off-by: Florian Schepers <florian.schepers@tngtech.com>

In AttributionView at multiple occations '!=' was used instead of '!=='. In order to be type safe the occurences of '!=' were changed to '!=='.


benedikt-richter (approved):
👍

PR #815, opened 2022-07-11
fix: Enable context isolation

nicarl:
### Summary of changes

- Activate context isolation
- Use a context bridge to expose the `ipcRenderer`
- Instead of exposing the full `ipcRenderer` API a subset of functions is exposed
    - for the ocassions where `.invoke` is called an API is defined, which describes the actual action, e.g. `openLink`
    - when listeners are used (`.on` and `.removeListener`) the respective part of the API is exposed but an allow list of channels is applied in the backend before passing anything through  

Deactivating context isolation and exposing `ipcRenderer` globally is discouraged. An attacker could access it from the frontend with `window.ipcRenderer` and would have a rather powerful API. Instead it is recommended to use a preload script where a subset of APIs is exposed.
Details can be found [here](https://www.electronjs.org/docs/latest/tutorial/context-isolation).

- run all tests
- more importantly open the app and check the actions invoking the `ipcRenderer` are working (saving files, opening links, ...)

### DIFF
--- a/src/shared/ipc-channels.ts
+++ b/src/shared/ipc-channels.ts
@@ -1,20 +1,24 @@
 // SPDX-FileCopyrightText: Meta Platforms, Inc. and its affiliates
 // SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+// SPDX-FileCopyrightText: Nico Carl <nicocarl@protonmail.com>
 //
 // SPDX-License-Identifier: Apache-2.0
 
 export enum IpcChannel {
   ExportFile = 'export-file',
+  OpenFile = 'open-file',
+  OpenLink = 'open-link',
+  SaveFile = 'save-file',
+  SendErrorInformation = 'send-error-information',
+}
+
+export enum AllowedFrontendChannels {
   ExportFileRequest = 'export-file-request',
   FileLoaded = 'file-loaded',
   Logging = 'logging',
-  OpenFile = 'open-file',
-  OpenLink = 'open-link',
   ResetLoadedFile = 'reset-loaded-file',
   RestoreFrontend = 'restore-frontend',
-  SaveFile = 'save-file',
   SaveFileRequest = 'save-file-request',
-  SendErrorInformation = 'send-error-information',
   ToggleHighlightForCriticalSignals = 'toggle-highlight-for-critical-signals',
   ShowSearchPopup = 'show-search-pop-up',
   ShowProjectMetadataPopup = 'show-project-metadata-pop-up',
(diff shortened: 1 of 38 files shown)


src/shared/shared-types.ts
```diff
@@ -187,3 +190,26 @@ export interface ExternalAttributionSources {
     priority: number;
   };
 }
+
+// eslint-disable-next-line  @typescript-eslint/no-explicit-any
+export type Listener = (event: IpcRendererEvent, ...args: any[]) => void;
```
benedikt-richter, line 195: `any[]` -> `Array<any>`

src/shared/ipc-channels.ts
```diff
@@ -21,3 +22,30 @@ export enum IpcChannel {
   ShowProjectStatisticsPopup = 'show-project-statistics-pop-up',
   SetBaseURLForRoot = 'set-base-url-for-root',
 }
+
```
benedikt-richter, line 25: For the channels I'd suggest to split the enum containing all channels into two
* frontend channels
* other channels

This could look like (maybe naming might be improved):

```

export enum IpcChannel {
  ExportFile = 'export-file',
  OpenFile = 'open-file',
  OpenLink = 'open-link',
  SaveFile = 'save-file',
  SendErrorInformation = 'send-error-information',
}

export enum AllowedFrontendChannels {
  ExportFileRequest = 'export-file-request',
  FileLoaded = 'file-loaded',
  Logging = 'logging',
  ResetLoadedFile = 'reset-loaded-file',
  RestoreFrontend = 'restore-frontend',
  SaveFileRequest = 'save-file-request',
  ToggleHighlightForCriticalSignals = 'toggle-highlight-for-critical-signals',
  ShowSearchPopup = 'show-search-pop-up',
  ShowProjectMetadataPopup = 'show-project-metadata-pop-up',
  ShowProjectStatisticsPopup = 'show-project-statistics-pop-up',
  SetBaseURLForRoot = 'set-base-url-for-root',
}


```

Then the checks in the backend
` if (allowedFrontendChannels.includes(channel)) {`
could read
` if (Object.values(AllowedFrontendChannels).includes(channel)) {`
and the constant
`export const allowedFrontendChannels`
would not be required.


This would avoid to add new channels to two enums in case of "allowed for frontend" and to the constant as well. Further, it would make the distinction between the two enums more explicit.

benedikt-richter (changes requested):
Great to takle the context isolation! Looks good to me and works.
I added an idea on how to refactor the enums for the IpcChannels.

src/shared/ipc-channels.ts
nicarl: Good idea I changed it according to your suggestion.

benedikt-richter (approved):

leslielazzarino (approved):

Issue #825, opened 2022-07-18
Follow-up export triggered multiple times


benedikt-richter:
Opening the example file `opossum_input.json` in OpossumUI and triggering the Follow-up export, the respective export is triggered multiple times
![Screenshot 2022-07-18 at 16 35 19](https://user-images.githubusercontent.com/46576389/179535810-f4707f21-51d7-4cf8-bf1c-504756a1912e.png)

Sometimes this causes the UI to be stuck in the following state:
![Screenshot 2022-07-18 at 16 35 40](https://user-images.githubusercontent.com/46576389/179535948-90faf3a3-89e2-4ffe-abb8-37d76aafd4f2.png)

This behavior seems to have been introduced in #815

PR #828, opened 2022-07-18
Fix removal of listeners

benedikt-richter:
### Summary of changes

Always remove all listeners of a channel.

When we introduced the contextBridge the removal of listeners broke, causing e.g. the export of follow-ups to be triggered multiple times.  For details see #825.

Carefully test file exports and saving operations (everything that relys on listeners).


leslielazzarino (approved):

antonbauhofer (approved):
Looks good to me.
I tested it in the app and it works like expected.

PR #843, opened 2022-07-26
The menu point "Highlight critical signals" grayed out if no critical signals

vasily-pozdnyakov:
### Summary of changes

If the input file has no critical signals (none of the "medium" and "high"), the menu point "Highlight critical signals" is disabled.

The menu point "Highlight critical signals" should be grayed out if none exists.

Open an input file without critical signals, e.g. `large_opossum_input.json` from example files.

### DIFF
--- a/src/ElectronBackend/input/importFromFile.ts
+++ b/src/ElectronBackend/input/importFromFile.ts
@@ -61,9 +61,8 @@ export async function loadJsonFromFilePath(
   }
 
   log.info('... Successfully parsed input file.');
-  const externalAttributions = parseRawAttributions(
-    parsingResult.externalAttributions
-  );
+  const [externalAttributions, inputContainsCriticalSignals] =
+    parseRawAttributions(parsingResult.externalAttributions);
 
   const manualAttributionFilePath = getFilePathWithAppendix(
     filePath,
@@ -84,7 +83,7 @@ export async function loadJsonFromFilePath(
 
   log.info(`Starting to parse output file ${manualAttributionFilePath} ...`);
   const opossumOutputData = parseOpossumOutputFile(manualAttributionFilePath);
-  const manualAttributions = parseRawAttributions(
+  const [manualAttributions] = parseRawAttributions(
     opossumOutputData.manualAttributions
   );
   log.info('... Successfully parsed output file.');
@@ -151,6 +150,7 @@ export async function loadJsonFromFilePath(
     spdxYamlFilePath: getFilePathWithAppendix(filePath, '.spdx.yaml'),
     spdxJsonFilePath: getFilePathWithAppendix(filePath, '.spdx.json'),
     projectTitle: parsingResult.metadata.projectTitle,
+    inputContainsCriticalSignals,
   };
   setGlobalBackendState(newGlobalBackendState);
 
--- a/src/ElectronBackend/main/listeners.ts
+++ b/src/ElectronBackend/main/listeners.ts
@@ -4,7 +4,7 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
-import { BrowserWindow, shell, WebContents } from 'electron';
+import { BrowserWindow, Menu, shell, WebContents } from 'electron';
 import { AllowedFrontendChannels } from '../../shared/ipc-channels';
 import {
   ExportArgsType,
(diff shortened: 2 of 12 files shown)


src/ElectronBackend/main/menu.ts
```diff
@@ -14,9 +14,18 @@ import {
 } from './notice-document-helpers';
 import { ExportType } from '../../shared/shared-types';
 
-export function createMenu(mainWindow: BrowserWindow): Menu {
+export function createMenu(
+  mainWindow: BrowserWindow,
+  criticalSignalsFlag: boolean
```
benedikt-richter, line 19: Can't we use the `GlobalBackendState` instead of explicitly parsing the flag?

benedikt-richter (changes requested):
The menu item is ticked even if it is disabled.

Screenshot 2022-07-27 at 08 52 33

antonbauhofer:
> The menu item is ticked even if it is disabled.
> 
> Screenshot 2022-07-27 at 08 52 33

In your Screenshot, the menu item is greyed out, but has a checkmark. The checkmark does not make sense, I think. If I test it myself, there is no checkmark, however. Has this already been solved and the screenshot is outdated?

src/ElectronBackend/input/importFromFile.ts
```diff
@@ -155,6 +155,7 @@ export async function loadJsonFromFilePath(
   setGlobalBackendState(newGlobalBackendState);
 
   log.info('File import finished successfully');
+  return criticalSignalsFlag;
```
antonbauhofer, line 158: We use the name `criticalSignalsFlag` quite often. I think we could use a more descriptive name. Maybe something along the lines of `loadedFileContainsCriticalSignals`.

antonbauhofer (commented):
It looks good all together. But I left some smaller comments to consider. No strong opinions, just suggestions.

benedikt-richter:
> In your Screenshot, the menu item is greyed out, but has a checkmark. The checkmark does not make sense, I think. If I test it myself, there is no checkmark, however. Has this already been solved and the screenshot is outdated?

I also think this checkmark should not be there. This happens if the menu item is checked and then you open a new file that does not have critical signals

src/ElectronBackend/main/menu.ts
```diff
@@ -16,6 +16,23 @@ import { ExportType } from '../../shared/shared-types';
...
+    : false;
+
+  if (ischecked && !inputContainsCriticalSignals) {
+    ischecked = false;
+    webContents.send(
+      AllowedFrontendChannels.ToggleHighlightForCriticalSignals,
+      {
+        toggleHighlightForCriticalSignals: true,
+      }
+    );
```
benedikt-richter, line 34: I'd suggest to extract the logic as a function, like

```

function getCheckedStatusAndSyncWithFrontend(
  inputContainsCriticalSignals: boolean,
  webContents: WebContents
): boolean {
  let isHighlightCriticalSignalMenuItemChecked = Boolean(
    Menu.getApplicationMenu()?.getMenuItemById('highlightCritical')?.checked
  );

  if (
    isHighlightCriticalSignalMenuItemChecked &&
    !inputContainsCriticalSignals
  ) {
    webContents.send(
      AllowedFrontendChannels.ToggleHighlightForCriticalSignals,
      {
        toggleHighlightForCriticalSignals: true,
      }
    );
    isHighlightCriticalSignalMenuItemChecked = false;
  }

  return isHighlightCriticalSignalMenuItemChecked;
}

```

such that in `createMenu` we just have to call it

```
...

  const inputContainsCriticalSignals = Boolean(
    getGlobalBackendState().inputContainsCriticalSignals
  );
  const isHighlightCriticalSignalMenuItemChecked =
    getCheckedStatusAndSyncWithFrontend(
      inputContainsCriticalSignals,
      webContents
    );

...

```

src/ElectronBackend/main/listeners.ts
```diff
@@ -137,6 +138,8 @@ export async function openFile(
     setTitle(mainWindow, filePath);
   } finally {
     loadingWindow.close();
+    mainWindow.removeMenu();
+    Menu.setApplicationMenu(createMenu(mainWindow));
```
benedikt-richter, line 142: This code should be moved into the `try` block

benedikt-richter (changes requested):
Works nicely 👍 

Just a small refactor and moving code out of the `finally` block

benedikt-richter (approved):

antonbauhofer (approved):

PR #849, opened 2022-07-29
Use @tanstack/react-query 4.0.0

benedikt-richter:
### Summary of changes

Use @tanstack/react-query 4.0.0

react-query is deprecated: https://www.npmjs.com/package/react-query


MarkusObendrauf (approved):

antonbauhofer (approved):
Runs as expected

PR #873, opened 2022-08-11
Add a search to the attribution column in the audit view.

MarkusObendrauf:
### Summary of changes

Add a search functionality for signals. This is accessed by clicking on the search icon, which is located next to the tabs. This opens a textbox that allows a user to search attributions by package name, license name and copyright.

As a user I would like to search attributions for problematic licenses.


MarkusObendrauf:
keycloak_11 0 2_resources json 2022-08-10 16_17_38

vasily-pozdnyakov:
Some points regarding the visual:
1.  When no signals found, the user can still see sources titles, this is confusing: 
image

2. User can close the search window when there is some text and move to another file/folder/view, but the filtering is still active: 
image
you can either disable closing the search field when it is not empty or automatically clear it when it is closing.

3. It would be nice to make some margin between tabs local/global and the search field, as the text "Search" covers the LOCAL tab. 
image

src/Frontend/Components/PackageList/AttributionsViewPackageList.tsx
```diff
@@ -0,0 +1,50 @@
...
+const CARD_VERTICAL_DISTANCE = 41;
+
+interface AttributionsViewPackageListProps {
+  attributions: Attributions;
+  attributionIds: Array<string>;
+  getAttributionCard(attributionId: string): ReactElement | null;
+  max: NumberOfDisplayedItems | Height;
+}
+
+export function AttributionsViewPackageList(
```
vasily-pozdnyakov, line 22: This function is 80% identical with PackageList function, can we somehow merge them?

src/Frontend/Components/PackagePanel/PackagePanel.tsx
```diff
@@ -107,37 +172,23 @@ export function PackagePanel(
       {sortedSources.map((sourceName) => (
         <div key={`PackageListForSource-${sourceName}`}>
           {sourceName ? (
-            <MuiTypography variant={'body2'}>
-              {prettifySource(sourceName, attributionSources)}
-            </MuiTypography>
+            <>
```
vasily-pozdnyakov, line 175: Do we need the additional <></>? I think it is not necessary, as we use only one component MuiTypography.

src/Frontend/Components/ResourceDetailsTabs/ResourceDetailsTabs.tsx
```diff
@@ -88,8 +112,24 @@ export function ResourceDetailsTabs(
     [Tabs.Global]: 'Global',
   };
 
+  function onSearchToggleClick(): void {
+    dispatch(toggleAccordionSearchField());
+  }
+
+  const isAccordionSearchFieldDisplayed = useAppSelector(
```
vasily-pozdnyakov, line 119: Can we move all the constants closer to the top?

src/Frontend/Components/PackageList/PackageList.tsx
```diff
@@ -26,102 +20,37 @@ const classes = {
...
-  isExternalAttribution: boolean;
-  preSelectedExternalAttributionIdsForSelectedResource: Array<string>;
-  isAddToPackageEnabled: boolean;
-  selectedResourceId: string;
+  attributionIds: Array<string>;
+  getAttributionCard(attributionId: string): ReactElement | null;
+  maxNumberOfDisplayedItems: number;
 }
 
 export function PackageList(props: PackageListProps): ReactElement {
```
vasily-pozdnyakov, line 29: Should we test this function?

vasily-pozdnyakov (changes requested):
Nice job.

src/Frontend/shared-styles.ts
```diff
@@ -63,6 +63,21 @@ export const disabledIcon = {
   color: OpossumColors.disabledButtonGrey,
 };
 
+export const largeIcon = {
```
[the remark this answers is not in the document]
MarkusObendrauf: fixed

src/Frontend/Components/PackageList/AttributionsViewPackageList.tsx
MarkusObendrauf: I discussed this with Benedikt. Context: the PackageList is used in the tabs in the audit view, while the AttributionsViewPackageList is used in the attributions view. If in the future we will make changes to both lists at once, then its better to merge them. If there will be more differences between the two components, then its better to keep them separate. Right now there are already some slight differences between these components, especially with how the search works, and we might make changes to the attibution view, so in the end we decided to keep them separate.

src/Frontend/Components/PackagePanel/PackagePanel.tsx
MarkusObendrauf: fixed

src/Frontend/Components/ResourceDetailsTabs/ResourceDetailsTabs.tsx
MarkusObendrauf: good point, moved it. there were some more `useAppSelector`s higher up, it looks much better when they're all next to each other

src/Frontend/Components/PackageList/PackageList.tsx
MarkusObendrauf: i believe it is sufficiently tested in `src\Frontend\Components\ResourceDetailsTabs\__tests__\ResourceDetailsTabs.test.tsx` (tests for the parent component)

src/Frontend/shared-styles.ts
```diff
@@ -63,6 +63,21 @@ export const disabledIcon = {
...
+  width: '26px',
+  height: '26px',
+  padding: '2px',
+  margin: '0 2px',
+};
+
+export const largeClickableIcon = {
```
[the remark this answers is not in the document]
MarkusObendrauf: not sure, the question is if we will use this style again in the future

leslielazzarino (approved):

src/Frontend/shared-styles.ts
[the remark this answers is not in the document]
MarkusObendrauf: ok

MarkusObendrauf:
Still working on Vasily's first comment

benedikt-richter:
Would be nice to indicate the user that there are no results in case that there are no results. Possible using some text.

Screenshot 2022-08-16 at 09 37 07

benedikt-richter:
Is it intended that the search is restored when navigating from audit view to attribution view and back?

benedikt-richter:
I fear it is not transparent for the user how search works. Which fields are included in search and which are not. For example, the user might try to search for a version number and might get zero results even if that version is included in the list of packages.

benedikt-richter:
It would be nice to have a "clear search" button, like we have on attribution view

vasily-pozdnyakov (commented):
The code looks nice, please check the comments from Benedikt.

vasily-pozdnyakov (approved):

Issue #759, opened 2022-06-20
Multiple popup windows


vasily-pozdnyakov:
User can open multiple pop up windows (e.g. "Show Project Metadata") at the same time one above the others and have to close them one by one later.

benedikt-richter:
In principle, this is a intended feature. It is, for example, used when the EDIT popup in REPORT view is open, data is modified and subsequently the CANCEL button is hit. In this case the UNSAVED CHANGES ppoup is opened on top of the EDIT popup.
Do you have specific use cases in mind where this should not be possible?

vasily-pozdnyakov:
I think it is ok to open different popups one above other, but maybe it should not be possible to open multiple exactly the same popups (e.g. multiple press CTRL+F opens multiple search windows with exactly the same content).

benedikt-richter:
Ok, I understand. That's a good point! That should not be allowed to happen.

PR #882, opened 2022-08-19
Allow only one popup of a kind

benedikt-richter:
### Summary of changes

Fix bug of allowing for several instances of a single popup kind to be open.

Fix: #759

Open the same popup several times. E.g. press ctrl+F to open search popup several times and observe that now it closes once you click away once.

### DIFF
--- a/src/Frontend/state/reducers/view-reducer.ts
+++ b/src/Frontend/state/reducers/view-reducer.ts
@@ -57,9 +57,13 @@ export function viewState(
         popupInfo: state.popupInfo.slice(0, -1),
       };
     case ACTION_OPEN_POPUP:
+      const openPopups = state.popupInfo.map((popupInfo) => popupInfo.popup);
+      const newPopupInfo = openPopups.includes(action.payload.popup)
+        ? state.popupInfo
+        : state.popupInfo.concat(action.payload);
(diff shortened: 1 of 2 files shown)


vasily-pozdnyakov (approved):
well done.

src/Frontend/state/reducers/view-reducer.ts
```diff
@@ -57,9 +57,13 @@ export function viewState(
         popupInfo: state.popupInfo.slice(0, -1),
       };
     case ACTION_OPEN_POPUP:
+      const openPopups = state.popupInfo.map((popupInfo) => popupInfo.popup);
```
leslielazzarino, line 60: Shouldn't the logic be that we just want to show one popup, no matter the type?

leslielazzarino (changes requested):

src/Frontend/state/reducers/view-reducer.ts
vasily-pozdnyakov: @leslielazzarino Please check the comment https://github.com/opossum-tool/OpossumUI/issues/759#issuecomment-1174785709

leslielazzarino (approved):

Issue #740, opened 2022-06-13
Package card for first party attributions without content is empty


benedikt-richter:
When opening an input file that has signals that consist of only a first party flag, the corresponding package card displays the "first party icon" but no text (see screenshot). Instead it should display some standard placeholder text like e.g. "First Party" or "is first party".

![Screenshot 2022-06-13 at 11 28 28](https://user-images.githubusercontent.com/46576389/173323640-4540448e-8fee-4cc5-9bb8-d4303405703c.png)

PR #883, opened 2022-08-19
Display text if only first party flag

benedikt-richter:
### Summary of changes

In the past we did not display any text on package cards that had only the first party label set. Now a respective text is displayed in this case.

Fix: #740

Before:
Screenshot 2022-08-19 at 16 58 27

After:
Screenshot 2022-08-19 at 16 58 07


antonbauhofer (approved):
Looks good to me

leslielazzarino (approved):

PR #899, opened 2022-08-25
Add opening pull requests as trigger for Github workflow runs

JakobSchubert:
### Summary of changes

Add pull requests as triggers for Github workflow runs.

Before, the tests in the CI ran only on push. The issue was that tests were not running for pull requests from forks.

Check that tests are executed on opening of pull requests (after merging this PR).

Note: Please review the [guidelines for contributing](../CONTRIBUTING.md) to this repository.

### DIFF
--- a/.github/workflows/reuse.yml
+++ b/.github/workflows/reuse.yml
@@ -5,7 +5,10 @@
 
 name: REUSE Compliance Check
 
-on: [push, pull_request]
+on:
+  push:
+  pull_request:
+    types: [opened, reopened, synchronize]
 
 jobs:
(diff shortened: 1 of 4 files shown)


.github/workflows/reuse.yml
```diff
@@ -5,7 +5,10 @@
 
 name: REUSE Compliance Check
 
-on: [push, pull_request]
```
JakobSchubert, line 8: I made this change just to be consistent with the other files.

FlorianSchepers (approved):
lgtm

PR #913, opened 2022-08-30
Use locator in e2e tests

benedikt-richter:
### Summary of changes

Use locator in e2e tests

The previous pattern that was used is discouraged.

### DIFF
--- a/src/e2e-tests/__tests__/e2e-loading-large-file.test.ts
+++ b/src/e2e-tests/__tests__/e2e-loading-large-file.test.ts
@@ -4,7 +4,11 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { ElectronApplication, Page } from 'playwright';
-import { E2E_LARGE_TEST_TIMEOUT, getApp } from '../test-helpers/test-helpers';
+import {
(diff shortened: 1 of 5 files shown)


leslielazzarino (approved):

src/e2e-tests/__tests__/e2e-loading-large-file.test.ts
```diff
@@ -26,6 +30,6 @@ describe('Open large zipped file via command line', () => {
 
   // The test was flaky and is now disabled.
```
vasily-pozdnyakov, line 31: Is it still so?

vasily-pozdnyakov (approved):

src/e2e-tests/__tests__/e2e-loading-large-file.test.ts
benedikt-richter: I'd expect so. Still one could look into that one ...

vasily-pozdnyakov: Maybe we should add it to the issues.

PR #923, opened 2022-08-31
Display text if only first party flag on attribution view

benedikt-richter:
### Summary of changes

Display text if only first party flag on attribution view.

In the past we did not display any text on package cards that had only the
 first party label set on attribution view. Now a respective text is displayed in this case
 also on on attribution view.


src/Frontend/Components/AttributionList/AttributionList.tsx
```diff
@@ -42,48 +41,29 @@ export function AttributionList(props: AttributionListProps): ReactElement {
...
-    }
-
     function onClick(): void {
       props.onCardClick(attributionId);
     }
 
-    const cardConfig: ListCardConfig = {
+    const cardConfig: PackageCardConfig = {
       isSelected: isSelected(),
-      isMarkedForReplacement: isMarkedForReplacement(),
```
benedikt-richter, line 61: this was overwritten later anyway

src/Frontend/util/get-card-labels.ts
```diff
@@ -3,97 +3,113 @@
...
 
 const PRIORITIZED_PACKAGE_INFO_ATTRIBUTES: Array<
-  'name' | 'copyright' | 'licenseName' | 'licenseText' | 'comment' | 'url'
-> = ['name', 'copyright', 'licenseName', 'licenseText', 'comment', 'url'];
+  | 'packageName'
+  | 'copyright'
+  | 'licenseName'
+  | 'licenseText'
+  | 'comment'
+  | 'url'
```
MarkusObendrauf, line 14: Could we replace this complicated type with `keyof PackageInfo`? Or otherwise extract it. Right now it doesn't look nice and the logic is duplicated.

MarkusObendrauf (approved):

leslielazzarino (approved):

PR #954, opened 2022-09-09
Replace library for md5

benedikt-richter:
### Summary of changes

The library calculating md5 hashes is replaced. 

The former one caused the application to crash on opening particular files.


maxhbr (approved):
Code looks good, but not tested

vasily-pozdnyakov (approved):
looks nice

PR #959, opened 2022-09-09
Add input file with duplicated licenses

vasily-pozdnyakov:
We want to merge duplicated licenses from project statistics table, therefore we need an example input file, which is closer to reality (e.g. contains extra spaces and dashes in licenses names).


benedikt-richter (changes requested):
It would be good to include the new data into our "standard" example file `opossum_input.json` instead of creating a new example file.

PR #969, opened 2022-09-12
Fix opening app via cli

benedikt-richter:
### Summary of changes

* extract required logic for opening a file into `handleOpeningFile`
* use `handleOpeningFile` when opening a file via cli

When starting the app via cli the backend state was not updated correctly such that the app failed when attempting to save.

E.g. replace line 87 of `package.json` by 

```  
  "start:darwin:linux": "concurrently \"yarn build:dev; BROWSER=none ESLINT_NO_DEV_ERRORS=true react-scripts start\" \"wait-on http://127.0.0.1:3000 && electron . example-files/opossum_input.json\"",
```

and run `yarn start`.


maxhbr (approved):
LGTM

leslielazzarino (approved):

Issue #938, opened 2022-09-07
Terminal is flooded with warnings when running tests 


benedikt-richter:
When running tests, e.g. using `yarn test:unit` the terminal is flooded with warnings like:

```
  Warning: An update to $Component inside a test was not wrapped in act(...).
      
      When testing, code that causes React state updates should be wrapped into act(...):
      
      act(() => {
        /* fire events that update state */
      });
      /* assert on the output */
```

Some of these warnings can be fixed by wrapping the `testStore.dispatch(...)` into an `act(...)` in the tests. However, not all of these warnings are fixable by this approach.

The goal of the current ticket is to get rid of the warnings.

 Note: this might be related to our upgrade to React 18.

benedikt-richter:
Superceded by #1273

PR #984, opened 2022-09-13
Suppress console output in tests

benedikt-richter:
### Summary of changes

Mock `console.error` for the relevant cases by matching substrings.

Currently, the terminal is spammed when running tests. Here we suppress output to the terminal related to issue #938.

`yarn test:unit`

### DIFF
--- a/src/Frontend/Components/ResourceDetailsTabs/__tests__/ResourceDetailsTabs.test.tsx
+++ b/src/Frontend/Components/ResourceDetailsTabs/__tests__/ResourceDetailsTabs.test.tsx
@@ -136,7 +136,6 @@ describe('The ResourceDetailsTabs', () => {
         })
       )
     );
-    console.log(store.getState().resourceState.allViews.manualData);
     act(() => {
       store.dispatch(setSelectedResourceId('/root/fileWithoutAttribution'));
     });
--- a/setupTests.ts
+++ b/setupTests.ts
@@ -7,10 +7,23 @@
 import '@testing-library/jest-dom/extend-expect';
 
 const TEST_TIMEOUT = 15000;
+const SUBSTRINGS_TO_SUPPRESS_IN_CONSOLE_INFO = [
+  'Web worker error in workers context',
+  'Error in rendering folder progress bar',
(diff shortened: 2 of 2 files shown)


MarkusObendrauf:
With these changes, errors now point to `originalConsoleError.call(console, args)` which was added to `setupTests.tx`.  Example:
```  ● Console

    console.error
      [
        "Warning: ReactDOM.render is no longer supported in React 18. Use createRoot instead. Until you switch to the new API, your app will behave as if it's running React 17. Learn more: https://reactjs.org/link/switch-to-createroot"
      ]

      72 |     }
      73 |     return type === 'error'
    > 74 |       ? originalConsoleError.call(console, args)
         |                              ^
      75 |       : originalConsoleInfo.call(console, args);
      76 |   };
      77 | }

      at setupTests.ts:74:30
      at console.error (node_modules/@testing-library/react-hooks/lib/core/console.js:19:7)
      at printWarning (node_modules/react-dom/cjs/react-dom.development.js:86:30)
      at error (node_modules/react-dom/cjs/react-dom.development.js:60:7)
      at Object.render (node_modules/react-dom/cjs/react-dom.development.js:29670:5)```

Notice that the error message is pointing to `setupTests.ts` instead of the real source of the error. It would be nicer to keep the original error if possible.

File: src/Frontend/Components/ResourceDetailsTabs/__tests__/ResourceDetailsTabs.test.tsx [hunk 1]
```diff
@@ -136,7 +136,6 @@ describe('The ResourceDetailsTabs', () => {
         })
       )
     );
-    console.log(store.getState().resourceState.allViews.manualData);
```
leslielazzarino, line 139: Nice catch.

leslielazzarino (approved):

benedikt-richter:
> 

Good point. I'll see whether there is an easy fix. Should not be of huge impact, though, as the stack traces for failing tests are preserved.

MarkusObendrauf (approved):

PR #992, opened 2022-09-16
Optimize accordion web worker memory usage

leslielazzarino:
### Summary of changes

attributionToResources fields in AttributionData can be huge. As hey are not used by the accordion web workers now they are not sent to them, making the memory copy operations faster and increasing the threshold for out of memory errors.

See above.


src/Frontend/Components/AggregatedAttributionsPanel/AggregatedAttributionsPanel.tsx
```diff
@@ -55,10 +56,15 @@ export function AggregatedAttributionsPanel(
     [selectedResourceId, externalData, resolvedExternalAttributions]
   );
 
+  const manualDataToGetContained: AttributionDataToGetContained = {
```
benedikt-richter, line 59: Not sure I understand the variable name `manualDataToGetContained`.

benedikt-richter (commented):
Looks good. I'd just rename `manualDataToGetContained` to something like `PackagePanelData` or similar to emphasize more what it is used for than where it comes from.

benedikt-richter (changes requested):
I'd also rename `AttributionDataToGetContained` accordingly

benedikt-richter (approved):

vasily-pozdnyakov (approved):

Issue #947, opened 2022-09-08
Add table showing critical licenses


benedikt-richter:
Add a table showing critical licenses (clustered by criticality high or medium).  The table should have two columns: One should contain the license name and the other one total number. 
Make sure to use deduplicated data as described in #943.

someshkhandelia:
Currently the statistics popup looks like this:
![image](https://user-images.githubusercontent.com/9270691/191931782-3f5719bc-73b3-4395-b52d-ce42344e733e.png)
After a verbal discussion with @vasily-pozdnyakov and @Billwuzl, it was understood that the goal of this issue is to essentially extract a table, out of the **Signals per Sources** table, that contains the **high** (red) and **medium** (orange) critical license names and their **total** numbers.

benedikt-richter:
Exactly, this would be the goal. In the long run one could also image to merge both tables into a more complex one with sorting and filtering. However, for the moment a new one like the one you describe would be nice.

Issue #1011, opened 2022-09-23
Clean up project statistics popup


vasily-pozdnyakov:
After we introduce new graphs, tables and styling on project statistic popup it would be nice to finalize and clean up:
* check that styling is consistent
* check if there are any bugs or any things to improve
* refactor and clean up the code e.g. remove (or merge) duplicating styles...

someshkhandelia:
@benedikt-richter gave a valuable suggestion:
We can have a separate table cell component for License Names, since they require different colors for criticality and are always for example left-aligned

benedikt-richter:
In my opinion, it also would be an improvement to not display table header in ALL_CAPS. To me this looks a bit odd:

![Image](https://user-images.githubusercontent.com/46576389/192505835-2ec643ed-56ce-4db1-a60f-8bfd86aa47ae.png)

PR #1013, opened 2022-09-23
Add critical licenses table to statistics popup

someshkhandelia:
### Summary of changes

Add critical licenses table to statistics popup

- The table shows only high and medium critical licenses.
- The high critical licenses are placed above the medium critical ones.
- Both groups are sorted within themselves.

Fix: #947

### DIFF
--- a/src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
+++ b/src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
@@ -1,0 +1,193 @@
+// SPDX-FileCopyrightText: Meta Platforms, Inc. and its affiliates
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+
+import React, { ReactElement } from 'react';
+import {
+  AttributionCountPerSourcePerLicense,
+  LICENSE_COLUMN_NAME_IN_TABLE_HEADER,
+  LicenseNamesWithCriticality,
+  PLACEHOLDER_ATTRIBUTION_COUNT,
+  SOURCE_TOTAL_HEADER,
+  TOTAL_COLUMN_NAME_IN_TABLE_HEADER,
+} from './project-statistics-popup-helpers';
+import MuiTypography from '@mui/material/Typography';
+import MuiBox from '@mui/material/Box';
+import MuiTable from '@mui/material/Table';
+import MuiTableBody from '@mui/material/TableBody';
+import MuiTableCell from '@mui/material/TableCell';
+import MuiTableContainer from '@mui/material/TableContainer';
+import MuiTableHead from '@mui/material/TableHead';
+import MuiTableFooter from '@mui/material/TableFooter';
+import MuiTableRow from '@mui/material/TableRow';
+import { projectStatisticsPopupClasses } from './shared-project-statistics-popup-styles';
+import { OpossumColors } from '../../shared-styles';
+import { Criticality } from '../../../shared/shared-types';
+
+function getLicenseNamesByCriticality(
+  allLicensesWithCriticality: Array<{
+    licenseName: string;
+    criticality: Criticality | undefined;
+  }>,
+  criticality: Criticality
+): Array<string> {
+  return allLicensesWithCriticality
+    .map((licenseNameAndCriticality) =>
+      licenseNameAndCriticality.criticality === criticality
+        ? licenseNameAndCriticality.licenseName
+        : ''
+    )
+    .filter((licenseName) => licenseName !== '');
+}
+
+function getCriticalLicenseNamesWithTheirTotalAttributions(
+  attributionCountPerSourcePerLicense: AttributionCountPerSourcePerLicense,
+  criticalLicenseNames: Array<string>
+): Array<{ licenseName: string; totalNumberOfAttributions: number }> {
+  const licenseNamesAndTheirTotalAttributions = criticalLicenseNames.map(
+    (criticalLicenseName) => {
+      return {
+        licenseName: criticalLicenseName,
+        totalNumberOfAttributions:
+          attributionCountPerSourcePerLicense[criticalLicenseName][
+            SOURCE_TOTAL_HEADER
+          ],
+      };
+    }
+  );
+  return licenseNamesAndTheirTotalAttributions.sort();
+}
+
+function getTotalNumberOfAttributions(
+  licenseNamesAndTheirTotalAttributions: Array<{
+    licenseName: string;
+    totalNumberOfAttributions: number;
+  }>
+): number {
+  return licenseNamesAndTheirTotalAttributions
+    .map(
+      (licenseNameAndTotalAttributions) =>
+        licenseNameAndTotalAttributions.totalNumberOfAttributions
+    )
+    .reduce((total, value) => total + value, 0);
+}
+
+interface CriticalLicensesTableProps {
+  attributionCountPerSourcePerLicense: AttributionCountPerSourcePerLicense;
+  licenseNamesWithCriticality: LicenseNamesWithCriticality;
+  title: string;
+}
+
+export function CriticalLicensesTable(
+  props: CriticalLicensesTableProps
+): ReactElement {
+  const allLicensesWithCriticality = Object.entries(
+    props.licenseNamesWithCriticality
+  ).map((licenseNameAndCriticality) => {
+    return {
+      licenseName: licenseNameAndCriticality[0],
+      criticality: licenseNameAndCriticality[1],
+    };
+  });
+  const highCriticalityLicenseNames = getLicenseNamesByCriticality(
+    allLicensesWithCriticality,
+    Criticality.High
+  );
+  const mediumCriticalityLicenseNames = getLicenseNamesByCriticality(
+    allLicensesWithCriticality,
+    Criticality.Medium
+  );
+  const highCriticalityLicensesTotalAttributions =
+    getCriticalLicenseNamesWithTheirTotalAttributions(
+      props.attributionCountPerSourcePerLicense,
+      highCriticalityLicenseNames
+    );
+  const mediumCriticalityLicensesTotalAttributions =
+    getCriticalLicenseNamesWithTheirTotalAttributions(
+      props.attributionCountPerSourcePerLicense,
+      mediumCriticalityLicenseNames
+    );
+  const criticalLicensesTotalAttributions =
+    highCriticalityLicensesTotalAttributions.concat(
+      mediumCriticalityLicensesTotalAttributions
+    );
+  const tableColumnNames = [
+    LICENSE_COLUMN_NAME_IN_TABLE_HEADER,
+    TOTAL_COLUMN_NAME_IN_TABLE_HEADER,
+  ];
+  const tableFooter = [SOURCE_TOTAL_HEADER].concat(
+    getTotalNumberOfAttributions(criticalLicensesTotalAttributions).toString()
+  );
+
+  return (
+    <MuiBox>
+      <MuiTypography variant="subtitle1">{props.title}</MuiTypography>
+      <MuiTableContainer
+        sx={projectStatisticsPopupClasses.criticalLicensesTable}
+      >
+        <MuiTable size="small" stickyHeader>
+          <MuiTableHead>
+            <MuiTableRow>
+              {tableColumnNames.map((columnName, index) => (
+                <MuiTableCell
+                  sx={projectStatisticsPopupClasses.head}
+                  key={index}
+                  align={index === 0 ? 'left' : 'center'}
+                >
+                  {columnName.toUpperCase()}
+                </MuiTableCell>
+              ))}
+            </MuiTableRow>
+          </MuiTableHead>
+
+          <MuiTableBody>
+            {criticalLicensesTotalAttributions.map(
+              ({ licenseName, totalNumberOfAttributions }, rowIndex) => (
+                <MuiTableRow key={rowIndex}>
+                  {tableColumnNames.map((columnName, index) => (
+                    <MuiTableCell
+                      sx={{
+                        ...projectStatisticsPopupClasses.body,
+                        ...(index === 0
+                          ? props.licenseNamesWithCriticality[licenseName] ===
+                            Criticality.High
+                            ? { color: OpossumColors.orange }
+                            : props.licenseNamesWithCriticality[licenseName] ===
+                              Criticality.Medium
+                            ? { color: OpossumColors.mediumOrange }
+                            : {}
+                          : {}),
+                      }}
+                      key={index}
+                      align={index === 0 ? 'left' : 'center'}
+                    >
+                      {index === 0
+                        ? licenseName
+                        : totalNumberOfAttributions ||
+                          PLACEHOLDER_ATTRIBUTION_COUNT}
+                    </MuiTableCell>
+                  ))}
+                </MuiTableRow>
+              )
+            )}
+          </MuiTableBody>
+
+          <MuiTableFooter>
+            <MuiTableRow>
+              {tableFooter.map((total, index) => (
+                <MuiTableCell
+                  sx={projectStatisticsPopupClasses.footer}
+                  key={index}
+                  align={index === 0 ? 'left' : 'center'}
+                >
+                  {total}
+                </MuiTableCell>
+              ))}
+            </MuiTableRow>
+          </MuiTableFooter>
+        </MuiTable>
+      </MuiTableContainer>
+    </MuiBox>
+  );
+}
--- a/src/Frontend/Components/ProjectStatisticsPopup/project-statistics-popup-helpers.ts
+++ b/src/Frontend/Components/ProjectStatisticsPopup/project-statistics-popup-helpers.ts
@@ -30,6 +30,10 @@ const ATTRIBUTION_PROPERTIES_TO_DISPLAY: Array<keyof PackageInfo> = [
 export const SOURCE_TOTAL_HEADER = 'Total';
 export const LICENSE_TOTAL_HEADER = 'Total';
 export const ATTRIBUTION_TOTAL_HEADER = 'Total Attributions';
+export const LICENSE_COLUMN_NAME_IN_TABLE_HEADER = 'License';
+export const TOTAL_COLUMN_NAME_IN_TABLE_HEADER = 'Total';
+export const PLACEHOLDER_ATTRIBUTION_COUNT = '-';
+
 const UNKNOWN_SOURCE_PLACEHOLDER = '-';
 const ATTRIBUTION_PROPERTIES_ID_TO_DISPLAY_NAME: {
   [attributionProperty: string]: string;
--- a/src/Frontend/Components/ProjectStatisticsPopup/shared-project-statistics-popup-styles.ts
+++ b/src/Frontend/Components/ProjectStatisticsPopup/shared-project-statistics-popup-styles.ts
@@ -15,10 +15,14 @@ export const projectStatisticsPopupClasses = {
     fontWeight: 'bold',
     fontSize: 12,
     background: OpossumColors.lightBlue,
+    position: 'sticky',
+    bottom: 0,
   },
   body: {
     fontSize: 11,
     background: OpossumColors.lightestBlue,
+    maxWidth: '100px',
+    overflow: 'auto',
   },
   attributionCountPerSourcePerLicenseTable: {
     maxHeight: '400px',
@@ -29,8 +33,9 @@ export const projectStatisticsPopupClasses = {
     maxWidth: '400px',
     marginBottom: 3,
   },
-  tableFooter: {
-    position: 'sticky',
-    bottom: 0,
+  criticalLicensesTable: {
+    maxHeight: '400px',
+    maxWidth: '400px',
+    marginBottom: 3,
   },
 };
--- a/src/Frontend/Components/ProjectStatisticsPopup/AttributionCountPerSourcePerLicenseTable.tsx
+++ b/src/Frontend/Components/ProjectStatisticsPopup/AttributionCountPerSourcePerLicenseTable.tsx
@@ -6,8 +6,9 @@
 import React, { ReactElement } from 'react';
 import {
   AttributionCountPerSourcePerLicense,
-  LicenseNamesWithCriticality,
   LICENSE_TOTAL_HEADER,
+  LicenseNamesWithCriticality,
+  PLACEHOLDER_ATTRIBUTION_COUNT,
   SOURCE_TOTAL_HEADER,
 } from './project-statistics-popup-helpers';
 import MuiTypography from '@mui/material/Typography';
@@ -21,8 +22,7 @@ import MuiTableFooter from '@mui/material/TableFooter';
 import MuiTableRow from '@mui/material/TableRow';
 import { projectStatisticsPopupClasses } from './shared-project-statistics-popup-styles';
 import { OpossumColors } from '../../shared-styles';
-
-const PLACEHOLDER_ATTRIBUTION_COUNT = '-';
+import { Criticality } from '../../../shared/shared-types';
 
 interface AttributionCountPerSourcePerLicenseTableProps {
   attributionCountPerSourcePerLicense: AttributionCountPerSourcePerLicense;
@@ -85,10 +85,10 @@ export function AttributionCountPerSourcePerLicenseTable(
                       ...projectStatisticsPopupClasses.body,
                       ...(index === 0
                         ? props.licenseNamesWithCriticality[licenseName] ===
-                          'high'
+                          Criticality.High
                           ? { color: OpossumColors.orange }
                           : props.licenseNamesWithCriticality[licenseName] ===
-                            'medium'
+                            Criticality.Medium
                           ? { color: OpossumColors.mediumOrange }
                           : {}
                         : {}),
@@ -107,7 +107,7 @@ export function AttributionCountPerSourcePerLicenseTable(
             ))}
           </MuiTableBody>
 
-          <MuiTableFooter sx={projectStatisticsPopupClasses.tableFooter}>
+          <MuiTableFooter>
             <MuiTableRow>
               {totalsRow.map((total, index) => (
                 <MuiTableCell
--- a/src/Frontend/Components/ProjectStatisticsPopup/AttributionPropertyCountTable.tsx
+++ b/src/Frontend/Components/ProjectStatisticsPopup/AttributionPropertyCountTable.tsx
@@ -53,7 +53,7 @@ export function AttributionPropertyCountTable(
               )}
             </MuiTableRow>
           </MuiTableHead>
-          <MuiTableFooter sx={projectStatisticsPopupClasses.tableFooter}>
+          <MuiTableFooter>
             <MuiTableRow>
               {attributionPropertyCounts.map(
                 (attributionPropertyCount, index) => (
--- a/src/Frontend/Components/ProjectStatisticsPopup/ProjectStatisticsPopup.tsx
+++ b/src/Frontend/Components/ProjectStatisticsPopup/ProjectStatisticsPopup.tsx
@@ -21,10 +21,12 @@ import {
 } from './project-statistics-popup-helpers';
 import { AttributionCountPerSourcePerLicenseTable } from './AttributionCountPerSourcePerLicenseTable';
 import { AttributionPropertyCountTable } from './AttributionPropertyCountTable';
+import { CriticalLicensesTable } from './CriticalLicensesTable';
 
 const attributionCountPerSourcePerLicenseTableTitle = 'Signals per Sources';
 const attributionPropertyCountTableTitle =
   'First Party and Follow Up Attributions';
+const criticalLicensesTableTitle = 'Critical Licenses';
 
 export function ProjectStatisticsPopup(): ReactElement {
   const dispatch = useAppDispatch();
@@ -64,6 +66,13 @@ export function ProjectStatisticsPopup(): ReactElement {
               sortedManualAttributionPropertyCountsEntries
             }
             title={attributionPropertyCountTableTitle}
+          />
+          <CriticalLicensesTable
+            attributionCountPerSourcePerLicense={
+              attributionCountPerSourcePerLicense
+            }
+            licenseNamesWithCriticality={licenseNamesWithCriticalities}
+            title={criticalLicensesTableTitle}
           />
           <AttributionCountPerSourcePerLicenseTable
             attributionCountPerSourcePerLicense={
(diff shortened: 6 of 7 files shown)


src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
```diff
@@ -0,0 +1,148 @@
...
+
+const PLACEHOLDER_ATTRIBUTION_COUNT = '-';
+
+interface CriticalLicensesTableProps {
+  attributionCountPerSourcePerLicense: AttributionCountPerSourcePerLicense;
+  licenseNamesWithCriticality: LicenseNamesWithCriticality;
+  title: string;
+}
+
+export function CriticalLicensesTable(
```
benedikt-richter, line 33: Generally, it seems the comments could be replaced by introducing functions with good names that do the respective calculations described in the comments.

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
```diff
@@ -0,0 +1,148 @@
...
+
+  // The following will construct the data for the table footer.
+  const totalAttributions = criticalLicensesTotalAttributions
+    .map((license) => license[1])
+    .reduce((total1, total2) => +total1 + +total2, 0);
+  const tableFooter: Array<string> = [SOURCE_TOTAL_HEADER].concat(
+    totalAttributions.toString()
+  );
+
+  return (
```
benedikt-richter, line 81: There are some parts of this component that are used in all three current tables (mainly header and footer). Seems to be good to extract these.

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
```diff
@@ -0,0 +1,148 @@
...
+                      ...(index === 0
+                        ? props.licenseNamesWithCriticality[license[0]] ===
+                          'high'
+                          ? { color: OpossumColors.orange }
+                          : props.licenseNamesWithCriticality[license[0]] ===
+                            'medium'
+                          ? { color: OpossumColors.mediumOrange }
+                          : {}
+                        : {}),
+                    }}
```
benedikt-richter, line 118: This logic seems related to the one in the other table showing licenses highlighted by criticality.

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
```diff
@@ -0,0 +1,148 @@
...
+      .filter((license) => mediumCriticalityLicenseNames.includes(license[0]))
+      .map((license) => [license[0], license[1][SOURCE_TOTAL_HEADER]])
+      .sort();
+  const criticalLicensesTotalAttributions =
+    highCriticalityLicensesTotalAttributions.concat(
+      mediumCriticalityLicensesTotalAttributions
+    );
+
+  // The following will construct the data for the table header.
+  const tableColumnNames = ['LICENSE'].concat(SOURCE_TOTAL_HEADER);
```
benedikt-richter, line 71: This table has no relation to "sources". Seems the constant `SOURCE_TOTAL_HEADER` should not be used in the current context.
Further, `'LICENSE'` should be a constant.

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
```diff
@@ -0,0 +1,148 @@
...
+): ReactElement {
+  // Obtain license names that are of interest (ones with high and medium criticality)
+  const allLicensesWithCriticality = Object.entries(
+    props.licenseNamesWithCriticality
+  );
+  const highCriticalityLicenseNames = allLicensesWithCriticality
+    .map((licenseAndCriticality) =>
+      licenseAndCriticality[1] === 'high' ? licenseAndCriticality[0] : null
+    )
+    .filter((licenseName) => !isNull(licenseName));
```
benedikt-richter, line 44: filtering should happen before mapping.

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
```diff
@@ -0,0 +1,148 @@
...
+    props.attributionCountPerSourcePerLicense
+  );
+  const highCriticalityLicensesTotalAttributions = allAttributionsOfAllLicenses
+    .filter((license) => highCriticalityLicenseNames.includes(license[0]))
+    .map((license) => [license[0], license[1][SOURCE_TOTAL_HEADER]])
+    .sort();
+  const mediumCriticalityLicensesTotalAttributions =
+    allAttributionsOfAllLicenses
+      .filter((license) => mediumCriticalityLicenseNames.includes(license[0]))
+      .map((license) => [license[0], license[1][SOURCE_TOTAL_HEADER]])
```
benedikt-richter, line 63: here logic is duplicated and should be extracted.

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
```diff
@@ -0,0 +1,148 @@
...
+    .map((licenseAndCriticality) =>
+      licenseAndCriticality[1] === 'medium' ? licenseAndCriticality[0] : null
+    )
+    .filter((licenseName) => !isNull(licenseName));
+
+  // Obtain total number of attributions of interesting licenses.
+  // The following will basically construct the data for the table body.
+  const allAttributionsOfAllLicenses = Object.entries(
+    props.attributionCountPerSourcePerLicense
+  );
```
benedikt-richter, line 55: Why do we do this conversion? Can't we just look up the relevant data for the licenses we are interested in (i.e. the critical ones) in `attributionCountPerSourcePerLicense`?

Here we convert a huge object into nested arrays and then filter it two times with subsequent mapping and sorting. That seems an unnecessarily expensive calculation.

benedikt-richter (changes requested):

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
someshkhandelia: Yes, I agree with you.

What I had thought was to extract the content of the current issue, and create functions when I work on https://github.com/opossum-tool/OpossumUI/issues/1011 so that I can use those functions in the other table as well.

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
someshkhandelia: Should be addressed in https://github.com/opossum-tool/OpossumUI/issues/1011

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
someshkhandelia: Should probably be addressed in https://github.com/opossum-tool/OpossumUI/issues/1011

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
someshkhandelia: I am filtering out the null values that are being produced as a result of the map function. Thus, filtering should follow mapping in this case.

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
someshkhandelia: I received an error that `attributionCountPerSourcePerLicense` could be undefined (it being a prop that is received from another component) and therefore I could not access its contents unless I used Object.entries()

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
```diff
@@ -0,0 +1,148 @@
...
+    .sort();
+  const mediumCriticalityLicensesTotalAttributions =
+    allAttributionsOfAllLicenses
+      .filter((license) => mediumCriticalityLicenseNames.includes(license[0]))
+      .map((license) => [license[0], license[1][SOURCE_TOTAL_HEADER]])
+      .sort();
+  const criticalLicensesTotalAttributions =
+    highCriticalityLicensesTotalAttributions.concat(
+      mediumCriticalityLicensesTotalAttributions
+    );
```
vasily-pozdnyakov, line 68: Can we use just one for-loop through licenseNamesWithCriticality instead of all these duplicated reduce-map pairs?

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
vasily-pozdnyakov: We will always have two columns: 'LICENSE' and 'TOTAL'. Not sure that we need to use concat here. Just create a list with two elements.

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
```diff
@@ -0,0 +1,148 @@
...
+                          ? { color: OpossumColors.mediumOrange }
+                          : {}
+                        : {}),
+                    }}
+                    key={index}
+                    align={index === 0 ? 'left' : 'center'}
+                  >
+                    {index === 0
+                      ? license[0]
+                      : license[1] || PLACEHOLDER_ATTRIBUTION_COUNT}
```
vasily-pozdnyakov, line 124: What do you mean by "license"? Is it like licenseNameWithCount? This is all through the code.
Also [0] and [1] are not really clear. Can we use dictionary instead of array?

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
```diff
@@ -0,0 +1,148 @@
...
+import MuiTableCell from '@mui/material/TableCell';
+import MuiTableContainer from '@mui/material/TableContainer';
+import MuiTableHead from '@mui/material/TableHead';
+import MuiTableFooter from '@mui/material/TableFooter';
+import MuiTableRow from '@mui/material/TableRow';
+import { projectStatisticsPopupClasses } from './shared-project-statistics-popup-styles';
+import { OpossumColors } from '../../shared-styles';
+import { isNull } from 'lodash';
+
+const PLACEHOLDER_ATTRIBUTION_COUNT = '-';
```
vasily-pozdnyakov, line 25: There is the same constant in AttributionCountPerSourcePerLicenseTable.tsx.  Do we really need it here?

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
```diff
@@ -0,0 +1,148 @@
...
+          <MuiTableBody>
+            {criticalLicensesTotalAttributions.map((license, rowIndex) => (
+              <MuiTableRow key={rowIndex}>
+                {tableColumnNames.map((columnName, index) => (
+                  <MuiTableCell
+                    sx={{
+                      ...projectStatisticsPopupClasses.body,
+                      ...(index === 0
+                        ? props.licenseNamesWithCriticality[license[0]] ===
+                          'high'
```
vasily-pozdnyakov, line 111: Btw, it would be nice to use Criticality enum from shared-types instead of hard-coded strings 'high' and 'medium'. Could you please also change it in AttributionCountPerSourcePerLicenseTable.tsx?

src/Frontend/Components/ProjectStatisticsPopup/shared-project-statistics-popup-styles.ts
```diff
@@ -29,6 +31,11 @@ export const projectStatisticsPopupClasses = {
     maxWidth: '400px',
     marginBottom: 3,
   },
+  criticalLicensesTable: {
+    maxHeight: '600px',
+    maxWidth: '500px',
+    marginBottom: 3,
+  },
   tableFooter: {
```
vasily-pozdnyakov, line 39: Can we merge footer and tableFooter styles?

src/Frontend/Components/ProjectStatisticsPopup/shared-project-statistics-popup-styles.ts
```diff
@@ -29,6 +31,11 @@ export const projectStatisticsPopupClasses = {
     maxWidth: '400px',
     marginBottom: 3,
   },
+  criticalLicensesTable: {
+    maxHeight: '600px',
```
vasily-pozdnyakov, line 35: Why does it have maxHeight of 600px, when the main table attributionCountPerSourcePerLicenseTable has maxHeight of 400px?

vasily-pozdnyakov (changes requested):

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
someshkhandelia: I have introduced functions and better variable names that will clarify what (and what type) is being contained by such variables.

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
someshkhandelia: For loop was not essential. Function calls have removed the need for duplicated reduce-map pairs.

src/Frontend/Components/ProjectStatisticsPopup/shared-project-statistics-popup-styles.ts
```diff
@@ -29,6 +31,11 @@ export const projectStatisticsPopupClasses = {
     maxWidth: '400px',
     marginBottom: 3,
   },
+  criticalLicensesTable: {
```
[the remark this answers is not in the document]
someshkhandelia: Done.

src/Frontend/Components/ProjectStatisticsPopup/shared-project-statistics-popup-styles.ts
someshkhandelia: changed it to 400px.

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
someshkhandelia: Removed and moved to a common place. Done.

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
someshkhandelia: Done.

src/Frontend/Components/ProjectStatisticsPopup/shared-project-statistics-popup-styles.ts
someshkhandelia: Done.

src/Frontend/Components/ProjectStatisticsPopup/project-statistics-popup-helpers.ts
```diff
@@ -30,6 +30,10 @@ const ATTRIBUTION_PROPERTIES_TO_DISPLAY: Array<keyof PackageInfo> = [
 export const SOURCE_TOTAL_HEADER = 'Total';
 export const LICENSE_TOTAL_HEADER = 'Total';
 export const ATTRIBUTION_TOTAL_HEADER = 'Total Attributions';
+export const LICENSE_COLUMN_NAME_IN_TABLE_HEADER = 'License';
+export const TOTAL_COLUMN_NAME_IN_TABLE_HEADER = 'Total';
+export const PLACEHOLDER_ATTRIBUTION_COUNT = '-';
+
 const UNKNOWN_SOURCE_PLACEHOLDER = '-';
```
vasily-pozdnyakov, line 37: Can we merge them?

src/Frontend/Components/ProjectStatisticsPopup/project-statistics-popup-helpers.ts
```diff
@@ -30,6 +30,10 @@ const ATTRIBUTION_PROPERTIES_TO_DISPLAY: Array<keyof PackageInfo> = [
 export const SOURCE_TOTAL_HEADER = 'Total';
 export const LICENSE_TOTAL_HEADER = 'Total';
 export const ATTRIBUTION_TOTAL_HEADER = 'Total Attributions';
+export const LICENSE_COLUMN_NAME_IN_TABLE_HEADER = 'License';
+export const TOTAL_COLUMN_NAME_IN_TABLE_HEADER = 'Total';
```
vasily-pozdnyakov, line 34: Would be nice to merge all "Total", as it is always a column name.

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
```diff
@@ -0,0 +1,172 @@
...
+  licenseNamesAndTheirTotalAttributions: Array<[string, number]>
+): number {
+  return licenseNamesAndTheirTotalAttributions
+    .map(
+      (licenseNameAndTotalAttributions) => licenseNameAndTotalAttributions[1]
+    )
+    .reduce((total1, total2) => +total1 + +total2, 0);
+}
+
+interface CriticalLicensesTableProps {
```
vasily-pozdnyakov, line 63: Interface declaration should be before functions.

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
vasily-pozdnyakov: However, we can do all calculations in one run instead of mapping multiple times over the array. In this case you won't need to transform dictionary into array, and you can directly write licenses into separate lists based on criticality (plus: you do not need to use reduce).

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
vasily-pozdnyakov: Could we do array of dictionaries? In this case it will be clear that [0]=licenseName, [1]=count

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
vasily-pozdnyakov: SOURCE_TOTAL_HEADER is not an appropriate name here, I think.

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
```diff
@@ -0,0 +1,172 @@
...
+import MuiTableFooter from '@mui/material/TableFooter';
+import MuiTableRow from '@mui/material/TableRow';
+import { projectStatisticsPopupClasses } from './shared-project-statistics-popup-styles';
+import { OpossumColors } from '../../shared-styles';
+import { Criticality } from '../../../shared/shared-types';
+
+function getLicenseNamesByCriticality(
+  allLicensesWithCriticality: Array<[string, Criticality | undefined]>,
+  criticality: Criticality
+): Array<string> {
```
vasily-pozdnyakov, line 31: Btw, you can use set.

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
```diff
@@ -0,0 +1,172 @@
...
+}
+
+function getTotalNumberOfAttributions(
+  licenseNamesAndTheirTotalAttributions: Array<[string, number]>
+): number {
+  return licenseNamesAndTheirTotalAttributions
+    .map(
+      (licenseNameAndTotalAttributions) => licenseNameAndTotalAttributions[1]
+    )
+    .reduce((total1, total2) => +total1 + +total2, 0);
```
vasily-pozdnyakov, line 60: See reduce usage in project-statistics-popup-helpers.ts. Would be nice to make it consistent.

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
```diff
@@ -0,0 +1,172 @@
...
+  const highCriticalityLicensesTotalAttributions =
+    getCriticalLicenseNamesWithTheirTotalAttributions(
+      props.attributionCountPerSourcePerLicense,
+      highCriticalityLicenseNames
+    );
+  const mediumCriticalityLicensesTotalAttributions =
+    getCriticalLicenseNamesWithTheirTotalAttributions(
+      props.attributionCountPerSourcePerLicense,
+      mediumCriticalityLicenseNames
+    );
```
vasily-pozdnyakov, line 92: Can we do both calculations in one function?

vasily-pozdnyakov (changes requested):

src/Frontend/Components/ProjectStatisticsPopup/project-statistics-popup-helpers.ts
someshkhandelia: Depending on the context these variables are used in, we are having different names, despite having the same values. That's what I understood from @benedikt-richter's comment here: https://github.com/opossum-tool/OpossumUI/pull/1013#discussion_r979607836

src/Frontend/Components/ProjectStatisticsPopup/project-statistics-popup-helpers.ts
someshkhandelia: Please see https://github.com/opossum-tool/OpossumUI/pull/1013#discussion_r980130507

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
someshkhandelia: The interface here refers to the props of the function that is just after it. I think it is more readable if props and function are close to each other.

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
someshkhandelia: I am trying to have a more generic table construct overall, where we have an array of table rows, instead of key, value pairs that say that license name is a key and number of attributions is a value.
Besides, because of the variable naming and function definition, it should be clear what does the variable contain.

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
someshkhandelia: I would prefer to have separate functions rather than have a single function that does too many things.
Regarding  multiple runs, the complexity will stay O(n) in both cases, and in this case the code is more readable

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
someshkhandelia: I am trying to calculate a grand total of several totals, that is why I have the variables named the way they are. The initial value starts with 0, which is more clear here than the reference you have pointed to.

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
someshkhandelia: It is no longer being used in the recent version of the code.

src/Frontend/Components/ProjectStatisticsPopup/project-statistics-popup-helpers.ts
someshkhandelia: Also it is not always a column name. Depending on context, it is used as a key to access different parts of a dictionary in another table.

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
```diff
@@ -0,0 +1,172 @@
...
+import { OpossumColors } from '../../shared-styles';
+import { Criticality } from '../../../shared/shared-types';
+
+function getLicenseNamesByCriticality(
+  allLicensesWithCriticality: Array<[string, Criticality | undefined]>,
+  criticality: Criticality
+): Array<string> {
+  return allLicensesWithCriticality
+    .map((licenseNameAndCriticality) =>
+      licenseNameAndCriticality[1] === criticality
```
benedikt-richter, line 34: Having used an object that would be `licenseNameAndCriticality.criticality === criticality`

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
```diff
@@ -0,0 +1,172 @@
...
+        ? licenseNameAndCriticality[0]
+        : ''
+    )
+    .filter((licenseName) => licenseName != '');
```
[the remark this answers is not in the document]
benedikt-richter: Could also be `.filter(Boolean)`

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
```diff
@@ -0,0 +1,172 @@
...
+}
+
+function getCriticalLicenseNamesWithTheirTotalAttributions(
+  attributionCountPerSourcePerLicense: AttributionCountPerSourcePerLicense,
+  criticalLicenseNames: Array<string>
+): Array<[string, number]> {
```
benedikt-richter, line 44: as above, using objects make it easier to understand the semantics from the signature of a function

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
benedikt-richter: I also find the name `license` confusing. I'd not expect it to be a tuple.

benedikt-richter: > I am trying to have a more generic table construct overall, where we have an array of table rows, instead of key, value pairs that say that license name is a key and number of attributions is a value. Besides, because of the variable naming and function definition, it should be clear what does the variable contain.

I like that idea. However, I'd prefer to have all calculations using objects (as the names of the keys help understanding what's going on: the business logic) and then to have a final step where the data is transformed into what the component needs (this might be arrays) where the display logic lives.

benedikt-richter (changes requested):

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
someshkhandelia: Done

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
someshkhandelia: Done

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
someshkhandelia: Done

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
someshkhandelia: Done.

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
someshkhandelia: Done.

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
someshkhandelia: Done.

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
someshkhandelia: Done.

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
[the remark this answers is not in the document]
someshkhandelia: Done

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
```diff
@@ -0,0 +1,172 @@
...
+import MuiTableContainer from '@mui/material/TableContainer';
+import MuiTableHead from '@mui/material/TableHead';
+import MuiTableFooter from '@mui/material/TableFooter';
+import MuiTableRow from '@mui/material/TableRow';
+import { projectStatisticsPopupClasses } from './shared-project-statistics-popup-styles';
+import { OpossumColors } from '../../shared-styles';
+import { Criticality } from '../../../shared/shared-types';
+
+function getLicenseNamesByCriticality(
+  allLicensesWithCriticality: Array<[string, Criticality | undefined]>,
```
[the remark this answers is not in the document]
someshkhandelia: Done.

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
someshkhandelia: Done.

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
```diff
@@ -0,0 +1,172 @@
...
+    allLicensesWithCriticality,
+    Criticality.Medium
+  );
+  const highCriticalityLicensesTotalAttributions =
+    getCriticalLicenseNamesWithTheirTotalAttributions(
+      props.attributionCountPerSourcePerLicense,
+      highCriticalityLicenseNames
+    );
+  const mediumCriticalityLicensesTotalAttributions =
+    getCriticalLicenseNamesWithTheirTotalAttributions(
```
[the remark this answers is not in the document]
someshkhandelia: Thanks.

src/Frontend/Components/ProjectStatisticsPopup/CriticalLicensesTable.tsx
someshkhandelia: Used Array of Dictionaries as suggested by you and @benedikt-richter. Keeping other things the same.

benedikt-richter (approved):

vasily-pozdnyakov (approved):

Issue #944, opened 2022-09-08
Introduce pie chart for most frequent licenses in signals in statistics popup


benedikt-richter:
Add a pie chart for most frequent licenses in signals statistics popup.
The 5 most frequent licenses should be presented in a pie chart using the same deduplication as used in #943. Show the remaining ones as one more sector.

Note: for pie charts the following library might be well suited: https://devexpress.github.io/devextreme-reactive/react/chart/demos/pie/pie/

PR #1016, opened 2022-09-26
Introduce pie chart for most frequent licenses in signals in statisti…

Billwuzl:
### Summary of changes

Introduce pie chart for most frequent licenses in signals in statistic popup

Created a function in project-statistics-popup-helpers.ts to obtain the most frequent licenses
Pie chart is created in PieCharts.tsx

Fix: #944

Signed-off-by: Billwuzl <wu.zhuolin2000@gmail.com>

### DIFF
--- a/src/Frontend/Components/ProjectStatisticsPopup/PieCharts.tsx
+++ b/src/Frontend/Components/ProjectStatisticsPopup/PieCharts.tsx
@@ -1,0 +1,90 @@
+// SPDX-FileCopyrightText: Meta Platforms, Inc. and its affiliates
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+
+import React, { ReactElement } from 'react';
+import {
+  PieChart,
+  Pie,
+  Cell,
+  ResponsiveContainer,
+  PieLabelRenderProps,
+} from 'recharts';
+import MuiBox from '@mui/material/Box';
+import MuiTypography from '@mui/material/Typography';
+import { PieChartProps } from './AttributionCountPerSourcePerLicenseTable';
+import { getMostFrequentLicenses } from './project-statistics-popup-helpers';
+import { OpossumColors } from '../../shared-styles';
+
+const COLORS = [
+  OpossumColors.orange,
+  OpossumColors.mediumOrange,
+  OpossumColors.darkBlue,
+  OpossumColors.brown,
+  OpossumColors.lightBlue,
+  OpossumColors.green,
+];
+
+const RADIAN = Math.PI / 180;
+
+function renderCustomizedLabel(props: PieLabelRenderProps): ReactElement {
+  const iRadius = Number(props.innerRadius) || 0;
+  const oRadius = Number(props.outerRadius) || 0;
+  const mAngle = Number(props.midAngle) || 0;
+  const chartX = Number(props.cx) || 0;
+  const chartY = Number(props.cy) || 0;
+  const radius = iRadius + (oRadius - iRadius) * 1.5;
+  const x = chartX + radius * Math.cos(-mAngle * RADIAN);
+  const y = chartY + radius * Math.sin(-mAngle * RADIAN);
+
+  return (
+    <text
+      x={x}
+      y={y}
+      fill="black"
+      textAnchor={x > Number(props.cx) ? 'start' : 'end'}
+      dominantBaseline="central"
+      fontFamily="Arial"
+    >
+      {`${props.payload.payload.licenseName} (${props.payload.payload.count})`}
+    </text>
+  );
+}
+
+export function MostFrequentLicensesPieChart(
+  props: PieChartProps
+): ReactElement | null {
+  const sortedMostFrequentLicenses = getMostFrequentLicenses(
+    props.attributionCountPerSourcePerLicense
+  );
+  if (sortedMostFrequentLicenses.length === 0) {
+    return null;
+  }
+
+  return (
+    <MuiBox>
+      <MuiTypography variant="subtitle1">{props.title}</MuiTypography>
+      <ResponsiveContainer width="100%" maxHeight={300} aspect={1}>
+        <PieChart>
+          <Pie
+            data={sortedMostFrequentLicenses}
+            dataKey="count"
+            nameKey="licenseName"
+            cx="50%"
+            cy="50%"
+            paddingAngle={1}
+            minAngle={5}
+            outerRadius="65%"
+            isAnimationActive={false}
+            label={renderCustomizedLabel}
+          >
+            {sortedMostFrequentLicenses.map((entry, index) => (
+              <Cell key={`cell-${index}`} fill={COLORS[index]} />
+            ))}
+          </Pie>
+        </PieChart>
+      </ResponsiveContainer>
+    </MuiBox>
+  );
+}
--- a/src/Frontend/Components/ProjectStatisticsPopup/__tests__/ProjectStatisticsPopup.test.tsx
+++ b/src/Frontend/Components/ProjectStatisticsPopup/__tests__/ProjectStatisticsPopup.test.tsx
@@ -13,6 +13,9 @@ import { Attributions } from '../../../../shared/shared-types';
 import { loadFromFile } from '../../../state/actions/resource-actions/load-actions';
 import { getParsedInputFileEnrichedWithTestData } from '../../../test-helpers/general-test-helpers';
 import { screen } from '@testing-library/react';
+import { mockResizeObserver } from '../../../test-helpers/popup-test-helpers';
+
+mockResizeObserver();
 
 describe('The ProjectStatisticsPopup', () => {
   it('displays license names and source names', () => {
@@ -23,14 +26,14 @@ describe('The ProjectStatisticsPopup', () => {
           name: 'scancode',
           documentConfidence: 10,
         },
-        licenseName: 'test-license-name',
+        licenseName: 'Apache License Version 2.0',
       },
       uuid_2: {
         source: {
           name: 'reuser',
           documentConfidence: 90,
         },
-        licenseName: 'test-license-name_1',
+        licenseName: 'The MIT License (MIT)',
       },
     };
     store.dispatch(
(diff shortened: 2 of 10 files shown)


Billwuzl:
I will try to fix the problems of overlapping texts in the pie chart and the failing test cases now

src/Frontend/Components/ProjectStatisticsPopup/PieCharts.tsx
```diff
@@ -0,0 +1,83 @@
...
+          cx="30%"
+          cy="45%"
+          paddingAngle={2}
+          minAngle={3}
+          outerRadius={70}
+          fill="#8884d8"
+          label={renderCustomizedLabel}
+        >
+          {sortedMostFrequentLicenses.map((entry, index) => (
+            <Cell key={`cell-${index}`} fill={COLORS[index % COLORS.length]} />
```
vasily-pozdnyakov, line 77: Can we just use COLOR[index]?

src/Frontend/Components/ProjectStatisticsPopup/PieCharts.tsx
```diff
@@ -0,0 +1,83 @@
...
+  '#FFBB28',
+  '#55ff42',
+  '#9bb4b1',
+  '#7ea4a8',
+  '#0088FE',
+];
+
+const RADIAN = Math.PI / 180;
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
vasily-pozdnyakov, line 32: Can we somehow avoid usage of it?

src/Frontend/Components/ProjectStatisticsPopup/PieCharts.tsx
```diff
@@ -0,0 +1,83 @@
...
+): ReactElement {
+  const sortedMostFrequentLicenses = getMostFrequentLicenses(
+    props.attributionCountPerSourcePerLicense,
+    props.licenseNamesWithCriticality
+  );
+
+  return (
+    <MuiBox>
+      <MuiTypography variant="subtitle1">{props.title}</MuiTypography>
+      <PieChart width={730} height={250}>
```
vasily-pozdnyakov, line 63: Such hard-coded values do not look good to me:) Can we use relative values?

vasily-pozdnyakov (changes requested):
* Can we have more space between license names?
image


* If there are no licenses (very unlikely, but still) we see a lot of empty space
image

* Would be nice to add some testing.

src/Frontend/Components/ProjectStatisticsPopup/PieCharts.tsx
```diff
@@ -0,0 +1,81 @@
...
+        <Pie
+          data={sortedMostFrequentLicenses}
+          dataKey="value"
+          nameKey="name"
+          cx="50%"
+          cy="50%"
+          outerRadius={80}
+          fill="#8884d8"
```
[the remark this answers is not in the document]
Billwuzl: actually we don't need this fill

benedikt-richter:
The warning of the kind 
```

FAIL src/Frontend/Components/AppContainer/__tests__/AppContainer.test.tsx
  ● Test suite failed to run

    Jest encountered an unexpected token

    Jest failed to parse a file. This happens e.g. when your code or its dependencies use non-standard JavaScript syntax, or when Jest is not configured to support such syntax.

    Out of the box Jest supports Babel, which will be used to transform your files into valid JS based on your Babel configuration.

    By default "node_modules" folder is ignored by transformers.

    Here's what you can do:
     • If you are trying to use ECMAScript Modules, see https://jestjs.io/docs/ecmascript-modules for how to enable it.
     • If you are trying to use TypeScript, see https://jestjs.io/docs/getting-started#using-typescript
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/configuration
    For information about custom transformations, see:
    https://jestjs.io/docs/code-transformation

    Details:

    /home/runner/work/OpossumUI/OpossumUI/node_modules/d3-shape/src/index.js:1
    ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,jest){export {default as arc} from "./arc.js";


```

that cause the tests to fail, can be fixed by modifying the value of the key `"jest"` in `package.json` to be

```

  "jest": {
    "resetMocks": false,
    "transformIgnorePatterns": [
      "/node_modules/(?!d3-*|internmap)"
    ],
    "transform": {
      "/node_modules/(?!d3-*|internmap)": "babel-jest"
    }
  },

```

Billwuzl:
Now larger gaps between license names are created.

If there is no license, the pie chart will disappear now

Billwuzl:
Test Cases added

vasily-pozdnyakov:
General things:
* Margins before and after the pie chart are too big 
image

* It would be nice to have a positive test for the pie chart (when the ui actually renders it)

antonbauhofer:
The labels appear with a delay. Also, as pointed out before, the gaps between the fractions seem to be a bit big.
See screenshot:

Screenshot 2022-09-28 at 10 12 51

src/Frontend/Components/ProjectStatisticsPopup/PieCharts.tsx
```diff
@@ -0,0 +1,88 @@
...
+  return (
+    <MuiBox>
+      <MuiTypography variant="subtitle1">{props.title}</MuiTypography>
+      <ResponsiveContainer width="100%" maxHeight={350} aspect={1}>
+        <PieChart>
+          <Pie
+            data={sortedMostFrequentLicenses}
+            dataKey="count"
+            nameKey="licenseName"
+            cx="40%"
```
antonbauhofer, line 73: I would either put this to 50% to have it horizontally centered, or have it left aligned.

src/Frontend/Components/ProjectStatisticsPopup/PieCharts.tsx
```diff
@@ -0,0 +1,88 @@
...
+            cy="50%"
+            paddingAngle={3}
```
antonbauhofer, line 75: I would prefer to have this set to 1, so that the distances between the fractions are reduced.

src/Frontend/Components/ProjectStatisticsPopup/PieCharts.tsx
```diff
@@ -0,0 +1,88 @@
...
+  OpossumColors.mediumOrange,
+  OpossumColors.darkBlue,
+  OpossumColors.brown,
+  OpossumColors.lightBlue,
+  OpossumColors.green,
+];
+
+const RADIAN = Math.PI / 180;
+
+function renderCustomizedLabel(props: PieLabelRenderProps): ReactElement {
```
antonbauhofer, line 31: Can you adjust the font to be the same as in the rest of the app?

antonbauhofer (changes requested):
Looks good so far.
I think the styling could be improved and some tests are missing.

src/Frontend/Components/ProjectStatisticsPopup/project-statistics-popup-helpers.ts
```diff
@@ -22,6 +22,10 @@ export interface LicenseNamesWithCriticality {
 interface UniqueLicenseNameToAttributions {
   [strippedLicenseName: string]: Array<string>;
 }
+export interface PieChartData {
```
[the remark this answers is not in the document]
Billwuzl: I think we might keep all pie chart data formats the same, as the current format seems to be the only data format accepted by Recharts PieCharts. :)

vasily-pozdnyakov:
Good job. See comments from Toni. Need to do something with gaps before and after the chart: if there are no labels, the margins are too big.

src/Frontend/Components/ProjectStatisticsPopup/PieCharts.tsx
```diff
@@ -0,0 +1,88 @@
...
+      textAnchor={x > Number(props.cx) ? 'start' : 'end'}
+      dominantBaseline="central"
+    >
+      {`${props.payload.payload.licenseName} (${props.payload.payload.count})`}
+    </text>
+  );
+}
+
+export function MostFrequentLicensesPieChart(
+  props: AttributionCountPerSourcePerLicenseTableProps
```
benedikt-richter, line 55: Looks strange to use here the same props as for the table. The PieChart should have its own props.

benedikt-richter:
Currently, the PieChart component is customized to display the most common licenses. Before, introducing the next pie chart in project overview we should extract a generic component `PieChart` that lives at the same level as `ProjectStatisticsPopup` and has a interface that can also be used for following pie charts. Otherwise we would duplicate the code that is relevant for the display logic three times. This does not have to happen in this PR, though.

src/Frontend/Components/ProjectStatisticsPopup/PieCharts.tsx
Billwuzl: yeah. This will be fixed in the following PR if more pie charts are added

src/Frontend/Components/ProjectStatisticsPopup/__tests__/ProjectStatisticsPopup.test.tsx
```diff
@@ -48,6 +51,53 @@ describe('The ProjectStatisticsPopup', () => {
...
+    store.dispatch(
+      loadFromFile(
+        getParsedInputFileEnrichedWithTestData({
+          externalAttributions: testExternalAttributions,
+        })
+      )
+    );
+
+    renderComponentWithStore(<ProjectStatisticsPopup />, { store });
+    expect(screen.getByText('Most Frequent Licenses')).toBeInTheDocument();
```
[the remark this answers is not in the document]
Billwuzl: Actually, this might be a bit hard. We can discuss this later 👍

src/Frontend/Components/GlobalPopup/__tests__/GlobalPopup.test.tsx
```diff
@@ -17,6 +17,9 @@ import { loadFromFile } from '../../../state/actions/resource-actions/load-actio
 import { getParsedInputFileEnrichedWithTestData } from '../../../test-helpers/general-test-helpers';
 import { setMultiSelectSelectedAttributionIds } from '../../../state/actions/resource-actions/attribution-view-simple-actions';
 import { act } from 'react-dom/test-utils';
+import { mockResizeObserver } from '../../ProjectStatisticsPopup/__tests__/project-statistics-popup-helpers.test';
```
benedikt-richter, line 20: we should not import from one test to another.

vasily-pozdnyakov (approved):
From my side it looks fine, just please check ProjectStatisticsPopup.test.tsx for duplicating tests.
Final styling can be done later, when we have all functionality implemented.

src/Frontend/Components/GlobalPopup/__tests__/GlobalPopup.test.tsx
Billwuzl: I just realized that ProjectStatisticsPopup.test.tsx is the only file that needs to use mockResizeObserver.

Billwuzl: but I think I will put mockResizeObserver in src/Frontend/test-helpers/popup-test-helpers.ts. Just in case we need this function again.

benedikt-richter (approved):

antonbauhofer (approved):
Looks good now and works well. Nice job!

Issue #1007, opened 2022-09-22
Add huge OpossumUI input file


benedikt-richter:
For testing purposes it would be good to have a large OpossumUI input file that contains a lot of signals from (many) different sources where the field criticality is non-empty.
This would be, in particular, helpful for testing the project statistics popup and performance in general.

PR #1017, opened 2022-09-26
Add example file

benedikt-richter:
### Summary of changes

Add example input file containing many signals with licenses.

The file was creating by scanning OpossumUI source code using scancode. In partiuclar, some folders have been scanned repeatedly. Subsequently, the output files where converted to OpossumUI input files using [opossum-tool/opossum.lib.hs](https://github.com/opossum-tool/opossum.lib.hs). The resulting Opossum input files then where modified to change relevant attributes like "sources", "criticality", etc. Finally, all files where merged using [opossum-tool/opossum.lib.hs](https://github.com/opossum-tool/opossum.lib.hs).

Screenshot 2022-09-26 at 09 44 33

Fix: #1007


vasily-pozdnyakov (approved):
I would keep large_opossum_input.json as it is useful in performance testing.

benedikt-richter:
> I would keep large_opossum_input.json as it is useful in performance testing.

Added it back. However, it's now gzipped, as it was rather large in size.

leslielazzarino (approved):

Issue #1026, opened 2022-09-28
Always highlight criticality of signals


benedikt-richter:
Currently, highlighting of criticality in signals is an opt-in feature that can be enabled in the menu.

Instead it should always be enabled.

PR #1027, opened 2022-09-28
Always highlight criticality

benedikt-richter:
### Summary of changes

Remove opt-in for highlighting criticality.

Fix: #1026

### DIFF
--- a/src/ElectronBackend/main/menu.ts
+++ b/src/ElectronBackend/main/menu.ts
@@ -4,7 +4,7 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
-import { app, BrowserWindow, Menu, shell, WebContents } from 'electron';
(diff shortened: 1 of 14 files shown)


FlorianSchepers (approved):
The code changes look good and its working for me.

Why do we want to delete this feature instead of setting as default. This would enable the user to deactivate it if he doesn't want it.

MarkusObendrauf (approved):
Looks good to me

benedikt-richter:
> Why do we want to delete this feature instead of setting as default. This would enable the user to deactivate it if he doesn't want it.

It's true that we could with low effort change the feature to make it opt-out. However, it's probably unlikely for the user to opt-out of that.
The main motivation of having this feature toggle initally was to introduce the new feature and allow users to test it. By now it is accepted and we can remove the feature toggle.

PR #1031, opened 2022-09-28
Improve highlighting of criticality in statistics popup

benedikt-richter:
### Summary of changes

In the tables in statistics popup we were very conservative and marked a license as critical if at least one of its variants was critical. Now we increased this threshold.

So far if there were 1000 signals with license MIT and a single one was of some criticality the license in general got this criticality. The logic turned out to be too conservative.

### DIFF
--- a/src/Frontend/Components/ProjectStatisticsPopup/project-statistics-popup-helpers.ts
+++ b/src/Frontend/Components/ProjectStatisticsPopup/project-statistics-popup-helpers.ts
@@ -48,9 +48,9 @@ export function aggregateLicensesAndSourcesFromAttributions(
   attributionSources: ExternalAttributionSources
 ): {
   attributionCountPerSourcePerLicense: AttributionCountPerSourcePerLicense;
-  licenseNamesWithCriticalities: LicenseNamesWithCriticality;
+  licenseNamesWithCriticality: LicenseNamesWithCriticality;
 } {
-  const { attributionCountPerSourcePerLicense, licenseNamesWithCriticalities } =
+  const { attributionCountPerSourcePerLicense, licenseNamesWithCriticality } =
     getLicenseDataFromAttributionsAndSources(
       strippedLicenseNameToAttribution,
       attributions,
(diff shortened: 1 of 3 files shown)


src/Frontend/Components/ProjectStatisticsPopup/project-statistics-popup-helpers.ts
```diff
@@ -174,6 +176,19 @@ function getLicenseDataFromVariants(
   };
 }
 
+function getLicenseCriticality(licenseCriticalityCounts: {
+  high: number;
+  medium: number;
+  none: number;
+}): Criticality | undefined {
+  return licenseCriticalityCounts['medium'] + licenseCriticalityCounts['high'] >
```
vasily-pozdnyakov, line 184: Should we maybe just return criticality with max count?

src/Frontend/Components/ProjectStatisticsPopup/project-statistics-popup-helpers.ts
```diff
@@ -145,12 +145,14 @@ function getLicenseDataFromVariants(
...
-      if (
-        variantCriticality === Criticality.High ||
-        (variantCriticality === Criticality.Medium &&
-          licenseCriticality === undefined)
-      ) {
-        licenseCriticality = variantCriticality;
+      if (variantCriticality === Criticality.High) {
+        licenseCriticalityCounts['high']++;
+      }
+      if (variantCriticality === Criticality.Medium) {
```
vasily-pozdnyakov, line 151: how about "else if"?

src/Frontend/Components/ProjectStatisticsPopup/project-statistics-popup-helpers.ts
```diff
@@ -145,12 +145,14 @@ function getLicenseDataFromVariants(
...
+        licenseCriticalityCounts['medium']++;
+      }
+      if (variantCriticality === undefined) {
```
vasily-pozdnyakov, line 154: "else"? to be 100% safe.

vasily-pozdnyakov (approved):

src/Frontend/Components/ProjectStatisticsPopup/project-statistics-popup-helpers.ts
benedikt-richter: We could. That would be a different heuristic. Here, it is a bit more conservative and for more cases returns a (high) criticality. I guess we will further fine tune in the future.

leslielazzarino (approved):

Issue #966, opened 2022-09-12
Improve loading spinner


vasily-pozdnyakov:
The current implementation of a loading spinner is not the best. As users we want the following:
- center window on macOs
- center spinner in Window
- make it smaller
Now: 
![image](https://user-images.githubusercontent.com/85183359/189623043-bc9196d4-9568-4f49-88a0-97844f73b004.png)

PR #1037, opened 2022-09-28
Improve loading spinner

benedikt-richter:
### Summary of changes

Instead of opening a new window to display a loading spinner, the loading spinner is displayed in the same window. 

Before:
Screenshot 2022-09-28 at 21 06 55

After:
Screenshot 2022-09-28 at 21 04 20

Fix: #966

Test that the new spinner is shown while files are opened and during file export.


benedikt-richter:
With the most recent change the loading spinner does also work properly for file exports. Before, for most of the duration of the export there was no spinner shown but the app was just frozen.

public/loading.html
```diff
@@ -1,43 +0,0 @@
-<!DOCTYPE html>
```
antonbauhofer, line 1: Where is this file coming from / what is it's purpose?

antonbauhofer (approved):
Well done. The new spinner looks much better than before!
The code looks good and it works as expected.

public/loading.html
benedikt-richter: that's the old spinner

Issue #1079, opened 2022-10-17
Refactor project-statistics-popup-helpers.test.tsx


leslielazzarino:
1. One describe per function
2. One test for each expected behaviour, with precise description
    a. no duplicate tests with slightly different test data
3. Reference example: src/Frontend/Components/ResourcesList/__tests__/resource-list-helpers.test.ts

benedikt-richter:
Currently the statistics popup tests are very fragile.
Consider improving on the tests e.g. by using aria labels for components and then to assert their presence instead of asserting their titles (and names of columns of tables).

PR #1165, opened 2022-11-17
Refactor Tests for statistics popup helpers

someshkhandelia:
### Summary of changes

Refactor Tests for statistics popup helpers

 - Have separate tests to check different behaviours of a function.
 - Refactor existing tests to have a more clear workflow.
 - Add dedicated tests for all functions that exist in the helper file. 

Fix: #1079

### DIFF
--- a/src/Frontend/Components/ProjectStatisticsPopup/ProjectStatisticsPopup.tsx
+++ b/src/Frontend/Components/ProjectStatisticsPopup/ProjectStatisticsPopup.tsx
@@ -22,7 +22,6 @@ import {
   getIncompleteAttributionsCount,
   getMostFrequentLicenses,
   getUniqueLicenseNameToAttribution,
-  sortAttributionPropertiesEntries,
 } from './project-statistics-popup-helpers';
 import { AttributionCountPerSourcePerLicenseTable } from './AttributionCountPerSourcePerLicenseTable';
 import { AttributionPropertyCountTable } from './AttributionPropertyCountTable';
@@ -53,10 +52,6 @@ export function ProjectStatisticsPopup(): ReactElement {
 
   const manualAttributionPropertyCounts =
     aggregateAttributionPropertiesFromAttributions(manualAttributions);
-  const sortedManualAttributionPropertyCountsEntries =
-    sortAttributionPropertiesEntries(
-      Object.entries(manualAttributionPropertyCounts)
-    );
 
   const mostFrequentLicenseCountData = getMostFrequentLicenses(
     attributionCountPerSourcePerLicense
@@ -86,9 +81,9 @@ export function ProjectStatisticsPopup(): ReactElement {
           <MuiBox style={classes.panels}>
             <MuiBox style={classes.leftPanel}>
               <AttributionPropertyCountTable
-                attributionPropertyCountsEntries={
-                  sortedManualAttributionPropertyCountsEntries
-                }
+                attributionPropertyCountsEntries={Object.entries(
+                  manualAttributionPropertyCounts
+                )}
                 title={
                   ProjectStatisticsPopupTitle.AttributionPropertyCountTable
                 }
--- a/src/Frontend/Components/ProjectStatisticsPopup/__tests__/project-statistics-popup-helpers.test.tsx
+++ b/src/Frontend/Components/ProjectStatisticsPopup/__tests__/project-statistics-popup-helpers.test.tsx
@@ -6,19 +6,26 @@
 import {
   aggregateAttributionPropertiesFromAttributions,
   aggregateLicensesAndSourcesFromAttributions,
+  ATTRIBUTION_TOTAL,
+  AttributionCountPerSourcePerLicense,
+  getAttributionPropertyDisplayNameFromId,
   getColorsForPieChart,
   getCriticalSignalsCount,
   getIncompleteAttributionsCount,
+  getLicenseCriticality,
   getMostFrequentLicenses,
   getUniqueLicenseNameToAttribution,
+  LicenseNamesWithCriticality,
 } from '../project-statistics-popup-helpers';
 import {
   Attributions,
   Criticality,
+  ExternalAttributionSources,
   FollowUp,
 } from '../../../../shared/shared-types';
 import { OpossumColors } from '../../../shared-styles';
 import { ProjectStatisticsPopupTitle } from '../../../enums/enums';
+import { PieChartData } from '../../PieChart/PieChart';
 
 const testAttributions_1: Attributions = {
   uuid1: {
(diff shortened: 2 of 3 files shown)


src/Frontend/Components/ProjectStatisticsPopup/project-statistics-popup-helpers.ts
```diff
@@ -40,7 +40,7 @@ export const LICENSE_COLUMN_NAME_IN_TABLE = 'License name';
 export const AMOUNT_COLUMN_NAME_IN_TABLE = 'Amount';
 export const PLACEHOLDER_ATTRIBUTION_COUNT = '-';
 
-const UNKNOWN_SOURCE_PLACEHOLDER = '-';
+export const UNKNOWN_SOURCE_PLACEHOLDER = '-';
```
benedikt-richter, line 43: This constant should not be exported for testing. Better to hardcode it in tests.

src/Frontend/Components/ProjectStatisticsPopup/project-statistics-popup-helpers.ts
```diff
@@ -183,7 +183,7 @@ function getLicenseCriticality(licenseCriticalityCounts: {
     : undefined;
 }
 
-function getSourceDisplayNameFromId(
+export function getSourceDisplayNameFromId(
```
benedikt-richter, line 186: We should not export functions only to test them (except there are really good reasons). 
Functions that are not exported can be considered implementation details. These functions are tested implicitly in the tests of the functions that are exported.

src/Frontend/Components/ProjectStatisticsPopup/__tests__/project-statistics-popup-helpers.test.tsx
```diff
@@ -577,13 +456,44 @@ describe('The ProjectStatisticsPopup helper', () => {
...
+        count: 4,
+      },
+    ];
+    const incompleteAttributionCount =
+      getIncompleteAttributionsCount(testAttributions_1);
+    expect(incompleteAttributionCount).toEqual(
+      expectedIncompleteAttributionCount
+    );
+  });
+  it('counts only incomplete attributions', () => {
```
benedikt-richter, line 486: Typically we have empty lines between tests

src/Frontend/Components/ProjectStatisticsPopup/ProjectStatisticsPopup.tsx
```diff
@@ -53,10 +52,6 @@ export function ProjectStatisticsPopup(): ReactElement {
 
   const manualAttributionPropertyCounts =
     aggregateAttributionPropertiesFromAttributions(manualAttributions);
-  const sortedManualAttributionPropertyCountsEntries =
```
benedikt-richter, line 56: 👍

src/Frontend/Components/ProjectStatisticsPopup/__tests__/project-statistics-popup-helpers.test.tsx
```diff
@@ -111,224 +118,15 @@ const testAttributions_2: Attributions = {
...
-      name: 'SC',
-      documentConfidence: 10,
-    },
-    criticality: Criticality.Medium,
-    licenseName: 'Apache License Version 1.0.0.0.1',
-    firstParty: true,
-  },
-};
-
 const attributionSources = {
```
benedikt-richter, line 121: All constants in tests should have an explicit type: `const attributionSources: Type = {`
This helps a lot during refactorings.

benedikt-richter (changes requested):
This is a first round of review. Will look at the actual test implementations later.

src/Frontend/Components/ProjectStatisticsPopup/__tests__/project-statistics-popup-helpers.test.tsx
```diff
@@ -347,100 +144,241 @@ describe('The ProjectStatisticsPopup helper', () => {
...
+    );
+    expect(sourceName).toEqual(expectedSourceName);
+  });
+});
+
+describe('getUniqueLicenseNameToAttribution', () => {
+  it('aggregates attributions by stripped license names', () => {
+    const expectedStrippedLicenseNameToAttribution = {
+      'apachelicenseversion2.0': ['uuid1', 'uuid3', 'uuid4'],
+      'themitlicense(mit)': ['uuid2', 'uuid5'],
```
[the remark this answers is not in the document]
MarkusObendrauf: There should definitely be a test the grouping though. Eg. test that "Apache License 2.0" is the same as "aPaChE lIcEnSe 2.0" but not "Apache License 3.0"

src/Frontend/Components/ProjectStatisticsPopup/__tests__/project-statistics-popup-helpers.test.tsx
```diff
@@ -347,100 +144,241 @@ describe('The ProjectStatisticsPopup helper', () => {
...
+      expectedAttributionPropertyCounts
+    );
+    expect(
+      attributionPropertyCounts[attributionPropertyCounts.length - 1][0]
+    ).toEqual(ATTRIBUTION_TOTAL);
+  });
+});
+
+describe('getAttributionPropertyDisplayNameFromId', () => {
+  it('get valid display name, for example firstParty', () => {
```
MarkusObendrauf, line 295: remove "for example firstParty"

MarkusObendrauf (changes requested):
Looks a lot better!

**Answered by @someshkhandelia**: Thank you @MarkusObendrauf

src/Frontend/Components/ProjectStatisticsPopup/project-statistics-popup-helpers.ts
someshkhandelia: Had not hardcoded it and reused it so that if the value is changed, the test is auto-updated. But I understand your point, the test should then fail and we should not mask any changes done in the code.

someshkhandelia: Fixed.

src/Frontend/Components/ProjectStatisticsPopup/__tests__/project-statistics-popup-helpers.test.tsx
someshkhandelia: Fixed.

src/Frontend/Components/ProjectStatisticsPopup/ProjectStatisticsPopup.tsx
someshkhandelia: Thanks.

src/Frontend/Components/ProjectStatisticsPopup/project-statistics-popup-helpers.ts
```diff
@@ -247,19 +245,6 @@ export function getAttributionPropertyDisplayNameFromId(
...
-  // Move ATTRIBUTION_TOTAL to the end of the list
-  /* eslint-disable @typescript-eslint/no-unused-vars */
-  return attributionPropertiesOrTotalEntries
-    .slice()
-    .sort(([property1, _count1], [_property2, _count2]) =>
-      property1 === ATTRIBUTION_TOTAL ? 1 : 0
-    );
-  /* eslint-enable @typescript-eslint/no-unused-vars */
-}
-
```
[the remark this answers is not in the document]
someshkhandelia: Yeah. Thanks.

src/Frontend/Components/ProjectStatisticsPopup/project-statistics-popup-helpers.ts
someshkhandelia: Do not completely agree with you here. It may go against the spirit of unit testing where every function should be tested.
What can be done is to follow a convention (like we do in Python), whereby we can begin a function name with underscore '_' to indicate it is private and should never be exported. If this is not the case, all other functions should be part of unit testing.

src/Frontend/Components/ProjectStatisticsPopup/__tests__/project-statistics-popup-helpers.test.tsx
```diff
@@ -347,100 +144,241 @@ describe('The ProjectStatisticsPopup helper', () => {
...
     );
     expect(licenseNamesWithCriticality).toEqual(
       expectedLicenseNamesWithCriticality
     );
   });
+});
 
-  it('counts sources for licenses, criticalities - testAttributions_1', () => {
+describe('getLicenseDataFromVariants', () => {
+  it('Obtain license information from attribution id', () => {
```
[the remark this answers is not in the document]
someshkhandelia: Makes sense. Good point. Fixed. Thanks.

src/Frontend/Components/ProjectStatisticsPopup/__tests__/project-statistics-popup-helpers.test.tsx
someshkhandelia: Further improved: kept separation between between 3 distinct parts of a test, namely "test data", "call to tested function", and the "expect statements". All tests now follow the same structure.

src/Frontend/Components/ProjectStatisticsPopup/project-statistics-popup-helpers.ts
benedikt-richter: I do not agree that testing ALL functions is the best way to write tests. Actually testing implementation details has serious drawbacks. Maybe you can convince me that I'm wrong and that the advantages outweight the disadvantages. Otherwise, I would like to stick to the typical pattern for writing tests that we use (almost) consistently in the project.

benedikt-richter (changes requested):
see comment on testing

**Answered by @someshkhandelia**: can appreciate the point made by you. Will find some optimal solution instead of extreme one (don't test any/ test ALL). Please see reply to comment as well.

src/Frontend/Components/ProjectStatisticsPopup/project-statistics-popup-helpers.ts
someshkhandelia: Ok. I agree with you. Maybe ALL functions is not a good idea. Will try to find an optimal solution (test some and not test others). I think some functions might need to be tested (even if they are not being explicitly exported)

src/Frontend/Components/ProjectStatisticsPopup/__tests__/project-statistics-popup-helpers.test.tsx
```diff
@@ -347,100 +144,241 @@ describe('The ProjectStatisticsPopup helper', () => {
...
+  it('obtain high criticality', () => {
+    const expectedLicenseCriticality = Criticality.High;
+    const licenseCriticalityCounts = { high: 1, medium: 3, none: 1 };
+    const licenseCriticality = getLicenseCriticality(licenseCriticalityCounts);
+    expect(licenseCriticality).toEqual(expectedLicenseCriticality);
+  });
+});
+
+describe('getSourceDisplayNameFromId', () => {
+  it('obtain ScanCode display name', () => {
```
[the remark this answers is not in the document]
someshkhandelia: Fixed.

src/Frontend/Components/ProjectStatisticsPopup/__tests__/project-statistics-popup-helpers.test.tsx
someshkhandelia: Fixed.

src/Frontend/Components/ProjectStatisticsPopup/__tests__/project-statistics-popup-helpers.test.tsx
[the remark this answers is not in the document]
someshkhandelia: That is already being tested I think.

"Apache License 2.0" is different from "Apache License 3.0" in the same way as "Apache License 2.0" is different from "The MIT License" : the difference in characters basically, **which is tested in different keys of the test data**.

"Apache License 2.0" is the same as "aPaChE lIcEnSe 2.0": **tested in the same key of the test data, with different uuid**

src/Frontend/Components/ProjectStatisticsPopup/project-statistics-popup-helpers.ts
```diff
@@ -217,9 +217,7 @@ export function getUniqueLicenseNameToAttribution(
 
 export function aggregateAttributionPropertiesFromAttributions(
   attributions: Attributions
-): {
-  [attributionPropertyOrTotal: string]: number;
-} {
+): Array<Array<string | number>> {
```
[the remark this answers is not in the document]
someshkhandelia: Actually there wasn't a type change. There used to be a function that took in some values and sorted them and returned them. However, the input used to be sorted already, and so the need for the function vanished. The function was therefore removed.

However the removed function's output was used as a prop in a component, and therefore to keep the API usage as it is, this function's output type was modified to be the same as the removed function's output type.

src/Frontend/Components/ProjectStatisticsPopup/project-statistics-popup-helpers.ts
```diff
@@ -82,7 +82,7 @@ export function aggregateLicensesAndSourcesFromAttributions(
   return { attributionCountPerSourcePerLicense, licenseNamesWithCriticality };
 }
 
-function getLicenseDataFromAttributionsAndSources(
+export function getLicenseDataFromAttributionsAndSources(
```
benedikt-richter, line 85: This function has no branches. It will always be tested in the tests of the function that uses it. It seems not necessary to export it.

src/Frontend/Components/ProjectStatisticsPopup/project-statistics-popup-helpers.ts
```diff
@@ -120,7 +120,7 @@ function getLicenseDataFromAttributionsAndSources(
   return { attributionCountPerSourcePerLicense, licenseNamesWithCriticality };
 }
 
-function getLicenseDataFromVariants(
+export function getLicenseDataFromVariants(
```
benedikt-richter, line 123: This function has almost no branches. It will always be tested in the tests of the function that uses it. It seems not necessary to export it.

src/Frontend/Components/ProjectStatisticsPopup/project-statistics-popup-helpers.ts
```diff
@@ -170,7 +170,7 @@ function getLicenseDataFromVariants(
   };
 }
 
-function getLicenseCriticality(licenseCriticalityCounts: {
+export function getLicenseCriticality(licenseCriticalityCounts: {
```
benedikt-richter, line 173: it would be good to add a comment that the function is exported only for testing

src/Frontend/Components/ProjectStatisticsPopup/project-statistics-popup-helpers.ts
benedikt-richter, line 186: I think this one should not be exported for testing. Actually, part of it's logic is "hidden" at the place it is invoked...

```

const sourceName = getSourceDisplayNameFromId(
        attributions[attributionId].source?.name ?? UNKNOWN_SOURCE_PLACEHOLDER,
        attributionSources
);

```

To be honest, I think this function should not exist and be inlined or at least it should take `attributions[attributionId].source` as input insterad of `attributions[attributionId].source?.name ?? UNKNOWN_SOURCE_PLACEHOLDER`

benedikt-richter (changes requested):

src/Frontend/Components/ProjectStatisticsPopup/__tests__/project-statistics-popup-helpers.test.tsx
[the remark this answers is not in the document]
someshkhandelia: Test removed, as per discussion, since it is an implementation detail and is being tested elsewhere.

src/Frontend/Components/ProjectStatisticsPopup/project-statistics-popup-helpers.ts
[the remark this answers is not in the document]
someshkhandelia: Fixed as per verbal discussion.

src/Frontend/Components/ProjectStatisticsPopup/project-statistics-popup-helpers.ts
someshkhandelia: Fixed.

src/Frontend/Components/ProjectStatisticsPopup/project-statistics-popup-helpers.ts
someshkhandelia: Fixed.

src/Frontend/Components/ProjectStatisticsPopup/project-statistics-popup-helpers.ts
someshkhandelia: Done.

src/Frontend/Components/ProjectStatisticsPopup/project-statistics-popup-helpers.ts
someshkhandelia: Removed export.

someshkhandelia: Inlined the function behaviour where it was being called.

someshkhandelia: Fixed.

src/Frontend/Components/ProjectStatisticsPopup/__tests__/project-statistics-popup-helpers.test.tsx
someshkhandelia: Fixed.

MarkusObendrauf (approved):
Nice. I personally think that some more tests could have been kept, lets see what Benedikt thinks.

**Answered by @someshkhandelia**: yes, it was a little painful for me to have written so many new tests for functions and then to have deleted them. More so, because of our single commit squashing policy, all that work is now lost and cannot be recovered in future :(  
But on the other hand it is also true that those functions were implementation details and are being tested as part of other functions. So I guess it's ok in a way.

src/Frontend/Components/ProjectStatisticsPopup/__tests__/project-statistics-popup-helpers.test.tsx
```diff
@@ -111,336 +115,204 @@ const testAttributions_2: Attributions = {
...
-        count: 3,
-      },
-      {
-        name: 'The MIT License (MIT)',
-        count: 3,
-      },
-    ];
+describe('aggregateLicensesAndSourcesFromAttributions', () => {
+  it('counts sources for licenses, criticality', () => {
+    const expectedAttributionCountPerSourcePerLicense = {
```
benedikt-richter, line 127: there should be type annotations for all constants used in the tests, whenever there exist defined types that we can use.

src/Frontend/Components/ProjectStatisticsPopup/ProjectStatisticsPopup.tsx
```diff
@@ -86,9 +81,9 @@ export function ProjectStatisticsPopup(): ReactElement {
           <MuiBox style={classes.panels}>
             <MuiBox style={classes.leftPanel}>
               <AttributionPropertyCountTable
-                attributionPropertyCountsEntries={
-                  sortedManualAttributionPropertyCountsEntries
-                }
+                attributionPropertyCountsEntries={Object.entries(
```
benedikt-richter, line 84: This should be done in `aggregateAttributionPropertiesFromAttributions`.

src/Frontend/Components/ProjectStatisticsPopup/__tests__/project-statistics-popup-helpers.test.tsx
```diff
@@ -111,336 +115,204 @@ const testAttributions_2: Attributions = {
...
-      'apachelicenseversion2.0': ['uuid1', 'uuid3', 'uuid4'],
-      'themitlicense(mit)': ['uuid2', 'uuid5'],
+describe('aggregateAttributionPropertiesFromAttributions', () => {
+  it('counts attribution properties', () => {
+    const expectedAttributionPropertyCounts = {
+      followUp: 1,
+      firstParty: 2,
+      incomplete: 4,
+      [ATTRIBUTION_TOTAL]: 6,
     };
```
benedikt-richter, line 191: why was the return type changed from Array?

src/Frontend/Components/ProjectStatisticsPopup/__tests__/project-statistics-popup-helpers.test.tsx
```diff
@@ -111,336 +115,204 @@ const testAttributions_2: Attributions = {
...
+    const attributionPropertyDisplayName =
+      getAttributionPropertyDisplayNameFromId(propertyID);
+
+    expect(attributionPropertyDisplayName).toEqual(expectedDisplayName);
   });
+});
 
-  it('counts sources for licenses, criticalities - testAttributions_1', () => {
+describe('getMostFrequentLicenses', () => {
+  it('obtains most frequent licenses without Other accumulation', () => {
```
benedikt-richter, line 241: "Other" -> "other"

src/Frontend/Components/ProjectStatisticsPopup/__tests__/project-statistics-popup-helpers.test.tsx
```diff
@@ -111,336 +115,204 @@ const testAttributions_2: Attributions = {
...
       attributionCountPerSourcePerLicense
     );
+
     expect(sortedMostFrequentLicenses).toEqual(
       expectedSortedMostFrequentLicenses
     );
   });
 
-  it('counts sources for licenses, criticalities - testAttributions_1', () => {
+  it('obtains most frequent licenses with Other accumulation', () => {
```
benedikt-richter, line 267: "Other" -> "other"

benedikt-richter (changes requested):
some last comments.

Tests look much better now 👍

src/Frontend/Components/ProjectStatisticsPopup/ProjectStatisticsPopup.tsx
someshkhandelia: Had made changes as per: https://github.com/opossum-tool/OpossumUI/pull/1165#discussion_r1028080961 and a following verbal discussion with @MarkusObendrauf

src/Frontend/Components/ProjectStatisticsPopup/__tests__/project-statistics-popup-helpers.test.tsx
someshkhandelia: Had made changes as per: https://github.com/opossum-tool/OpossumUI/pull/1165#discussion_r1028080961 and a following verbal discussion with @MarkusObendrauf

benedikt-richter: It would be great to test at some place that the order (that is important for the display) is correct. Currently there is no test verifying that the order is correct.

src/Frontend/Components/ProjectStatisticsPopup/__tests__/project-statistics-popup-helpers.test.tsx
someshkhandelia: Fixed.

src/Frontend/Components/ProjectStatisticsPopup/__tests__/project-statistics-popup-helpers.test.tsx
someshkhandelia: Fixed.

src/Frontend/Components/ProjectStatisticsPopup/ProjectStatisticsPopup.tsx
someshkhandelia: Made changes according to verbal discussion today.

src/Frontend/Components/ProjectStatisticsPopup/__tests__/project-statistics-popup-helpers.test.tsx
someshkhandelia: Made changes according to verbal discussion today.

src/Frontend/Components/ProjectStatisticsPopup/__tests__/project-statistics-popup-helpers.test.tsx
someshkhandelia: Fixed.

benedikt-richter (approved):

MarkusObendrauf (approved):

PR #1171, opened 2022-11-18
Make license pie chart more stylish

leslielazzarino:

### Summary of changes
- shades of blue instead of unrelated, high saturation colors
- moved to the pie chart, as in the app we want to use as much as possible a limited number of shades, to have a more coherent look
- removed unused colors from the palette

New
NewPieChart
Old
OldPieChart

I think the previous color scheme did look unpolished.

Go to the  projectstatistics popup.


src/Frontend/shared-styles.ts
```diff
@@ -14,31 +16,22 @@ export const OpossumColors = {
   middleBlueOnHover: 'hsl(220, 41%, 60%)',
   darkBlue: 'hsl(220, 41%, 41%)',
   darkBlueOnHover: 'hsl(220, 41%, 65%)',
-  white: 'hsl(0, 0%, 100%)',
-  whiteOnHover: 'hsl(220, 41%, 65%)',
+  disabledButtonGrey: 'hsla(0, 0%, 0%, 0.13)',
+  lightGrey: 'hsla(0, 0%, 0%, 0.16)',
```
leslielazzarino, line 20: I wonder is we really need both line 19 and 20.

benedikt-richter: would be nice to get rid of one of these two

benedikt-richter (changes requested):
For me as a user it is very difficult to map the segments of the pie charts to the elements of the legend. Actually, I had to compare the ordering on both to identify or to hover. 
Usability (for me) certainly dropped.

vasily-pozdnyakov:
For me it is quite difficult to distinguish between the colors for complete/incomplete atrributions.
image

sebathomas:
Just looking at the screenshot, why keep the critical licenses in that yellow-orange color? It's a bit hard to read and doesn't provide any information if they're all the same color.

PR #1205, opened 2022-11-30
Do not call the resource popup when it is not shown.

leslielazzarino:
### Summary of changes
Do not call the resource popup when it is not shown.

It just makes the app slow. Comparison: 
OpossumUIvsResourcePopupLogsComparison


benedikt-richter (approved):
nice catch 👍

lennartclaas:
Very nice.

lennartclaas (approved):

PR #1206, opened 2022-11-30
Package cards should not depend on temporary package info

leslielazzarino:
### Summary of changes

Remove temporary package info dependency from Package cards.

Package cards should not depend on temporary package info. That is just display information, until it gets saved it should not influence anything. That also triggers a re-render of all package cards every time a character is written in a field, making the input slow.


benedikt-richter:
nice 👍

benedikt-richter (approved):

vasily-pozdnyakov (approved):
Nice!

PR #1208, opened 2022-11-30
Fix 1198 preselected attributions high confidence

### LINKED ISSUE #1198: Pre-selected attributions should have high confidence
Currently, signals that are pre-selected lead to pre-selected attributions. These attributions have the same confidence as the corresponding signal. Therefore, the confidence can be higher than 80 (for us that corresponds to "high").
Actually, pre-selected attributions should always have high (80) confidence.
See screenshot below for a pre-selected attribution with confidence 100 (should be 80).

Screenshot 2022-11-29 at 09 45 04

vasily-pozdnyakov:
### Summary of changes

Every pre-selected attribution have confidence of 80, which is saved in attribution file.

Fix #1198

yarn test:changed
Check confidence of pre-selected attributions.


src/ElectronBackend/input/importFromFile.ts
```diff
@@ -189,6 +189,12 @@ function createOutputFile(
     )
   );
 
+  for (const attributionId of Object.keys(preselectedAttributions)) {
+    if (preselectedAttributions[attributionId].preSelected) {
```
benedikt-richter, line 193: shouldn't this always be true for pre-selected attributions?

src/ElectronBackend/input/importFromFile.ts
```diff
@@ -189,6 +189,12 @@ function createOutputFile(
...
+      preselectedAttributions[attributionId].attributionConfidence = 80;
```
benedikt-richter, line 194: `80` -> `DiscreteConfidence.High`

benedikt-richter (changes requested):
Just two small points

src/ElectronBackend/input/importFromFile.ts
vasily-pozdnyakov: right.

benedikt-richter (changes requested):
Looks good. Just squash commits.

FlorianSchepers (approved):
lgtm

benedikt-richter (approved):

PR #1215, opened 2022-12-02
Run tests only on ubuntu

benedikt-richter:
### Summary of changes

Run tests only on ubuntu in Github actions.

The unit and integration tests run in our Github actions on mac, ubuntu, and windows. However, we do not expect a lot of differences there. For this reason and to use less resources, the Ci is changed.


lennartclaas (approved):
It looks reasonable.  👍

Issue #993, opened 2022-09-19
Show small resource tree instead of plain list of resources


benedikt-richter:
Currently, we show a plain list of resources in the "Resources for selected attribution" popup:

![Screenshot 2022-09-19 at 15 11 14](https://user-images.githubusercontent.com/46576389/191025273-cbe5549f-1b10-4047-8338-1e49c2c04975.png)

Instead a small resource tree containing just these resources should be displayed (basically a filtered tree of all resources). This is similar to the case for the attribution view: #962.

PR #1224, opened 2022-12-05
Replace resource popup list with a tree.

lennartclaas:
### Summary of changes
The resource popup list was replaced with a tree. 

The list shown when opening the resource popup looked quite unappealing. The new tree view provides improved navigation. By default, the tree opens with all affected nodes expanded.

### Further comments
The icon design is still rather sparse and could be improved in another ticket, if necessary.

By default, all nodes in the tree are expanded when the popup is opened. This strategy has been chosen to minimize clicking work. Therefore, the tree can become very large if many resources are affected by the underlying signal/attribution. However, the whole tree can be contracted by clicking the arrow of the root node. Afterwards, the tree can be expanded manually.

Before:
0_opossumui_resources_popup_list

After:
0_opossumui_resources_popup_tree_final

Sorry for the large number of changed lines. However, some stuff was just copied from ResourceBrowser.tsx to create the VistualizedTree.

Fix: #993

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,5 @@ yarn-error.log*
 .idea
 /opossumUI.iml
 /notices
+
+/.vscode
--- a/src/Frontend/Components/ResourcePathPopup/ResourcePathPopup.tsx
+++ b/src/Frontend/Components/ResourcePathPopup/ResourcePathPopup.tsx
@@ -3,31 +3,40 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
-import React, { ReactElement } from 'react';
+import React, { ReactElement, useState } from 'react';
 import { NotificationPopup } from '../NotificationPopup/NotificationPopup';
-import { ResourcesList } from '../ResourcesList/ResourcesList';
 import {
   getExternalAttributionsToResources,
   getManualAttributionsToResources,
 } from '../../state/selectors/all-views-resource-selectors';
 import { useWindowHeight } from '../../util/use-window-height';
-import { getSelectedResourceId } from '../../state/selectors/audit-view-resource-selectors';
-import { splitResourceIdsToCurrentAndOtherFolder } from './resource-path-popup-helpers';
-import { ResourcesListBatch } from '../../types/types';
-import { useAppSelector } from '../../state/hooks';
+import {
+  getInitialExpandedIds,
+  getResourcesFromResourcePaths,
+} from './resource-path-popup-helpers';
+import { useAppDispatch, useAppSelector } from '../../state/hooks';
 import { ButtonText } from '../../enums/enums';
 import MuiBox from '@mui/material/Box';
+import { VirtualizedTree } from '../../extracted/VirtualisedTree/VirtualizedTree';
+import { treeClasses } from '../../shared-styles';
+import {
+  getAttributionBreakpoints,
+  getFilesWithChildren,
+} from '../../state/selectors/all-views-resource-selectors';
+import { getFileWithChildrenCheck } from '../../util/is-file-with-children';
+import { getTreeItemLabel } from '../ResourceBrowser/get-tree-item-label';
+import { getAttributionBreakpointCheck } from '../../util/is-attribution-breakpoint';
+import { Resources } from '../../../shared/shared-types';
+import { navigateToSelectedPathOrOpenUnsavedPopup } from '../../state/actions/popup-actions/popup-actions';
+import { remove } from 'lodash';
+import { getSelectedResourceId } from '../../state/selectors/audit-view-resource-selectors';
+import { EMPTY_ATTRIBUTION_DATA } from '../../shared-constants';
 
-const heightOffset = 300;
-
-const classes = {
-  header: {
-    whiteSpace: 'nowrap',
-  },
-  resourceListContainer: {
-    overflowY: 'hidden',
-  },
-};
+const ROOT_FOLDER_LABEL = '';
+const TREE_ROW_HEIGHT = 20;
+const VERTICAL_SPACE_BETWEEN_TREE_AND_VIEWPORT_EDGES = 236;
+const HORIZONTAL_SPACE_BETWEEN_TREE_AND_VIEWPORT_EDGES = 112;
+const POPUP_CONTENT_PADDING = 48;
 
 interface ResourcePathPopupProps {
   closePopup(): void;
@@ -37,44 +46,78 @@ interface ResourcePathPopupProps {
 }
 
 export function ResourcePathPopup(props: ResourcePathPopupProps): ReactElement {
-  const folderPath = useAppSelector(getSelectedResourceId);
-
   const externalAttributionsToResources = useAppSelector(
     getExternalAttributionsToResources
   );
   const manualAttributionsToResources = useAppSelector(
     getManualAttributionsToResources
   );
+  const selectedResourceId = useAppSelector(getSelectedResourceId);
+  const attributionBreakpoints = useAppSelector(getAttributionBreakpoints);
+  const filesWithChildren = useAppSelector(getFilesWithChildren);
+
+  function getTreeItemLabelGetter() {
+    return (
+      resourceName: string,
+      resource: Resources | 1,
+      nodeId: string
+    ): ReactElement =>
+      getTreeItemLabel(
+        resourceName,
+        resource,
+        nodeId,
+        {},
+        {},
+        {},
+        {},
+        {},
+        new Set<string>(),
+        getAttributionBreakpointCheck(attributionBreakpoints),
+        getFileWithChildrenCheck(filesWithChildren),
+        EMPTY_ATTRIBUTION_DATA
+      );
+  }
+
   const allResourceIds = props.isExternalAttribution
     ? externalAttributionsToResources[props.attributionId]
     : manualAttributionsToResources[props.attributionId];
 
+  const initialExpandedIds = getInitialExpandedIds(allResourceIds);
+
+  const [expandedIds, setExpandedIds] = useState<string[]>(initialExpandedIds);
+
+  function handleToggle(nodeIdsToExpand: Array<string>): void {
+    let newExpandedNodeIds = [...expandedIds];
+    if (expandedIds.includes(nodeIdsToExpand[0])) {
+      remove(newExpandedNodeIds, (nodeId: string): boolean =>
+        nodeId.startsWith(nodeIdsToExpand[0])
+      );
+    } else {
+      newExpandedNodeIds = newExpandedNodeIds.concat(nodeIdsToExpand);
+    }
+    setExpandedIds(newExpandedNodeIds);
+  }
+
+  const dispatch = useAppDispatch();
+
+  function handleSelect(
+    event: React.ChangeEvent<unknown>,
+    nodeId: string
+  ): void {
+    dispatch(navigateToSelectedPathOrOpenUnsavedPopup(nodeId));
+  }
+
+  const maxTreeHeight: number =
+    useWindowHeight() - VERTICAL_SPACE_BETWEEN_TREE_AND_VIEWPORT_EDGES;
   const header = `Resources for selected ${
     props.isExternalAttribution ? 'signal' : 'attribution'
   }`;
-
-  function getResourcesListBatches(): Array<ResourcesListBatch> {
-    const resourcesListBatches: Array<ResourcesListBatch> = [];
-
-    const { currentFolderResourceIds, otherFolderResourceIds } =
-      splitResourceIdsToCurrentAndOtherFolder(allResourceIds, folderPath);
-
-    resourcesListBatches.push({ resourceIds: currentFolderResourceIds });
-    if (otherFolderResourceIds.length > 0) {
-      resourcesListBatches.push({
-        header: 'Resources in Other Folders',
-        resourceIds: otherFolderResourceIds,
-      });
-    }
-    return resourcesListBatches;
-  }
-
-  const resourcesListBatches = getResourcesListBatches();
+  const resources = getResourcesFromResourcePaths(allResourceIds);
 
   return (
     <NotificationPopup
       header={header}
-      headerSx={classes.header}
+      headerSx={treeClasses.header(POPUP_CONTENT_PADDING)}
       rightButtonConfig={{
         onClick: props.closePopup,
         buttonText: ButtonText.Close,
@@ -82,15 +125,41 @@ export function ResourcePathPopup(props: ResourcePathPopupProps): ReactElement {
       onBackdropClick={props.closePopup}
       onEscapeKeyDown={props.closePopup}
       content={
-        <MuiBox sx={classes.resourceListContainer}>
-          <ResourcesList
-            resourcesListBatches={resourcesListBatches}
-            maxHeight={useWindowHeight() - heightOffset}
+        <MuiBox
+          sx={treeClasses.treeContainer(
+            VERTICAL_SPACE_BETWEEN_TREE_AND_VIEWPORT_EDGES
+          )}
+        >
+          <VirtualizedTree
+            expandedIds={expandedIds}
+            isFakeNonExpandableNode={getFileWithChildrenCheck(
+              filesWithChildren
+            )}
+            onSelect={handleSelect}
+            onToggle={handleToggle}
+            nodes={{ [ROOT_FOLDER_LABEL]: resources }}
+            selectedNodeId={selectedResourceId}
+            ariaLabel={'resource browser'}
+            getTreeNodeLabel={getTreeItemLabelGetter()}
+            cardHeight={TREE_ROW_HEIGHT}
+            maxHeight={maxTreeHeight}
+            sx={treeClasses.tree(
+              'popup',
+              HORIZONTAL_SPACE_BETWEEN_TREE_AND_VIEWPORT_EDGES,
+              POPUP_CONTENT_PADDING
+            )}
+            alwaysShowHorizontalScrollBar={true}
+            treeNodeStyle={{
+              root: treeClasses.treeItemLabel,
+              childrenOfSelected: treeClasses.treeItemLabelChildrenOfSelected,
+              selected: treeClasses.treeItemLabelSelected,
+              treeExpandIcon: treeClasses.treeExpandIcon,
+            }}
           />
         </MuiBox>
       }
       isOpen={true}
-      fullWidth={true}
+      fullWidth={false}
     />
   );
 }
--- a/src/Frontend/Components/ResourcePathPopup/__tests__/ResourcePathPopup.test.tsx
+++ b/src/Frontend/Components/ResourcePathPopup/__tests__/ResourcePathPopup.test.tsx
@@ -4,10 +4,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import React, { ReactElement } from 'react';
-import {
-  createTestAppStore,
-  renderComponentWithStore,
-} from '../../../test-helpers/render-component-with-store';
+import { renderComponentWithStore } from '../../../test-helpers/render-component-with-store';
 import { doNothing } from '../../../util/do-nothing';
 import { ResourcePathPopup } from '../ResourcePathPopup';
 import {
@@ -18,7 +15,6 @@ import {
   setExternalData,
   setManualData,
 } from '../../../state/actions/resource-actions/all-views-simple-actions';
-import { setSelectedResourceId } from '../../../state/actions/resource-actions/audit-view-simple-actions';
 import { screen } from '@testing-library/react';
 import { useAppDispatch } from '../../../state/hooks';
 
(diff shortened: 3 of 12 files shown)


src/Frontend/Components/ResourcePathPopup/ResourcePathPopup.tsx
```diff
@@ -49,27 +100,74 @@ export function ResourcePathPopup(props: ResourcePathPopupProps): ReactElement {
     ? externalAttributionsToResources[props.attributionId]
     : manualAttributionsToResources[props.attributionId];
 
+  console.time('convertResourcesList');
+  const resources = getResourcesFromResourcePaths(allResourceIds);
+  console.timeEnd('convertResourcesList');
```
lennartclaas, line 105: Already removed the time measurements.

src/Frontend/shared-styles.ts
```diff
@@ -33,7 +33,7 @@ export const OpossumColors = {
   brown: 'hsl(33, 55%, 44%)',
 };
 
-export const resourceBrowserWidthInPixels = 420;
+export const resourceBrowserWidthInPixels = 640;
```
lennartclaas, line 36: This does not seem to be very elegant. I will make a workaround for this and leave the value at 420.

vasily-pozdnyakov: It also influences resource browser in Audit view.

src/Frontend/Components/ResourcePathPopup/ResourcePathPopup.tsx
```diff
@@ -3,29 +3,80 @@
...
 
 const classes = {
   header: {
     whiteSpace: 'nowrap',
   },
   resourceListContainer: {
-    overflowY: 'hidden',
+    overflow: 'hidden',
+    height: '75vh',
+    maxHeight: 'calc(100vh - 220px)',
```
lennartclaas, line 46: Fix by Vasily to remove the unintended second vertical scroll bar when extremely reducing the window size.

src/Frontend/Components/ResourcePathPopup/ResourcePathPopup.tsx
```diff
@@ -3,29 +3,80 @@
...
-import { splitResourceIdsToCurrentAndOtherFolder } from './resource-path-popup-helpers';
-import { ResourcesListBatch } from '../../types/types';
-import { useAppSelector } from '../../state/hooks';
+import {
+  getInitialExpandedIds,
+  getResourcesFromResourcePaths,
+} from './resource-path-popup-helpers';
+import { useAppDispatch, useAppSelector } from '../../state/hooks';
 import { ButtonText } from '../../enums/enums';
 import MuiBox from '@mui/material/Box';
```
vasily-pozdnyakov, line 20: Delete the line

src/Frontend/Components/ResourcePathPopup/ResourcePathPopup.tsx
```diff
@@ -36,9 +87,10 @@ interface ResourcePathPopupProps {
   displayedAttributionName: string;
 }
 
-export function ResourcePathPopup(props: ResourcePathPopupProps): ReactElement {
-  const folderPath = useAppSelector(getSelectedResourceId);
+const TREE_ROW_HEIGHT = 20;
+const ROOT_FOLDER_LABEL = '';
```
vasily-pozdnyakov, line 91: Please move the constants above the styling classes.

src/Frontend/Components/ResourcePathPopup/ResourcePathPopup.tsx
```diff
@@ -36,9 +87,10 @@ interface ResourcePathPopupProps {
...
 
+export function ResourcePathPopup(props: ResourcePathPopupProps): ReactElement {
   const externalAttributionsToResources = useAppSelector(
```
vasily-pozdnyakov, line 94: It is very difficult to follow, could you please group the constants maybe? Like all useAppSelector next to each other, etc.

src/Frontend/Components/ResourcePathPopup/__tests__/ResourcePathPopup.test.tsx
```diff
@@ -69,31 +65,42 @@ describe('ResourcePathPopup', () => {
...
+            ['file.tsx']: 1,
+          },
+        },
+        ['abc.test.tsx']: 1,
+      },
+      ['DCO.md']: 1,
+    },
+  };
+
+  test('check structure of returned Resources object', () => {
```
vasily-pozdnyakov, line 95: test -> it

src/Frontend/Components/ResourcePathPopup/__tests__/ResourcePathPopup.test.tsx
```diff
@@ -69,31 +65,42 @@ describe('ResourcePathPopup', () => {
...
+    );
   });
 
-  it('renders no subheader, if no resources in other folders exist', () => {
-    const testStore = createTestAppStore();
-    testStore.dispatch(setSelectedResourceId('/thirdParty'));
-    renderComponentWithStore(
-      <HelperComponent isExternalAttribution={false} />,
-      { store: testStore }
+  test('check type of returned Resources object', () => {
```
vasily-pozdnyakov, line 101: You can merge the tests in one test.

src/Frontend/test-helpers/popup-test-helpers.ts
```diff
@@ -18,6 +18,15 @@ export function clickOnPathInPopupWithResources(
   fireEvent.click(getByText(popupWithResources, path));
 }
```
vasily-pozdnyakov, line 20: I guess you do not use the function clickOnPathInPopupWithResources anymore and therefore do not need it.

vasily-pozdnyakov (changes requested):
Why do we have files "/" there?
image

.gitignore
```diff
@@ -34,3 +34,5 @@ yarn-error.log*
 .idea
 /opossumUI.iml
 /notices
+
+/.vscode
```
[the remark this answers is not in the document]
lennartclaas: To prevent that auto-generated stuff from VSCode is considered for commits.

src/Frontend/Components/ResourcePathPopup/ResourcePathPopup.tsx
lennartclaas: I feel that grouping into functional chunks is more meaningful. This means that the values that we get from invoking useAppSelector should be used directly afterwards, if possible, to not "lose" the underlying functional logic / the process. However, I adjusted the order a little bit.

src/Frontend/test-helpers/popup-test-helpers.ts
lennartclaas: It felt a little bit unpleasant to remove entire functions. :D But yeah, we don't need it anymore.

lennartclaas:
@vasily-pozdnyakov I cannot reproduce the tree with "/"-files. What was your approach?

src/Frontend/Components/ResourcePathPopup/ResourcePathPopup.tsx
benedikt-richter, line 46: What is the meaning of `220px`? Is it the size of something? Would be good not to have magic numbers.

src/Frontend/Components/ResourcePathPopup/ResourcePathPopup.tsx
```diff
@@ -3,29 +3,80 @@
...
     whiteSpace: 'nowrap',
   },
   resourceListContainer: {
-    overflowY: 'hidden',
+    overflow: 'hidden',
+    height: '75vh',
+    maxHeight: 'calc(100vh - 220px)',
+  },
+  tree: {
+    width: resourceBrowserWidthInPixels + 220,
```
benedikt-richter, line 49: as above

src/Frontend/Components/ResourcePathPopup/ResourcePathPopup.tsx
```diff
@@ -3,29 +3,80 @@
...
+      backgroundColor: `${OpossumColors.lightBlueOnHover}`,
+      cursor: 'pointer',
+    },
+  },
+  treeItemLabelChildrenOfSelected: {
+    backgroundColor: `${OpossumColors.lightestBlue}`,
+    borderBottom: `1px ${OpossumColors.lightestBlue} solid`,
+  },
+  treeItemLabelSelected: {
+    backgroundColor: `${OpossumColors.lightestBlue} !important`,
```
benedikt-richter, line 65: If possible avoid "!important"

src/Frontend/Components/ResourcePathPopup/ResourcePathPopup.tsx
```diff
@@ -49,27 +101,71 @@ export function ResourcePathPopup(props: ResourcePathPopupProps): ReactElement {
...
+  const resourcesWithExternalAttributedChildren = {};
+  const resolvedExternalAttributions = new Set<string>();
 
-    resourcesListBatches.push({ resourceIds: currentFolderResourceIds });
-    if (otherFolderResourceIds.length > 0) {
-      resourcesListBatches.push({
-        header: 'Resources in Other Folders',
-        resourceIds: otherFolderResourceIds,
-      });
+  const attributionBreakpoints = useAppSelector(getAttributionBreakpoints); // vielleicht
```
benedikt-richter, line 122: remove comment

src/Frontend/Components/ResourcePathPopup/ResourcePathPopup.tsx
```diff
@@ -49,27 +101,71 @@ export function ResourcePathPopup(props: ResourcePathPopupProps): ReactElement {
...
+  const [expandedIds, setExpandedIds] = useState<string[]>(initialExpandedIds);
 
-    const { currentFolderResourceIds, otherFolderResourceIds } =
-      splitResourceIdsToCurrentAndOtherFolder(allResourceIds, folderPath);
+  const manualAttributions = {};
+  const resourcesToManualAttributions = {};
+  const resourcesWithManualAttributedChildren = {};
+  const resourcesToExternalAttributions = {};
+  const resourcesWithExternalAttributedChildren = {};
+  const resolvedExternalAttributions = new Set<string>();
```
benedikt-richter, line 120: what going on here?

src/Frontend/Components/ResourcePathPopup/ResourcePathPopup.tsx
```diff
@@ -49,27 +101,71 @@ export function ResourcePathPopup(props: ResourcePathPopupProps): ReactElement {
...
 
-    resourcesListBatches.push({ resourceIds: currentFolderResourceIds });
-    if (otherFolderResourceIds.length > 0) {
-      resourcesListBatches.push({
-        header: 'Resources in Other Folders',
-        resourceIds: otherFolderResourceIds,
-      });
+  const attributionBreakpoints = useAppSelector(getAttributionBreakpoints); // vielleicht
+  const filesWithChildren = useAppSelector(getFilesWithChildren);
+  const externalData = {} as AttributionData;
```
benedikt-richter, line 124: why do we do this?

src/Frontend/Components/ResourcePathPopup/ResourcePathPopup.tsx
```diff
@@ -49,27 +101,71 @@ export function ResourcePathPopup(props: ResourcePathPopupProps): ReactElement {
...
+        resourcesWithManualAttributedChildren,
+        resolvedExternalAttributions,
+        getAttributionBreakpointCheck(attributionBreakpoints),
+        getFileWithChildrenCheck(filesWithChildren),
+        externalData
+      );
   }
 
-  const resourcesListBatches = getResourcesListBatches();
+  const maxTreeHeight: number = useWindowHeight() - topBarHeight - 200;
```
benedikt-richter, line 168: what does the magic number `200`stand for?

src/Frontend/Components/ResourcePathPopup/__tests__/ResourcePathPopup.test.tsx
benedikt-richter, line 101: Remove this test. We should not test types.

Furthermore, this test is green for `desiredResources = {}` and `getResourcesFromResourcePaths(testResourcesList) = null` ...

src/Frontend/Components/ResourcePathPopup/resource-path-popup-helpers.ts
```diff
@@ -23,3 +25,44 @@ export function splitResourceIdsToCurrentAndOtherFolder(
...
+function addPathToResources(resourcePath: string, resources: Resources): void {
+  const resourcePathSplit = resourcePath.split('/').slice(1);
+  resourcePathSplit.reduce(
+    (parent: Resources, childName: string, index: number) => {
+      if (index == resourcePathSplit.length - 1) {
+        parent[childName] = 1;
+      } else {
+        parent[childName] = parent[childName] || {};
+      }
+      return parent[childName] as Resources;
```
benedikt-richter, line 49: avoid type casts. There is alonst no case where we should use `as`. Why is it required here?

src/Frontend/test-helpers/popup-test-helpers.ts
benedikt-richter: If it is not used, it should die :)

src/Frontend/Components/ResourcePathPopup/ResourcePathPopup.tsx
vasily-pozdnyakov: I think it should be somehow synchronized with maxTreeHeight

src/Frontend/Components/ResourcePathPopup/ResourcePathPopup.tsx
vasily-pozdnyakov: All of them are empty and used in getTreeItemLabel. Why?

src/Frontend/Components/ResourcePathPopup/ResourcePathPopup.tsx
leslielazzarino: This number should just be extracted to a constant, in order not to be a magic number. The name of the constant should explain what it is.

src/Frontend/Components/ResourcePathPopup/ResourcePathPopup.tsx
leslielazzarino: This is and will always be not what you want to do. Casting can be used, sparringly, to help the compiler when it doesn't understand what's happening with the types in your code. Never to circumvent the typing.

leslielazzarino (changes requested):

lennartclaas:
> Why do we have files "/" there? image

Resolved. Now, it looks like this:
res_tree_folders_correct

src/Frontend/Components/ResourcePathPopup/ResourcePathPopup.tsx
```diff
@@ -3,29 +3,80 @@
...
+  },
+  treeExpandIcon: {
+    width: '16px',
+    height: '20px',
+    padding: '0px',
+    margin: '0px',
+    color: OpossumColors.darkBlue,
+    '&:hover': {
+      background: OpossumColors.middleBlue,
+    },
```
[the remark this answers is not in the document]
lennartclaas: Done in latest commit.

vasily-pozdnyakov:
I see different resources lists for e.g. /OpossumUI/build/index.html. Not sure which one is correct.
image
image

src/Frontend/shared-styles.ts
```diff
@@ -93,3 +93,91 @@ export const tableClasses = {
     color: OpossumColors.black,
   },
 };
+
+interface TreeStyleBrowserVars {
```
vasily-pozdnyakov, line 97: Can the interfaces be merged?

src/Frontend/test-helpers/popup-test-helpers.ts
```diff
@@ -9,13 +9,13 @@ export function expectShowResourcesPopupVisible(screen: Screen): void {
   expect(getPopupWithResources(screen)).toBeTruthy();
 }
 
-export function clickOnPathInPopupWithResources(
+export function clickOnNodeInPopupWithResources(
   screen: Screen,
-  path: string
+  node: string
```
vasily-pozdnyakov, line 14: is it nodeId? Would be nice to be more specific.

src/Frontend/Components/ResourcePathPopup/resource-path-popup-helpers.ts
```diff
@@ -23,3 +25,47 @@ export function splitResourceIdsToCurrentAndOtherFolder(
...
+
+function addPathToResources(resourcePath: string, resources: Resources): void {
+  const isFolder = resourcePath.endsWith('/');
+  const resourcePathSplit = isFolder
+    ? resourcePath.split('/').slice(1, -1)
+    : resourcePath.split('/').slice(1);
+  resourcePathSplit.reduce(
+    (parent: Resources, childName: string, index: number) => {
+      if (index === resourcePathSplit.length - 1) {
+        parent[childName] = isFolder ? {} : 1;
```
vasily-pozdnyakov, line 48: Why can it be 1?

leslielazzarino:
Screen_Capture_-_Dec_8__11_43_AM
I don't like this behavior, the tree should use all the horizontal space it can before the need to scroll horizontally arises.

vasily-pozdnyakov (approved):
Just minor things. Well done!

src/Frontend/Components/ResourcePathPopup/resource-path-popup-helpers.ts
```diff
@@ -23,3 +25,47 @@ export function splitResourceIdsToCurrentAndOtherFolder(
...
+      } else {
+        parent[childName] = parent[childName] || {};
+      }
+      return parent[childName] as Resources;
```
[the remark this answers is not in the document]
benedikt-richter: Actually, this function can return `1`. Typescript is right. To reproduce, execute the following: `getResourcesFromResourcePaths(testResourcesList)`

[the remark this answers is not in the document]
benedikt-richter: `  return parent[childName] as Resources;` could be

```
      const returnValue = parent[childName];
      return returnValue !== 1 ? returnValue : {};
```

src/Frontend/Components/ResourcePathPopup/__tests__/ResourcePathPopup.test.tsx
```diff
@@ -69,31 +65,36 @@ describe('ResourcePathPopup', () => {
...
-    renderComponentWithStore(
-      <HelperComponent isExternalAttribution={false} />,
-      { store: testStore }
-    );
-    expect(
-      screen.queryByText(resourcesInOtherFoldersHeader)
-    ).not.toBeInTheDocument();
-    expect(screen.getByText('/thirdParty')).toBeInTheDocument();
-  });
+it('correctly converts a list of resource ids (paths) into a Resources object', () => {
```
benedikt-richter, line 96: helpers should be tested in a different file (to be created): `resource-path-popup-heklpers.test.ts`

src/Frontend/test-helpers/popup-test-helpers.ts
lennartclaas: It's not the nodeId, i.e., the full path to a node. It's just the label (or name) of a node. I changed it to `nodeLabel`.

src/Frontend/Components/ResourcePathPopup/resource-path-popup-helpers.ts
lennartclaas: This is a characteristic of the Resources interface. An objects indicates a folder, and a 1 indicates a file.

src/Frontend/shared-styles.ts
lennartclaas: I found the SxProps type and used this instead. Looks way better.

src/Frontend/Components/ResourcePathPopup/resource-path-popup-helpers.ts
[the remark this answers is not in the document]
lennartclaas: Nice solution. Thanks. 👍

src/Frontend/shared-styles.ts
```diff
@@ -93,3 +95,84 @@ export const tableClasses = {
...
+          width: resourceBrowserWidthInPixels,
+          padding: '4px 0',
+          background: OpossumColors.white,
+          height: '100%',
+        };
+      }
+      case 'popup': {
+        if (horizontalSpaceBetweenTreeAndViewportEdges !== null) {
+          return {
+            width: `calc(100vw - ${horizontalSpaceBetweenTreeAndViewportEdges}px)`,
```
leslielazzarino, line 135: Doesn't work properly with wide form factors, as the popup has a max width. Either remove the max width from the popup or introduce it here too.

leslielazzarino (changes requested):

leslielazzarino (approved):

Issue #1217, opened 2022-12-02
Disable fetch data button if no url


vasily-pozdnyakov:
Fetch data button is still clickable and has tooltip if it is greyed out (e.g. when there is no url), compare to "open url" button next to it.
![image](https://user-images.githubusercontent.com/85183359/205334139-f2fb8fde-ec33-4421-97b0-254e88132b6f.png)

benedikt-richter:
Good point! Actually, a tooltip could be shown for both. It can contain information why the respective action is not possible.

PR #1237, opened 2022-12-07
Adjust buttons of URL field in attribution column.

lennartclaas:
### Summary of changes
If disabled, the fetch button is not clickable. The fetch button and the open button do show a tooltip if they are disabled. The tooltips indicate that fetching or opening is not possible because the URL is missing or unsuitable.

Previously, the fetch button was clickable, although disabled. Furthermore, disabled buttons did not show a tooltip. However, a tooltip with information on why the buttons are disabled is helpful for unexperienced users of the app.

Tooltip of the enabled fetch button:
url_tooltip_fetch_enabled

Tooltip of the enabled open button:
url_tooltip_open_enabled

Tooltip of the disabled fetch button:
fetch

Tooltip of the disabled open button:
open

Fix: #1217

### DIFF
--- a/src/Frontend/Components/FetchLicenseInformationButton/FetchLicenseInformationButton.tsx
+++ b/src/Frontend/Components/FetchLicenseInformationButton/FetchLicenseInformationButton.tsx
@@ -52,7 +52,9 @@ const classes = {
   },
 };
 
-const FETCH_DATA_TOOLTIP = 'Fetch data';
+export const FETCH_DATA_TOOLTIP = 'Fetch data';
+export const FETCH_DATA_BUTTON_DISABLED_TOOLTIP =
+  'Fetching data is not possible. Please enter a URL with one of the following domains: pypi.org, npmjs.com, github.com.';
 
 export enum FetchStatus {
   Idle = 'Idle',
@@ -125,16 +127,16 @@ export function useFetchPackageInfo(props: LicenseFetchingInformation): {
 function DisabledFetchLicenseInformationButton(): ReactElement {
   return (
     <IconButton
-      tooltipTitle={FETCH_DATA_TOOLTIP}
+      tooltipTitle={FETCH_DATA_BUTTON_DISABLED_TOOLTIP}
       tooltipPlacement="right"
-      disabled={false}
+      disabled={true}
       onClick={doNothing}
       icon={<CachedIcon sx={classes.disabledIcon} />}
     />
   );
 }
 
-function EnabledFetchLicenseInformationButton(
+export function EnabledFetchLicenseInformationButton(
   props: LicenseFetchingInformation
 ): ReactElement {
   const { fetchStatus, errorMessage, fetchData } = useFetchPackageInfo(props);
--- a/src/Frontend/Components/FetchLicenseInformationButton/__tests__/FetchLicenseInformationButton.test.tsx
+++ b/src/Frontend/Components/FetchLicenseInformationButton/__tests__/FetchLicenseInformationButton.test.tsx
@@ -3,11 +3,13 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
-import { render, screen } from '@testing-library/react';
+import { fireEvent, render, screen, waitFor } from '@testing-library/react';
 import React, { ReactElement, ReactNode } from 'react';
 import {
   FetchLicenseInformationButton,
   FetchStatus,
+  FETCH_DATA_BUTTON_DISABLED_TOOLTIP,
+  FETCH_DATA_TOOLTIP,
   useFetchPackageInfo,
 } from '../FetchLicenseInformationButton';
 import {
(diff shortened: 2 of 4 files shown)


src/Frontend/Components/IconButton/IconButton.tsx
```diff
@@ -21,7 +21,13 @@ interface IconButtonProps {
 export function IconButton(props: IconButtonProps): ReactElement {
   function wrapInTooltip(children: ReactElement): ReactElement {
     return props.disabled ? (
-      <span>{children}</span>
+      <MuiTooltip
```
benedikt-richter, line 24: the function `wrapInTooltip`seems to be no longer required. The return value is in case of `props.disabled` being `true`or `false`the same (modulo unimportant `span`).

src/Frontend/Components/FetchLicenseInformationButton/FetchLicenseInformationButton.tsx
```diff
@@ -125,9 +125,11 @@ export function useFetchPackageInfo(props: LicenseFetchingInformation): {
 function DisabledFetchLicenseInformationButton(): ReactElement {
   return (
     <IconButton
-      tooltipTitle={FETCH_DATA_TOOLTIP}
+      tooltipTitle={
```
benedikt-richter, line 128: Would be nice to extract the tooltip text as a constant like it was the case before for `FETCH_DATA_TOOLTIP`

src/Frontend/Components/FetchLicenseInformationButton/FetchLicenseInformationButton.tsx
```diff
@@ -125,9 +125,11 @@ export function useFetchPackageInfo(props: LicenseFetchingInformation): {
...
+        'Fetching data is not possible. Please enter a suitable URL.'
```
benedikt-richter, line 129: not sure "suitable" is the correct term here.

benedikt-richter (changes requested):

benedikt-richter (changes requested):
Nice 👍 Just some minor comments

vasily-pozdnyakov:
Could you please attach screenshots of how it looks now?

someshkhandelia:
Looks nice! One small comment about the tests.

For the disabled Fetch button, you have a test to check the text of the tooltip, but the same is not true for the enabled button. Also there is a switch case for the enabled button depending on the state of loading URL. Maybe that can be tested?

src/Frontend/Components/FetchLicenseInformationButton/__tests__/FetchLicenseInformationButton.test.tsx
```diff
@@ -26,7 +26,11 @@ import { convertGithubPayload } from '../github-fetching-helpers';
...
     render(<FetchLicenseInformationButton isDisabled={true} url={''} />);
-    expect(screen.getByLabelText('Fetch data')).toBeInTheDocument();
+    expect(
+      screen.getByLabelText(
+        'Fetching data is not possible. Please enter a suitable URL.'
+      )
+    ).toBeInTheDocument();
   });
 
   it('renders enabled button', () => {
```
someshkhandelia, line 36: Looks nice! One small comment about the tests.

For the disabled Fetch button, you have a test to check the text of the tooltip, but the same is not true for the enabled button. Also there is a switch case for the enabled button depending on the state of loading URL. Maybe that can be tested?

src/Frontend/Components/IconButton/IconButton.tsx
lennartclaas: In fact, the span tag is necessary. It facilitates tooltips for disabled buttons.

src/Frontend/Components/FetchLicenseInformationButton/FetchLicenseInformationButton.tsx
lennartclaas: Suitable means, that the URL links to pypi, npm, or github. Otherwise, the button is disabled. This more specific information could also be given in the tooltip. But it would get quite lengthy.

lennartclaas:
Regarding the tests: Any idea how `setTimeout()` can be avoided? `async`, `await` and `find` instead of `get` queries do not seem to help.

lennartclaas:
The tooltip of the disabled fetch data button now looks like this:
tooltip_disabled_new

benedikt-richter:
> Regarding the tests: Any idea how `setTimeout()` can be avoided? `async`, `await` and `find` instead of `get` queries do not seem to help.

Jest fake timers might be helpful: https://jestjs.io/docs/timer-mocks

src/Frontend/Components/FetchLicenseInformationButton/__tests__/FetchLicenseInformationButton.test.tsx
lennartclaas: I added multiple tests for different tooltips.

src/Frontend/Components/IconButton/IconButton.tsx
benedikt-richter: Good point! In this case, I'd suggest to always include the span in order to reduce complexity. The component could be something like:

```
export function IconButton(props: IconButtonProps): ReactElement {
  return (
    <MuiTooltip
      sx={tooltipStyle}
      title={props.tooltipTitle}
      placement={props.tooltipPlacement}
    >
      <span>
        {
          // this span enables tooltips for disabled buttons
        }
        <MuiButtonBase
          sx={props.sx}
          onClick={(event): void => {
            event.stopPropagation();
            props.onClick();
          }}
          disabled={props.disabled}
        >
          {props.icon}
        </MuiButtonBase>
      </span>
    </MuiTooltip>
  );
}
```

benedikt-richter (changes requested):

benedikt-richter (approved):
👍

vasily-pozdnyakov:
See failed tests.

someshkhandelia (approved):

Issue #1194, opened 2022-11-28
Decouple project statistics popup from its constituents


benedikt-richter:
Currently, in `project-statistics-popup-helpers.ts` we define the following constants:

```
export const SOURCE_TOTAL = 'Total';
export const LICENSE_TOTAL = 'Total';
```
These couple these helpers to `AttributionCountPerSourcePerLicenseTable.tsx` and `CriticalLicensesTable.tsx`.

To reduce technical debt this coupling should be removed.

someshkhandelia:
- I would also draw attention to the fact that `ProjectLicensesTable` is the generic table which is already decoupled from data logic of the statistic helper. 
- The tables `AttributionCountPerSourcePerLicenseTable` and `CriticalLicensesTable` are specific to statistics popup and therefore don't seem to have the need to be decoupled.  
- `ProjectLicensesTable` was decoupled (made generic) because it was reused across `AttributionCountPerSourcePerLicenseTable` and `CriticalLicensesTable`, but the latter two tables are used in the popup only and won't be reused with other data in future as well. If other data is used in future, it would be used directly with `ProjectLicensesTable`.
- Also reiterating a convention policy: "Don't make something too generic until it is used 3 times". `AttributionCountPerSourcePerLicenseTable` and `CriticalLicensesTable` are not used more than 1 time, and even if they are reused in future > 1 time, they will be used with the same data.

benedikt-richter:
The main motivation of this ticket is to remove the coupling by "magic" keys in the data structure consumed by the two tables in question. 

In particular, the following interface causes this issue:

```
export interface AttributionCountPerSourcePerLicense {
  [licenseNameOrTotal: string]: { [sourceNameOrTotal: string]: number };
}
```

We use the constants to memorize the keys we want to treat differently. Instead the data structure should be more complex to explicitly handle the data for total values and therefore the "magic" constants are not needed anymore.

PR #1238, opened 2022-12-07
Decouple project statistics popup helper

someshkhandelia:
### Summary of changes

Decouple project statistics popup helper

Issue: #1194

### DIFF
--- a/src/Frontend/Components/AttributionCountPerSourcePerLicenseTable/AttributionCountPerSourcePerLicenseTable.tsx
+++ b/src/Frontend/Components/AttributionCountPerSourcePerLicenseTable/AttributionCountPerSourcePerLicenseTable.tsx
@@ -4,15 +4,12 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import React, { ReactElement } from 'react';
-import {
-  LICENSE_TOTAL,
-  SOURCE_TOTAL,
-} from '../ProjectStatisticsPopup/project-statistics-popup-helpers';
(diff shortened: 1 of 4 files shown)


benedikt-richter (changes requested):
Actually, these components and the helpers for the project statistics popup are still coupled (not technically anymore but logically).
If you change the value of one of the constants in one of the files the logic breaks. The logic currently relies on these constants to have equal value. E.g. in `src/Frontend/Components/AttributionCountPerSourcePerLicenseTable/AttributionCountPerSourcePerLicenseTable.tsx` in line 37 and 38 the sorting of the sources relies on this fact.

**Answered by @someshkhandelia**: The logic won't break now, since the constants are shared now.  I would also draw attention to the fact that `ProjectLicensesTable` is the generic table which is already decoupled from data logic of the statistic helper. The tables `AttributionCountPerSourcePerLicenseTable` and `CriticalLicensesTable` are specific to statistics popup and therefore don't seem to have the need to be decoupled.

benedikt-richter:
Regarding the reply made as an edit to https://github.com/opossum-tool/OpossumUI/pull/1238#pullrequestreview-1208228482.
Let's discuss the question whether to decouple these components in the respective issue https://github.com/opossum-tool/OpossumUI/issues/1194 and reserve the space here for discussions on the actual implementation.

benedikt-richter (approved):

leslielazzarino (approved):

Issue #962, opened 2022-09-12
Show file tree in Attribution view


vasily-pozdnyakov:
As a user I want to see the resource tree in the attribution view. This is to see where linked resources are located. Probably it should be "read only" and look similar to the resource tree in the Audit view.
Currently: 
![image](https://user-images.githubusercontent.com/85183359/197339583-406afb5a-e160-4938-bc8f-7b8016e10c77.png)

PR #1245, opened 2022-12-12
Replace attribution view resource list with a tree.

MarkusObendrauf:
### Summary of changes

The attribution view list was replaced with a tree, [similar to the recent changes done to the resources popup.
](https://github.com/opossum-tool/OpossumUI/pull/1224)

The list shown in the attribution view looked quite unappealing. The new tree view provides improved navigation. By default, the tree opens with all affected nodes expanded.

See https://github.com/opossum-tool/OpossumUI/issues/962

Before:
image

After:
image


src/Frontend/Components/AttributionDetailsViewer/AttributionDetailsViewer.tsx
```diff
@@ -91,17 +89,18 @@ export function AttributionDetailsViewer(): ReactElement | null {
     }
   }
 
+  const maxTreeHeight: number = useWindowHeight() - 120;
```
lennartclaas, line 92: Same as above regarding the 120. This value has some kind of origin (usually padding or margins). Would make sense to define a constant, so that this value can be comprehended in the future.

src/Frontend/shared-styles.ts
```diff
@@ -109,23 +114,29 @@ export const treeClasses = {
   },
   treeContainer: (
     verticalSpaceBetweenTreeAndViewportEdges: number
-  ): SxProps => {
+  ): TreeContainerStyleVars => {
```
lennartclaas, line 117: Although I get the idea, I don't quite understand why this is necessary. If one would be consistent, corresponding interfaces should also be defined for the functions returning the SxProps for header and tree.  Since the tree function can return three different objects, this would be quite a mess. However, this is more like an opinion rather than a change request.

src/Frontend/Components/AttributionDetailsViewer/AttributionDetailsViewer.tsx
MarkusObendrauf: extracted it to VERTICAL_SPACE_OF_LINKED_RESOURCE_HEADER_AND_FOOTER

src/Frontend/shared-styles.ts
MarkusObendrauf: you are right, reverted it

src/Frontend/Components/AttributionDetailsViewer/AttributionDetailsViewer.tsx
```diff
@@ -40,7 +40,7 @@ const classes = {
     display: 'flex',
     flexDirection: 'column',
     width: '30%',
-    height: '100%',
+    height: 'calc(100% - 48px)',
```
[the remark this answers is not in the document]
MarkusObendrauf: deined two consts:
`const VERTICAL_RESOURCE_COLUMN_PADDING = 48;
const VERTICAL_RESOURCE_HEADER_AND_FOOTER_SIZE = 72;`

The const `VERTICAL_RESOURCE_COLUMN_PADDING` is used for the `resourceColumn` class, while `VERTICAL_RESOURCE_COLUMN_PADDING + VERTICAL_RESOURCE_HEADER_AND_FOOTER_SIZE ` is used for the height of the graph itself. I think it makes sense to add these two variables together, because the graph size is dependant on both - this way you can increase `VERTICAL_RESOURCE_COLUMN_PADDING` and the graph height will change appropriately.

lennartclaas (approved):

leslielazzarino (changes requested):
The white field should not leave an extra wide margin at the bottom, just the same the buttons in the right column.

src/Frontend/Components/AttributionView/__tests__/AttributionView.test.tsx
```diff
@@ -55,7 +55,7 @@ describe('The Attribution View', () => {
     store.dispatch(
       loadFromFile(
         getParsedInputFileEnrichedWithTestData({
-          resources: { ['test resource']: 1 },
+          resources: { ['/test resource']: 1 },
```
leslielazzarino, line 58: This change is wrong, wer do not write the slash in Resources

src/Frontend/Components/AttributionView/__tests__/AttributionView.test.tsx
```diff
@@ -134,7 +134,7 @@ describe('The Attribution View', () => {
...
```
leslielazzarino, line 137: Same for the othewr three here

src/Frontend/Components/ResourcesTree/ResourcesTree.tsx
```diff
@@ -0,0 +1,115 @@
...
+} from './resources-tree-helpers';
+
+interface ResourcesTreeProps {
+  resourcePaths: Array<string>;
+  maxHeight?: number;
+  sx?: SxProps;
+}
+
+const ROOT_FOLDER_LABEL = '';
+const TREE_ROW_HEIGHT = 20;
```
leslielazzarino, line 36: These two constants should be extracted to shared constants, as they are also used in the other tree.

src/Frontend/Components/ResourcesTree/ResourcesTree.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+export function ResourcesTree(props: ResourcesTreeProps): ReactElement {
+  const dispatch = useAppDispatch();
+
+  const initialExpandedIds = getInitialExpandedIds(props.resourcePaths);
```
leslielazzarino, line 41: Thinking more about this issue, I think you should remove this line. The calculation in the useMemo will be triggered anyhow on the first time, and as only a shallow comparison happens a rerender will also be triggered, even if the two objects are the same. Just use an empty array as initial state.

src/Frontend/Components/ResourcesTree/__tests__/ResourcesTree.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+    '/',
+    '/folder1/',
+    '/folder1/folder2/',
+    '/folder1/folder2/resource_1',
+  ];
+
+  it('component renders', () => {
+    renderComponentWithStore(<ResourcesTree resourcePaths={resourcePaths} />);
+    expect(screen.getByText('resource_2')).toBeInTheDocument();
+  });
```
leslielazzarino, line 30: You do not need this test, as you already have the next one checking the initial state.

leslielazzarino (changes requested):

MarkusObendrauf:
> The white field should not leave an extra wide margin at the bottom, just the same the buttons in the right column.

Adjusted the vertical spacing to match the height of the "CONFIRM" button

src/Frontend/Components/ResourcesTree/ResourcesTree.tsx
MarkusObendrauf: done

src/Frontend/Components/ResourcesTree/ResourcesTree.tsx
MarkusObendrauf: done

src/Frontend/Components/AttributionView/__tests__/AttributionView.test.tsx
MarkusObendrauf: a resource is a path, so it should contain a leading slash

MarkusObendrauf: ok, I see the issue now. Fixed.

benedikt-richter:
Please, also test with a real world example before merging to ensure huge trees are rendered correctly.

benedikt-richter (changes requested):
Please, sqash before merging

leslielazzarino:
The selected resource on audit view should not cause resources in the panel in attribution view to be highlighted. In the resource popup it is ok though.

leslielazzarino (approved):

lennartclaas:
Highlighting of selected resources is now disabled in the attribution view tree.

benedikt-richter (changes requested):

src/Frontend/shared-constants.ts
```diff
@@ -38,3 +38,6 @@ export const MUI_BREAKPOINTS_TO_PIXELS_MAPPING = {
   lg: 1200,
   xl: 1536,
 };
+
+export const TREE_ROW_HEIGHT = 20;
```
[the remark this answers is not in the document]
lennartclaas: Then, I think, the same is true for the last 4 constants in this file.

src/Frontend/Components/SimpleTreeItemLabel/SimpleTreeItemLabel.tsx
```diff
@@ -0,0 +1,44 @@
...
+import MuiBox from '@mui/material/Box';
+
+interface StyledTreeItemProps {
+  labelText: string;
+  canHaveChildren: boolean;
+  isAttributionBreakpoint: boolean;
+  showFolderIcon: boolean;
+}
+
+export function SimpleTreeItemLabel(props: StyledTreeItemProps): ReactElement {
```
benedikt-richter, line 19: By convention the properties of a component are named `ComponentNameProps`.

src/Frontend/Components/ResourcesTree/get-tree-item-label-simple.tsx
```diff
@@ -0,0 +1,36 @@
...
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+
+import { Resources } from '../../../shared/shared-types';
+import { PathPredicate } from '../../types/types';
+import React, { ReactElement } from 'react';
+import { SimpleTreeItemLabel } from '../SimpleTreeItemLabel/SimpleTreeItemLabel';
+
+export function getTreeItemLabelSimple(
```
benedikt-richter, line 11: Why not call it `getSimpleTreeItemLabel`. In my opinion that would make more sense.

benedikt-richter (changes requested):
We now have `SimpleTreeItemLabel` and `StyledTreeItemLabel`. Both of which are styled. Would be good to clarify naming a bit more. Maybe:
`ResourceBrowserTreeItemLabel` and `GeneralResourceTreeItemLabel`?

benedikt-richter (approved):

PR #1246, opened 2022-12-13
Reset the progress bar cache when opening a new file

MarkusObendrauf:
Signed-off-by: Markus Obendrauf <markus.obendrauf@tngtech.com>

### Summary of changes

Make explicit which messages sent to the progress bar worker are used for cache initialisation, and correct the behaviour on opening a new file.

[Another PR](https://github.com/opossum-tool/OpossumUI/commit/2dfff7b80cd5a5b59a20d75e2582a7fb1ee74ec6) introduced a bug where the progress bar would not update if you open two different input files. This PR fixes the bug.

Open two different input files and hover over the progress bar to check that it is correct.

### DIFF
--- a/src/Frontend/Components/WorkersContextProvider/WorkersContextProvider.tsx
+++ b/src/Frontend/Components/WorkersContextProvider/WorkersContextProvider.tsx
@@ -74,8 +74,16 @@ export const ProgressBarWorkersContextProvider: FC<{
 
   useMemo(() => {
     try {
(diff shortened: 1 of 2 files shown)


vasily-pozdnyakov (approved):
Honestly, I do not understand how the workers work, but the issue is fixed and the code looks ok.
Do we have no tests for workers? Probably it would be nice to add such tests. Or at least add another test for TopProgressBar, which covers the case. You could try to do it within the fix (if it is easy), or create a new issue.

PR #1256, opened 2022-12-16
Remove improper selector returns

leslielazzarino:
### Summary of changes

Selectors have been rewritten to return null instead of {} or [] created in the selector itself.

Empty object or arrays created in the selector themselves should never be returned. They cause unnecessary re-renders and degraded performances. For examplle by navigation in the resource tree.

### DIFF
--- a/src/Frontend/Components/AttributionDetailsViewer/AttributionDetailsViewer.tsx
+++ b/src/Frontend/Components/AttributionDetailsViewer/AttributionDetailsViewer.tsx
@@ -10,7 +10,6 @@ import { PackageInfo } from '../../../shared/shared-types';
 import { getTemporaryPackageInfo } from '../../state/selectors/all-views-resource-selectors';
 import { AttributionColumn } from '../AttributionColumn/AttributionColumn';
 import { ResourcesList } from '../ResourcesList/ResourcesList';
-import { isEqual } from 'lodash';
 import {
   deleteAttributionGloballyAndSave,
   savePackageInfo,
@@ -53,10 +52,8 @@ const classes = {
 export function AttributionDetailsViewer(): ReactElement | null {
   const selectedAttributionId = useAppSelector(getSelectedAttributionId);
   const temporaryPackageInfo = useAppSelector(getTemporaryPackageInfo);
-  const resourceIdsOfSelectedAttributionId: Array<string> = useAppSelector(
-    getResourceIdsOfSelectedAttribution,
-    isEqual
-  );
+  const resourceIdsOfSelectedAttributionId: Array<string> =
+    useAppSelector(getResourceIdsOfSelectedAttribution) || [];
 
   const resourceListMaxHeight = useWindowHeight() - 112;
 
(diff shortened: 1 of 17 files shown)


src/Frontend/Components/AttributionDetailsViewer/AttributionDetailsViewer.tsx
```diff
@@ -53,10 +52,8 @@ const classes = {
 export function AttributionDetailsViewer(): ReactElement | null {
   const selectedAttributionId = useAppSelector(getSelectedAttributionId);
   const temporaryPackageInfo = useAppSelector(getTemporaryPackageInfo);
-  const resourceIdsOfSelectedAttributionId: Array<string> = useAppSelector(
-    getResourceIdsOfSelectedAttribution,
-    isEqual
```
leslielazzarino, line 58: This is expensive, it should be avoided.

src/Frontend/state/selectors/all-views-resource-selectors.ts
```diff
@@ -121,26 +121,29 @@ export function getCurrentAttributionId(state: State): string | null {
...
 }
 
-export function getPackageInfoOfSelectedAttribution(state: State): PackageInfo {
+export function getPackageInfoOfSelectedAttribution(
+  state: State
+): PackageInfo | null {
   const selectedAttributionId = getSelectedAttributionId(state);
 
   if (!selectedAttributionId) {
-    return {};
```
leslielazzarino, line 128: We should have caugth these arrors during reviews.

src/Frontend/state/selectors/all-views-resource-selectors.ts
```diff
@@ -36,8 +36,8 @@ export function getManualData(state: State): AttributionData {
   return state.resourceState.allViews.manualData;
 }
 
-export function getManualAttributions(state: State): Attributions {
-  return state.resourceState.allViews.manualData.attributions || {};
+export function getManualAttributions(state: State): Attributions | null {
```
benedikt-richter, line 39: this function will never return `null`. The type can be removed and a lot of the `|| {}`else where can be removed

src/Frontend/state/selectors/view-selector.ts
```diff
@@ -38,7 +39,8 @@ export function getActiveFilters(state: State): Set<FilterType> {
 
 export function getPopupAttributionId(state: State): string | null {
   const popup = state.viewState.popupInfo.slice(-1);
-  return popup.length === 1 ? popup[0].attributionId ?? null : null;
+
+  return (popup.length === 1 && popup[0].attributionId) || null;
```
benedikt-richter, line 43: 👍

benedikt-richter (changes requested):
Nice catch! 👍 

There is one more to fix: `getAttributionIdsOfSelectedResource`. Also see comment.

lennartclaas (approved):

src/Frontend/Components/ProjectStatisticsPopup/ProjectStatisticsPopup.tsx
```diff
@@ -37,7 +38,9 @@ const classes = {
 export function ProjectStatisticsPopup(): ReactElement {
   const dispatch = useAppDispatch();
 
-  const manualAttributions = useAppSelector(getManualAttributions);
+  const manualAttributions: Attributions = useAppSelector(
```
benedikt-richter, line 41: why do we always specify the type explicitly? Would be better to stay consistent with other cases, e.g. `const externalAttributions = useAppSelector(getExternalAttributions);`

benedikt-richter (changes requested):

benedikt-richter (approved):

Issue #1260, opened 2022-12-19
Fix errors when fetching license data from github.com or npmjs.com


lennartclaas:
By entering a URL with a domain from github.com, pypi.org, or npmjs.com, license data of packages or repos can automatically be fetched. While fetching from pypi.org works like a charm, fetching from github.com and npmjs.com does not.

Github.com:
Although license data can be fetched from Github.com, a 'Buffer is not defined' error is thrown. See the following images.

OpossumUI:
![buffer_not_defined_error_github](https://user-images.githubusercontent.com/83081698/208463397-630edd77-87e6-44e8-ab84-08466e36f6dd.png)

opencv:
![buffer_not_defined_error_github_2](https://user-images.githubusercontent.com/83081698/208464683-24a3ba2d-de3a-4ebd-a1ab-92a4d7b2ac9d.png)

Npmjs.com:
Fetching from npmjs.com results in a 'Network error'. See the following images.

Typescript:
![network_error_npmjs](https://user-images.githubusercontent.com/83081698/208467768-c9abd925-bee5-420f-9b11-af4eefba73c4.png)

Axios mock adapter:
![network_error_npmjs_2](https://user-images.githubusercontent.com/83081698/208467827-5c2448ea-8f47-499d-9bff-ddfadca93000.png)

The startpoint for fixing this issue is FetchLicenseInformationButton.tsx

PR #1274, opened 2022-12-22
Fix bugs that prevent package data to be fetched from github and npmjs.

lennartclaas:
### Summary of changes

Github: the `buffer` package was installed and imported to facilitate conversion from base64 to utf-8.
Npmjs: a backend function was removed that manually extended the http response header to allow cross-origin resource sharing. This manual extension of the header is not required anymore.

When trying to fetch package data from github, a 'Buffer is not defined' error is thrown. Fetching data from npmjs results in a 'Network error'.

Fix: #1260

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
     "@tanstack/react-query": "^4.20.4",
     "@types/recharts": "^1.8.24",
     "axios": "^1.2.1",
+    "buffer": "^6.0.3",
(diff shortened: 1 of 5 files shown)


benedikt-richter (approved):

vasily-pozdnyakov (approved):
Do not know how, but it works, nice!

PR #1291, opened 2022-12-28
Fixed shifted header by changing padding

### LINKED ISSUE #1220: Shifted table header in Report view
The header of the table seems to be shifted sometimes.
image

vasily-pozdnyakov:
### Summary of changes

The problem appeared when the table contained a vertical scrollbar, which shifted the table to the right, but not the header.
Now the first empty cell of the header has padding from all sides.

Fix #1220

image

### DIFF
--- a/src/Frontend/Components/ReportTableHeader/ReportTableHeader.tsx
+++ b/src/Frontend/Components/ReportTableHeader/ReportTableHeader.tsx
@@ -30,10 +30,6 @@ export const reportTableClasses = {
     padding: '10px',
     flex: '1 1 auto',
   },
-  emptyTableCell: {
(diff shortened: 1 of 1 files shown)


lennartclaas (approved):
Looks good. If there are no situations where the `emptyTableCell` could be useful, it should be safe to delete it.
**Vasily**: I found no other usages.

benedikt-richter:
Noted that the issue still exists:

Screenshot 2023-01-02 at 07 34 06
Screenshot 2023-01-02 at 07 34 23

benedikt-richter (approved):
The issue still exists as commented above. However, it now looks better and the issue is minor. It should be good enough for now.

Issue #1110, opened 2022-10-27
Improve performance of save operation


benedikt-richter:
When saving attributions in OpossumUI, the user has to wait for the app to become reactive again.
There have been reports that this time increased recently. Unfortunately, there is no conclusive data on that.
It would be good to check how performance evolved and whether we have to improve here. In case it is necessary, improvements should be implemented.

benedikt-richter:
We decided to move the calculation of the progress bar data to a web worker

PR #1296, opened 2023-01-03
Improve performance of package panel

benedikt-richter:
### Summary of changes

Before all external attributions where filtered after converting the object into an array. In cases where there are several 100k of attributions, this takes a long time and affects performance.

The performance of the PackagePanel is improved for large numbers of external attributions. In test some cases, the performance was improved by a factor of 5 for the action of clicking on a file in OpossumUI. 
The time for adding an attribution to a root folder for a huge example went down from approx 4.6s to approx 3.1s.

This work is in the context of [#1110](https://github.com/opossum-tool/OpossumUI/issues/1110).


vasily-pozdnyakov (approved):

lennartclaas (approved):
Nice! No comments.

Issue #1279, opened 2022-12-23
[1275] Create scrollable list 


benedikt-richter:
This ticket is in the context of #1275.

  Create a new component that can be used as component to select e.g. the package name:

It should have the following features:
* items can be selected
* there is highlighting for name/version

The interface might be something like:
```
{
        items: Array< {name: string; attributes: Array<{name: string, highlighted: boolean}>; highlighted: boolean }>
        showAddNewInput: boolean (added in next ticket)
}
```

PR #1297, opened 2023-01-03
Attribution Wizard: Implement scrollable list component

lennartclaas:
### Summary of changes

Add AttributionWizardList component, helpers, and corresponding tests.

#1279 
The attribution wizard will be populated with scrollable lists that contain information about package namespace, package name, and package version. Individual items can be selected. Attributes of each individual item can be highlighted with a box.

Screenshot with dummy data:
attribution_wizard_list

### DIFF
--- a/src/Frontend/Components/AttributionWizardPopup/AttributionWizardPopup.tsx
+++ b/src/Frontend/Components/AttributionWizardPopup/AttributionWizardPopup.tsx
@@ -3,9 +3,8 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
-import React, { ReactElement } from 'react';
+import React, { ReactElement, useState } from 'react';
 import MuiBox from '@mui/material/Box';
-import MuiTypography from '@mui/material/Typography';
 import { doNothing } from '../../util/do-nothing';
 import { OpossumColors } from '../../shared-styles';
 import { NotificationPopup } from '../NotificationPopup/NotificationPopup';
@@ -13,8 +12,9 @@ import { ButtonText } from '../../enums/enums';
 import { useAppDispatch } from '../../state/hooks';
 import { closePopup } from '../../state/actions/view-actions/view-actions';
 import { PathBar } from '../PathBar/PathBar';
+import { ListWithAttributes } from '../ListWithAttributes/ListWithAttributes';
 
-// TODO: const POPUP_CONTENT_PADDING = 48; monitored in upcoming tickets
+// const POPUP_CONTENT_PADDING = 48; // TODO: monitored in upcoming tickets
 const attributionWizardPopupHeader = 'Attribution Wizard';
 
 const classes = {
@@ -23,19 +23,21 @@ const classes = {
   },
   dialogHeader: {
     whiteSpace: 'nowrap',
-    // TODO: width: `calc(100% - ${POPUP_CONTENT_PADDING}px)`,  monitored in upcoming tickets
+    // width: `calc(100% - ${POPUP_CONTENT_PADDING}px)`, // TODO: monitored in upcoming tickets
   },
-
   mainContent: {
     borderRadius: 2,
     paddingTop: '0px',
     background: OpossumColors.white,
   },
   mainContentBox: {
-    padding: '4px',
-    borderRadius: 2,
+    key: 'mainContent',
+    display: 'flex',
+    gap: '30px',
+    width: 'fit-content',
     marginTop: '8px',
-    background: OpossumColors.lightBlue,
+    maxHeight: '70vh',
+    minWidth: '60vh',
   },
   pathBar: {
     paddingLeft: '5px',
@@ -69,26 +71,46 @@ export function AttributionWizardPopup(): ReactElement {
     isDisabled: false,
   };
 
-  const testContent = (
-    <MuiBox style={classes.dialogContent}>
-      <MuiBox sx={classes.pathBarBox}>
-        <PathBar sx={classes.pathBar} />
-      </MuiBox>
-      <MuiBox sx={classes.mainContentBox}>
-        <MuiBox sx={classes.mainContent}>
-          <MuiTypography
-            sx={{
-              textAlign: 'center',
-              width: '300px',
-              height: '300px',
-            }}
-          >
-            Content dummy
-          </MuiTypography>
-        </MuiBox>
-      </MuiBox>
-    </MuiBox>
-  );
+  // TODO: streamline and integrate this logic later
+  const [selectedIdList1, setSelectedIdList1] = useState<string>('');
+  const [selectedIdList2, setSelectedIdList2] = useState<string>('');
+  const handleListItemClickList1 = (id: string): void => {
+    setSelectedIdList1(id);
+  };
+  const handleListItemClickList2 = (id: string): void => {
+    setSelectedIdList2(id);
+  };
+
+  // create dummy data
+  const N = 15;
+  const items = [];
+  const highlightedAttributeIds = [];
+  for (let i = 0; i < N; i++) {
+    items.push({
+      text: `package${i}`,
+      id: `testItemId${i}`,
+      attributes: [
+        {
+          text: `attrib${4 * i}`,
+          id: `testAttributeId${4 * i}`,
+        },
+        {
+          text: `attrib${4 * i + 1}`,
+          id: `testAttributeId${4 * i + 1}`,
+        },
+        {
+          text: `attrib${4 * i + 2}`,
+          id: `testAttributeId${4 * i + 2}`,
+        },
+        {
+          text: `attrib${4 * i + 3}`,
+          id: `testAttributeId${4 * i + 3}`,
+        },
+      ],
+    });
+    highlightedAttributeIds.push(`testAttributeId${4 * i + (i % 4)}`);
+  }
+
   return (
     <NotificationPopup
       header={attributionWizardPopupHeader}
@@ -99,7 +121,29 @@ export function AttributionWizardPopup(): ReactElement {
       isOpen={true}
       fullWidth={false}
       headerSx={classes.dialogHeader}
-      content={testContent}
+      content={
+        <>
+          <MuiBox sx={classes.pathBarBox}>
+            <PathBar sx={classes.pathBar} />
+          </MuiBox>
+          <MuiBox sx={classes.mainContentBox}>
+            <ListWithAttributes
+              listItems={items}
+              selectedListItemId={selectedIdList1}
+              highlightedAttributeIds={highlightedAttributeIds}
+              handleListItemClick={handleListItemClickList1}
+              showAddNewInput={false}
+            />
+            <ListWithAttributes
+              listItems={items}
+              selectedListItemId={selectedIdList2}
+              highlightedAttributeIds={highlightedAttributeIds}
+              handleListItemClick={handleListItemClickList2}
+              showAddNewInput={false}
+            />
+          </MuiBox>
+        </>
+      }
     />
   );
 }
--- a/src/Frontend/Components/ListWithAttributes/ListWithAttributes.tsx
+++ b/src/Frontend/Components/ListWithAttributes/ListWithAttributes.tsx
@@ -1,0 +1,102 @@
+// SPDX-FileCopyrightText: Meta Platforms, Inc. and its affiliates
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+
+import React, { ReactElement } from 'react';
+import MuiBox from '@mui/material/Box';
+import MuiList from '@mui/material/List';
+import MuiListItem from '@mui/material/ListItem';
+import MuiListItemText from '@mui/material/ListItemText';
+import MuiListItemButton from '@mui/material/ListItemButton';
+import { OpossumColors } from '../../shared-styles';
+import { getAttributesWithHighlighting } from './list-with-attributes-helpers';
+import { ListWithAttributesItem } from '../../types/types';
+
+const LIST_WITH_ATTRIBUTES_VERTICAL_BORDER_AND_MARGIN = 10; // 10px = margin + border
+
+const classes = {
+  listBox: {
+    width: 'fit-content',
+    maxHeight: '100%',
+    background: OpossumColors.lightBlue,
+  },
+  list: {
+    padding: '0px',
+    width: 'fit-content',
+    backgroundColor: OpossumColors.white,
+    border: `1px ${OpossumColors.grey} solid`,
+    maxHeight: `calc(100% - ${LIST_WITH_ATTRIBUTES_VERTICAL_BORDER_AND_MARGIN}px)`,
+    overflowY: 'auto',
+    margin: '4px',
+  },
+  listItem: {
+    padding: '0px',
+    borderBottom: 2,
+    borderColor: OpossumColors.lightBlue,
+  },
+  listItemButton: {
+    padding: '0px 0px 0px 4px',
+    '&:hover': {
+      backgroundColor: OpossumColors.lightestBlueOnHover,
+    },
+    '&.Mui-selected': {
+      '&:hover': {
+        backgroundColor: OpossumColors.middleBlueOnHover,
+      },
+      backgroundColor: OpossumColors.middleBlue,
+    },
+  },
+  listItemTextAttributesBox: {
+    display: 'flex',
+    // TODO: Now: Single scroll bar for each list item.
+    // Remove 'overflowX' property to get single scroll bar
+    // for the whole list (also if only a single
+    // list item exceeds maxWidth)
+    overflowX: 'auto',
+    maxWidth: '50vw',
+    marginLeft: '20px',
+    paddingTop: '1px',
+  },
+};
+
+interface ListWithAttributesProps {
+  listItems: Array<ListWithAttributesItem>;
+  selectedListItemId: string;
+  highlightedAttributeIds: Array<string>;
+  handleListItemClick: (id: string) => void;
+  showAddNewInput: boolean; // TODO: required later
+}
+
+export function ListWithAttributes(
+  props: ListWithAttributesProps
+): ReactElement {
+  return (
+    <MuiBox sx={classes.listBox}>
+      <MuiList sx={classes.list}>
+        {props.listItems.map((item) => (
+          <MuiListItem key={`itemId-${item.id}`} sx={classes.listItem}>
+            <MuiListItemButton
+              sx={classes.listItemButton}
+              selected={item.id === props.selectedListItemId}
+              onClick={(): void => props.handleListItemClick(item.id)}
+            >
+              <MuiListItemText
+                primary={item.text}
+                secondary={
+                  <MuiBox sx={classes.listItemTextAttributesBox}>
+                    {getAttributesWithHighlighting(
+                      item.attributes,
+                      props.highlightedAttributeIds
+                    )}
+                  </MuiBox>
+                }
+                secondaryTypographyProps={{ component: 'span' }}
+              />
+            </MuiListItemButton>
+          </MuiListItem>
+        ))}
+      </MuiList>
+    </MuiBox>
+  );
+}
--- a/src/Frontend/Components/ListWithAttributes/__tests__/ListWithAttributes.test.tsx
+++ b/src/Frontend/Components/ListWithAttributes/__tests__/ListWithAttributes.test.tsx
@@ -1,0 +1,57 @@
+// SPDX-FileCopyrightText: Meta Platforms, Inc. and its affiliates
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+
+import React from 'react';
+import { render, within } from '@testing-library/react';
+import { ListWithAttributes } from '../ListWithAttributes';
+import { screen } from '@testing-library/react';
+import { ListWithAttributesItem } from '../../../types/types';
+import { doNothing } from '../../../util/do-nothing';
+
+describe('ListWithAttributes', () => {
+  it('renders with items containing name and attributes', () => {
+    const testItems: Array<ListWithAttributesItem> = [
+      {
+        text: 'package_0',
+        id: 'test_item_id_0',
+        attributes: [
+          { text: 'attrib_00', id: 'test_attribute_id_00' },
+          { text: 'attrib_01', id: 'test_attribute_id_01' },
+        ],
+      },
+      {
+        text: 'package_1',
+        id: 'test_item_id_1',
+        attributes: [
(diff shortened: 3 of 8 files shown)


src/Frontend/Components/ListWithAttributes/ListWithAttributes.tsx
```diff
@@ -0,0 +1,121 @@
...
+import MuiListItemText from '@mui/material/ListItemText';
+import MuiListItemButton from '@mui/material/ListItemButton';
+import { OpossumColors } from '../../shared-styles';
+import {
+  concatenateAttributeNames,
+  getInitialHighlightedListIndex,
+} from './list-with-attributes-helpers';
+import { ListWithAttributesItem } from '../../types/types';
+
+const ATTRIBUTION_WIZARD_LIST_VERTICAL_BORDER_AND_MARGIN = 10; // 10px = margin + border
```
benedikt-richter, line 20: Should not contain `ATTRIBUTION_WIZARD` as it is a generic list

src/Frontend/Components/ListWithAttributes/ListWithAttributes.tsx
```diff
@@ -0,0 +1,121 @@
...
+}
+
+export function ListWithAttributes(
+  props: ListWithAttributesProps
+): ReactElement {
+  const initialHighlightedIndex = useMemo(
+    () => getInitialHighlightedListIndex(props.items),
+    [props.items]
+  );
+  const [selectedIndex, setSelectedIndex] = useState<number | null>(
```
benedikt-richter, line 75: This component should be a controlled component and its state should be managed in its parent.

src/Frontend/Components/ListWithAttributes/ListWithAttributes.tsx
```diff
@@ -0,0 +1,121 @@
...
+    initialHighlightedIndex
+  );
+
+  const handleListItemClick = (index: number): void => {
```
benedikt-richter, line 79: Should be provided as `prop`

src/Frontend/Components/ListWithAttributes/ListWithAttributes.tsx
```diff
@@ -0,0 +1,121 @@
...
+    setSelectedIndex(index);
+  };
+
+  return (
+    <MuiBox sx={classes.listBox}>
+      <MuiList sx={classes.list}>
+        {props.items.map((item, index) => (
+          <React.Fragment key={`listFragment${index}`}>
+            <MuiListItem key={`listItem${index}`} sx={classes.listItem}>
```
benedikt-richter, line 88: why do we use a key here?

src/Frontend/Components/ListWithAttributes/ListWithAttributes.tsx
```diff
@@ -0,0 +1,121 @@
...
+              <MuiListItemButton
+                key={`listItemButton${index}`}
```
benedikt-richter, line 90: why do we use a key here?

src/Frontend/Components/ListWithAttributes/ListWithAttributes.tsx
```diff
@@ -0,0 +1,121 @@
...
+                sx={classes.listItemButton}
+                selected={selectedIndex === index}
+                onClick={(): void => handleListItemClick(index)}
+              >
+                <MuiListItemText
+                  key={`listItemText${index}`}
```
benedikt-richter, line 96: why do we use a key here?

src/Frontend/Components/ListWithAttributes/ListWithAttributes.tsx
```diff
@@ -0,0 +1,121 @@
...
+                  primary={item.name}
+                  secondary={
+                    <MuiBox
+                      key={`listBox${index}`}
```
benedikt-richter, line 100: why do we use a key here?

src/Frontend/Components/ListWithAttributes/ListWithAttributes.tsx
```diff
@@ -0,0 +1,121 @@
...
+                      sx={classes.listItemTextSecondaryBox}
+                    >
+                      {'\xa0\xa0\xa0'}
```
benedikt-richter, line 103: Why do we use this? Seems to solve an issue that should be solved with CSS.

src/Frontend/Components/ListWithAttributes/ListWithAttributes.tsx
```diff
@@ -0,0 +1,121 @@
...
+                        item.attributes,
+                        `listItemAttributes${index}`
+                      )}
+                    </MuiBox>
+                  }
+                />
+              </MuiListItemButton>
+            </MuiListItem>
+            {index < props.items.length - 1 && (
+              <MuiDivider key={`listDivider${index}`} variant="middle" />
```
benedikt-richter, line 114: I'd suggest not to use a divider but to give a border to the list items as we do in other lists

src/Frontend/Components/ListWithAttributes/list-with-attributes-helpers.tsx
```diff
@@ -0,0 +1,66 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import React, { ReactElement } from 'react';
+import MuiBox from '@mui/material/Box';
+import {
+  ListWithAttributesItem,
+  ListWithAttributesItemAttribute,
+} from '../../types/types';
+
+export function getInitialHighlightedListIndex(
```
benedikt-richter, line 13: IF we provide highlighting via a property of the list items, all should be highlighted. There should be no logic "secretly" overwriting configuration.

src/Frontend/Components/ListWithAttributes/list-with-attributes-helpers.tsx
```diff
@@ -0,0 +1,66 @@
...
+  attributes: Array<ListWithAttributesItemAttribute>,
+  keyPrefix: string
+): Array<ReactElement> {
+  const styleBasic = {
+    padding: '0px 1px 0px 2px',
+  };
+  const styleHighlighted = {
+    border: 1,
+  };
+  const attributeNameList = new Array<ReactElement>();
```
benedikt-richter, line 44: we never used this pattern so far.

src/Frontend/Components/ListWithAttributes/list-with-attributes-helpers.tsx
```diff
@@ -0,0 +1,66 @@
...
+  for (
+    let attributeIndex = 0;
+    attributeIndex < attributes.length;
+    attributeIndex++
+  ) {
+    attributeNameList.push(
+      <React.Fragment key={keyPrefix + `_fragment${attributeIndex}`}>
+        {attributeIndex === 0 ? '' : ',\xa0'}
+        <MuiBox
+          key={keyPrefix + `_box${attributeIndex}`}
```
benedikt-richter, line 54: why do we use a key here?

src/Frontend/Components/ListWithAttributes/list-with-attributes-helpers.tsx
```diff
@@ -0,0 +1,66 @@
...
+  const styleHighlighted = {
+    border: 1,
+  };
+  const attributeNameList = new Array<ReactElement>();
+  for (
+    let attributeIndex = 0;
+    attributeIndex < attributes.length;
+    attributeIndex++
+  ) {
+    attributeNameList.push(
```
benedikt-richter, line 50: instead of this loop, this could use `.map()` on `attributes` to improve readability

src/Frontend/types/types.ts
```diff
@@ -138,3 +138,14 @@ export interface LicenseNamesWithCriticality {
   [licenseName: string]: Criticality | undefined;
 }
 export type ProgressBarType = 'FolderProgressBar' | 'TopProgressBar';
+
+export interface ListWithAttributesItemAttribute {
+  name: string;
+  highlighted: boolean;
+}
```
benedikt-richter, line 145: Reconsidering now, this could be:
```
export interface ListWithAttributesItemAttribute {
  text: string;
  id: string;
}

```

The highlighting could then happen by specifying the ids that should be highlighted.

benedikt-richter (changes requested):

src/Frontend/Components/ListWithAttributes/ListWithAttributes.tsx
lennartclaas: I got a corresponding warning. All children of a list component should be assigned a key.

src/Frontend/Components/ListWithAttributes/ListWithAttributes.tsx
lennartclaas: see above

src/Frontend/Components/ListWithAttributes/ListWithAttributes.tsx
lennartclaas: see above

src/Frontend/Components/ListWithAttributes/ListWithAttributes.tsx
lennartclaas: see above

src/Frontend/Components/ListWithAttributes/list-with-attributes-helpers.tsx
lennartclaas: see above

src/Frontend/Components/AttributionWizardPopup/AttributionWizardPopup.tsx
```diff
@@ -69,26 +71,46 @@ export function AttributionWizardPopup(): ReactElement {
...
-            }}
-          >
-            Content dummy
-          </MuiTypography>
-        </MuiBox>
-      </MuiBox>
-    </MuiBox>
-  );
+  // TODO: streamline and integrate this logic later
+  const [selectedIdList1, setSelectedIdList1] = useState<string | null>(null);
```
benedikt-richter, line 75: Naming should be improved. However, can  happen in one of the follow up PRs.

src/Frontend/Components/ListWithAttributes/ListWithAttributes.tsx
```diff
@@ -0,0 +1,108 @@
...
+    overflowX: 'auto',
+    maxWidth: '25vh',
+    marginLeft: '20px',
+    paddingTop: '1px',
+  },
+};
+
+interface ListWithAttributesProps {
+  listWithAttributesItems: Array<ListWithAttributesItem>;
+  selectedListWithAttributesItemId: string | null;
```
benedikt-richter, line 65: Instead of `null`we can take the empty string in the case nothing is selected. Then we avoid unnecessary `null` checks

src/Frontend/Components/ListWithAttributes/ListWithAttributes.tsx
```diff
@@ -0,0 +1,108 @@
...
+        {props.listWithAttributesItems.map((item, index) => (
+          <MuiListItem key={`listItem${index}`} sx={classes.listItem}>
+            <MuiListItemButton
+              key={`listItemButton${index}`}
+              sx={classes.listItemButton}
+              selected={item.itemId === props.selectedListWithAttributesItemId}
+              onClick={(): void => props.handleListItemClick(item.itemId)}
+            >
+              <MuiListItemText
+                key={`listItemText${index}`}
```
benedikt-richter, line 86: This key should not be necessary

src/Frontend/Components/ListWithAttributes/ListWithAttributes.tsx
```diff
@@ -0,0 +1,108 @@
...
+export function ListWithAttributes(
+  props: ListWithAttributesProps
+): ReactElement {
+  return (
+    <MuiBox sx={classes.listBox}>
+      <MuiList sx={classes.list}>
+        {props.listWithAttributesItems.map((item, index) => (
+          <MuiListItem key={`listItem${index}`} sx={classes.listItem}>
+            <MuiListItemButton
+              key={`listItemButton${index}`}
```
benedikt-richter, line 80: This key should not be necessary

src/Frontend/Components/ListWithAttributes/ListWithAttributes.tsx
```diff
@@ -0,0 +1,108 @@
...
+}
+
+export function ListWithAttributes(
+  props: ListWithAttributesProps
+): ReactElement {
+  return (
+    <MuiBox sx={classes.listBox}>
+      <MuiList sx={classes.list}>
+        {props.listWithAttributesItems.map((item, index) => (
+          <MuiListItem key={`listItem${index}`} sx={classes.listItem}>
```
benedikt-richter, line 78: Better to take the `id` of the `item` instead of the index. This can help e.g. debugging.

src/Frontend/Components/ListWithAttributes/ListWithAttributes.tsx
```diff
@@ -0,0 +1,108 @@
...
+              sx={classes.listItemButton}
+              selected={item.itemId === props.selectedListWithAttributesItemId}
+              onClick={(): void => props.handleListItemClick(item.itemId)}
+            >
+              <MuiListItemText
+                key={`listItemText${index}`}
+                primary={item.itemText}
+                secondary={
+                  <MuiBox
+                    key={`listBox${index}`}
```
benedikt-richter, line 90: This key should not be necessary

src/Frontend/Components/ListWithAttributes/ListWithAttributes.tsx
```diff
@@ -0,0 +1,108 @@
...
+                    sx={classes.listItemTextSecondaryBox}
```
benedikt-richter, line 91: `SecondaryBox` sounds strange. Would have no idea what it is.

src/Frontend/Components/ListWithAttributes/ListWithAttributes.tsx
```diff
@@ -0,0 +1,108 @@
...
+                  >
+                    {concatenateAttributeNames(
+                      item.attributes,
+                      props.highlightedListWithAttributesItemAttributeIds,
+                      `listItemAttributes${index}`
```
benedikt-richter, line 96: This key should not be necessary

src/Frontend/Components/ListWithAttributes/__tests__/ListWithAttributes.test.tsx
```diff
@@ -0,0 +1,59 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import React from 'react';
+import { render, within } from '@testing-library/react';
+import { ListWithAttributes } from '../ListWithAttributes';
+import { screen } from '@testing-library/react';
+import { ListWithAttributesItem } from '../../../types/types';
+import { doNothing } from '../../../util/do-nothing';
+
+describe('ListWithAttributes', () => {
```
benedikt-richter, line 13: nice 👍

src/Frontend/Components/ListWithAttributes/__tests__/list-with-attributes-helpers.test.tsx
```diff
@@ -0,0 +1,63 @@
...
+              ...(testHighlightedAttributeIds.includes(attribute.attributeId)
+                ? styleHighlighted
+                : {}),
+            }}
+          >
+            {attribute.attributeText}
+          </MuiBox>
+        </React.Fragment>
+      )
+    );
```
benedikt-richter, line 45: Seems we are copying the entire logic of the function we test to the test to test the function. This test feels not very useful.

src/Frontend/Components/ListWithAttributes/list-with-attributes-helpers.tsx
```diff
@@ -0,0 +1,39 @@
...
+  const styleBasic = {
+    padding: '0px 1px 0px 2px',
+    marginLeft: '5px',
+  };
+  const styleHighlighted = {
+    border: 1,
+  };
+
+  return attributes.map((attribute, attributeIndex) => (
+    <React.Fragment key={keyPrefix + `_fragment${attributeIndex}`}>
```
benedikt-richter, line 24: This key might not be necessary

src/Frontend/Components/ListWithAttributes/list-with-attributes-helpers.tsx
```diff
@@ -0,0 +1,39 @@
+// SPDX-FileCopyrightText: Meta Platforms, Inc. and its affiliates
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+
+import React, { ReactElement } from 'react';
+import MuiBox from '@mui/material/Box';
+import { ListWithAttributesItemAttribute } from '../../types/types';
+
+export function concatenateAttributeNames(
```
benedikt-richter, line 10: From `concatenateAttributeNames` I'd expect to get a string or similar. Maybe a better name could be `getAttributesWithHighlighting`

src/Frontend/types/types.ts
```diff
@@ -138,3 +138,14 @@ export interface LicenseNamesWithCriticality {
   [licenseName: string]: Criticality | undefined;
 }
 export type ProgressBarType = 'FolderProgressBar' | 'TopProgressBar';
+
+export interface ListWithAttributesItemAttribute {
+  attributeText: string;
+  attributeId: string;
```
benedikt-richter, line 144: In my opinion `text` and `id` would be sufficient

src/Frontend/types/types.ts
```diff
@@ -138,3 +138,14 @@ export interface LicenseNamesWithCriticality {
...
+}
+
+export interface ListWithAttributesItem {
+  itemText: string;
+  itemId: string;
```
benedikt-richter, line 149: In my opinion `text` and `id` would be sufficient

benedikt-richter (changes requested):
Just saome minor points. Looks good in general 👍

vasily-pozdnyakov:
Too many scrollbars 
image

src/Frontend/Components/ListWithAttributes/ListWithAttributes.tsx
```diff
@@ -0,0 +1,108 @@
...
+    paddingTop: '1px',
+  },
+};
+
+interface ListWithAttributesProps {
+  listWithAttributesItems: Array<ListWithAttributesItem>;
+  selectedListWithAttributesItemId: string | null;
+  highlightedListWithAttributesItemAttributeIds: Array<string | null>;
+  handleListItemClick: (id: string) => void;
+  showAddNewInput: boolean; // TODO: required later
```
vasily-pozdnyakov, line 68: Should it be always shown?

src/Frontend/Components/ListWithAttributes/ListWithAttributes.tsx
```diff
@@ -0,0 +1,108 @@
...
+    display: 'flex',
+    // TODO: Now: Single scroll bar for each list item.
+    // Remove 'overflowX' property to get single scroll bar
+    // for the whole list (also if only a single
+    // list item exceeds maxWidth)
+    overflowX: 'auto',
+    maxWidth: '25vh',
+    marginLeft: '20px',
+    paddingTop: '1px',
+  },
```
lennartclaas, line 60: https://github.com/opossum-tool/OpossumUI/pull/1297#issuecomment-1370869819

Regarding the scroll bars for individual list items: currently, this is intended behavior and can be changed by removing `overflowX: 'auto'` in the respective CSS class. Can be changed later.

vasily-pozdnyakov (changes requested):
I also miss fields titles.

src/Frontend/Components/ListWithAttributes/__tests__/list-with-attributes-helpers.test.tsx
lennartclaas: True. I propose to completely remove this part. Hardcoding the expected component without `map()` seems also not to be very reasonable.

src/Frontend/Components/ListWithAttributes/list-with-attributes-helpers.tsx
lennartclaas: This key is actually necessary as we have multiple components in an array which seems to be quite similar to a list.

src/Frontend/Components/ListWithAttributes/ListWithAttributes.tsx
lennartclaas: Changed to `listItemTextAttributesBox`.

src/Frontend/Components/AttributionWizardPopup/AttributionWizardPopup.tsx
```diff
@@ -99,7 +121,33 @@ export function AttributionWizardPopup(): ReactElement {
       isOpen={true}
       fullWidth={false}
       headerSx={classes.dialogHeader}
-      content={testContent}
+      content={
+        <React.Fragment>
```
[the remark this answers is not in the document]
lennartclaas: Good point. Was a remnant from when a Divider was used.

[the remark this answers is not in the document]
lennartclaas: It is needed since `content` expects a single component.

src/Frontend/Components/ListWithAttributes/ListWithAttributes.tsx
lennartclaas: To my knowledge, this has not been decided so far. But I suppose that at least in the case of copyright texts this prop should be false.

src/Frontend/Components/ListWithAttributes/ListWithAttributes.tsx
lennartclaas: > Should it be possible to choose attributions? Or just the most frequent is highlighted? In this case it would be nice to have some tooltip which explains the border around attribution.

We have to be careful and not confuse attributes with attributions. Happens for me all the time. :D However, the attributes are not meant to be clickable as the highlighting is controlled by business logic.

lennartclaas: https://github.com/opossum-tool/OpossumUI/pull/1297#issuecomment-1370869819

The list witdth being dependent on the window height was a bug and is now fixed. Thanks. 👍

benedikt-richter (approved):
lgtm

Styling certainly could be imporved but can happen later

benedikt-richter:
Regarding https://github.com/opossum-tool/OpossumUI/pull/1297#pullrequestreview-1235810731:
I'm not sure we will have these, as in the current case we do not have input fields but a list of items.

src/Frontend/Components/AttributionWizardPopup/AttributionWizardPopup.tsx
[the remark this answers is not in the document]
vasily-pozdnyakov: Why not MuiBox? Or empty tags <></>?

vasily-pozdnyakov (approved):
Looks nice

Issue #1290, opened 2022-12-28
Hide "open resource in browser" tooltip


vasily-pozdnyakov:
The tooltip "open resource in browser" is still shown when the button is disabled.

![image](https://user-images.githubusercontent.com/85183359/209807060-5a593ead-67db-44dc-87b4-0e671fc40442.png)

lennartclaas:
A possible solution could be to apply `props.sx` of the `IconButton` not only to the `MuiButtonBase` but also to the `MuiTooltip` component. Background: if the `IconButton` is to be hidden, it is usually done by `props.sx`.

PR #1318, opened 2023-01-12
[BugFix] Do not show tooltip for hidden button

someshkhandelia:
### Summary of changes

[BugFix] Do not show tooltip for hidden button

- We keep the text shown in tooltip as empty to not let the tooltip render, when IconButton has the hidden prop set to true.

Fix: #1290

- Load the large_opossum_input.json.gz file in the application
- In the right corner of PathBar of AuditView, the tooltip is no longer visible since the button is hidden. 

### DIFF
--- a/src/Frontend/Components/IconButton/IconButton.tsx
+++ b/src/Frontend/Components/IconButton/IconButton.tsx
@@ -8,6 +8,13 @@ import MuiButtonBase from '@mui/material/ButtonBase';
 import MuiTooltip from '@mui/material/Tooltip';
 import { tooltipStyle } from '../../shared-styles';
 import { SxProps } from '@mui/material';
+import { getSxFromPropsAndClasses } from '../../util/get-sx-from-props-and-classes';
+
+const classes = {
+  hidden: {
+    visibility: 'hidden',
+  },
+};
 
 interface IconButtonProps {
   tooltipTitle: string;
@@ -16,6 +23,7 @@ interface IconButtonProps {
   onClick(): void;
   icon: ReactElement;
   disabled?: boolean;
+  hidden?: boolean;
 }
 
 export function IconButton(props: IconButtonProps): ReactElement {
@@ -23,7 +31,7 @@ export function IconButton(props: IconButtonProps): ReactElement {
     <MuiTooltip
       describeChild={true}
       sx={tooltipStyle}
-      title={props.tooltipTitle}
+      title={props.hidden ? '' : props.tooltipTitle}
       placement={props.tooltipPlacement}
     >
       <span>
@@ -32,7 +40,14 @@ export function IconButton(props: IconButtonProps): ReactElement {
         }
         <MuiButtonBase
           aria-label={props.tooltipTitle}
-          sx={props.sx}
+          sx={
+            props.hidden
+              ? getSxFromPropsAndClasses({
+                  styleClass: classes.hidden,
+                  sxProps: props.sx,
+                })
+              : props.sx
+          }
           onClick={(event): void => {
             event.stopPropagation();
             props.onClick();
--- a/src/Frontend/Components/IconButton/__tests__/IconButton.test.tsx
+++ b/src/Frontend/Components/IconButton/__tests__/IconButton.test.tsx
@@ -8,7 +8,7 @@ import React from 'react';
 import { render, screen } from '@testing-library/react';
 import { doNothing } from '../../../util/do-nothing';
 
-describe('Button', () => {
+describe('IconButton', () => {
   it('renders a button', () => {
     render(
       <IconButton
@@ -22,5 +22,50 @@ describe('Button', () => {
 
     expect(screen.getByLabelText('Test'));
     expect(screen.getByText('Test Icon'));
+  });
+
+  it('renders tooltip when hidden flag is inactive', () => {
+    render(
+      <IconButton
+        tooltipTitle={'Test'}
+        tooltipPlacement="left"
+        disabled={false}
+        onClick={doNothing}
+        icon={<div>Test Icon</div>}
+        hidden={false}
+      />
+    );
+
+    expect(screen.getByTitle('Test')).toBeInTheDocument();
+  });
+
+  it('does not render tooltip when hidden flag is active', () => {
+    render(
+      <IconButton
+        tooltipTitle={'Test'}
+        tooltipPlacement="left"
+        disabled={false}
+        onClick={doNothing}
+        icon={<div>Test Icon</div>}
+        hidden={true}
+      />
+    );
+
+    expect(screen.queryByTitle('Test')).not.toBeInTheDocument();
+  });
+
+  it('checks for successful application of hidden visibility style', () => {
+    render(
+      <IconButton
+        tooltipTitle={'Test'}
+        tooltipPlacement="left"
+        disabled={false}
+        onClick={doNothing}
+        icon={<div>Test Icon</div>}
+        hidden={true}
+      />
+    );
+
+    expect(screen.getByLabelText('Test')).toHaveStyle('visibility: hidden');
   });
 });
--- a/src/Frontend/Components/GoToLinkButton/GoToLinkButton.tsx
+++ b/src/Frontend/Components/GoToLinkButton/GoToLinkButton.tsx
@@ -20,13 +20,6 @@ import { useAppDispatch, useAppSelector } from '../../state/hooks';
 import { openPopup } from '../../state/actions/view-actions/view-actions';
 import { PopupType } from '../../enums/enums';
 
-const classes = {
-  hidden: {
-    visibility: 'hidden',
-  },
-  clickableIcon,
-};
-
 export function GoToLinkButton(): ReactElement {
   const path = useAppSelector(getSelectedResourceId);
   const baseUrlsForSources = useAppSelector(getBaseUrlsForSources);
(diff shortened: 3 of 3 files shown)


someshkhandelia:
I still have to add tests. I can add them if the implementation of the bug-fix is considered satisfactory by the reviewers.

src/Frontend/Components/IconButton/IconButton.tsx
```diff
@@ -19,11 +19,21 @@ interface IconButtonProps {
...
 export function IconButton(props: IconButtonProps): ReactElement {
+  const sxProps = Object.entries(props.sx ?? {});
+  let isIconButtonHidden = false;
+  sxProps.map((style) => {
+    const styleName = style[0];
+    const styleValue = style[1];
+    if (styleName === 'visibility' && styleValue === 'hidden') {
+      isIconButtonHidden = true;
+    }
+  });
```
benedikt-richter, line 30: It is not a good practice to parse styling in order to set variables in code. If we provide styling to a component it should just be applied and not have other effects.

Better: set the tooltip title to an empty string where it is used, i.e. in `GoToLinkButton`.

benedikt-richter (changes requested):

src/Frontend/Components/IconButton/IconButton.tsx
lennartclaas: @benedikt-richter What about adding a new prop `hidden` instead of hiding the `IconButton` via sx?

benedikt-richter: That's a good idea!

someshkhandelia: Working on adding a new prop `hidden`. Thanks for the comments guys!

src/Frontend/Components/IconButton/IconButton.tsx
```diff
@@ -13,17 +13,20 @@ interface IconButtonProps {
   tooltipTitle: string;
   tooltipPlacement: 'left' | 'right';
   sx?: SxProps;
+
```
benedikt-richter, line 16: Please, remove empty lines

src/Frontend/Components/IconButton/IconButton.tsx
```diff
@@ -32,7 +35,7 @@ export function IconButton(props: IconButtonProps): ReactElement {
         }
         <MuiButtonBase
           aria-label={props.tooltipTitle}
-          sx={props.sx}
+          sx={props.hidden ? { visibility: 'hidden', ...props.sx } : props.sx}
```
benedikt-richter, line 38: Would be good to extract the styling like 
```

const classes = {
  hidden: {
    visibility: 'hidden',
  }
};
```

benedikt-richter:
Works nicely 👍 
Just two minor points

lennartclaas (changes requested):
Elegant solution. 👍

src/Frontend/Components/IconButton/IconButton.tsx
someshkhandelia: I actually did not touch those parts of the code. I just added the last line `hidden?: boolean`. Maybe the pre-commit hook added it or something. I will remove it manually now and create a commit, but not sure if the hook will re-add it.

someshkhandelia: Linter is not allowing me to do this. If I push without running pre-commit hooks, the tests in the online GIT CLI fail.

src/Frontend/Components/IconButton/IconButton.tsx
someshkhandelia: Me, Leslie and Lennart found a weird thing. For `visibility` it only works inline. The `classes` approach works for everything except `visibility`

src/Frontend/Components/IconButton/IconButton.tsx
someshkhandelia: I committed without running pre-commit hooks. But I think if someone touches those lines, those blank lines will come back again from the Linter.

src/Frontend/Components/IconButton/IconButton.tsx
benedikt-richter: Interesting. In this case it is fine. However, I would add a comment in code, as otherwise somebody will refactor it and thus introduce a bug.

benedikt-richter: Or better: In this special situation we could have a test testing in particular the styling, i.e. that the hidden flag works.

someshkhandelia: Done.

src/Frontend/Components/IconButton/__tests__/IconButton.test.tsx
```diff
@@ -23,4 +23,34 @@ describe('Button', () => {
...
+        onClick={doNothing}
+        icon={<div>Test Icon</div>}
+        hidden={false}
+      />
+    );
+
+    expect(screen.getByTitle('Test'));
+  });
+
+  it('does not render tooltip when hidden flag is active', () => {
```
benedikt-richter, line 42: here we should also check that the styling `visibility: 'hidden'` is applied.

src/Frontend/Components/IconButton/__tests__/IconButton.test.tsx
```diff
@@ -23,4 +23,34 @@ describe('Button', () => {
...
+        tooltipTitle={'Test'}
+        tooltipPlacement="left"
+        disabled={false}
+        onClick={doNothing}
+        icon={<div>Test Icon</div>}
+        hidden={false}
+      />
+    );
+
+    expect(screen.getByTitle('Test'));
```
benedikt-richter, line 39: -> `expect(screen.getByTitle('Test')).toBeInTheDocument();`

src/Frontend/Components/IconButton/__tests__/IconButton.test.tsx
```diff
@@ -23,4 +23,34 @@ describe('Button', () => {
...
+        tooltipTitle={'Test'}
+        tooltipPlacement="left"
+        disabled={false}
+        onClick={doNothing}
+        icon={<div>Test Icon</div>}
+        hidden={true}
+      />
+    );
+
+    expect(() => screen.getByTitle('Test')).toThrow();
```
benedikt-richter, line 54: -> `expect(() => screen.queryByTitle('Test')).toBeInTheDocument()`

src/Frontend/Components/IconButton/IconButton.tsx
```diff
@@ -32,7 +33,9 @@ export function IconButton(props: IconButtonProps): ReactElement {
         }
         <MuiButtonBase
           aria-label={props.tooltipTitle}
-          sx={props.sx}
+          // TypeScript issue:
```
benedikt-richter, line 36: not sure this is a typescript issue. Might be a MUI issue as well...

benedikt-richter (changes requested):
Add a test for the styling (element being hidden), see https://github.com/opossum-tool/OpossumUI/pull/1318/files#r1068995740

src/Frontend/Components/IconButton/IconButton.tsx
someshkhandelia: True.

src/Frontend/Components/IconButton/__tests__/IconButton.test.tsx
someshkhandelia: Fixed.

src/Frontend/Components/IconButton/__tests__/IconButton.test.tsx
someshkhandelia: This command wasn't working. (This is a negative test, we are not supposed to find the element with title "Test")

Used this instead:
`expect(screen.queryByTitle('Test')).toBeNull();`

src/Frontend/Components/IconButton/IconButton.tsx
benedikt-richter: Just remembered. Maybe try using our util function `getSxFromPropsAndClasses`

src/Frontend/Components/IconButton/__tests__/IconButton.test.tsx
```diff
@@ -23,4 +23,49 @@ describe('Button', () => {
...
+        tooltipTitle={'Test'}
+        tooltipPlacement="left"
+        disabled={false}
+        onClick={doNothing}
+        icon={<div>Test Icon</div>}
+        hidden={true}
+      />
+    );
+
+    expect(screen.getByLabelText('Test')).toHaveStyle('visibility: hidden');
```
benedikt-richter, line 69: 👍

benedikt-richter (approved):

src/Frontend/Components/IconButton/IconButton.tsx
someshkhandelia: Fixed.

src/Frontend/Components/IconButton/IconButton.tsx
```diff
@@ -32,7 +40,14 @@ export function IconButton(props: IconButtonProps): ReactElement {
...
           aria-label={props.tooltipTitle}
-          sx={props.sx}
+          sx={
+            props.hidden
+              ? getSxFromPropsAndClasses({
+                  styleClass: classes.hidden,
+                  sxProps: props.sx,
+                })
+              : props.sx
+          }
```
lennartclaas, line 50: Nice. 👍

lennartclaas (approved):

Issue #1281, opened 2022-12-23
[1275] Complete implementation of "first step"


benedikt-richter:
This ticket is in the context of #1275.

Complete implementation of "first step".

* Add list components and populate with relevant data
* Note that the relevant data is the data displayed in the package panels (reuse functionality from there to get that data, `getContainedExternalPackages` might be a good entry point)
* Store the user's inputs in local state of popup

![first_step](https://user-images.githubusercontent.com/46576389/209332405-5c3271dd-f58d-4700-a5a0-df7f73f48d72.png)

PR #1319, opened 2023-01-12
Attribution Wizard: Finalize first wizard step

lennartclaas:
### Summary of changes

Relevant data is taken from the redux state and processed in order to fill the lists of the first wizard step (package namespace and version). The second wizard step isn't touched. 

### Further comments
Only the large `input_from_ort.json.gz` provides reasonable data for the wizard but the performance is poor. Maybe we should create a new more appropriate file.

Example image captured with `input_from_ort.json.gz` and resource `package.json`
Unbenannt

Fix: #1281

### DIFF
--- a/src/Frontend/Components/AttributionWizardPackageStep/AttributionWizardPackageStep.tsx
+++ b/src/Frontend/Components/AttributionWizardPackageStep/AttributionWizardPackageStep.tsx
@@ -4,9 +4,12 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import React, { ReactElement } from 'react';
+import { ListWithAttributesItem } from '../../types/types';
 import { ListWithAttributes } from '../ListWithAttributes/ListWithAttributes';
 
 interface AttributionWizardPackageStepProps {
+  attributedPackageNamespaces: Array<ListWithAttributesItem>;
+  attributedPackageNames: Array<ListWithAttributesItem>;
   selectedPackageNamespaceId: string;
   selectedPackageNameId: string;
   handlePackageNamespaceListItemClick: (id: string) => void;
@@ -16,50 +19,18 @@ interface AttributionWizardPackageStepProps {
 export function AttributionWizardPackageStep(
   props: AttributionWizardPackageStepProps
 ): ReactElement {
-  // create dummy data
-  const N = 15;
-  const items = [];
-  const highlightedAttributeIds = [];
-  for (let i = 0; i < N; i++) {
-    items.push({
-      text: `package${i}`,
-      id: `testItemId${i}`,
-      attributes: [
-        {
-          text: `attrib${4 * i}`,
-          id: `testAttributeId${4 * i}`,
-        },
-        {
-          text: `attrib${4 * i + 1}`,
-          id: `testAttributeId${4 * i + 1}`,
-        },
-        {
-          text: `attrib${4 * i + 2}`,
-          id: `testAttributeId${4 * i + 2}`,
-        },
-        {
-          text: `attrib${4 * i + 3}`,
-          id: `testAttributeId${4 * i + 3}`,
-        },
-      ],
-    });
-    highlightedAttributeIds.push(`testAttributeId${4 * i + (i % 4)}`);
-  }
-
   return (
     <>
       <ListWithAttributes
-        listItems={items}
+        listItems={props.attributedPackageNamespaces}
         selectedListItemId={props.selectedPackageNamespaceId}
-        highlightedAttributeIds={highlightedAttributeIds}
         handleListItemClick={props.handlePackageNamespaceListItemClick}
         showAddNewInput={false}
         title={'Package namespace'}
       />
       <ListWithAttributes
-        listItems={items}
+        listItems={props.attributedPackageNames}
         selectedListItemId={props.selectedPackageNameId}
-        highlightedAttributeIds={highlightedAttributeIds}
         handleListItemClick={props.handlePackageNameListItemClick}
         showAddNewInput={false}
         title={'Package name'}
--- a/src/Frontend/Components/AttributionWizardPopup/AttributionWizardPopup.tsx
+++ b/src/Frontend/Components/AttributionWizardPopup/AttributionWizardPopup.tsx
@@ -8,21 +8,26 @@ import MuiBox from '@mui/material/Box';
 import { NotificationPopup } from '../NotificationPopup/NotificationPopup';
 import { ButtonText } from '../../enums/enums';
 import { Breadcrumbs } from '../Breadcrumbs/Breadcrumbs';
-import { useAppDispatch } from '../../state/hooks';
+import { useAppDispatch, useAppSelector } from '../../state/hooks';
 import { closePopup } from '../../state/actions/view-actions/view-actions';
 import { OpossumColors } from '../../shared-styles';
 import { PathBar } from '../PathBar/PathBar';
 import { AttributionWizardPackageStep } from '../AttributionWizardPackageStep/AttributionWizardPackageStep';
 import { AttributionWizardVersionStep } from '../AttributionWizardVersionStep/AttributionWizardVersionStep';
 import { ButtonConfig } from '../../types/types';
+import { getExternalData } from '../../state/selectors/all-views-resource-selectors';
+import {
+  getSelectedResourceId,
+  getResolvedExternalAttributions,
+} from '../../state/selectors/audit-view-resource-selectors';
+import { getContainedExternalPackages } from '../../util/get-contained-packages';
+import { getAttributionWizardPackageListsItems } from './attribution-wizard-popup-helpers';
 
-// const POPUP_CONTENT_PADDING = 48; // TODO: monitored in upcoming tickets
 const attributionWizardPopupHeader = 'Attribution Wizard';
 
 const classes = {
   dialogHeader: {
     whiteSpace: 'nowrap',
-    // width: `calc(100% - ${POPUP_CONTENT_PADDING}px)`, // TODO: monitored in upcoming tickets
   },
   pathFieldAndBreadcrumbsBox: {
     display: 'flex',
@@ -38,6 +43,7 @@ const classes = {
     heigth: 'fit-content',
     marginTop: '12px',
     maxHeight: '70vh',
+    minHeight: '50vh',
     minWidth: '60vw',
   },
   pathBar: {
@@ -50,6 +56,18 @@ const classes = {
 };
 
 export function AttributionWizardPopup(): ReactElement {
+  const selectedResourceId = useAppSelector(getSelectedResourceId);
+  const externalData = useAppSelector(getExternalData);
+  const resolvedExternalAttributions = useAppSelector(
+    getResolvedExternalAttributions
+  );
+  // const popupAttributionId = useAppSelector(getPopupAttributionId);  // TODO: required later
+
+  const dispatch = useAppDispatch();
+  function closeAttributionWizardPopup(): void {
+    dispatch(closePopup());
+  }
+
   const wizardStepIdsToDisplayValues: Array<[string, string]> = [
     ['packageNamespaceAndName', 'package'],
     ['packageVersion', 'version'],
@@ -58,7 +76,6 @@ export function AttributionWizardPopup(): ReactElement {
     (idAndDisplayValue) => idAndDisplayValue[0]
   );
 
-  // TODO: streamline and integrate this logic later
   const [selectedPackageNamespaceId, setSelectedPackageNamespaceId] =
     useState<string>('');
   const [selectedPackageNameId, setSelectedPackageNameId] =
@@ -69,13 +86,20 @@ export function AttributionWizardPopup(): ReactElement {
     wizardStepIds[0]
   );
 
-  const packageNamespaceAndNameSelected: boolean =
+  const packageNamespaceAndNameSelected =
     selectedPackageNamespaceId !== '' && selectedPackageNameId !== '';
 
-  const dispatch = useAppDispatch();
-  function closeAttributionWizardPopup(): void {
-    dispatch(closePopup());
-  }
+  const containedExternalPackages = getContainedExternalPackages({
+    selectedResourceId,
+    externalData,
+    resolvedExternalAttributions,
+  });
+
+  const { attributedPackageNamespaces, attributedPackageNames } =
+    getAttributionWizardPackageListsItems(
+      containedExternalPackages,
+      externalData.attributions
+    );
 
   const handleBreadcrumbsClick = function (wizardStepId: string): void {
     setSelectedWizardStepId(wizardStepId);
@@ -129,6 +153,36 @@ export function AttributionWizardPopup(): ReactElement {
     isDark: false,
   };
 
+  // create dummy version data
+  const N = 15;
+  const dummyPackageVersionListItems = [];
+  const dummyHighlightedPackageIds = [];
+  for (let i = 0; i < N; i++) {
+    dummyPackageVersionListItems.push({
+      text: `version${i}`,
+      id: `versionId-${i}`,
+      attributes: [
+        {
+          text: `package${4 * i}`,
+          id: `packageId-${4 * i}`,
+        },
+        {
+          text: `package${4 * i + 1}`,
+          id: `packageId-${4 * i + 1}`,
+        },
+        {
+          text: `package${4 * i + 2}`,
+          id: `packageId-${4 * i + 2}`,
+        },
+        {
+          text: `package${4 * i + 3}`,
+          id: `packageId-${4 * i + 3}`,
+        },
+      ],
+    });
+    dummyHighlightedPackageIds.push(`packageId-${4 * i + (i % 4)}`);
+  }
+
   return (
     <NotificationPopup
       header={attributionWizardPopupHeader}
@@ -161,6 +215,8 @@ export function AttributionWizardPopup(): ReactElement {
           <MuiBox sx={classes.mainContentBox}>
             {selectedWizardStepId === wizardStepIds[0] ? (
               <AttributionWizardPackageStep
+                attributedPackageNamespaces={attributedPackageNamespaces}
+                attributedPackageNames={attributedPackageNames}
                 selectedPackageNamespaceId={selectedPackageNamespaceId}
                 selectedPackageNameId={selectedPackageNameId}
                 handlePackageNamespaceListItemClick={
@@ -170,6 +226,10 @@ export function AttributionWizardPopup(): ReactElement {
               />
             ) : selectedWizardStepId === wizardStepIds[1] ? (
               <AttributionWizardVersionStep
+                packageVersionListItems={dummyPackageVersionListItems}
+                highlightedPackageNameIds={dummyHighlightedPackageIds}
+                selectedPackageNamespaceId={selectedPackageNamespaceId}
+                selectedPackageNameId={selectedPackageNameId}
                 selectedPackageVersionId={selectedPackageVersionId}
                 handlePackageVersionListItemClick={
                   handlePackageVersionListItemClick
--- a/src/Frontend/Components/AttributionWizardPopup/__tests__/AttributionWizardPopup.test.tsx
+++ b/src/Frontend/Components/AttributionWizardPopup/__tests__/AttributionWizardPopup.test.tsx
@@ -14,8 +14,13 @@ import { setSelectedResourceId } from '../../../state/actions/resource-actions/a
 import { ButtonText, PopupType } from '../../../enums/enums';
 import { getOpenPopup } from '../../../state/selectors/view-selector';
 import { openPopup } from '../../../state/actions/view-actions/view-actions';
+import {
+  Attributions,
+  ResourcesToAttributions,
+} from '../../../../shared/shared-types';
+import { setExternalData } from '../../../state/actions/resource-actions/all-views-simple-actions';
 
-describe('Attribution Wizard Popup', () => {
+describe('AttributionWizardPopup', () => {
   it('renders with header, resource path, and buttons', () => {
     const testStore = createTestAppStore();
     testStore.dispatch(setSelectedResourceId('/thirdParty'));
@@ -39,10 +44,10 @@ describe('Attribution Wizard Popup', () => {
     );
 
     fireEvent.click(screen.getByText(ButtonText.Cancel));
-    expect(getOpenPopup(testStore.getState())).toBe(null);
+    expect(getOpenPopup(testStore.getState())).toBeNull();
   });
 
-  it('renders breadcrumbs in first wizard step', () => {
+  it('renders breadcrumbs', () => {
     const testStore = createTestAppStore();
     testStore.dispatch(setSelectedResourceId('/thirdParty'));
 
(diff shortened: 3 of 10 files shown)


src/Frontend/Components/AttributionWizardPopup/attribution-wizard-popup-helpers.tsx
```diff
@@ -0,0 +1,133 @@
...
+}
+
+export function getAttributionWizardPackageListsItems(
+  containedExternalPackages: Array<AttributionIdWithCount>,
+  externalAttributions: Attributions
+): [Array<ListWithAttributesItem>, Array<ListWithAttributesItem>] {
+  let totalCountAttributions = 0;
+  containedExternalPackages.forEach((containedExternalPackage) => {
+    totalCountAttributions +=
+      containedExternalPackage['childrenWithAttributionCount'] ?? 0;
```
MarkusObendrauf, line 25: functional might be nicer, but its a matter of taste. something like
```
let totalCountAttributions = containedExternalPackages
  .map(containedExternalPackage => containedExternalPackage['childrenWithAttributionCount'] ?? 0)
  .reduce((accumulator, current) => accumulator + current, 0);
```

src/Frontend/Components/AttributionWizardPopup/attribution-wizard-popup-helpers.tsx
```diff
@@ -0,0 +1,133 @@
...
+    .sort(compareTextAndCount);
+}
+
+function compareTextAndCount(a: TextAndCount, b: TextAndCount): number {
+  if (a.count > b.count) {
+    return -1;
+  } else if (a.count < b.count) {
+    return 1;
+  } else {
+    return a.text.localeCompare(b.text, undefined, {
```
MarkusObendrauf, line 102: localeCompare is extremely slow and has led to performance issues in the past, see https://github.com/opossum-tool/OpossumUI/pull/889. dont use it.

src/Frontend/Components/AttributionWizardPopup/attribution-wizard-popup-helpers.tsx
```diff
@@ -0,0 +1,133 @@
...
+      count,
+    }))
+    .sort(compareTextAndCount);
+}
+
+function compareTextAndCount(a: TextAndCount, b: TextAndCount): number {
+  if (a.count > b.count) {
+    return -1;
+  } else if (a.count < b.count) {
+    return 1;
```
MarkusObendrauf, line 100: there is nothing special about returning -1 or 1, all that matters is if the number is greater than 0, less than 0, or equal to 0. So you can write
`if (a.count != b.count) return b.count - a.count`

src/Frontend/Components/AttributionWizardPopup/attribution-wizard-popup-helpers.tsx
```diff
@@ -0,0 +1,133 @@
...
+        : 'empty'; // for monitoring during development
+
+    packageNamespacesAndCounts[packageNamespace] =
+      (packageNamespacesAndCounts[packageNamespace] ?? 0) + packageCount;
+    packageNamesAndCounts[packageName] =
+      (packageNamesAndCounts[packageName] ?? 0) + packageCount;
+  }
+
+  return [packageNamespacesAndCounts, packageNamesAndCounts];
+}
```
MarkusObendrauf, line 83: this function just does the same thing twice. can you instead make a function that you call two times?

src/Frontend/Components/ListWithAttributes/ListWithAttributes.tsx
```diff
@@ -40,6 +41,13 @@ const classes = {
     padding: '0px',
     borderBottom: 2,
     borderColor: OpossumColors.lightBlue,
+    // TODO: Now: Single scroll bar for each list item.
```
MarkusObendrauf, line 44: assuming this one is left intentionally

src/shared/shared-types.ts
```diff
@@ -214,3 +214,14 @@ declare global {
     electronAPI: IElectronAPI;
   }
 }
+
+export interface AttributionWizardListItemAttribute {
```
MarkusObendrauf, line 218: please come up with a better name

MarkusObendrauf, line 218: are these used anywhere?

MarkusObendrauf (changes requested):
I would guess the performance issues you described are caused by `localeCompare` and are trivial to fix :)

src/Frontend/Components/AttributionWizardPackageStep/AttributionWizardPackageStep.tsx
```diff
@@ -4,9 +4,12 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import React, { ReactElement } from 'react';
+import { ListWithAttributesItem } from '../../types/types';
 import { ListWithAttributes } from '../ListWithAttributes/ListWithAttributes';
 
 interface AttributionWizardPackageStepProps {
+  packageNamespaceListItems: Array<ListWithAttributesItem>;
+  packageNameListItems: Array<ListWithAttributesItem>;
```
benedikt-richter, line 12: maybe the "ListItem" part could be removed in all. It is technical and does not provide a lot of information. Suggestion "packageNameListItems" -> "attributedPackageNames" and also others

src/Frontend/Components/AttributionWizardPackageStep/AttributionWizardPackageStep.tsx
```diff
@@ -16,53 +19,21 @@ interface AttributionWizardPackageStepProps {
...
     <>
       <ListWithAttributes
-        listItems={items}
+        listItems={props.packageNamespaceListItems}
         selectedListItemId={props.selectedPackageNamespaceId}
-        highlightedAttributeIds={highlightedAttributeIds}
         handleListItemClick={props.handlePackageNamespaceListItemClick}
         showAddNewInput={false}
-        title={'Package namespace'}
+        title={'Package namespaces'}
```
benedikt-richter, line 29: I'd leave the singular, as we are only looking for a single one and we expect a single one for a package. Alternatively "Package namespace options".

src/Frontend/Components/AttributionWizardPopup/AttributionWizardPopup.tsx
```diff
@@ -8,13 +8,20 @@ import MuiBox from '@mui/material/Box';
...
 import { ButtonConfig } from '../../types/types';
+import { getExternalData } from '../../state/selectors/all-views-resource-selectors';
+import {
+  getSelectedResourceId,
+  getResolvedExternalAttributions,
+} from '../../state/selectors/audit-view-resource-selectors';
+import { getContainedExternalPackages } from '../../util/get-contained-packages';
+import { getAttributionWizardPackageListsItems } from './attribution-wizard-popup-helpers';
 
 // const POPUP_CONTENT_PADDING = 48; // TODO: monitored in upcoming tickets
```
benedikt-richter, line 26: is there a result of monitoring?

src/Frontend/Components/AttributionWizardPopup/AttributionWizardPopup.tsx
```diff
@@ -71,11 +90,19 @@ export function AttributionWizardPopup(): ReactElement {
 
   const packageNamespaceAndNameSelected: boolean =
```
benedikt-richter, line 91: `: boolean` is not necessary

src/Frontend/Components/AttributionWizardPopup/AttributionWizardPopup.tsx
```diff
@@ -71,11 +90,19 @@ export function AttributionWizardPopup(): ReactElement {
...
     selectedPackageNamespaceId !== '' && selectedPackageNameId !== '';
+  const externalAttributions = externalData['attributions'];
```
benedikt-richter, line 93: this can be inlined. Just use `externalData.attributions` instead of `externalAttributions` or move this line up to the selectors to have the "data fetching" at one place.

src/Frontend/Components/AttributionWizardPopup/__tests__/AttributionWizardPopup.test.tsx
```diff
@@ -51,5 +56,83 @@ describe('Attribution Wizard Popup', () => {
     expect(screen.getByText('package')).toBeInTheDocument();
     expect(screen.getByText('version')).toBeInTheDocument();
   });
-  // TODO: More logic required to test for navigation as it requires item selection of dispatched data
+});
+
+describe('AttributionWizardPoup navigation', () => {
+  const testStore = createTestAppStore();
```
benedikt-richter, line 62: This is very very dangerous! All tests use the same store and thus there can be side effects that are very difficult to debug. Use a different store for each test.

src/Frontend/Components/ListWithAttributes/ListWithAttributes.tsx
```diff
@@ -53,14 +61,12 @@ const classes = {
       backgroundColor: OpossumColors.middleBlue,
     },
   },
+  listItemText: {
+    // overflowX: 'visible',
+    // whiteSpace: 'nowrap',
```
benedikt-richter, line 66: remove comment

src/Frontend/Components/ListWithAttributes/ListWithAttributes.tsx
```diff
@@ -84,13 +91,17 @@ export function ListWithAttributes(
       <MuiBox sx={classes.listBox}>
         <MuiList sx={classes.list}>
           {props.listItems.map((item) => (
-            <MuiListItem key={`itemId-${item.id}`} sx={classes.listItem}>
+            <MuiListItem
+              key={`itemId-${item.id}`}
+              sx={Object.assign(classes.listItem, props.sx)}
```
benedikt-richter, line 96: Typically we use `{...classes.listItem, ...props.sx}` instead of `Object.assign(classes.listItem, props.sx)`

src/Frontend/Components/ListWithAttributes/ListWithAttributes.tsx
```diff
@@ -69,10 +75,11 @@ const classes = {
 interface ListWithAttributesProps {
   listItems: Array<ListWithAttributesItem>;
   selectedListItemId: string;
-  highlightedAttributeIds: Array<string>;
+  highlightedAttributeIds?: Array<string>;
   handleListItemClick: (id: string) => void;
   showAddNewInput: boolean; // TODO: required later
   title?: string;
+  sx?: SxProps;
```
benedikt-richter, line 82: As this does not get attached to the root element of this component, I'd prefer a more descriptive name like e..g `listItemSx`

benedikt-richter (changes requested):
Still have to review the new helper functions

src/Frontend/Components/AttributionWizardPopup/attribution-wizard-popup-helpers.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+function sortPackageData(packageDataAndCounts: {
```
benedikt-richter, line 85: Why is this called `sortPackageData`? 
If I look at the function I do not see `packageData` and  `Counts`.

src/Frontend/Components/AttributionWizardPopup/attribution-wizard-popup-helpers.tsx
```diff
@@ -0,0 +1,133 @@
...
+  [namespace: string]: number;
+}): Array<TextAndCount> {
+  return Object.entries(packageDataAndCounts)
+    .map(([dataName, count]) => ({
```
benedikt-richter, line 89: in my opinion `dataName` is too abstact. It is almost never good to have `Data` as part of a variable name.

src/Frontend/Components/AttributionWizardPopup/attribution-wizard-popup-helpers.tsx
```diff
@@ -0,0 +1,133 @@
...
+    getPackageNamespacesAndNamesWithCounts(
+      containedExternalPackages,
+      externalAttributions
+    );
+
+  const packageNamespacesAndCountsArray = sortPackageData(
+    packageNamespacesAndCounts
+  );
+
+  const packageNamesAndCountsArray = sortPackageData(packageNamesAndCounts);
```
benedikt-richter, line 38: as above

src/Frontend/Components/AttributionWizardPopup/attribution-wizard-popup-helpers.tsx
```diff
@@ -0,0 +1,133 @@
...
+      (packageNamesAndCounts[packageName] ?? 0) + packageCount;
+  }
+
+  return [packageNamespacesAndCounts, packageNamesAndCounts];
+}
+
+function sortPackageData(packageDataAndCounts: {
+  [namespace: string]: number;
+}): Array<TextAndCount> {
+  return Object.entries(packageDataAndCounts)
```
benedikt-richter, line 88: this seems very expensive for a huge number of attributed children.

src/Frontend/Components/AttributionWizardPopup/attribution-wizard-popup-helpers.tsx
```diff
@@ -0,0 +1,133 @@
...
+            (count / totalCountAttributes) *
+            100
+          ).toFixed(1)}%)`,
+          id: `${idPrefix}-attribute-${packageAttribute}`,
+        },
+      ],
+    });
+  }
+
+  return packageAttributesListItems;
```
benedikt-richter, line 132: Could be something like the following with much improved naming

```
function getWizardListItems(
  packageAttributesAndCountsArray: Array<TextAndCount>,
  totalCountAttributes: number,
  idPrefix: string
): Array<ListWithAttributesItem> {
  return packageAttributesAndCountsArray.map((packageAttributeAndCount) =>
    getWizardListItem(packageAttributeAndCount, totalCountAttributes, idPrefix)
  );
}

function getWizardListItem(
  packageAttributeAndCount: TextAndCount,
  totalCountAttributes: number,
  idPrefix: string
): ListWithAttributesItem {
  const count = packageAttributeAndCount.count;
  const packageAttribute = packageAttributeAndCount.text;
  return {
    text: packageAttribute,
    id: `${idPrefix}-${packageAttribute}`,
    attributes: [
      {
        text: `count: ${count} (${(
          (count / totalCountAttributes) *
          100
        ).toFixed(1)}%)`,
        id: `${idPrefix}-attribute-${packageAttribute}`,
      },
    ],
  };
}

```

benedikt-richter (changes requested):

src/Frontend/Components/AttributionWizardPopup/AttributionWizardPopup.tsx
```diff
@@ -50,6 +58,18 @@ const classes = {
 };
 
 export function AttributionWizardPopup(): ReactElement {
+  const selectedResourceId = useAppSelector(getSelectedResourceId);
+  const externalData = useAppSelector(getExternalData);
+  const resolvedExternalAttributions = useAppSelector(
+    getResolvedExternalAttributions
+  );
+  // const popupAttributionId = useAppSelector(getPopupAttributionId);  // TODO: required later
```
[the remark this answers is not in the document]
lennartclaas: In an upcoming ticket, we need the current attribution id to update the package info in the redux state. This is just a reminder.

src/Frontend/Components/AttributionWizardPopup/attribution-wizard-popup-helpers.tsx
lennartclaas: I like it. Although `reduce `usually decreases readability. But in this case, it is quite clear.

src/Frontend/Components/AttributionWizardPopup/attribution-wizard-popup-helpers.tsx
lennartclaas: Nice.

src/Frontend/Components/AttributionWizardPopup/attribution-wizard-popup-helpers.tsx
lennartclaas: I did some logging and `localCompare`  seems not to be the issue. It's the rendering process itself, if we have long lists. However, replacing `localCompare` with a simple `return a.text.toLowerCase() < b.text.toLowerCase() ? -1 : 1;` reduced sorting time from about 20ms down to 0.8ms for the large ort file. 👍

lennartclaas: It could make sense to add a new ticket for checking whether `localeCompare` can be replaced with a simpler expression at other locations of the project.

src/Frontend/Components/AttributionWizardPopup/attribution-wizard-popup-helpers.tsx
lennartclaas: That would be possible. But in this case, we would have two `for` loops instead of just a single one. Since containedExternalPackages could become quite large, it could affect the performance.

src/Frontend/Components/ListWithAttributes/ListWithAttributes.tsx
lennartclaas: Yes. Each comment that is marked with `TODO` is intentional.

src/shared/shared-types.ts
lennartclaas: We discussed a long time about this naming. :D I would not like to change it again unless there is a really convincing reason.

lennartclaas: Good point. Strange remnant... The interfaces are defined in `src/Frontend/types/types.ts`.

package.json
```diff
@@ -92,6 +92,7 @@
     "build:dev:win32": "yarn generate-notice && cross-env DISABLE_ESLINT_PLUGIN=true react-scripts build && tsc -p src/ElectronBackend",
     "build:prod": "yarn generate-notice && react-scripts build && tsc -p src/ElectronBackend",
     "compile-all": "tsc -p ./ && tsc --noEmit -p src/ElectronBackend",
+    "test:single": "react-scripts test src/Frontend/Components/AttributionWizardPopup/__tests__/AttributionWizardPopup.test.tsx --watchAll=false --setupFilesAfterEnv=./setupTests.ts",
```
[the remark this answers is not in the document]
lennartclaas: It's hard to remember. 😄

src/Frontend/Components/AttributionWizardPopup/AttributionWizardPopup.tsx
lennartclaas: After having populated the list of the second step with real data (next ticket).

src/Frontend/Components/ListWithAttributes/ListWithAttributes.tsx
lennartclaas: This produces an error "No overload matches this call.". `{ ...classes.listItem, ...props.sx } as SxProps` does the trick but we don't like castings.

src/Frontend/Components/AttributionWizardPopup/attribution-wizard-popup-helpers.tsx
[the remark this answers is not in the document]
lennartclaas: Very good point. Haven't thought about this yet. I also changed this at other locations.

src/Frontend/Components/AttributionWizardPopup/attribution-wizard-popup-helpers.tsx
lennartclaas: Sorting does take less than 1ms for the large ort file.

src/Frontend/Components/AttributionWizardPopup/attribution-wizard-popup-helpers.tsx
lennartclaas: This function can be used to sort `packageNamespacesAndCounts` and `packageNamesAndCounts`. We therefore need an expression that includes namespaces and names and counts. But it's true that data is rather generic. WHat about `sortPackageAttributesAndCounts`?

lennartclaas: But when I think about it, this could lead to confusion with the `attributes` in the `ListWithAttributesItem`.

src/Frontend/Components/AttributionWizardPopup/attribution-wizard-popup-helpers.tsx
lennartclaas: see above

src/Frontend/Components/AttributionWizardPopup/attribution-wizard-popup-helpers.tsx
```diff
@@ -0,0 +1,133 @@
...
+  for (const containedExternalPackage of containedExternalPackages) {
+    const packageInfo: PackageInfo =
+      externalAttributions[containedExternalPackage['attributionId']];
+    const packageCount: number =
+      containedExternalPackage['childrenWithAttributionCount'] ?? 0;
+
+    const packageNamespace: string =
+      packageInfo['packageNamespace'] !== ''
+        ? packageInfo['packageNamespace'] ?? 'none'
+        : 'empty'; // for monitoring during development
```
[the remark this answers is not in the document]
lennartclaas: I see. I experienced that namespaces and names in the `PackageInfo` of our test files are sometimes undefined (read as 'none') or just empty strings (read as 'empty'). So far, I can only guess if this can also happen with real data.

src/Frontend/Components/ListWithAttributes/ListWithAttributes.tsx
lennartclaas: Solved with `getSxFromPropsAndClasses`.

src/Frontend/Components/AttributionWizardPackageStep/AttributionWizardPackageStep.tsx
lennartclaas: Used singular.

src/Frontend/Components/AttributionWizardPopup/AttributionWizardPopup.tsx
```diff
@@ -71,11 +90,19 @@ export function AttributionWizardPopup(): ReactElement {
...
-  function closeAttributionWizardPopup(): void {
-    dispatch(closePopup());
-  }
+  const containedExternalPackages = getContainedExternalPackages({
+    selectedResourceId,
+    externalData,
+    resolvedExternalAttributions,
+  });
+
+  const [packageNamespaceListItems, packageNameListItems] =
```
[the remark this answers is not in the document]
lennartclaas: Okay, I changed it to `attributedPackageNamespaces `and `attributedPackageNames`.

src/Frontend/Components/AttributionWizardPopup/attribution-wizard-popup-helpers.tsx
MarkusObendrauf: was this measured?

MarkusObendrauf (approved):
Looks good, I would still like to see a refactor of `getPackageNamespacesAndNamesWithCounts` but maybe it's not so simple to do.

src/Frontend/Components/AttributionWizardPopup/attribution-wizard-popup-helpers.tsx
lennartclaas: Actually, the difference is just a fracture of a millisecond. I will refactor it.

benedikt-richter (approved):

PR #1323, opened 2023-01-13
Refactor sx props usage

benedikt-richter:
### Summary of changes

The helper `getSxFromPropsAndClasses` that merges style classes and sx props is used though out the code.

We did not handle applying style classes and sx props to components consistently. This, in particular, caused some confusions: https://github.com/opossum-tool/OpossumUI/pull/1318#discussion_r1068128769

### DIFF
--- a/src/Frontend/Components/Icons/Icons.tsx
+++ b/src/Frontend/Components/Icons/Icons.tsx
@@ -26,6 +26,7 @@ import {
 import { SxProps } from '@mui/material';
 import RectangleIcon from '@mui/icons-material/Rectangle';
 import { Criticality } from '../../../shared/shared-types';
+import { getSxFromPropsAndClasses } from '../../util/get-sx-from-props-and-classes';
 
 const classes = {
   clickableIcon,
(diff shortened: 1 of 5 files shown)


src/Frontend/Components/Icons/Icons.tsx
```diff
@@ -135,15 +148,22 @@ export function DirectoryIcon({
   sx,
   labelDetail,
 }: LabelDetailIconProps): ReactElement {
```
lennartclaas, line 150: What's the point of not using `props` but instead explicitly defining `sx` and `labelDetail`?

src/Frontend/Components/Icons/Icons.tsx
```diff
@@ -161,13 +181,20 @@ export function FileIcon({
...
```
lennartclaas, line 183: as above

lennartclaas (approved):
Nice. Definitely cleaner now. Does the refactoring also affect tests?

leslielazzarino (approved):

src/Frontend/Components/Icons/Icons.tsx
benedikt-richter: I think there is no good point. Actually, the way you are suggesting would improve the current state.

PR #1326, opened 2023-01-16
Update yarn.lock

benedikt-richter:
### Summary of changes

Remove argument "modern" from jest.useFakeTimers calls. Update yarn.lock.

`yarn compile-all` failed after deleting the `package.lock` and running `yarn install`.

Run `yarn compile-all` .


src/ElectronBackend/spdxTools/__tests__/spdxTools.test.ts
```diff
@@ -19,7 +19,7 @@ jest.mock('uuid', () => ({
   v4: (): string => 'testUUID',
 }));
 
-jest.useFakeTimers('modern').setSystemTime(new Date('2020-01-01').getTime());
+jest.useFakeTimers().setSystemTime(new Date('2020-01-01').getTime());
```
antonbauhofer, line 22: We could pass in `FakeTimersConfig` instead of leaving it empty. This way we would ensure that it does not use the legacy option.

antonbauhofer (commented):
I tested it and `yarn start` works. However, if possible, I would prefer if we could use `FakeTimersConfig` instead of not passing an argument to `jest.useFakeTimers()`.

antonbauhofer (approved):

lennartclaas (approved):

PR #1344, opened 2023-01-24
Replace deprecated keypress event

benedikt-richter:
### Summary of changes

The keypress event is replaced.

The keypress event is deprecated.

Select license from dropdown using "Enter" key.


lennartclaas (approved):

antonbauhofer (approved):
Looks good.
I tested it with some logging and it behaves as expected.

PR #1351, opened 2023-01-24
Sort license dropdown

benedikt-richter:
### Summary of changes

* Sort licenses in license dropdown alphabetically.
* For performance reasons `localCompare` is replaced everywhere.

Licenses in license dropdown were not sorted alphabetically.

Check that in UI packages etc are still sorted correctly.

### DIFF
--- a/src/Frontend/Components/AttributionColumn/LicenseField.tsx
+++ b/src/Frontend/Components/AttributionColumn/LicenseField.tsx
@@ -44,7 +44,9 @@ export function LicenseField(props: LicenseFieldProps): ReactElement {
     getFormattedLicenseNamesToShortNameMapping(props.frequentLicenseNames);
   const sortedLicenses = Object.keys(
     formattedLicenseNamesToShortNameMapping
-  ).sort((a, b) => (a.toLowerCase() < b.toLowerCase() ? -1 : 1));
+  ).sort((license, otherLicense) =>
+    license.toLowerCase() < otherLicense.toLowerCase() ? -1 : 1
+  );
 
   function formatOptionForDisplay(option: string): string {
     return formattedLicenseNamesToShortNameMapping[option]
--- a/src/Frontend/Components/AttributionColumn/LicenseField.tsx
+++ b/src/Frontend/Components/AttributionColumn/LicenseField.tsx
@@ -17,18 +17,18 @@ function isPresentInOptions(
   return frequentLicenseNames.some(matchesValue);
 }
 
-function combineLicenseNames(licenseName: FrequentLicenseName): string {
-  return licenseName.shortName + ' - ' + licenseName.fullName;
-}
-
 function getFormattedLicenseNamesToShortNameMapping(
   frequentLicenseNames: Array<FrequentLicenseName>
 ): {
   [key: string]: string;
 } {
+  function formatLicenseName(licenseName: FrequentLicenseName): string {
+    return licenseName.shortName + ' - ' + licenseName.fullName;
+  }
+
   return Object.fromEntries(
     frequentLicenseNames.map((option: FrequentLicenseName) => [
-      combineLicenseNames(option),
+      formatLicenseName(option),
       option.shortName,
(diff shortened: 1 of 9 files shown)


src/Frontend/Components/PackagePanel/package-panel-helpers.ts
```diff
@@ -61,9 +61,10 @@ function sortSources(
         attributionSources[sourceA]?.priority -
         attributionSources[sourceB]?.priority
       ) ||
-      attributionSources[sourceA]?.name.localeCompare(
```
antonbauhofer, line 64: Why do you prefer your implementation to `localeCompare`?

src/Frontend/Components/AttributionColumn/__tests__/LicenseField.test.tsx
```diff
@@ -58,4 +58,36 @@ describe('The LicenseField', () => {
       'GPL - General Public License',
     ]);
   });
+
+  it('sorts entries', () => {
```
antonbauhofer, line 62: We could have a few more entries to sort and include some more exotic names (i.e. containing numbers / umlaute)

src/Frontend/util/get-alphabetical-comparer.ts
```diff
@@ -41,9 +41,9 @@ export function getAlphabeticalComparer(attributions: Attributions) {
     if (elementTitleIsAlphabetical && !otherElementTitleIsAlphabetical)
       return -1;
 
-    return elementTitle.localeCompare(otherElementTitle, undefined, {
-      sensitivity: 'base',
```
antonbauhofer, line 45: what is `sensitivity: 'base'`? Please make sure it's ok to drop that.

antonbauhofer (approved):
Looks good and works as expected.
Just consider my (optional) comments.

src/Frontend/Components/AttributionColumn/LicenseField.tsx
```diff
@@ -42,6 +42,9 @@ interface LicenseFieldProps extends InputElementProps {
 export function LicenseField(props: LicenseFieldProps): ReactElement {
   const formattedLicenseNamesToShortNameMapping =
     getFormattedLicenseNamesToShortNameMapping(props.frequentLicenseNames);
+  const sortedLicenses = Object.keys(
+    formattedLicenseNamesToShortNameMapping
+  ).sort((a, b) => (a.toLowerCase() < b.toLowerCase() ? -1 : 1));
```
lennartclaas, line 47: One could think about using more meaningful names than `a` and `b` as long as these do not reduce readability.

src/Frontend/Components/PackagePanel/package-panel-helpers.ts
benedikt-richter: The performance of `localCompare` is much lower than with the current implementation.

src/Frontend/util/get-alphabetical-comparer.ts
benedikt-richter:
```
// a negative value: in German, ä sorts before z
'ä'.localeCompare('z', 'de')                               // -1// a positive value: in Swedish, ä sorts after z
'ä'.localeCompare('z', 'sv')                               // 1// in German, ä has a as the base letter
'ä'.localeCompare('a', 'de', { sensitivity: 'base' })      // 0// in Swedish, ä and a are separate base letters
'ä'.localeCompare('a', 'sv', { sensitivity: 'base' })      // 1
```

cf. https://medium.com/nerd-for-tech/basics-of-javascript-string-localecompare-method-b2aa50207706

Seems these edge cases are not relavant for us

src/Frontend/Components/AttributionColumn/__tests__/LicenseField.test.tsx
benedikt-richter: These cases seem not very relevant

Issue #1280, opened 2022-12-23
[1275] Add the "add new" functionality


benedikt-richter:
This ticket is in the context of #1275.

There should be a way to create new entries in the menu created in #1279

lennartclaas:
Mybe using MuiChip.

PR #1361, opened 2023-01-30
Attribution Wizard: Implement "add new" functionality.

lennartclaas:
### Summary of changes

A textbox is placed below each list for adding items manually. After having entered a string, the item can be added by clicking the icon button on the right.

Flexibility for the user.

### Further comments

Inserting the textbox below the list was done for better visual separation.

First wizard step:

Fix: #1280

### DIFF
--- a/src/Frontend/Components/AttributionWizardPopup/AttributionWizardPopup.tsx
+++ b/src/Frontend/Components/AttributionWizardPopup/AttributionWizardPopup.tsx
@@ -86,6 +86,16 @@ export function AttributionWizardPopup(): ReactElement {
   const popupAttributionId = useAppSelector(getPopupAttributionId);
   const manualAttributions = useAppSelector(getManualAttributions);
 
+  const [manuallyAddedNamespaces, setManuallyAddedNamespaces] = useState<
+    Array<string>
+  >([]);
+  const [manuallyAddedNames, setManuallyAddedNames] = useState<Array<string>>(
+    []
+  );
+  const [manuallyAddedVersions, setManuallyAddedVersions] = useState<
+    Array<string>
+  >([]);
+
   const dispatch = useAppDispatch();
   function closeAttributionWizardPopup(): void {
     dispatch(closePopup());
@@ -139,7 +149,9 @@ export function AttributionWizardPopup(): ReactElement {
     {
       ...externalData.attributions,
       ...manualData.attributions,
-    }
+    },
+    manuallyAddedNamespaces,
+    manuallyAddedNames
   );
 
   const selectedPackageNamespace = filterForPackageAttributeId(
@@ -156,7 +168,8 @@ export function AttributionWizardPopup(): ReactElement {
     selectedPackageName !== ''
       ? getAttributionWizardPackageVersionListItems(
           selectedPackageName,
-          packageNamesToVersions
+          packageNamesToVersions,
+          manuallyAddedVersions
         )
       : [];
 
@@ -198,7 +211,6 @@ export function AttributionWizardPopup(): ReactElement {
   const handleBreadcrumbsClick = function (wizardStepId: string): void {
     setSelectedWizardStepId(wizardStepId);
   };
-
   function handleNextClick(): void {
     if (selectedWizardStepId !== wizardStepIds[wizardStepIds.length - 1]) {
       setSelectedWizardStepId(
@@ -213,7 +225,6 @@ export function AttributionWizardPopup(): ReactElement {
       );
     }
   }
-
   function handlePackageNamespaceListItemClick(id: string): void {
     setSelectedPackageNamespaceId(id);
   }
@@ -223,7 +234,6 @@ export function AttributionWizardPopup(): ReactElement {
   function handlePackageVersionListItemClick(id: string): void {
     setSelectedPackageVersionId(id);
   }
-
   function handleApplyClick(): void {
     dispatch(
       setTemporaryPackageInfo({
@@ -305,6 +315,10 @@ export function AttributionWizardPopup(): ReactElement {
                   handlePackageNamespaceListItemClick
                 }
                 handlePackageNameListItemClick={handlePackageNameListItemClick}
+                manuallyAddedNamespaces={manuallyAddedNamespaces}
+                setManuallyAddedNamespaces={setManuallyAddedNamespaces}
+                manuallyAddedNames={manuallyAddedNames}
+                setManuallyAddedNames={setManuallyAddedNames}
                 listBoxSx={classes.listBox}
                 listSx={classes.list}
               />
@@ -317,6 +331,8 @@ export function AttributionWizardPopup(): ReactElement {
                 handlePackageVersionListItemClick={
                   handlePackageVersionListItemClick
                 }
+                manuallyAddedVersions={manuallyAddedVersions}
+                setManuallyAddedVersions={setManuallyAddedVersions}
                 listSx={classes.list}
               />
             ) : null}
--- a/src/Frontend/Components/AttributionWizardPopup/attribution-wizard-popup-helpers.tsx
+++ b/src/Frontend/Components/AttributionWizardPopup/attribution-wizard-popup-helpers.tsx
@@ -106,7 +106,9 @@ export function getPreSelectedPackageAttributeIds(popupPackage: PackageInfo): {
 
 export function getAttributionWizardPackageListsItems(
   externalAndManualAttributionIdsWithCounts: Array<AttributionIdWithCount>,
-  externalAndManualAttributions: Attributions
+  externalAndManualAttributions: Attributions,
+  manuallyAddedNamespaces: Array<string>,
+  manuallyAddedNames: Array<string>
 ): {
   attributedPackageNamespaces: Array<ListWithAttributesItem>;
   attributedPackageNames: Array<ListWithAttributesItem>;
@@ -130,7 +132,7 @@ export function getAttributionWizardPackageListsItems(
     'name'
   );
 
-  const packageNamesToVersions = getPackageNamesToVersions(
+  let packageNamesToVersions = getPackageNamesToVersions(
     externalAndManualAttributionIdsWithCounts,
     externalAndManualAttributions
   );
@@ -142,16 +144,34 @@ export function getAttributionWizardPackageListsItems(
     packageNamesAndCounts
   );
 
-  const attributedPackageNamespaces = getWizardListItems(
+  let attributedPackageNamespaces = getWizardListItems(
     packageNamespacesWithCounts,
     totalAttributionCount,
     'namespace'
   );
-  const attributedPackageNames = getWizardListItems(
+  let attributedPackageNames = getWizardListItems(
     packageNamesWithCounts,
     totalAttributionCount,
     'name'
   );
+
+  attributedPackageNamespaces =
+    concatenatePackageAttributesWithManuallyAddedOnes(
+      manuallyAddedNamespaces,
+      attributedPackageNamespaces,
+      'namespace'
+    );
+
+  attributedPackageNames = concatenatePackageAttributesWithManuallyAddedOnes(
+    manuallyAddedNames,
+    attributedPackageNames,
+    'name'
+  );
+
+  packageNamesToVersions = {
+    ...packageNamesToVersions,
+    ...getManuallyAddedPackageNamesToVersions(manuallyAddedNames),
+  };
 
   return {
     attributedPackageNamespaces,
@@ -270,7 +290,8 @@ function getCountText(count: number, totalAttributeCount: number): string {
 
 export function getAttributionWizardPackageVersionListItems(
   packageName: string,
-  packageNamesToVersions: { [name: string]: Set<string> }
+  packageNamesToVersions: { [name: string]: Set<string> },
+  manuallyAddedVersions: Array<string>
 ): Array<ListWithAttributesItem> {
   const packageVersionsToNames: { [version: string]: Set<string> } = {};
   for (const [name, versions] of Object.entries(packageNamesToVersions)) {
@@ -282,9 +303,9 @@ export function getAttributionWizardPackageVersionListItems(
   }
 
   const versions = Array.from(packageNamesToVersions[packageName]).sort();
-  const packageVersionListItems: Array<ListWithAttributesItem> = [];
+  let attributedPackageVersions: Array<ListWithAttributesItem> = [];
   for (const version of versions) {
-    packageVersionListItems.push({
+    attributedPackageVersions.push({
       text: version,
       id: `version-${version}`,
       attributes: Array.from(packageVersionsToNames[version])
@@ -295,7 +316,14 @@ export function getAttributionWizardPackageVersionListItems(
         })),
     });
   }
-  return packageVersionListItems;
+
+  attributedPackageVersions = concatenatePackageAttributesWithManuallyAddedOnes(
+    manuallyAddedVersions,
+    attributedPackageVersions,
+    'version'
+  );
+
+  return attributedPackageVersions;
 }
 
 export function getHighlightedPackageNameIds(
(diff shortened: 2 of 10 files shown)


vasily-pozdnyakov:
### Comments about styling
* It would be nice to have the labels consistent (e.g. see "manually added" tag at different steps)
image
image

package.json
```diff
@@ -92,6 +92,7 @@
     "build:dev:win32": "yarn generate-notice && cross-env DISABLE_ESLINT_PLUGIN=true react-scripts build && tsc -p src/ElectronBackend",
     "build:prod": "yarn generate-notice && react-scripts build && tsc -p src/ElectronBackend",
     "compile-all": "tsc -p ./ && tsc --noEmit -p src/ElectronBackend",
+    "test:single": "react-scripts test src/Frontend/Components/ListWithAttributes/__tests__/ListWithAttributes.test.tsx --watchAll=false --setupFilesAfterEnv=./setupTests.ts",
```
vasily-pozdnyakov, line 95: It should not be in the final version.

src/Frontend/Components/ListWithAttributes/ListWithAttributes.tsx
```diff
@@ -95,6 +99,41 @@ interface ListWithAttributesProps {
...
+    if (!addedInputs.includes(textBoxInput)) {
+      setAddedInputs([...addedInputs, textBoxInput]);
+      if (props.setAddNewListItem) {
+        props.setAddNewListItem([...addedInputs, textBoxInput]);
+      }
+    }
+    setTextBoxInput('');
+  }
+
+  function getAddNewInputIcon(): ReactElement {
```
vasily-pozdnyakov, line 121: There is a file with icons: \Frontend\Components\Icons\Icons.tsx
Please move the icon there.

vasily-pozdnyakov (changes requested):
Small things.

lennartclaas:
First revision update.

First step:


Second step:

src/Frontend/Components/AttributionWizardPopup/attribution-wizard-popup-helpers.tsx
```diff
@@ -317,3 +317,33 @@ export function filterForPackageAttributeId(
...
+export function convertManuallyAddedListEntriesToListItems(
+  manuallyAddedListEntries: Array<string>,
+  packageAttributeId: string
+): Array<ListWithAttributesItem> {
+  return manuallyAddedListEntries.map((item) => ({
+    text: item,
+    id: `${packageAttributeId}-${item}`,
+    attributes: [
+      {
+        text: 'manually added',
```
vasily-pozdnyakov, line 330: I think "manually added" should be a constant (probably in shared constants?) and also exported to src/Frontend/Components/ListWithAttributes/list-with-attributes-helpers.tsx

src/Frontend/Components/ListWithAttributes/list-with-attributes-helpers.tsx
```diff
@@ -28,6 +28,10 @@ export function getAttributesWithHighlighting(
     backgroundColor: OpossumColors.mediumGrey,
   };
 
+  if (attributes?.length === 1 && attributes[0].text === 'manually added') {
```
vasily-pozdnyakov, line 31: why "length===1" ?

vasily-pozdnyakov (approved):
A couple of minor things.

src/Frontend/Components/ListWithAttributes/list-with-attributes-helpers.tsx
lennartclaas: I wasn't sure if the second condition would be enough when we enhance the wizard later. But for now, the second condition should be enough. I'll change it.

src/Frontend/Components/AttributionWizardPopup/attribution-wizard-popup-helpers.tsx
```diff
@@ -317,3 +318,33 @@ export function filterForPackageAttributeId(
...
+    attributes: [
+      {
+        text: LIST_WITH_ATTRIBUTES_DEFAULT_ATTRIBUTE_WHEN_MANUALLY_ADDED_ITEM,
+        id: `${packageAttributeId}-attribute-${item}`,
+      },
+    ],
+  }));
+}
+
+export function getManuallyAddedPackageNamesToVersions(
```
benedikt-richter, line 338: I do not think that there is a lot of value for the user adding a manually added package name to the version field where there is no version...
Why would the user expect that?

src/Frontend/Components/AttributionWizardPopup/AttributionWizardPopup.tsx
```diff
@@ -142,24 +154,61 @@ export function AttributionWizardPopup(): ReactElement {
...
+    )
+  );
+  attributedPackageNames = attributedPackageNames.concat(
+    convertManuallyAddedListEntriesToListItems(manuallyAddedNames, 'name')
+  );
+
+  packageNamesToVersions = {
+    ...packageNamesToVersions,
+    ...getManuallyAddedPackageNamesToVersions(manuallyAddedNames),
+  };
```
benedikt-richter, line 170: It feels, this should happen in `getAttributionWizardPackageListsItems`

src/Frontend/Components/AttributionWizardPopup/AttributionWizardPopup.tsx
```diff
@@ -142,24 +154,61 @@ export function AttributionWizardPopup(): ReactElement {
...
+  console.log(manuallyAddedNamespaces);
+  console.log('manuallyAddedNames');
+  console.log(manuallyAddedNames);
+  console.log('manuallyAddedVersions');
+  console.log(manuallyAddedVersions);
+
+  console.log('selectedPackageNamespaceId');
+  console.log(selectedPackageNamespaceId);
+  console.log('attributedPackageNamespaces');
+  console.log(attributedPackageNamespaces);
```
benedikt-richter, line 183: do not forget to remove

src/Frontend/Components/AttributionWizardPopup/AttributionWizardPopup.tsx
```diff
@@ -142,24 +154,61 @@ export function AttributionWizardPopup(): ReactElement {
...
   const selectedPackageNamespace = filterForPackageAttributeId(
     selectedPackageNamespaceId,
     attributedPackageNamespaces
   );
 
+  // console.log('-------------------');
+  // console.log('selectedPackageNameId');
+  // console.log(selectedPackageNameId);
+  // console.log('attributedPackageNames');
+  // console.log(attributedPackageNames);
```
benedikt-richter, line 194: do not forget to remove

src/Frontend/Components/AttributionWizardPopup/AttributionWizardPopup.tsx
```diff
@@ -142,24 +154,61 @@ export function AttributionWizardPopup(): ReactElement {
...
     selectedPackageName !== ''
       ? getAttributionWizardPackageVersionListItems(
           selectedPackageName,
           packageNamesToVersions
         )
       : [];
 
+  attributedPackageVersions = attributedPackageVersions.concat(
+    convertManuallyAddedListEntriesToListItems(manuallyAddedVersions, 'version')
+  );
```
benedikt-richter, line 210: It feels this should be one function (as above)

src/Frontend/Components/ListWithAttributes/list-with-attributes-helpers.tsx
```diff
@@ -28,6 +31,13 @@ export function getAttributesWithHighlighting(
     backgroundColor: OpossumColors.mediumGrey,
   };
 
+  if (
+    attributes[0].text ===
+    LIST_WITH_ATTRIBUTES_DEFAULT_ATTRIBUTE_WHEN_MANUALLY_ADDED_ITEM
+  ) {
+    showChipsForAttributes = false;
+  }
```
benedikt-richter, line 39: This is a very custom logic to handle this edge case. We match the first element of an array (here we relay on the array only having a single entry). Further we match the display name that is exposed to the user.
Feels there should be a better way.

src/Frontend/shared-styles.ts
```diff
@@ -280,3 +280,6 @@ export const attributionWizardStepClasses = {
 };
 
 export const ATTRIBUTION_WIZARD_PURL_TOTAL_HEIGHT = 45;
+
+export const LIST_WITH_ATTRIBUTES_DEFAULT_ATTRIBUTE_WHEN_MANUALLY_ADDED_ITEM =
```
benedikt-richter, line 284: This should be a shared constant if we want to keep it.

benedikt-richter (changes requested):
Looks good and works nicely. Only some technical comments.

Regarding styling I'd suggest to use some icon to highlight manually added items. That way we do not have to abuse the attributes field and we would not display different information (whether something was added manually vs the package name) in the same spot.

src/Frontend/Components/AttributionWizardPopup/AttributionWizardPopup.tsx
lennartclaas: Is currently needed to prevent a TypeError. Will get obsolet when showing all versions (next ticket).

src/Frontend/shared-styles.ts
lennartclaas: Removed because not needed anymore.

src/Frontend/Components/ListWithAttributes/list-with-attributes-helpers.tsx
lennartclaas: Removed because not needed anymore. Functionality is achieved with new prop in `ListWithAttributesItem`.

src/Frontend/Components/AttributionWizardPopup/attribution-wizard-popup-helpers.tsx
lennartclaas: Is currently needed to prevent a TypeError. Will get obsolet when showing all versions (next ticket).

src/Frontend/Components/ListWithAttributes/list-with-attributes-helpers.tsx
```diff
@@ -11,7 +11,8 @@ import { OpossumColors } from '../../shared-styles';
 export function getAttributesWithHighlighting(
   attributes: Array<ListWithAttributesItemAttribute>,
   showChipsForAttributes: boolean,
-  highlightedAttributeIds?: Array<string>
+  highlightedAttributeIds?: Array<string>,
+  manuallyAddedItem?: boolean
```
benedikt-richter, line 15: Actually it would be nice to have that function being stupid. We should use `manuallyAddedItem` to calculate `showChipsForAttributes` before invoking this function. I.e. to move the logic of l32-34 out of this function

src/Frontend/Components/ListWithAttributes/ListWithAttributes.tsx
```diff
@@ -95,6 +102,26 @@ interface ListWithAttributesProps {
...
+
+  function handleTextBoxInputChange(
+    event: ChangeEvent<HTMLInputElement | HTMLTextAreaElement>
+  ): void {
+    setTextBoxInput(event.target.value);
+  }
+
+  function handleAddNewInputClick(): void {
+    if (!props.manuallyAddedListItems?.includes(textBoxInput)) {
+      if (props.setManuallyAddedListItems && props.manuallyAddedListItems) {
```
benedikt-richter, line 115: these two `if` statements can be merged

src/Frontend/Components/ListWithAttributes/ListWithAttributes.tsx
```diff
@@ -128,22 +155,60 @@ export function ListWithAttributes(
               onClick={(): void => props.handleListItemClick(item.id)}
             >
               <MuiListItemText
-                primary={item.text || (props.emptyTextFallback ?? '-')}
+                primary={
+                  Boolean(item.manuallyAdded) ? (
+                    <MuiBox
+                      sx={{ display: 'flex', justifyContent: 'space-between' }}
```
benedikt-richter, line 161: extract that styling as we typically do

src/Frontend/Components/ListWithAttributes/ListWithAttributes.tsx
```diff
@@ -128,22 +155,60 @@ export function ListWithAttributes(
...
+                    >
+                      {item.text || (props.emptyTextFallback ?? '-')}
```
benedikt-richter, line 163: extract `-` as constant

src/Frontend/Components/ListWithAttributes/ListWithAttributes.tsx
```diff
@@ -128,22 +155,60 @@ export function ListWithAttributes(
...
+                      <ManuallyAddedListItemAdornmentIcon
+                        sx={{
+                          ...baseIcon,
+                          ...{ color: OpossumColors.darkBlue },
```
benedikt-richter, line 167: extract that styling as we typically do

src/Frontend/Components/ListWithAttributes/ListWithAttributes.tsx
```diff
@@ -128,22 +155,60 @@ export function ListWithAttributes(
...
+                        props.showChipsForAttributes,
+                        props.highlightedAttributeIds,
+                        item.manuallyAdded
+                      )}
+                    </MuiBox>
+                  )
+                }
+                primaryTypographyProps={
+                  Boolean(item.manuallyAdded)
+                    ? { margin: '13px 5px 13px 0px' }
```
benedikt-richter, line 189: extract that styling as we typically do

src/Frontend/Components/Icons/Icons.tsx
```diff
@@ -71,6 +75,16 @@ interface LabelDetailIconProps extends IconProps {
...
 
+interface AddNewListItemIconProps {
+  onClick(): void;
+  disabled: boolean;
+}
+
+interface ManuallyAddedListItemAdornmentIconProps {
+  sx?: SxProps;
+  fontSize?: 'inherit' | 'small' | 'medium' | 'large';
+}
```
benedikt-richter, line 86: Should extend `IconProps` or - as described below  not being required

src/Frontend/Components/AttributionWizardPopup/attribution-wizard-popup-helpers.tsx
```diff
@@ -317,3 +345,48 @@ export function filterForPackageAttributeId(
...
+  packageAttributeId: string
+): Array<ListWithAttributesItem> {
+  return manuallyAddedListEntries.map((item) => ({
+    text: item,
+    id: `${packageAttributeId}-${item}`,
+    manuallyAdded: true,
+    attributes: [
+      {
+        text: '-',
+        id: `${packageAttributeId}-attribute-${item}`,
```
benedikt-richter, line 388: why do we set this?
If there is a reason, then why is `'-'` not the empty attribute?

src/Frontend/Components/AttributionWizardPopup/attribution-wizard-popup-helpers.tsx
```diff
@@ -317,3 +345,48 @@ export function filterForPackageAttributeId(
...
+  const manuallyAddedPackageNamesToVersions: {
+    [packageName: string]: Set<string>;
+  } = {};
+  manuallyAddedPackageNames.forEach(
+    (manuallyAddedPackageName) =>
+      (manuallyAddedPackageNamesToVersions[manuallyAddedPackageName] =
+        new Set<string>([emptyAttribute]))
+  );
+  return manuallyAddedPackageNamesToVersions;
+}
```
benedikt-richter, line 361: do we really want to add a manually created package name to be listed at the emtpy version?

benedikt-richter (changes requested):

src/Frontend/Components/AttributionWizardPopup/attribution-wizard-popup-helpers.tsx
lennartclaas: I made attributes optional and removed them here.

benedikt-richter (approved):
👍

Issue #1352, opened 2023-01-25
[1275] Show all available versions and sort by relevance


lennartclaas:
This ticket is in the context of https://github.com/opossum-tool/OpossumUI/issues/1275.

The goal of this ticket is to show all available package versions in the corresponding list of the second wizard step, not only the versions of the selected package name. The version list should then be sorted by relevance. That means:

1. Versions that correspond to the currently selected package name are ar the top.
2. Versions that correspond to package names that are quite similar to the currently selected package name follow.

A suggestion would be to define a measure of similarity or a distance measure between two package names. Needs discussion.

PR #1373, opened 2023-02-01
Attribution Wizard: Show all versions in second step and sort by relevance

lennartclaas:
### Summary of changes

The second step of the wizard shows all versions instead of only the versions
that correspond to the selected package name. Versions that correspond
to the selected package name are displayed at the top.

Second wizard step:

Fix: #1352

### DIFF
--- a/src/Frontend/Components/AttributionWizardPopup/attribution-wizard-popup-helpers.tsx
+++ b/src/Frontend/Components/AttributionWizardPopup/attribution-wizard-popup-helpers.tsx
@@ -3,6 +3,7 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
+import { max } from 'lodash';
 import {
   AttributionData,
   AttributionIdWithCount,
@@ -108,11 +109,17 @@ export function getAttributionWizardPackageListsItems(
   externalAndManualAttributionIdsWithCounts: Array<AttributionIdWithCount>,
   externalAndManualAttributions: Attributions,
   manuallyAddedNamespaces: Array<string>,
-  manuallyAddedNames: Array<string>
+  manuallyAddedNames: Array<string>,
+  manuallyAddedVersions: Array<string>,
+  selectedPackageNamespaceId: string,
+  selectedPackageNameId: string
 ): {
-  attributedPackageNamespaces: Array<ListWithAttributesItem>;
-  attributedPackageNames: Array<ListWithAttributesItem>;
-  packageNamesToVersions: { [packageName: string]: Set<string> };
+  sortedAttributedPackageNamespacesWithManuallyAddedOnes: Array<ListWithAttributesItem>;
+  sortedAttributedPackageNamesWithManuallyAddedOnes: Array<ListWithAttributesItem>;
+  sortedAttributedPackageVersionsWithManuallyAddedOnes: Array<ListWithAttributesItem>;
+  selectedPackageNamespace: string;
+  selectedPackageName: string;
+  highlightedPackageNameIds: Array<string>;
 } {
   const totalAttributionCount = externalAndManualAttributionIdsWithCounts
     .map((attributionIdWithCount) => attributionIdWithCount.count ?? 0)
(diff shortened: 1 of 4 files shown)


src/Frontend/Components/AttributionWizardPopup/attribution-wizard-popup-helpers.tsx
```diff
@@ -168,15 +175,43 @@ export function getAttributionWizardPackageListsItems(
...
+    packageNamesToVersions
+  );
+
+  const packageNameWasManuallyAdded =
+    selectedPackageName in packageNamesToVersions;
+  const highlightedPackageNameIds = packageNameWasManuallyAdded
+    ? getHighlightedPackageNameIds(selectedPackageName, packageNamesToVersions)
+    : [];
+
+  attributedPackageVersions = sortAttributedPackageVersions(
```
benedikt-richter, line 197: `attributedPackageVersions` -> `sortedAttributedPackageVersions`

src/Frontend/Components/AttributionWizardPopup/attribution-wizard-popup-helpers.tsx
```diff
@@ -168,15 +175,43 @@ export function getAttributionWizardPackageListsItems(
...
+    attributedPackageVersions,
+    highlightedPackageNameIds
+  );
+
+  attributedPackageVersions = concatenatePackageAttributesWithManuallyAddedOnes(
```
benedikt-richter, line 202: `attributedPackageVersions` -> `sortedAttributedPackageVersionsWithManuallyAddedOnes`

src/Frontend/Components/AttributionWizardPopup/attribution-wizard-popup-helpers.tsx
```diff
@@ -168,15 +175,43 @@ export function getAttributionWizardPackageListsItems(
...
+  const selectedPackageNamespace = filterForPackageAttributeId(
+    selectedPackageNamespaceId,
+    attributedPackageNamespaces
+  );
+  const selectedPackageName = filterForPackageAttributeId(
+    selectedPackageNameId,
+    attributedPackageNames
+  );
+
+  let attributedPackageVersions = getAttributionWizardPackageVersionListItems(
```
benedikt-richter, line 187: do not reassign and instead use several constants (see below)

src/Frontend/Components/AttributionWizardPopup/attribution-wizard-popup-helpers.tsx
```diff
@@ -302,8 +335,8 @@ export function getAttributionWizardPackageVersionListItems(
     }
   }
 
-  const versions = Array.from(packageNamesToVersions[packageName]).sort();
-  let attributedPackageVersions: Array<ListWithAttributesItem> = [];
+  const versions = Array.from(Object.keys(packageVersionsToNames)).sort();
```
benedikt-richter, line 338: `Array.from` can be dropped, I guess

src/Frontend/Components/AttributionWizardPopup/attribution-wizard-popup-helpers.tsx
```diff
@@ -385,3 +398,72 @@ export function convertManuallyAddedListEntriesToListItems(
...
+
+    const numberOfMatchesA = attributeIdsA?.filter((attributeIdA) =>
+      attributeIdA.match(new RegExp(`${highlightedPackageName}`, 'i'))
+    ).length;
+    matchCounterA += numberOfMatchesA ?? 0;
+
+    const numberOfMatchesB = attributeIdsB?.filter((attributeIdB) =>
+      attributeIdB.match(new RegExp(`${highlightedPackageName}`, 'i'))
+    ).length;
+    matchCounterB += numberOfMatchesB ?? 0;
```
benedikt-richter, line 442: (potentially later) We should not use the ids for any logic, like sorting, the ids should just be technical identifiers.

src/Frontend/Components/AttributionWizardPopup/attribution-wizard-popup-helpers.tsx
```diff
@@ -385,3 +398,72 @@ export function convertManuallyAddedListEntriesToListItems(
...
+  );
+
+  const attributeIdsB = attributedPackageVersionB.attributes?.map(
+    (attribute) => attribute.id
+  );
+
+  let matchCounterA = 0;
+  let matchCounterB = 0;
+  for (const highlightedPackageNameId of highlightedPackageNameIds) {
+    const highlightedPackageName = highlightedPackageNameId.split('-').at(-1);
```
benedikt-richter, line 432: As below, the ids are technical and should never be parsed, sorted or used other than for referencing.

src/Frontend/Components/ListWithAttributes/list-with-attributes-helpers.tsx
```diff
@@ -38,7 +39,11 @@ export function getAttributesWithHighlighting(
           size={'small'}
           sx={{
             ...styleChips,
-            ...(highlightedAttributeIds?.includes(attribute.id)
+            ...(highlightedAttributeIds?.some((item) =>
+              Boolean(caseInsensitiveHighlighting)
+                ? item.match(new RegExp(`${attribute.id}`, 'i'))
+                : item === attribute.id
+            )
```
benedikt-richter, line 46: This feature can be dropped

benedikt-richter (changes requested):

vasily-pozdnyakov:
It would be nice to have a tooltip or any kind of a legend for manually-added items. The current icon is not really self-explanatory.

src/Frontend/Components/AttributionWizardPopup/attribution-wizard-popup-helpers.tsx
lennartclaas: I don't like this renaming. Basically, all arrays that hold `ListWithAttributesItem`s are sorted when they are received in the `AttributionWizardPopup` component. Namespaces and names are sorted by count, versions by relevance and number. If we stick to the naming logic, all of them have to be named with `sorted` at the beginning. To be more precise, one should then add the sorting strategy so the name, like `ByCount` or `ByRelevanceAndNumber`. I don't think that this is necessary.

src/Frontend/Components/AttributionWizardPopup/attribution-wizard-popup-helpers.tsx
lennartclaas: Jep. I added it to the ticket about optimizing the business logic.

src/Frontend/Components/AttributionWizardPopup/attribution-wizard-popup-helpers.tsx
lennartclaas: See above.

lennartclaas:
@vasily-pozdnyakov regarding [comment](https://github.com/opossum-tool/OpossumUI/pull/1373#issuecomment-1411858243) on tooltip.

I added a tooltip to the icon.

src/Frontend/Components/Icons/Icons.tsx
```diff
@@ -72,6 +72,11 @@ interface LabelDetailIconProps extends IconProps {
   disabled?: boolean;
 }
 
+interface ManuallyAddedListItemIconProps extends IconProps {
+  tooltipTitle: string;
+  tooltipPlacement: 'top' | 'right' | 'bottom' | 'left';
```
benedikt-richter, line 77: For all other icons these values are hardcoded. Is there a reason not to do so in this case?

benedikt-richter (approved):
Just consider last comment

vasily-pozdnyakov (approved):
lgtm, very good!

Issue #1317, opened 2023-01-12
Fix icons for folder leaves in resource tree


benedikt-richter:
Currently, the icons for folder leaves in resource tree are clickable and change after click (see screenshot):
![Screenshot 2023-01-12 at 12 50 48](https://user-images.githubusercontent.com/46576389/212059580-c337ed3f-89ff-4abe-ac8f-5f083f015172.png)

Instead the icons of folders that are leaves of the tree should be non-clickable, as clicking has no impact other than changing the icon what is confusing.

someshkhandelia:
How to reproduce this bug?

1. Load "opossum_ui_scan_input.json.gz" file in the application
2. Click a signal "+" e.g. MIT to make it an atttribution
![image](https://user-images.githubusercontent.com/9270691/216002972-977df3c3-2bf2-4ecf-af5a-a94f1f133a83.png)
3. Click 'show resources' button on the attribution
![image](https://user-images.githubusercontent.com/9270691/216003540-d8ce44a4-b778-4862-b79d-adbc068372f0.png)
4. The resource tree that opens, should not have the highlighted icon clickable (currently it does, and that is the bug), because it is a leaf of the tree and clicking the icon doesn't do anything.
![image](https://user-images.githubusercontent.com/9270691/216004085-a9ea9cd3-8886-4800-91a8-5cb047b1a7cd.png)

PR #1378, opened 2023-02-02
Fix icon for folder leaves

someshkhandelia:
### Summary of changes

Fix icon for folder leaves

- If folder is a leaf, it won't have any more children to be found (no keys)
- isEmpty() function from lodash doesn't work for this use-case. (because of recursion and we need to check next level)
- The tests needed to be fixed because now both files and empty folders are treated as proper leaves, and an empty folder has no importance over a file.

Fix: #1317

https://github.com/opossum-tool/OpossumUI/issues/1317#issuecomment-1411735374

### DIFF
--- a/src/Frontend/Components/ResourceBrowser/__tests__/ResourceBrowser.test.tsx
+++ b/src/Frontend/Components/ResourceBrowser/__tests__/ResourceBrowser.test.tsx
@@ -246,10 +246,10 @@ describe('ResourceBrowser', () => {
     expect(screen.queryByText('doesntExist')).not.toBeInTheDocument();
 
     const expectedSequence: Array<string> = [
-      'c_package_folder',
-      'd_package_folder',
       'a_package.exe',
       'b_package.exe',
+      'c_package_folder',
+      'd_package_folder',
     ];
 
     const allPackages = screen.queryAllByText(/package/);
@@ -281,11 +281,11 @@ describe('ResourceBrowser', () => {
     expect(screen.queryByText('doesntExist')).not.toBeInTheDocument();
 
     const expectedSequence: Array<string> = [
-      'a_package_folder',
-      'z_package_folder',
       'a_package.exe',
+      'a_package_folder',
       'package.json',
       'z_package.exe',
+      'z_package_folder',
     ];
 
     const allPackages = screen.queryAllByText(/package/);
--- a/src/Frontend/extracted/VirtualisedTree/utils/get-tree-node-props.tsx
+++ b/src/Frontend/extracted/VirtualisedTree/utils/get-tree-node-props.tsx
@@ -205,7 +205,7 @@ function getSortFunction(
 }
 
 function canNodeHaveChildren(node: NodesForTree | 1): node is NodesForTree {
-  return node !== 1;
+  return node !== 1 && Object.keys(node).length !== 0;
 }
 
 function isIdOfNodeWithChildren(nodeId: string): boolean {


benedikt-richter:
> Signed-off-by: someshkhandelia [someshkhandelia@gmail.com](mailto:someshkhandelia@gmail.com)

This is not necessary for the description

src/Frontend/Components/ResourceBrowser/__tests__/ResourceBrowser.test.tsx
```diff
@@ -246,10 +246,10 @@ describe('ResourceBrowser', () => {
     expect(screen.queryByText('doesntExist')).not.toBeInTheDocument();
 
     const expectedSequence: Array<string> = [
-      'c_package_folder',
-      'd_package_folder',
       'a_package.exe',
       'b_package.exe',
+      'c_package_folder',
```
benedikt-richter, line 251: Why did you modify this test? 
It seems to be broken, as now files are sorted before folders...

src/Frontend/Components/ResourceBrowser/__tests__/ResourceBrowser.test.tsx
```diff
@@ -281,11 +281,11 @@ describe('ResourceBrowser', () => {
     expect(screen.queryByText('doesntExist')).not.toBeInTheDocument();
 
     const expectedSequence: Array<string> = [
-      'a_package_folder',
-      'z_package_folder',
       'a_package.exe',
+      'a_package_folder',
       'package.json',
       'z_package.exe',
+      'z_package_folder',
```
benedikt-richter, line 288: Why did you modify this test?

benedikt-richter (changes requested):
There is a bug. To reproduce:
click on a folder in attribution view that contains no files:
Screenshot 2023-02-02 at 12 39 06
The app then navigates to the audit view as expected but shows the incorrect right panel (incl package panel).

src/Frontend/Components/ResourceBrowser/__tests__/ResourceBrowser.test.tsx
someshkhandelia: The idea was that empty folders and files are treated similarly from now as "leaves" and so an empty folder should probably not have importance over a file.

src/Frontend/Components/ResourceBrowser/__tests__/ResourceBrowser.test.tsx
someshkhandelia: The idea was that empty folders and files are treated similarly from now as "leaves" and so an empty folder should probably not have importance over a file.

src/Frontend/Components/ResourceBrowser/__tests__/ResourceBrowser.test.tsx
benedikt-richter: In this case a test should be added that tests that non-empty folders are sorted corerctly

src/Frontend/Components/ResourceBrowser/__tests__/ResourceBrowser.test.tsx
benedikt-richter: In this case a test should be added that tests that non-empty folders are sorted corerctly

someshkhandelia:
Created a new PR. The condition modified in this PR was being used in several places and several popup components tests were breaking. Please find it here: https://github.com/opossum-tool/OpossumUI/pull/1380

PR #1380, opened 2023-02-02
Fix icon for folder leaves

someshkhandelia:
### Summary of changes

Fix icon for folder leaves

- If folder is a leaf, it won't have any more children to be found (no keys)

Fix: #1317

https://github.com/opossum-tool/OpossumUI/issues/1317#issuecomment-1411735374

### DIFF
--- a/src/Frontend/extracted/VirtualisedTree/utils/get-tree-node-props.tsx
+++ b/src/Frontend/extracted/VirtualisedTree/utils/get-tree-node-props.tsx
@@ -34,8 +34,9 @@ export function getTreeNodeProps(
   for (const nodeName of sortedNodeNames) {
     const node = nodes[nodeName];
     const nodeHeight = cardHeight - 1;
-    const isExpandable = canNodeHaveChildren(node);
-    const nodeId = getNodeId(nodeName, parentPath, isExpandable);
+    const isExpandable =
+      canNodeHaveChildren(node) && Object.keys(node).length !== 0;
+    const nodeId = getNodeId(nodeName, parentPath, canNodeHaveChildren(node));
     const isExpandedNode = isExpanded(nodeId, expandedNodes);
 
     const nodeIdsToExpand: Array<string> = getNodeIdsToExpand(nodeId, node);


src/Frontend/extracted/VirtualisedTree/VirtualizedTreeNode.tsx
```diff
@@ -62,11 +62,12 @@ export function VirtualizedTreeNode(
   const marginRight =
     ((props.nodeId.match(/\//g) || []).length - 1) * INDENT_PER_DEPTH_LEVEL +
     (!props.isExpandable ? SIMPLE_NODE_EXTRA_INDENT : 0);
+  const doesNodeHaveChildren = Object.keys(props.node).length !== 0;
```
benedikt-richter, line 65: I would suggest to move this logic to the pleace, where `props.isExpandable` is provided.
It seems a bit odd to call a component with the property `isExpandable` = `true`and then the component internally has logic to still make it non-expandable.

someshkhandelia:
Created a new PR here. The condition modified in the old PR was being used in several places and several popup components tests were breaking, which should be completely independent of this bug here. Closed the old PR (https://github.com/opossum-tool/OpossumUI/pull/1378).

src/Frontend/extracted/VirtualisedTree/VirtualizedTreeNode.tsx
someshkhandelia: I already tried that in the old PR (https://github.com/opossum-tool/OpossumUI/pull/1378). It broke multiple tests across the project (not just the ones I fixed by commit in the PR). Those tests should have been independent of this change since they did not involve the icon causing the bug here. Yet those tests broke, implying that `isExpandable` has some other implications than just expanding/non-expanding. 

Since this logic is completely coupled with the icon, I think it could also stay where the icon is. (Although I do understand your point completely, it somehow didn't work).

benedikt-richter: The reason why in the other PR some features broke is because the function `canNodeHaveChildren` was modified which is used in different places and has a different semantic meaning. If we introduce a new function `isNodeExpandable` that we just use to check whether a node is expandable, nothing should break. The `isExpandable` property has only a single effect - whether a node is expandable.

someshkhandelia: That makes total sense. Didn't think about `canNodeHaveChildren`. Thanks for pointing that out Benedikt!

benedikt-richter (approved):
👍

vasily-pozdnyakov (approved):

Issue #1374, opened 2023-02-01
Attribution wizard: bug when adding empty item


vasily-pozdnyakov:
When user enters only empty spaces in the add field, the add button should be disabled. If you click "add", the item is buggy.
![image](https://user-images.githubusercontent.com/85183359/216088721-9a01eb40-db2a-4e19-bab0-049f766557c7.png)

PR #1382, opened 2023-02-03
Fix list item icon position

benedikt-richter:
### Summary of changes

The position of icon indicating whether a list item is added manually is fixed for empty list items. Further, the list item is extracted into a separate component.

Fix: #1374

Before:
Screenshot 2023-02-03 at 12 07 11

After:
Screenshot 2023-02-03 at 11 47 03


lennartclaas (approved):
Looks good. Just two minor suggestions: 
1. We could prevent adding a new list item if only unsuitable characters are given, e.g., whitespace and dashes.
2. We could rename the new `ListWithAttributesListItem` component to `ListWithAttributesItem`. Having 'list' in the name twice sounds a bit redundant to me.

vasily-pozdnyakov (changes requested):
Items with the same name allow multiple choice, it is intended behaviour?
image

benedikt-richter:
> Looks good. Just two minor suggestions:
> 
>     1. We could prevent adding a new list item if only unsuitable characters are given, e.g., whitespace and dashes.
> 
>     2. We could rename the new `ListWithAttributesListItem` component to `ListWithAttributesItem`. Having 'list' in the name twice sounds a bit redundant to me.

Regarding the first one, I've implemented it for white spaces. Would not do for "-". We would have to provide an exact requirement for package names instead. E.g. what about "?" or "r-t"?
Regarding the second one, this would clash with the respective type

benedikt-richter:
@vasily-pozdnyakov created #1385

vasily-pozdnyakov (approved):
lgtm

lennartclaas:
👍

Issue #1345, opened 2023-01-24
Attribution Wizard: Clean business logic in helpers


lennartclaas:
Currently, a constant string is used as a fallback value for package attributes (like namespace, name, version, etc.) that are not defined in an attribution:

`export const emptyAttribute = '-';` in `attribution-wizard-popup-helpers.tsx`.

This fallback string is later replaced with `undefined` when defining `selectedPackageInfo` in `AttributionWizardPopup.tsx`. It would be better to completely separate business names from display names.

Furthermore, ids should not be used for logic like sorting. They should just be used as technical identifiers.

This ticket needs further discussion to determine a solution.

PR #1383, opened 2023-02-03
Remove shared constant

benedikt-richter:
### Summary of changes

The constant emptyAttribute is removed and a fallback for the display value is introduced.

Related to #1345

Check that the attribution wizard still shows "-" as fallback for empty attributes.


lennartclaas (approved):
Works nicely.

PR #1410, opened 2023-02-16
POC of .opossum file [DRAFT]

vasily-pozdnyakov:
POC to proceed with actual implementation.

### Summary of changes

Implemented POC of one opossum file (.zip) which contain both input.json and attribution.json. 

One file is more convenient to use.

Run OpossumUI and open any .json file -> find zipped file in the same folder as .json.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -42,7 +42,7 @@ module.exports = {
     '@typescript-eslint/ban-ts-ignore': 0,
     '@typescript-eslint/ban-ts-comment': 0,
     '@typescript-eslint/no-empty-function': 0,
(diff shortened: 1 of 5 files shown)


.eslintrc.js
```diff
@@ -42,7 +42,7 @@ module.exports = {
     '@typescript-eslint/ban-ts-ignore': 0,
     '@typescript-eslint/ban-ts-comment': 0,
     '@typescript-eslint/no-empty-function': 0,
-    '@typescript-eslint/no-explicit-any': 2,
+    '@typescript-eslint/no-explicit-any': 'off',
```
benedikt-richter, line 45: Please, do not commit this eventually.

PR #1411, opened 2023-02-16
POC of .opossum file [DRAFT]

vasily-pozdnyakov:
POC to proceed with actual implementation.

## Summary of changes
Implemented POC of one opossum file (.zip) which contain both input.json and attribution.json.

## Context and reason for change
One file is more convenient to use.

## How can the changes be tested
Run OpossumUI and open any .json file -> find zipped file in the same folder as .json.


benedikt-richter:
Not required anymore.

PR #1422, opened 2023-02-20
Support .opossum files: create info popup

lennartclaas:
### Summary of changes

Implement a new popup for providing information to the user if an outdated
filetype (other than .opossum) is opened. The corresponding component is
not wired.

Uncomment lines 58-60 in `importFromFile.ts`, restart application, and open a file. The popup should be displayed. Since it is not wired, the file will load anyway, independent of the user decision.

#### Screenshots of the popup
Initial version:
Unbenannt

Latest version:
Unbenannt

### DIFF
--- a/src/shared/ipc-channels.ts
+++ b/src/shared/ipc-channels.ts
@@ -24,6 +24,7 @@ export enum AllowedFrontendChannels {
   SaveFileRequest = 'save-file-request',
   SetBaseURLForRoot = 'set-base-url-for-root',
   ShowChangedInputFilePopup = 'show-changed-input-file-pop-up',
+  ShowFileSupportPopup = 'show-file-support-popup',
   ShowSearchPopup = 'show-search-pop-up',
   ShowProjectMetadataPopup = 'show-project-metadata-pop-up',
   ShowProjectStatisticsPopup = 'show-project-statistics-pop-up',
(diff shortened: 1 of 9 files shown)


vasily-pozdnyakov:
1. I think the changes should be based on the current main branch, not POC, because we are currently working on that.
2. The popup should be shown only if we open old format.

lennartclaas:
@vasily-pozdnyakov regarding your [comment](https://github.com/opossum-tool/OpossumUI/pull/1422#issuecomment-1437066901)

1. Makes sense when thinking about it. Rebased.
2. This should be taken care of in the upcoming ticket about wiring the popup. In the current ticket, the popup is not visible at all. Only when uncommenting the lines mentioned in the PR comment (for review only).
3. Good suggestion. I adapted text and buttons.

src/shared/ipc-channels.ts
```diff
@@ -24,6 +24,7 @@ export enum AllowedFrontendChannels {
   SaveFileRequest = 'save-file-request',
   SetBaseURLForRoot = 'set-base-url-for-root',
   ShowChangedInputFilePopup = 'show-changed-input-file-pop-up',
+  ShowDotOpossumFileSupportInfoPopup = 'show-dot-opossum-file-support-info-popup',
```
benedikt-richter, line 27: (Optional) could also be
`ShowFileSupportPopup = 'show-file-support-popup',`

benedikt-richter (approved):
lgtm

One could keep the names a bit shorter and replace `DotOpossumFileSupportInfoPopup` with `FileSupportPopup` throughout this pr. However, I'd consider that optional.

vasily-pozdnyakov (approved):
Looks good.

PR #1423, opened 2023-02-20
Add .opossum support: Add script to generate dot opossum files.

lennartclaas:
### Summary of changes

`yarn generate-dot-opossum` scans `example-files` and generates `.opossum` files. `.opossum` files are ignored by git.

Needed for dev. 

Run script and take a look into `example-files`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -106,6 +106,7 @@
     "update-commit-hash": "run-script-os",
     "update-commit-hash:darwin:linux": "COMMIT_INFO=$(git describe --exact-match --tags 2> /dev/null || git rev-parse --short HEAD); echo \"{\\\"commitInfo\\\" : \\\"$COMMIT_INFO\\\" }\" > \"src/commitInfo.json\";\n",
     "update-commit-hash:win32": "build_scripts/get_app_version_for_windows.bat",
+    "generate-dot-opossum": "yarn node example-files/scripts/generateDotOpossum.js",
     "generate-notice": "run-script-os",
     "generate-notice:darwin:linux": "mkdir -p notices && yarn licenses generate-disclaimer --ignore-platform --production > notices/notices.txt && yarn node build_scripts/generateNotices.js",
     "generate-notice:win32": "(if not exist notices (mkdir notices)) && yarn licenses generate-disclaimer --ignore-platform --production > notices/notices.txt && yarn node build_scripts/generateNotices.js",
--- a/.gitignore
+++ b/.gitignore
@@ -4,11 +4,13 @@
 # testing
 /coverage
 /src/commitInfo.json
+/example-files/*.opossum
 /example-files/*_attributions.json
 /example-files/*_follow_up.csv
 /example-files/*_component_list.csv
 /example-files/*.spdx.yaml
 /example-files/*.spdx.json
+/example-files/performance_tests/*.opossum
 /example-files/performance_tests/*_attributions.json
 /example-files/performance_tests/*_follow_up.csv
 /example-files/performance_tests/*_component_list.csv
--- a/example-files/scripts/generateDotOpossum.js
+++ b/example-files/scripts/generateDotOpossum.js
@@ -1,0 +1,79 @@
+// SPDX-FileCopyrightText: Meta Platforms, Inc. and its affiliates
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+
+const fs = require('fs');
+const JSZip = require('jszip');
+const glob = require('glob');
+const path = require('path');
+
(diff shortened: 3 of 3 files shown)


lennartclaas:
@vasily-pozdnyakov regarding your [comment](https://github.com/opossum-tool/OpossumUI/pull/1423#issuecomment-1437162545)

I revised the variable naming. Looks better to me now.

I refactored parts of the logic into functions to improve readability.

For creating a zip archive via node, I needed to install `archiver`, hence the additional lines in yarn.lock.

lennartclaas:
In general, the location of the script could be changed, as it is not a build script in the narrower sense. Do we have any convention that considers this case?

build_scripts/generateDotOpossum.js
```diff
@@ -0,0 +1,68 @@
...
+
+function addInputJsonToArchive(archive, inputFilePath) {
+  const inputJson = fs.createReadStream(inputFilePath);
+  archive.append(inputJson, { name: 'input.json' });
+}
+
+function addAttributionsJsonToArchive(archive, inputFilePath) {
+  const expectedAssociatedAttributionFilePath =
+    inputFilePath.slice(0, -5) + '_attributions.json';
+  if (attributionFilePaths.includes(expectedAssociatedAttributionFilePath)) {
```
vasily-pozdnyakov, line 62: `attributionFilePaths` is a global variable which is used in the function, so I think it would be nice to make it the function argument.

vasily-pozdnyakov (changes requested):
I do not know, but should we may be add some logging?

build_scripts/generateDotOpossum.js
lennartclaas: I also added logging about the create `.opossum` files.

build_scripts/generateDotOpossum.js
```diff
@@ -0,0 +1,77 @@
+// SPDX-FileCopyrightText: Meta Platforms, Inc. and its affiliates
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+
+const fs = require('fs');
+const archiver = require('archiver');
```
benedikt-richter, line 7: Would be nice to use the same library as we do in the productive code

benedikt-richter (changes requested):

vasily-pozdnyakov (approved):
Minor thing.

build_scripts/generateDotOpossum.js
lennartclaas: Changed it. Now using `JSZip`.

build_scripts/generateDotOpossum.js
```diff
@@ -0,0 +1,76 @@
...
+  dotOpossumArchive
+    .generateAsync({
+      type: 'nodebuffer',
+      streamFiles: true,
+      compression: 'DEFLATE',
+      compressionOptions: { level: 9 },
+    })
+    .then((output) => dotOpossumWriteStream.write(output));
+
+  console.log('Created ' + dotOpossumFilePath);
```
vasily-pozdnyakov, line 43: If the file already exists, the logging should be like "The file already exists".
Actually, it the ".opossum" file already exists it should not overwrite it.

vasily-pozdnyakov (changes requested):

package.json
```diff
@@ -106,6 +106,7 @@
     "update-commit-hash": "run-script-os",
     "update-commit-hash:darwin:linux": "COMMIT_INFO=$(git describe --exact-match --tags 2> /dev/null || git rev-parse --short HEAD); echo \"{\\\"commitInfo\\\" : \\\"$COMMIT_INFO\\\" }\" > \"src/commitInfo.json\";\n",
     "update-commit-hash:win32": "build_scripts/get_app_version_for_windows.bat",
+    "generate-dot-opossum": "yarn node build_scripts/generateDotOpossum.js",
```
benedikt-richter, line 109: This won't work on windows, I guess?

build_scripts/generateDotOpossum.js
```diff
@@ -0,0 +1,76 @@
+// SPDX-FileCopyrightText: Meta Platforms, Inc. and its affiliates
```
benedikt-richter, line 1: The script should not be located in this folder. Could be moved into the folder `example-files`.

benedikt-richter (changes requested):

package.json
lennartclaas: Yes.

build_scripts/generateDotOpossum.js
lennartclaas: moved into `/example-files/scripts/`

vasily-pozdnyakov (approved):
lgtm

benedikt-richter (approved):
👍

PR #1426, opened 2023-02-21
Opossum format

MarkusObendrauf:
### Summary of changes

Add support for a new .opossum file format. This file format combines the old `resources.json` and `resources_attributions.json` by zipping them together.

In the current implementation, if you use the old format, you will save to old format, and if you use the new .opossum format, you will save to .opossum format.

Make it easier for a user to work with OpossumUI by only needed one type of file.

* Run the new unit tests
* For manual testing, a new file `example-files/example.opossum` is provided. This is a converted version of `example-files/opossum_input`.


src/ElectronBackend/input/parseFile.ts
```diff
@@ -15,40 +15,77 @@ import * as OpossumInputFileSchema from './OpossumInputFileSchema.json';
...
+      try {
+        parsedInputData = parseAndValidateJson(content, OpossumInputFileSchema);
+      } catch (err) {
+        return {
+          message: `Error: ${opossumfilePath} is not a valid input file.\n${err}`,
+          type: 'jsonParsingError',
+        };
+      }
+    });
+    zip.files['output.json'].async('text').then(function (content) {
```
lennartclaas: It seems we run into trouble here if the `.opossum` file does not include an `output.json` (it is optional).

src/ElectronBackend/input/__tests__/importFromFile.test.ts
```diff
@@ -206,6 +210,47 @@ describe('Test of loading function', () => {
     deleteFolder(temporaryPath);
   });
 
+  it('loads .opossum file and parses jsons successfully', async () => {
```
lennartclaas, line 213: I suggest to add a test that checks whether loading a `.opossum` file without an `output.json` works. Could also be done in `parseFiles.test.ts`.

src/ElectronBackend/input/importFromFile.ts
```diff
@@ -34,80 +33,95 @@ import log from 'electron-log';
...
-  const parsingResult = await parseOpossumInputFile(filePath);
-
-  if (isJsonParsingError(parsingResult)) {
-    log.info('Invalid input file.');
-    await getMessageBoxForParsingError(parsingResult.message);
+  let parsedInputData: ParsedOpossumInputFile;
+  let parsedOutputData: ParsedOpossumOutputFile | null = null;
 
-    return;
+  if (path.extname(filePath) == '.opossum') {
```
lennartclaas, line 58: I guess using `===` would be cleaner.

src/ElectronBackend/input/importFromFile.ts
```diff
@@ -34,80 +33,95 @@ import log from 'electron-log';
...
+    log.info(`Starting to read .opossum file ${filePath} ...`);
+    const parsingResult = await parseOpossumFile(filePath);
+    if (isJsonParsingError(parsingResult)) {
+      log.info('Invalid input file.');
+      await getMessageBoxForParsingError(parsingResult.message);
+      return;
+    }
+    [parsedInputData, parsedOutputData] = parsingResult;
```
lennartclaas, line 66: I think we mostly use object destructuring for returning multiple variables in OpossumUI. Using an array could potentially lead to mixing both values if further changes are made in the future. Therefore, parsingResult could be made an object.

lennartclaas (changes requested):
It seems that opening a `.opossum` file without an` output.json` is currently throwing an error. Apart from that, only minor issues.

MarkusObendrauf:
> It seems that opening a `.opossum` file without an `output.json` is currently throwing an error.

Do we need to support that?

src/ElectronBackend/input/importFromFile.ts
MarkusObendrauf: fixed

src/ElectronBackend/input/parseFile.ts
```diff
@@ -98,11 +135,45 @@ export function parseOpossumInputFile(
   return promise;
 }
 
-function parseJson(filePath: fs.PathLike): unknown {
+export function parseOuputJsonFile(
```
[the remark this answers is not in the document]
MarkusObendrauf: corrected

src/ElectronBackend/input/importFromFile.ts
MarkusObendrauf: changed it to return an object

MarkusObendrauf:
> > It seems that opening a `.opossum` file without an `output.json` is currently throwing an error.
> 
> Do we need to support that?

We do, I will work on changes to allow this

src/ElectronBackend/input/parseFile.ts
MarkusObendrauf: fixed

src/ElectronBackend/input/__tests__/importFromFile.test.ts
MarkusObendrauf: I added a new test in `importFromFile.test.ts` and one in `parseFiles.test.ts`

src/ElectronBackend/input/importFromFile.ts
```diff
@@ -34,80 +33,96 @@ import log from 'electron-log';
...
+      await getMessageBoxForParsingError(parsingResult.message);
+      return;
+    }
+    parsedInputData = parsingResult.input;
+    parsedOutputData = parsingResult.output;
+    log.info('... Successfully read .opossum file.\n');
+  } else {
+    log.info(`Starting to parse input file ${filePath}`);
+    const parsingResult = await parseInputJsonFile(filePath);
+    if (isJsonParsingError(parsingResult)) {
```
leslielazzarino, line 72: We should extract this if into a function

src/ElectronBackend/input/importFromFile.ts
```diff
@@ -116,31 +131,55 @@ export async function loadJsonFromFilePath(
...
+function parseOrCreateOutputJsonFile(
+  filePath: string,
+  externalAttributions: Attributions,
+  resourcesToExternalAttributions: AttributionsToResources,
+  projectId: string,
+  inputFileMD5Checksum?: string
+): ParsedOpossumOutputFile {
+  if (!fs.existsSync(filePath)) {
+    log.info(`Starting to create output file, project ID is ${projectId}`);
+    createOutputFile(
```
leslielazzarino, line 167: Should be renamed to createJsonOutputFile imo

src/ElectronBackend/input/parseFile.ts
```diff
@@ -8,47 +8,90 @@ import zlib from 'zlib';
...
+      }
+    });
+    if ('output.json' in zip.files) {
+      zip.files['output.json'].async('text').then(function (content) {
+        parsedOutputData = parseOutputJsonContent(content, opossumfilePath);
+      });
+    }
+  });
+
+  await sleep(500);
```
leslielazzarino, line 52: Why do we have this sleep?

src/ElectronBackend/output/writeJsonToOpossumFile.ts
```diff
@@ -0,0 +1,31 @@
...
+  }
+  await zip
+    .generateAsync({
+      type: 'nodebuffer',
+      streamFiles: true,
+      compression: 'DEFLATE',
+      compressionOptions: { level: 1 },
+    })
+    .then((output) => writeStream.write(output));
+  await sleep(500);
```
leslielazzarino, line 30: Here too?

leslielazzarino (changes requested):

lennartclaas:
When I open a `test.opossum` file that only contains an `input.json`, a new `test_attributions.json` file is created next to the `test.opossum` file. Furthermore, it could be beneficial to throw a warning (no error) if a `.opossum` file is openend that does not contain JSONs with proper naming. But I guess this is out of the scope of this story.

src/ElectronBackend/input/parseFile.ts
MarkusObendrauf: We sleep in order to wait for some async stuff to finish. I was unable to write a sync version (JS zip library only supports aasync stuff it seems), and also failed to get the async version working correctly. This is a terrible quick fix so that I unblock other tickets.

MarkusObendrauf:
> When I open a `test.opossum` file that only contains an `input.json`, a new `test_attributions.json` file is created next to the `test.opossum` file.

Good find, I will work on a fix

src/ElectronBackend/input/importFromFile.ts
MarkusObendrauf: Not possible because we return early from the function `loadInputAndOutputFromFilePath`.

src/ElectronBackend/output/writeJsonToOpossumFile.ts
MarkusObendrauf: same as above

MarkusObendrauf:
I managed to eliminate the `sleep`s by wrapping stuff in promises. I hope that's a reasonable solution.

src/ElectronBackend/input/parseFile.ts
```diff
@@ -8,47 +8,96 @@ import zlib from 'zlib';
...
-      OpossumOutputFileSchema,
-      validationOptions
-    );
-  } catch (err) {
-    throw new Error(
-      `Error: ${attributionFilePath} is not a valid attribution file.\n${err}`
-    );
-  }
+  await new Promise<void>((resolve) => {
+    fs.readFile(opossumfilePath, function (err, data) {
```
leslielazzarino, line 35: We usually use arrow functions when inlining functions. Is there a specific reason here not to use them? Otherwise I would prefer to have them here too.

src/ElectronBackend/input/parseFile.ts
```diff
@@ -8,47 +8,96 @@ import zlib from 'zlib';
...
+}
+
+export async function addOutputToOpossumFile(
+  opossumfilePath: string,
+  outputfileData: unknown
+): Promise<void> {
+  const new_zip = new JSZip();
+
+  await new Promise<void>((resolve) => {
+    fs.readFile(opossumfilePath, function (err, data) {
```
leslielazzarino, line 78: Same here.

leslielazzarino (approved):

lennartclaas (approved):
lgtm

src/ElectronBackend/input/importFromFile.ts
```diff
@@ -15,17 +15,17 @@ import {
 import { getGlobalBackendState } from '../main/globalBackendState';
 import {
   cleanNonExistentAttributions,
-  cleanNonExistentResolvedExternalAttributions,
```
benedikt-richter, line 18: Why do we no longer use this one?

benedikt-richter:
Might be a good point to get rid of the namings `input` and `output` :)
Given that we have a single input file that also is the output file, we could consider renaming these to better describe what these are.
E.g. the input might be the "immutable" part, while the output might be the "mutable" part (that's kind of the defining properties of these files)

src/ElectronBackend/input/parseFile.ts
MarkusObendrauf: Switched them to arrows

src/ElectronBackend/input/importFromFile.ts
MarkusObendrauf: Good find. Added it back.

MarkusObendrauf:
> Might be a good point to get rid of the namings `input` and `output` :) Given that we have a single input file that also is the output file, we could consider renaming these to better describe what these are. E.g. the input might be the "immutable" part, while the output might be the "mutable" part (that's kind of the defining properties of these files)

I personally like this naming, I think it's something we should discuss.

vasily-pozdnyakov:
When you try to open .opossum file, the opossumUI still checks hash of the input file and compares it with the one from _attribution.json (if it exists next to .opossum file).

src/ElectronBackend/main/listeners.ts
```diff
@@ -173,17 +187,18 @@ function initializeGlobalBackendState(
...
-  },
-  mainWindow: BrowserWindow,
-  filePath: string
+  }
 ): Promise<void> {
   log.info('Checking input file checksum');
   if (
-    checksums.actualInputFileChecksum === checksums.parsedInputFileChecksum ||
-    !checksums.parsedInputFileChecksum
+    checksums &&
```
benedikt-richter, line 199: This logic is broken, as it implies that all .opossum files raise an error. This is because the variable `checksum` is always falsy in that case.

src/ElectronBackend/main/listeners.ts
```diff
@@ -111,18 +122,21 @@ export async function handleOpeningFile(
...
+  let checksums;
+  if (!isOpossumFileFormat(resourceFilePath)) {
+    checksums = getActualAndParsedChecksums(resourceFilePath);
+  }
 
   log.info('Initializing global backend state');
   initializeGlobalBackendState(
     resourceFilePath,
-    checksums.actualInputFileChecksum
+    checksums === undefined ? undefined : checksums.actualInputFileChecksum
```
benedikt-richter, line 133: suggestion: 
replace
`checksums === undefined ? undefined : checksums.actualInputFileChecksum`
by
`checksums.?actualInputFileChecksum`

benedikt-richter (changes requested):
Currently, the app always shows a warning when opening a .opossum file.

benedikt-richter:
Further observation:
Currently, when opening a .opossum file, we still set the variable `attributionFilePath` in the globalBackendState. The value  '/Users/.../opossum_input_attributions.json' is set in this case. This is confusing (this file does not exist and never will) and instead it should most likely not be set or being explicitly set to `null`.

src/ElectronBackend/main/listeners.ts
MarkusObendrauf: I cleaned up this section.

MarkusObendrauf:
> Further observation: Currently, when opening a .opossum file, we still set the variable `attributionFilePath` in the globalBackendState. The value '/Users/.../opossum_input_attributions.json' is set in this case. This is confusing (this file does not exist and never will) and instead it should most likely not be set or being explicitly set to `null`.

I explicitly set it to `null` now.

benedikt-richter (changes requested):
Unfortunately, saving fails now:
Screenshot 2023-02-24 at 11 01 25


origin: `src/ElectronBackend/main/listeners.ts` approx l.64

MarkusObendrauf:
> Unfortunately, saving fails now:
> 
> origin: `src/ElectronBackend/main/listeners.ts` approx l.64

Theres an error because `attributionFilePath == null`. Ideally, we should have a different field for this in the backend state, for now I did the quick fix of checking for `undefined` instead of `undefined | null`

benedikt-richter (approved):
works nicely 👍

PR #1427, opened 2023-02-21
Add E2E tests for .opossum format

MarkusObendrauf:
### Summary of changes

Add E2E tests for .opossum format

Check that it is possible to use the new file format from command line.

`yarn test:e2e`


vasily-pozdnyakov (approved):

src/e2e-tests/__tests__/e2e.test.ts
```diff
@@ -45,7 +45,60 @@ test.describe('The OpossumUI', () => {
...
+      window,
+      'ElectronBackend'
+    );
+    await electronBackendEntry.click();
+
+    await expect(window.locator(`text=${'jQuery, 16.13.1'}`)).toBeVisible({
+      timeout: EXPECT_TIMEOUT,
+    });
+
+    await expect(window.locator(`text=${'Apache'}`)).toHaveCount(1, {
```
benedikt-richter, line 87: The tests are duplicates of the exisiting one with the slight change 2 -> 1 in the current line with the difference being the input format.
However, I guess, that is fine. There seems to be no need to artificially modify.

benedikt-richter (approved):

PR #1436, opened 2023-02-23
Handle attributions with missing package name in terms of styling

lennartclaas:
### Summary of changes

If the package name is missing, the package cards in attribution view are highlighted with dark orange color. The background color of the corresponding text box changes accordingly.

The user gets more feedback and is alerted if the package name is missing.

Screenshot:
Unbenannt

### DIFF
--- a/src/Frontend/Components/AttributionCountsPanel/AttributionCountsPanel.tsx
+++ b/src/Frontend/Components/AttributionCountsPanel/AttributionCountsPanel.tsx
@@ -10,7 +10,8 @@ import { getManualAttributions } from '../../state/selectors/all-views-resource-
 import { OpossumColors } from '../../shared-styles';
 import {
   FollowUpIcon,
-  IncompletePackagesIcon,
+  IncompleteAttributionsIcon,
+  MissingPackageNameIcon,
   PreSelectedIcon,
 } from '../Icons/Icons';
 import pickBy from 'lodash/pickBy';
@@ -36,6 +37,9 @@ const classes = {
   incompleteAttributionIcon: {
     color: OpossumColors.lightOrange,
   },
+  missingPackageNameIcon: {
+    color: OpossumColors.darkOrange,
+  },
 };
 
 interface AttributionCountsPanelProps {
@@ -56,6 +60,10 @@ export function AttributionCountsPanel(
 
   const numberOfIncompleteAttributions = Object.keys(
     pickBy(attributions, (value: PackageInfo) => isPackageInfoIncomplete(value))
+  ).length;
+
+  const numberOfAttributionsWithoutPackageName = Object.keys(
+    pickBy(attributions, (value: PackageInfo) => !value.packageName)
   ).length;
 
   return (
@@ -80,13 +88,17 @@ export function AttributionCountsPanel(
         }}
       />
       {`, ${numberOfIncompleteAttributions}`}
-      <IncompletePackagesIcon
+      <IncompleteAttributionsIcon
         sx={{
           ...classes.incompleteAttributionIcon,
           ...classes.icons,
         }}
       />
-      )
+      {`, ${numberOfAttributionsWithoutPackageName}`}
+      <MissingPackageNameIcon
+        sx={{ ...classes.icons, ...classes.missingPackageNameIcon }}
+      />
+      {')'}
     </MuiTypography>
   );
 }
--- a/src/Frontend/Components/InputElements/TextBox.tsx
+++ b/src/Frontend/Components/InputElements/TextBox.tsx
@@ -9,6 +9,9 @@ import { inputElementClasses, InputElementProps } from './shared';
 import MuiInputAdornment from '@mui/material/InputAdornment';
 import MuiBox from '@mui/material/Box';
 import { SxProps } from '@mui/material';
+import { HighlightingColor } from '../../enums/enums';
+import { SystemStyleObject } from '@mui/system/styleFunctionSx';
+import { getSxFromPropsAndClasses } from '../../util/get-sx-from-props-and-classes';
 
 interface TextProps extends InputElementProps {
   textFieldSx?: SxProps;
@@ -17,20 +20,32 @@ interface TextProps extends InputElementProps {
   maxRows?: number;
   endIcon?: ReactElement;
   multiline?: boolean;
+  highlightingColor?: HighlightingColor;
 }
 
 export function TextBox(props: TextProps): ReactElement {
+  const isDefaultHighlighting =
+    props.highlightingColor === HighlightingColor.LightOrange ||
+    props.highlightingColor === undefined;
+
+  const highlightedStyling = isDefaultHighlighting
+    ? inputElementClasses.defaultHighlightedTextField
+    : props.highlightingColor === HighlightingColor.DarkOrange
+    ? inputElementClasses.strongHighlightedTextField
+    : {};
+
+  const textBoxSx = getSxFromPropsAndClasses({
+    sxProps: props.isHighlighted ? highlightedStyling : {},
+    styleClass: {
+      ...(props.textFieldSx as SystemStyleObject),
+      ...inputElementClasses.textField,
+    },
+  });
   return (
     <MuiBox sx={props.sx}>
       <MuiTextField
         disabled={!props.isEditable}
-        sx={{
-          ...props.textFieldSx,
-          ...inputElementClasses.textField,
-          ...(props.isHighlighted
-            ? inputElementClasses.highlightedTextField
-            : {}),
-        }}
+        sx={textBoxSx}
         label={props.title}
         InputProps={{
           inputProps: {
--- a/src/Frontend/Components/InputElements/shared.ts
+++ b/src/Frontend/Components/InputElements/shared.ts
@@ -35,13 +35,23 @@ export const inputElementClasses = {
       },
     },
   },
-  highlightedTextField: {
+  defaultHighlightedTextField: {
     '& div': {
       backgroundColor: OpossumColors.lightOrange,
       borderRadius: '0px',
     },
     '& label': {
       backgroundColor: OpossumColors.lightOrange,
+      padding: '1px 3px',
+    },
+  },
+  strongHighlightedTextField: {
+    '& div': {
+      backgroundColor: OpossumColors.darkOrange,
+      borderRadius: '0px',
+    },
+    '& label': {
+      backgroundColor: OpossumColors.darkOrange,
       padding: '1px 3px',
     },
   },
--- a/src/Frontend/Components/ListCard/ListCard.tsx
+++ b/src/Frontend/Components/ListCard/ListCard.tsx
@@ -11,25 +11,13 @@ import { Criticality } from '../../../shared/shared-types';
 import MuiBox from '@mui/material/Box';
 import { SxProps } from '@mui/material';
 import { merge } from 'lodash';
+import { HighlightingColor } from '../../enums/enums';
 
 const defaultCardHeight = '40px';
 const hoveredSelectedBackgroundColor = OpossumColors.middleBlueOnHover;
 const hoveredBackgroundColor = OpossumColors.lightestBlueOnHover;
 const defaultBackgroundColor = OpossumColors.lightestBlue;
 const packageBorder = `1px ${OpossumColors.white} solid`;
-
-const getHighlightedBackground = (
-  highlightColor: string,
-  backgroundColor: string
-): string => {
-  return (
-    'linear-gradient(225deg, ' +
-    highlightColor +
-    ' 44.5px, ' +
-    backgroundColor +
-    ' 0) 0 0/100% 40px no-repeat'
-  );
-};
 
 const classes = {
   root: {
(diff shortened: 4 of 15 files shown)


src/Frontend/Components/AttributionCountsPanel/AttributionCountsPanel.tsx
```diff
@@ -58,6 +62,8 @@ export function AttributionCountsPanel(
     pickBy(attributions, (value: PackageInfo) => isPackageInfoIncomplete(value))
   ).length;
 
+  const numberOfAttributionsWithoutPackageName = 1;
```
benedikt-richter, line 65: ;)

src/Frontend/Components/AttributionCountsPanel/AttributionCountsPanel.tsx
```diff
@@ -86,7 +92,10 @@ export function AttributionCountsPanel(
           ...classes.icons,
         }}
       />
-      )
+      {`, ${numberOfAttributionsWithoutPackageName}`}
```
benedikt-richter, line 95: as above

benedikt-richter, line 95: Not introduced here and therefore optional.
Can we rename `IncompletePackagesIcon` above to `IncompleteAttributionsIcon`

src/Frontend/Components/InputElements/TextBox.tsx
```diff
@@ -54,3 +60,25 @@ export function TextBox(props: TextProps): ReactElement {
     </MuiBox>
   );
 }
+
+function getTextBoxSx(
```
benedikt-richter, line 64: should be a constant instead of a function

src/Frontend/Components/InputElements/TextBox.tsx
```diff
@@ -54,3 +60,25 @@ export function TextBox(props: TextProps): ReactElement {
...
+): MuiSx {
+  return getSxFromPropsAndClasses({
+    sxProps: isHighlighted ? getHighlighting(highlightingColor) : {},
+    styleClass: {
+      ...(textFieldSxFromProps as SystemStyleObject),
+      ...textFieldSx,
+    },
+  });
+}
+function getHighlighting(color: HighlightingColor | undefined): SxProps {
```
benedikt-richter, line 78: as above

src/Frontend/Components/ListCard/ListCard.tsx
```diff
@@ -199,7 +162,7 @@ interface ListCardProps {
   leftIcon?: JSX.Element;
   rightIcons?: Array<JSX.Element>;
   leftElement?: JSX.Element;
-  highlightedCard?: boolean;
+  highlighting?: HighlightingColor | false;
```
benedikt-richter, line 165: typically, we refer to it as `isHighlighted`

src/Frontend/Components/InputElements/shared.ts
```diff
@@ -35,7 +35,7 @@ export const inputElementClasses = {
       },
     },
   },
-  highlightedTextField: {
+  highlightedTextFieldLightOrange: {
```
benedikt-richter, line 38: I'm not a fan of adding the actual color to the name of the class, as this means we have to rename the class in several files once we decide to simply change the color. Or even worse, we change the color and do not rename...
Better use semantic naming: `defaultHighlightedTextField` and below `strongHighlightedTextField`

src/Frontend/util/is-important-attribution-information-missing.ts
```diff
@@ -18,6 +19,17 @@ interface ExtendedAttributionInfo extends AttributionInfo {
   icons: unknown;
 }
 
+export function getPackageCardHighlighting(
+  packageInfo: PackageInfo
+): HighlightingColor | false {
```
benedikt-richter, line 24: Why does this function return `false`? This is surprising.

benedikt-richter (changes requested):

src/Frontend/Components/AttributionCountsPanel/AttributionCountsPanel.tsx
lennartclaas: Oops... Changed that part and adapted the corresponding test.

src/Frontend/Components/AttributionCountsPanel/AttributionCountsPanel.tsx
lennartclaas: Not sure what the first part means. Why should it be optional? 

I renamed the icon component.

src/Frontend/Components/InputElements/TextBox.tsx
lennartclaas: Why? I remember that the convention is to use functions whenever possible. However, I changed it.

src/Frontend/Components/InputElements/TextBox.tsx
lennartclaas: see above

src/Frontend/Components/ListCard/ListCard.tsx
lennartclaas: `isHighlighted` is not an appropriate name anymore, as it refers to a boolean. The new prop speciefies the highlighting color if highlighting should be applied, or is false if no highlighting should be applied. Alternatively, HighlightinhColor could have an entry like `None`. However, this would introduce a lot of logic to handle these cases.

src/Frontend/util/is-important-attribution-information-missing.ts
lennartclaas: As above.

src/Frontend/Components/AttributionCountsPanel/AttributionCountsPanel.tsx
benedikt-richter: "optional" had the meaning that you could do it if you want. Thanks for doing it!

benedikt-richter (changes requested):

src/Frontend/Components/InputElements/TextBox.tsx
```diff
@@ -54,3 +60,25 @@ export function TextBox(props: TextProps): ReactElement {
...
+): MuiSx => {
+  return getSxFromPropsAndClasses({
+    sxProps: isHighlighted ? getHighlighting(highlightingColor) : {},
+    styleClass: {
+      ...(textFieldSxFromProps as SystemStyleObject),
+      ...textFieldSx,
+    },
+  });
+};
+const getHighlighting = (color: HighlightingColor | undefined): SxProps => {
```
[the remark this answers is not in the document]
lennartclaas: Why? By convention or just simplicity? However, I changed it. But now, this constant is computed each time the component renders, even if it is not highlighted. Not a big deal in terms of performance, I think, but still unnecassary.

src/Frontend/Components/InputElements/TextBox.tsx
```diff
@@ -54,3 +60,25 @@ export function TextBox(props: TextProps): ReactElement {
...
+  highlightingColor?: HighlightingColor | undefined
+): MuiSx => {
+  return getSxFromPropsAndClasses({
+    sxProps: isHighlighted ? getHighlighting(highlightingColor) : {},
+    styleClass: {
+      ...(textFieldSxFromProps as SystemStyleObject),
+      ...textFieldSx,
+    },
+  });
+};
```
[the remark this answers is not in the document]
lennartclaas: Why? By convention or just simplicity? However, I changed it.

lennartclaas:
@benedikt-richter regarding your [comment](https://github.com/opossum-tool/OpossumUI/pull/1436#discussion_r1117246552)

> What is the difference between the values undefined and false?

It was intended for compatibility. However, using it turned out to be not necessary. I therefore removed `false`.

src/Frontend/Components/ListCard/ListCard.tsx
lennartclaas: see below

src/Frontend/Components/InputElements/TextBox.tsx
```diff
@@ -17,20 +20,30 @@ interface TextProps extends InputElementProps {
   maxRows?: number;
   endIcon?: ReactElement;
   multiline?: boolean;
+  highlightingColor?: HighlightingColor;
 }
 
 export function TextBox(props: TextProps): ReactElement {
+  const HIGHLIGHTED_STYLING =
```
benedikt-richter, line 27: Why is this ALL_CAPS?

src/Frontend/Components/InputElements/TextBox.tsx
benedikt-richter: Sorry for that confusion. If we can write something as a constant it is preferred to do so instead of writing a function returning that constant.

src/Frontend/Components/InputElements/TextBox.tsx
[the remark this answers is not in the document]
benedikt-richter: This is just for simplicity

src/Frontend/Components/InputElements/TextBox.tsx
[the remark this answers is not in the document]
benedikt-richter: This is just for simplicity

benedikt-richter (approved):

src/Frontend/Components/ListCard/ListCard.tsx
```diff
@@ -326,15 +292,94 @@ function getSx(cardConfig: ListCardConfig, highlightedCard?: boolean): SxProps {
...
+  selected: boolean
+): SxProps {
+  if (selected) {
+    if (color === HighlightingColor.LightOrange) {
+      return getHighlightedStyleClass(
+        OpossumColors.lightOrange,
+        OpossumColors.middleBlue,
+        OpossumColors.lightOrangeOnHover,
+        hoveredSelectedBackgroundColor,
+        packageBorder
```
MarkusObendrauf, line 322: I would prefer if the parameters were in some kind of object, so that we have 
`highlightColor: 'OpossumColors.lightOrange',
backgroundColor: 'OpossumColors.lightOrange'`
etc. Because currently, I just see four colors but can't see what they're used for without scrolling to the function definition of `getHighlightedStyleClass`

MarkusObendrauf: also, can we just write
```
getHighlightedStyleClass(
        useLightOrange ? OpossumColors.lightOrange : OpossumColors.darkOrange,
        selected ? OpossumColors.middleBlue : defaultBackgroundColor,
        useLightOrange  ? OpossumColors.lightOrangeOnHover : OpossumColors.darkOrangeOnHover,
        selected ? hoveredSelectedBackgroundColor : hoveredBackgroundColor,
        packageBorder
```

MarkusObendrauf: another idea: create 4 variables
```
const highlightColor = useLightOrange ? OpossumColors.lightOrange : OpossumColors.darkOrange;
const backgroundColor = selected ? OpossumColors.middleBlue : defaultBackgroundColor;
```
etc. and then just use one call
```
return getHighlightedStyleClass(
        highlightColor ,
        backgroundColor ,
        highlightColorHover,
        backgroundColorHover,
        packageBorder
      );
```
this also clarifies which color is used for which purpose

MarkusObendrauf (changes requested):

src/Frontend/Components/InputElements/TextBox.tsx
```diff
@@ -17,20 +20,30 @@ interface TextProps extends InputElementProps {
...
+    props.highlightingColor === HighlightingColor.LightOrange ||
+    props.highlightingColor === undefined
+      ? inputElementClasses.defaultHighlightedTextField
+      : props.highlightingColor === HighlightingColor.DarkOrange
```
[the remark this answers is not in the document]
lennartclaas: I could imagine that `HighlightingColor` is extended in the future and used in other parts of the code. In this case, we are safe.

src/Frontend/Components/ListCard/ListCard.tsx
lennartclaas: Good idea. I implemented the last approach.

src/Frontend/Components/InputElements/TextBox.tsx
```diff
@@ -17,20 +20,30 @@ interface TextProps extends InputElementProps {
...
+      ? inputElementClasses.strongHighlightedTextField
+      : {};
```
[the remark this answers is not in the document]
MarkusObendrauf: a boolean variable needs to have a name like `useDefaultHighlighting`, making it clear that its a boolean. the current name suggests it is a color.

src/Frontend/Components/ListCard/ListCard.tsx
MarkusObendrauf: the code seems to be written with a non-existent third color in mind, turning 4 lines of code into 30, and i dont like it. (premature optimization?)
and actually, if more colors need to be added, id expect it to be implemented via a map `highlighting style -> (highlightColor, highlightColorOnHover)`. i think it would be fine to use that approach now, but imo its also ok to write code like theres only two colors.

src/Frontend/Components/InputElements/TextBox.tsx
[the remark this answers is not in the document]
lennartclaas: Seems reasonable, however, I would not call such a variable `use...` as use is reserved for hooks. I have chosen `isDefaultHighlighting`.

src/Frontend/Components/ListCard/ListCard.tsx
lennartclaas: Okay, I changed it. But I still think this is dangerous. However, in `PackageCard` the possible colors are restricted to `lightOrange` and `darkOrange`.

MarkusObendrauf (approved):

PR #1442, opened 2023-02-27
Disallow magic numbers

benedikt-richter:
### Summary of changes

In order to improve readability of our code, magic numbers are now forbidden by the linter. Current errors are fixed. Many are just ignored and have to be fixed successively.

Improve readability of our code.

`yarn lint`

### DIFF
--- a/src/Frontend/Components/AttributionWizardPopup/attribution-wizard-popup-helpers.tsx
+++ b/src/Frontend/Components/AttributionWizardPopup/attribution-wizard-popup-helpers.tsx
@@ -102,7 +102,8 @@ export function getAttributionWizardListItems(
 }
 
 function getCountText(count: number, totalAttributeCount: number): string {
-  const percentage = (count / totalAttributeCount) * 100;
+  const percentageFactor = 100;
+  const percentage = (count / totalAttributeCount) * percentageFactor;
   return percentage < 1
     ? `${count} (< 1%)`
     : `${count} (${percentage.toFixed(0)}%)`;
--- a/src/Frontend/Components/BackendCommunication/__tests__/BackendCommunication.test.tsx
+++ b/src/Frontend/Components/BackendCommunication/__tests__/BackendCommunication.test.tsx
@@ -16,7 +16,8 @@ import { Attributions, ExportType } from '../../../../shared/shared-types';
 describe('BackendCommunication', () => {
   it('renders an Open file icon', () => {
     renderComponentWithStore(<BackendCommunication />);
-    expect(window.electronAPI.on).toHaveBeenCalledTimes(11);
+    const expectedNumberOfCalls = 11;
+    expect(window.electronAPI.on).toHaveBeenCalledTimes(expectedNumberOfCalls);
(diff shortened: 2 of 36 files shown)


src/Frontend/Components/AttributionWizardPopup/attribution-wizard-popup-helpers.tsx
```diff
@@ -312,7 +312,8 @@ function getWizardListItem(
 }
 
 function getCountText(count: number, totalAttributeCount: number): string {
-  const percentage = (count / totalAttributeCount) * 100;
+  const percentageFactor = 100;
+  const percentage = (count / totalAttributeCount) * percentageFactor;
```
MarkusObendrauf, line 316: would be in favour of allowing `100`

src/Frontend/Components/BackendCommunication/__tests__/BackendCommunication.test.tsx
```diff
@@ -16,7 +16,8 @@ import { Attributions, ExportType } from '../../../../shared/shared-types';
 describe('BackendCommunication', () => {
   it('renders an Open file icon', () => {
     renderComponentWithStore(<BackendCommunication />);
-    expect(window.electronAPI.on).toHaveBeenCalledTimes(11);
+    const expectedNumberOfCalls = 11;
+    expect(window.electronAPI.on).toHaveBeenCalledTimes(expectedNumberOfCalls);
```
MarkusObendrauf, line 20: extracting like this in an assert just hurts readability imo

lennartclaas (approved):
I agree that prohibiting magic numbers in tests feels somewhat odd. But I have no strong opinion.

src/Frontend/Components/AttributionWizardPopup/attribution-wizard-popup-helpers.tsx
benedikt-richter: That's fine with me.

src/Frontend/Components/BackendCommunication/__tests__/BackendCommunication.test.tsx
benedikt-richter: I think there is no reasonable way to disable a single lint rule for test files. 
In my opinion, it does also increase readablility here. However, would be fine with ignoring the lint in cases where it makes sense (like some in this diff).

MarkusObendrauf: the "before" version reads just like a sentence: "Expect window.electronAPI.on to have been called 11 times".
in the "after" version, you have to hold another value in your head as you're reading the code, and the code reads as "Expect window.electronAPI.on to have been called expected number of calls times". thats definitely harder to read.

MarkusObendrauf:
My thoughts are: There are some places where this has a huge benefit, and some where its a small detriment. So overall, I think it's a good change. I would just like to mitigate as much weird stuff as possible (see review comments). It might be nice to get more opinions.

leslielazzarino (approved):

PR #1452, opened 2023-03-02
Allow user to cancel opening file

benedikt-richter:
### Summary of changes

We allow the user to cancel the process by clicking away or by pressing ESC.

Before this change the user had no way to cancel a file opening operation after one of the new file support popups was shown. 

Open an outdated input file and check that the popup can be closed as described above.

### DIFF
--- a/src/Frontend/Components/FileSupportDotOpossumAlreadyExistsPopup/FileSupportDotOpossumAlreadyExistsPopup.tsx
+++ b/src/Frontend/Components/FileSupportDotOpossumAlreadyExistsPopup/FileSupportDotOpossumAlreadyExistsPopup.tsx
@@ -34,9 +34,13 @@ const classes = {
 export function FileSupportDotOpossumAlreadyExistsPopup(): ReactElement {
   const dispatch = useAppDispatch();
 
+  function close(): void {
+    dispatch(closePopup());
+  }
+
   const handleOpenDotOpossumButtonClick = (): void => {
     window.electronAPI.openDotOpossumFile();
-    dispatch(closePopup());
+    close();
   };
 
   const OpenDotOpossumButtonConfig: ButtonConfig = {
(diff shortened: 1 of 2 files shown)


src/Frontend/Components/FileSupportDotOpossumAlreadyExistsPopup/FileSupportDotOpossumAlreadyExistsPopup.tsx
```diff
@@ -56,6 +60,8 @@ export function FileSupportDotOpossumAlreadyExistsPopup(): ReactElement {
           <MuiTypography>{INFO_TEXT_PART_2}</MuiTypography>
         </MuiBox>
       }
+      onBackdropClick={close}
```
antonbauhofer, line 63: I'm not exactly sure if this is the expected behaviour, but it's fine with me.

antonbauhofer (approved):
I would prefer the popup to have an x- or a cancel button, but this solution is ok for now.
Works as expected.

src/Frontend/Components/FileSupportDotOpossumAlreadyExistsPopup/FileSupportDotOpossumAlreadyExistsPopup.tsx
benedikt-richter: The idea is that, if a user accidentally clicked a wrong file, they can still stop the conversion/opening at this point in time.

vasily-pozdnyakov (approved):
Looks good. Please consider testing (not sure if we already have any tests for closing popup).

benedikt-richter:
> Looks good. Please consider testing (not sure if we already have any tests for closing popup).

In principle, I agree. However, here we basically configure our standard notification popup component. Therefore, I'd rather not repeat the corresponding test.

Issue #1451, opened 2023-03-02
Using the wizard an empty attribution can be created and than saved


leslielazzarino:
It is possible using the wizard to select an empty namespace, an empty package name and an empty version and than save the resulting empty attribution. The saving button should be disabled instead.

PR #1465, opened 2023-03-07
Attribution Wizard: Prevent empty attributions

lennartclaas:
### Summary of changes

The apply button is disabled as long as the generated purl is invalid. This includes empty attributions but also ensures that a truthy package name is given. Additionally, booleans indicating whether a package name, namespace, or version is selected are removed since these are always true due to pre-selection. 

Empty attributions are forbidden and should be avoided.

Open the attribution wizard and select empty package name.

Fix: #1451

### DIFF
--- a/src/Frontend/Components/AttributionWizardPopup/AttributionWizardPopup.tsx
+++ b/src/Frontend/Components/AttributionWizardPopup/AttributionWizardPopup.tsx
@@ -102,6 +102,8 @@ export function AttributionWizardPopup(): ReactElement {
     selectedPackageAttributeIds
   );
 
+  const selectedPackageNameIsValid = selectedPackageName !== '';
+
   const dispatch = useAppDispatch();
   function closeAttributionWizardPopupHandler(): void {
     dispatch(closeAttributionWizardPopup());
@@ -118,11 +120,6 @@ export function AttributionWizardPopup(): ReactElement {
   const [selectedWizardStepId, setSelectedWizardStepId] = useState<string>(
     wizardStepIds[0]
   );
-
-  const isPackageNamespaceAndNameSelected =
-    selectedPackageAttributeIds.namespaceId !== '' &&
-    selectedPackageAttributeIds.nameId !== '';
-  const isPackageVersionSelected = selectedPackageAttributeIds.versionId !== '';
 
   const {
     attributedPackageNamespacesWithManuallyAddedOnes,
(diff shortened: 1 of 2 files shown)


src/Frontend/Components/AttributionWizardPopup/AttributionWizardPopup.tsx
```diff
@@ -246,10 +239,10 @@ export function AttributionWizardPopup(): ReactElement {
   const applyButtonConfig: ButtonConfig = {
     buttonText: ButtonText.Apply,
     onClick: handleApplyClick,
-    disabled: !isPackageVersionSelected,
+    disabled: !isValidPurl,
     isDark: true,
-    tooltipText: !isPackageVersionSelected
-      ? 'Please select package version to apply changes'
+    tooltipText: !isValidPurl
+      ? 'Please generate valid purl to apply changes'
```
benedikt-richter, line 245: I think, for some users it is difficult to know what is a valid PURL. This tooltip might be confusing, as it requires an understanding of the spec for PURLs.

src/Frontend/Components/AttributionWizardPopup/AttributionWizardPopup.tsx
```diff
@@ -224,12 +222,7 @@ export function AttributionWizardPopup(): ReactElement {
   const nextButtonConfig: ButtonConfig = {
     buttonText: ButtonText.Next,
```
benedikt-richter, line 223: Why do we now allow to navigate to the second step? Once in the second step with mit having selected a package name, the user has no way to make the `apply` button clickable.

src/Frontend/Components/AttributionWizardPopup/AttributionWizardPopup.tsx
lennartclaas: Changed the tooltip.

antonbauhofer (commented):
In the summary, you write that version etc. are always true due to pre-selection. Can you elaborate on that?

antonbauhofer (approved):

src/Frontend/Components/AttributionWizardPopup/AttributionWizardPopup.tsx
```diff
@@ -246,12 +243,7 @@ export function AttributionWizardPopup(): ReactElement {
   const applyButtonConfig: ButtonConfig = {
     buttonText: ButtonText.Apply,
     onClick: handleApplyClick,
-    disabled: !isPackageVersionSelected,
     isDark: true,
-    tooltipText: !isPackageVersionSelected
-      ? 'Please select package version to apply changes'
-      : '',
-    tooltipPlacement: 'top',
```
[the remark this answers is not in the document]
lennartclaas: It is not possible that no item in the version list is selected. In an older version this was possible but not anymore.

lennartclaas:
@antonbauhofer regarding your [comment](https://github.com/opossum-tool/OpossumUI/pull/1465#pullrequestreview-1334406278).

> In the summary, you write that version etc. are always true due to pre-selection. Can you elaborate on that?

The wizard is opened via the context menu of a manual attribution. The package attributes of this manual attribution like name, namespace, and version are pre-selected in the corresponding lists in the wizard (the - entry if the attribute is empty). Therefore, it is not possible that no item in a list is selected. The corresponding flags are therefore always true and hence unnecessary.

leslielazzarino (changes requested):
Yous should write a test for the behavior.

leslielazzarino (approved):

Issue #1441, opened 2023-02-27
OpossumUI crashes when opening a `.opossum` file with `input.json` not containing external attributions


lennartclaas:
`.opossum` input parsing does not work if the `input.json` does not contain external attributions. The resulting parsed object is null, leading to an error when the manual attributions are read

 Good entry points are `loadInputAndOutputFromFilePath` and `createOutputInOpossumFile` in `importFromFile.ts`. 

Can be tested by loading one of the files in `example-files/performance_tests` (as `.opossum`).

PR #1469, opened 2023-03-07
Fix the failing file import when no external attributions

vasily-pozdnyakov:
### Summary of changes

We do not read input and output files in parallel, instead, first read output, after that input.json and finish.

Issue #1441

Open one of the .opossum files in performance_tests folder.

### DIFF
--- a/src/ElectronBackend/input/parseFile.ts
+++ b/src/ElectronBackend/input/parseFile.ts
@@ -33,7 +33,13 @@ export async function parseOpossumFile(
   await new Promise<void>((resolve) => {
     fs.readFile(opossumfilePath, (err, data) => {
       if (err) throw err;
-      JSZip.loadAsync(data).then((zip) => {
+      JSZip.loadAsync(data).then(async (zip) => {
+        if ('output.json' in zip.files) {
+          await zip.files['output.json'].async('text').then((content) => {
+            parsedOutputData = parseOutputJsonContent(content, opossumfilePath);
+          });
+        }
+
         zip.files['input.json']
           .async('text')
           .then((content) => {
@@ -46,11 +52,6 @@ export async function parseOpossumFile(
             error = err;
           })
           .then(resolve);
-        if ('output.json' in zip.files) {
-          zip.files['output.json'].async('text').then((content) => {
-            parsedOutputData = parseOutputJsonContent(content, opossumfilePath);
-          });
-        }
       });
     });
   });


lennartclaas (changes requested):
Why does loading output before input actually solves the problem? If large a output.json is loaded, I would assume that loading time is drastically increased. Is there a possibility to load both files asynchronously but ensure that both are read completely before continueing?

vasily-pozdnyakov:
> Why does loading output before input actually solves the problem? If large a output.json is loaded, I would assume that loading time is drastically increased. Is there a possibility to load both files asynchronously but ensure that both are read completely before continueing?

What actually solved the problem: now we await for the output loading result and read input afterwards.
About performance: I tested it on a big file and see no difference. This is probably because the attribution file is usually relatively small and the most of the time is taken by input file reading. Btw, when we open old .json format the app reads resources and attribution files one after another (not async).

leslielazzarino (changes requested):
There should be a test playing the situation that was causing the bug, to avoid regressions in the future.

lennartclaas:
Still, I don't get WHY loading output before input actually solves the problem.

vasily-pozdnyakov:
> Still, I don't get WHY loading output before input actually solves the problem.

Before: start reading input and output simultaneously and return values as soon as input file reading is complete (no matter what happens to reading output file)
Now: first finish reading output file (if exists), and only after that starting to read input file and return both values (input and output) as soon as reading of input file is done.

vasily-pozdnyakov:
> Ah, okay. Now, I got it. Thanks.
> 
> But shouldn't it, in general, be possible to load `input.json` and `output.json` in parallel and wait for both to finish before continuing? Without the need to do it consecutively?

Honestly I do not know the way. Also, sometimes there is no `output.json`.

vasily-pozdnyakov:
> There should be a test playing the situation that was causing the bug, to avoid regressions in the future.

Unfortunately I can not reproduce the bug in tests.

lennartclaas:
Could maybe something like this work:

const [content1, content2] = await Promise.all([
    zip.file(file1)?.async('text'),
    zip.file(file2)?.async('text'),
  ]);

vasily-pozdnyakov:
> Could maybe something like this work:
> 
> const [content1, content2] = await Promise.all([ zip.file(file1)?.async('text'), zip.file(file2)?.async('text'), ]);

might work, but I see no point to make more changes as we have no performance issues.

lennartclaas:
Regarding your [statement](https://github.com/opossum-tool/OpossumUI/pull/1469#issuecomment-1468110563):

> Unfortunately I can not reproduce the bug in tests.

Before your fix, we got the error when loading a small file without external attributions, like `pt_100_opossum_input.opossum` (or  `pt_100_opossum_input.json`). Maybe just loading this file in a test is enough to prove that the fix works?

lennartclaas:
Regarding your [statement](https://github.com/opossum-tool/OpossumUI/pull/1469#issuecomment-1468117919).

> might work, but I see no point to make more changes as we have no performance issues.

I'm not sure whether this is true for HHC workers working on large projects with many manual attributions. However, I'm fine with postponing this topic until it gets relevant.

leslielazzarino (approved):

vasily-pozdnyakov:
> Regarding your [statement](https://github.com/opossum-tool/OpossumUI/pull/1469#issuecomment-1468110563):
> 
> > Unfortunately I can not reproduce the bug in tests.
> 
> Before your fix, we got the error when loading a small file without external attributions, like `pt_100_opossum_input.opossum` (or `pt_100_opossum_input.json`). Maybe just loading this file in a test is enough to prove that the fix works?

I tried that, but it does not fail in the test, no idea why.

lennartclaas (approved):
👍

PR #1477, opened 2023-03-13
Add text field stack for comments

FlorianSchepers:
### Summary of changes

Introduced new component that accepts an array of comments and displays them as an array of text fields for signals and attributions.

We want to enable a signal or attribution to have multiple comments (i.e. if two identical signals with different comments are merged).

- Checkout PR
- Run tests in: `src/Frontend/Components/CommentTextFieldArray/__tests__/CommentTextFieldArray.test.tsx`
- In `src/Frontend/Components/AttributionColumn/AuditingSubPanel.tsx` replace the value for `comments` in the call of the `CommentTextFieldArray` by an hard coded array of strings and check the result in the app.

### DIFF
--- a/src/Frontend/Components/AttributionColumn/AuditingSubPanel.tsx
+++ b/src/Frontend/Components/AttributionColumn/AuditingSubPanel.tsx
@@ -19,6 +19,7 @@ import { useAppSelector } from '../../state/hooks';
 import { checkboxClass } from '../../shared-styles';
 import { isImportantAttributionInformationMissing } from '../../util/is-important-attribution-information-missing';
 import MuiBox from '@mui/material/Box';
+import { TextFieldStack } from '../TextFieldStack/TextFieldStack';
 
 const classes = {
   ...checkboxClass,
@@ -38,7 +39,8 @@ interface AuditingSubPanelProps {
   isEditable: boolean;
   displayPackageInfo: PackageInfo;
   showManualAttributionData: boolean;
-  commentRows: number;
+  isCommentsBoxCollapsed: boolean;
+  commentBoxHeight: number;
   followUpChangeHandler(event: React.ChangeEvent<HTMLInputElement>): void;
   excludeFromNoticeChangeHandler(
     event: React.ChangeEvent<HTMLInputElement>
@@ -135,15 +137,16 @@ export function AuditingSubPanel(props: AuditingSubPanelProps): ReactElement {
           />
         ) : null}
       </MuiBox>
-      <TextBox
+      <TextFieldStack
         isEditable={props.isEditable}
-        sx={classes.textBox}
-        title={'Comment'}
-        text={props.displayPackageInfo.comment}
-        minRows={props.commentRows}
-        maxRows={props.commentRows}
-        multiline={true}
-        handleChange={props.setUpdateTemporaryPackageInfoFor('comment')}
+        comments={
+          props.displayPackageInfo.comment
+            ? [props.displayPackageInfo.comment]
+            : []
+        }
+        isCollapsed={props.isCommentsBoxCollapsed}
+        commentBoxHeight={props.commentBoxHeight}
+        handleChange={props.setUpdateTemporaryPackageInfoFor}
       />
     </MuiPaper>
   );
--- a/src/Frontend/Components/AttributionColumn/AttributionColumn.tsx
+++ b/src/Frontend/Components/AttributionColumn/AttributionColumn.tsx
@@ -106,7 +106,7 @@ export function AttributionColumn(props: AttributionColumnProps): ReactElement {
     setIsLicenseTextShown,
     licenseTextRows,
     copyrightRows,
-    commentRows,
+    commentBoxHeight,
   } = useRows(
     view,
     props.resetViewIfThisIdChanges,
@@ -285,7 +285,8 @@ export function AttributionColumn(props: AttributionColumnProps): ReactElement {
         showHighlight={showHighlight}
       />
       <AuditingSubPanel
-        commentRows={commentRows}
+        commentBoxHeight={commentBoxHeight}
+        isCommentsBoxCollapsed={isLicenseTextShown}
         setUpdateTemporaryPackageInfoFor={
           props.setUpdateTemporaryPackageInfoFor
         }
--- a/src/Frontend/Components/AttributionColumn/attribution-column-helpers.ts
+++ b/src/Frontend/Components/AttributionColumn/attribution-column-helpers.ts
@@ -27,6 +27,9 @@ import { PanelPackage } from '../../types/types';
 
 const PRE_SELECTED_LABEL = 'Attribution was pre-selected';
 const MARKED_FOR_REPLACEMENT_LABEL = 'Attribution is marked for replacement';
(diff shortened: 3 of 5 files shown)


src/Frontend/Components/CommentTextFieldArray/__tests__/CommentTextFieldArray.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+        commentBoxHight={commentBoxHight}
+        setUpdateTemporaryPackageInfoFor={(): (() => void) => doNothing}
+      />
+    );
+    comments.forEach((comment, index) => {
+      expect(screen.getByLabelText(`Comment ${index + 1}`));
+      expect(screen.getByDisplayValue(comment));
+    });
+  });
+  it('renders number of comments if collapsed', () => {
```
lennartclaas, line 39: Extreme nit: empty line between tests.

src/Frontend/Components/CommentTextFieldArray/__tests__/CommentTextFieldArray.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+    const comments = [
+      'This ist the first comment',
+      'This ist the second comment',
+      'This ist the third comment',
+      'This ist the forth comment',
+    ];
+    const isCollapsed = false;
+    const commentBoxHight = 300;
+
+    renderComponentWithStore(
```
lennartclaas, line 25: Nit: render() should also work.

lennartclaas (changes requested):
Looks good so far, but I think some features could be added to make this new component more useful.

lennartclaas:
Could the distance indicated by the arrow be decreased before introducing the vertical scroll bar?

Screenshot 2023-03-13 181037

src/Frontend/Components/CommentTextFieldArray/__tests__/CommentTextFieldArray.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+    );
+    comments.forEach((comment, index) => {
+      expect(screen.getByLabelText(`Comment ${index + 1}`));
+      expect(screen.getByDisplayValue(comment));
+    });
+  });
+  it('renders number of comments if collapsed', () => {
+    const isEditable = false;
+    const comments = [
+      'This ist the first comment',
```
MarkusObendrauf, line 42: same as above

src/Frontend/Components/CommentTextFieldArray/__tests__/CommentTextFieldArray.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+        commentBoxHight={commentBoxHight}
+        setUpdateTemporaryPackageInfoFor={(): (() => void) => doNothing}
+      />
+    );
+    expect(screen.getByLabelText('There are 4 comments'));
+  });
+
+  it('renders correct message in case of one comment', () => {
+    const isEditable = false;
+    const comments = ['This ist the first comment'];
```
MarkusObendrauf, line 64: as above

MarkusObendrauf (changes requested):

src/Frontend/Components/AttributionColumn/AuditingSubPanel.tsx
```diff
@@ -135,15 +137,18 @@ export function AuditingSubPanel(props: AuditingSubPanelProps): ReactElement {
           />
         ) : null}
       </MuiBox>
-      <TextBox
+      <CommentTextFieldArray
```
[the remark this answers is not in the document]
FlorianSchepers: Ok, i will rename it to `TextFieldStack`

FlorianSchepers:
> Could the distance indicated by the arrow be decreased before introducing the vertical scroll bar?
> 
> Screenshot 2023-03-13 181037

FlorianSchepers:
> Could the distance indicated by the arrow be decreased before introducing the vertical scroll bar?
> 
> Screenshot 2023-03-13 181037

There is a hard coded size for the whole attribution panel. So the size of the comment section is calculated based on the remaining size below the other components.

MarkusObendrauf (changes requested):
I tried adding a comment to a component which had none, and it didn't work well.

- Firstly, the comments box is now called "No comments". Instead, "Comments" would be more sensible.
- After typing one letter, the text box height changed, and was deselected. I can't just continue typing, instead I need to select the box a second time.
- The text box is named "Comment 1". This suggests that it is possible to have multiple comments. But this is not actually possible for a user from the outside world. If we do not allow a user to create multiple comments, then it should be named "Comment".  But I think we should support users adding new comments, otherwise it's impossible to undo a comment merge.
- Also, performance on adding text is unbelievably bad. Really hurts the user experiences, it feels necessary to improve that. (Not related to this PR)

src/Frontend/Components/CommentTextFieldArray/CommentTextFieldArray.tsx
```diff
@@ -0,0 +1,78 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+
+import React, { ChangeEvent, ReactElement } from 'react';
+import { TextBox } from '../InputElements/TextBox';
+import MuiBox from '@mui/material/Box';
+
+const classes = {
+  textBoxesBox: {
+    width: '100%',
```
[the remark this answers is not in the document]
lennartclaas: 👍

src/Frontend/Components/CommentTextFieldArray/CommentTextFieldArray.tsx
```diff
@@ -0,0 +1,78 @@
...
+  isEditable: boolean;
+  comments: string[];
+  setUpdateTemporaryPackageInfoFor(
+    propertyToUpdate: string
+  ): (event: ChangeEvent<HTMLInputElement | HTMLTextAreaElement>) => void;
+  showHighlight?: boolean;
+  commentBoxHight: number;
+}
+
+export function CommentTextFieldArray(
```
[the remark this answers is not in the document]
lennartclaas: 👍

lennartclaas (approved):
👍

FlorianSchepers:
> I tried adding a comment to a component which had none, and it didn't work well.
> 
>     * Firstly, the comments box is now called "No comments". Instead, "Comments" would be more sensible.
> 
>     * After typing one letter, the text box height changed, and was deselected. I can't just continue typing, instead I need to select the box a second time.
> 
>     * The text box is named "Comment 1". This suggests that it is possible to have multiple comments. But this is not actually possible for a user from the outside world. If we do not allow a user to create multiple comments, then it should be named "Comment".  But I think we should support users adding new comments, otherwise it's impossible to undo a comment merge.
> 
>     * Also, performance on adding text is unbelievably bad. Really hurts the user experiences, it feels necessary to improve that. (Not related to this PR)

I have fixed your issues as follows:

- If there is no comment an empty comment with the title 'comment' will be shown. 
- Text can be added to this comment without the neet to reselect this comment box.
- Only if there are more than one comment the title of the comment boxes will be numbered, e.g., 'comment 1', 'comment 2' ...
- In case the comment section is collapsed (i.e., extended license text) a non editable text box is shown whichs tates the number of non empty comments.

MarkusObendrauf (approved):

PR #1509, opened 2023-03-27
Fix package card highlighting

### LINKED ISSUE #1503: 1st party flagged attributions should not be marked with dark orange color if package name is missing
In attribution view, attributions with missing package names are highlighted with dark orange color on the right side of the package card. The highlighting should be removed for 1st party attributions (indicated by the corresponding checkbox and the "1" icon on the package card).

Here, it is correctly highlighted (not 1st party).
Screenshot 2023-03-23 111453

Here, the highlighting should be removed.
Screenshot 2023-03-23 111514

vasily-pozdnyakov:
### Summary of changes

Added check it the package is first party or excluded from notices.

#1503

Check it in the UI


benedikt-richter (changes requested):
Please, add a test for the new feature.

benedikt-richter (approved):
👍

lennartclaas (approved):

PR #1527, opened 2023-03-31
Deduplicate Signals: Add originIds to PackageInfo

lennartclaas:
### Summary of changes

In PackageInfo, replace originId (string) with originIds (Array<string>). If the input file contains originId, write it into originIds. Same when loading output files. When saving, the output file only contains originIds.

Part of the story 'Deduplicate reused signals that are identical'.

Run newly added tests or open test files.


src/ElectronBackend/input/OpossumOutputFileSchema.json
```diff
@@ -95,14 +92,19 @@
...
           "originId": {
             "type": "string",
             "description": "Can be set to track a signal from the tooling that generated the input file. Copied from the input file"
+          },
+          "originIds": {
+            "type": "array",
+            "description": "Like originId but considers that a signal can have more than a single origin.",
+            "items": {
+              "type": "string"
+            }
```
MarkusObendrauf, line 107: i *really* dont think we should be changing the output schema. this would break tools that read opossum output files

src/ElectronBackend/input/OpossumInputFileSchema.json
```diff
@@ -141,6 +141,13 @@
           "originId": {
             "type": "string",
             "description": "Can be set to track a signal from the tooling that generated the input file. Will be copied into the output when the signal is converted into an attribution."
+          },
+          "originIds": {
+            "type": "array",
+            "description": "Like originId but considers that a signal can have more than a single origin.",
+            "items": {
+              "type": "string"
+            }
```
MarkusObendrauf, line 150: i don't think we should be changing the input file schema (does not seem to be in the ticket)

MarkusObendrauf (changes requested):

src/ElectronBackend/input/__tests__/parseInputData.test.ts
```diff
@@ -191,6 +191,56 @@ describe('sanitizeRawAttributions', () => {
...
+  it('removes originId if originIds exists', () => {
+    const testRawAttributions: RawAttributions = {
+      uuid: {
+        originId: 'abc',
+        originIds: ['def', 'ghi'],
+      },
+    };
+    const expectedParsedRawAttributions: RawAttributions = {
+      uuid: {
+        originIds: ['def', 'ghi'],
```
MarkusObendrauf, line 204: i would expect 'abc' to be added to this list

src/ElectronBackend/input/importFromFile.ts
```diff
@@ -67,6 +67,9 @@ export async function loadInputAndOutputFromFilePath(
   mainWindow.webContents.send(AllowedFrontendChannels.ResetLoadedFile, {
     resetState: true,
   });
+  console.log(
```
MarkusObendrauf, line 70: remove (+other `console.log`s in this file)

MarkusObendrauf (changes requested):

src/ElectronBackend/input/__tests__/parseInputData.test.ts
lennartclaas: We decided to drop `originId` in case `originIds`  exists as well, since `originId` should be outdated in that case. However, in practice, it should not be possible to have both fields set, unless someone adds it manually.

src/ElectronBackend/input/parseInputData.ts
```diff
@@ -142,10 +142,17 @@ export function parseRawAttributions(
     if (criticality && !Object.values(Criticality).includes(criticality)) {
       delete rawAttributions[attributionId].criticality;
     }
-
     if (rawAttributions[attributionId]?.criticality) {
       inputContainsCriticalExternalAttributions = true;
     }
+    if (rawAttributions[attributionId]?.originId) {
+      if (!rawAttributions[attributionId]?.originIds) {
+        rawAttributions[attributionId].originIds = new Array<string>(
```
benedikt-richter, line 150: Would prefer not to use the constructor for the array, as we typically do not do this

src/ElectronBackend/input/parseInputData.ts
```diff
@@ -142,10 +142,17 @@ export function parseRawAttributions(
...
+          rawAttributions[attributionId].originId ?? ''
```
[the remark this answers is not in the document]
lennartclaas: Changed as discussed. Execution time changed from 0.7 ms to 0.85 ms for the ort file... 😅

src/Frontend/Components/ReportTableItem/report-table-item-helpers.ts
```diff
@@ -13,7 +13,7 @@ export function getFormattedCellData(
   attributionInfo: AttributionInfo,
   isFileWithChildren: PathPredicate
 ): string | number {
-  let cellData: string | number;
+  let cellData: string | number = '';
```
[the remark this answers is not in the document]
lennartclaas: Just for typing reasons. We should discuss this in more detail.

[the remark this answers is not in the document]
lennartclaas: Changed according to discussion.

benedikt-richter (approved):

src/ElectronBackend/input/importFromFile.ts
```diff
@@ -188,7 +188,6 @@ export async function loadInputAndOutputFromFilePath(
   );
 
   log.info('Updating global backend state');
-
```
MarkusObendrauf, line 191: i would keep the line break (looks like an accident)

src/ElectronBackend/input/parseInputData.ts
```diff
@@ -142,10 +142,18 @@ export function parseRawAttributions(
...
     }
-
     if (rawAttributions[attributionId]?.criticality) {
       inputContainsCriticalExternalAttributions = true;
     }
+
+    const originId = rawAttributions[attributionId]?.originId;
+    if (originId) {
+      const originIds = rawAttributions[attributionId]?.originIds;
+      rawAttributions[attributionId].originIds = originIds
```
MarkusObendrauf, line 152: i had to do a double take when reading this line. the `= originIds` looks like an assignment but is actually a condition for a ternary. i would find this easier to read:
```
const originIds = rawAttributions[attributionId].originIds ?? [];
originIds.push(originId);
rawAttributions[attributionId].originIds = originIds;
```
although here we still change `originIds` in place which isnt perfect

MarkusObendrauf (changes requested):
just some minor comments

src/ElectronBackend/input/parseInputData.ts
benedikt-richter: The `= originIds` actually is an assignment. The ternary is `originIds ? [...originIds, originId] : [originId];`

However, I agree that this part of the code was not the easiest to read.

src/ElectronBackend/input/__tests__/parseInputData.test.ts
```diff
@@ -191,6 +191,56 @@ describe('sanitizeRawAttributions', () => {
...
+      },
+    };
+    const expectedParsedRawAttributions: RawAttributions = {
+      uuid: {
+        originIds: ['def', 'ghi', 'abc'],
+      },
+    };
+    const expectedCriticalExternalAttributionsFlag = false;
+    expect(parseRawAttributions(testRawAttributions)).toEqual([
+      expectedParsedRawAttributions,
```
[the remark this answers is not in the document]
lennartclaas: Why? To not having a red test when the implementation with respect to order is changed? I think it would be nice to know if changes like this happen. Maybe it is important in the future.

src/ElectronBackend/input/parseInputData.ts
lennartclaas: I agree that you need to take a glance at the next line to realize it is a ternary.  I changed the implementation according to Markus' suggestion. It's even slightly faster this way. :)

MarkusObendrauf (approved):
Nice

PR #1533, opened 2023-04-03
Deduplicate Signals: Populate externalAttributionsToHashes

lennartclaas:
### Summary of changes

After the parsed input data is sent to the frontend, calculate externalAttributionsToHashes and write it to the redux state. The following fields are ignored when hashing attributions: 'comment', 'attributionConfidence',  'originIds', and 'preSelected'. Furthermore, the following values are considered equal: null, undefined, and ''.

Part of the story 'Deduplicate reused signals that are identical'.

Run new test in `action-and-reducer-helpers.test.ts` or log the data when opening a new file.

### DIFF
--- a/src/Frontend/state/actions/resource-actions/load-actions.ts
+++ b/src/Frontend/state/actions/resource-actions/load-actions.ts
@@ -4,11 +4,13 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { ParsedFileContent } from '../../../../shared/shared-types';
+import { createExternalAttributionsToHashes } from '../../helpers/load-action-helpers';
 import { AppThunkAction, AppThunkDispatch } from '../../types';
 import {
   setAttributionBreakpoints,
   setBaseUrlsForSources,
   setExternalAttributionSources,
+  setExternalAttributionsToHashes,
   setExternalData,
   setFilesWithChildren,
   setFrequentLicenses,
@@ -59,5 +61,11 @@ export function loadFromFile(
     parsedFileContent.resolvedExternalAttributions.forEach((attribution) =>
       dispatch(addResolvedExternalAttribution(attribution))
     );
+
+    const externalAttributionsToHashes = createExternalAttributionsToHashes(
+      parsedFileContent.externalAttributions.attributions
+    );
+
+    dispatch(setExternalAttributionsToHashes(externalAttributionsToHashes));
   };
 }
--- a/src/Frontend/state/helpers/load-action-helpers.ts
+++ b/src/Frontend/state/helpers/load-action-helpers.ts
@@ -1,0 +1,64 @@
+// SPDX-FileCopyrightText: Meta Platforms, Inc. and its affiliates
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+
+import objectHash from 'object-hash';
+import {
+  Attributions,
+  AttributionsToHashes,
+  PackageInfo,
+} from '../../../shared/shared-types';
+
+const FIELDS_TO_IGNORE_WHEN_HASHING_ATTRIBUTIONS = new Set<string>([
+  'comment',
+  'attributionConfidence',
+  'originIds',
+  'preSelected',
+]);
+
+export function createExternalAttributionsToHashes(
+  externalAttributions: Attributions
+): AttributionsToHashes {
+  const hashOptions = {
+    excludeKeys: (key: string): boolean =>
+      FIELDS_TO_IGNORE_WHEN_HASHING_ATTRIBUTIONS.has(key),
+  };
+
+  const externalAttributionsToHashes: AttributionsToHashes = {};
+  const hashesToExternalAttributions: { [hash: string]: Array<string> } = {};
+
+  for (const [attributionId, attribution] of Object.entries(
+    externalAttributions
+  )) {
+    if (attribution.firstParty || attribution.packageName) {
+      const attributionKeys = Object.keys(attribution) as Array<
+        keyof PackageInfo
+      >;
+      attributionKeys.forEach(
+        (key) =>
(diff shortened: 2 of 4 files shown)


src/Frontend/state/helpers/action-and-reducer-helpers.ts
```diff
@@ -114,3 +116,56 @@ export function removeResolvedAttributionsFromResourcesWithAttributedChildren(
     );
   });
 }
+
+export function createExternalAttributionsToHashes(
```
benedikt-richter, line 120: `create` sounds a bit odd. Why not e.g. `get`?

src/Frontend/state/helpers/action-and-reducer-helpers.ts
```diff
@@ -114,3 +116,56 @@ export function removeResolvedAttributionsFromResourcesWithAttributedChildren(
...
+  externalAttributions: Attributions
+): AttributionsToHashes {
+  const excludeKeys = function (key: string): boolean {
+    return [
+      'comment',
+      'attributionConfidence',
+      'originIds',
+      'preSelected',
+    ].includes(key);
```
benedikt-richter, line 129: This config should be extracted into a constant

src/Frontend/state/helpers/action-and-reducer-helpers.ts
```diff
@@ -114,3 +116,56 @@ export function removeResolvedAttributionsFromResourcesWithAttributedChildren(
     );
   });
 }
+
+export function createExternalAttributionsToHashes(
+  externalAttributions: Attributions
+): AttributionsToHashes {
+  const excludeKeys = function (key: string): boolean {
```
benedikt-richter, line 123: We typically write `function excludeKeys(...`

src/Frontend/state/actions/resource-actions/load-actions.ts
```diff
@@ -59,5 +61,11 @@ export function loadFromFile(
     parsedFileContent.resolvedExternalAttributions.forEach((attribution) =>
       dispatch(addResolvedExternalAttribution(attribution))
     );
+
+    const externalAttributionsToHashes = createExternalAttributionsToHashes(
+      parsedFileContent.externalAttributions.attributions
+    );
+
+    dispatch(setExternalAttributionsToHashes(externalAttributionsToHashes));
```
benedikt-richter, line 69: this code should be tested

src/Frontend/state/helpers/action-and-reducer-helpers.ts
```diff
@@ -114,3 +116,56 @@ export function removeResolvedAttributionsFromResourcesWithAttributedChildren(
...
+  for (const [attributionId, attribution] of Object.entries(
+    externalAttributions
+  )) {
+    if (attribution.firstParty || attribution.packageName) {
+      const attributionKeys = Object.keys(attribution) as Array<
+        keyof PackageInfo
+      >;
+      attributionKeys.forEach(
+        (key) =>
+          (attribution[key] == null || attribution[key] === '') &&
```
benedikt-richter, line 147: which attribution[key] can actually be `null`?

benedikt-richter: btw using `==` is unsafe

src/Frontend/state/helpers/action-and-reducer-helpers.ts
benedikt-richter: In this function we loop two times over the in general very huge object `Object.entries(externalAttributions)`. Performance might be a real issue here.

benedikt-richter (changes requested):

src/Frontend/state/helpers/action-and-reducer-helpers.ts
lennartclaas: `getExternalAttributionsToHashes` is already used for the corresponding selector. Here, it's more about calculation. 

2nd comment as discusses.

src/Frontend/state/actions/resource-actions/load-actions.ts
lennartclaas: It is tested in `all-views-simple-actions.test.ts`. I noticed that in this file, actions and corresponding selectors are tested jointly.

src/Frontend/state/helpers/action-and-reducer-helpers.ts
lennartclaas: We used `attribution[key] == null` to consider both `(attribution[key] === null || attribution[key] === undefined)`. The reason is to consider `null`, `undefined`, and `''` the same with respect to hashing.

src/Frontend/state/actions/resource-actions/load-actions.ts
benedikt-richter: You added the present code, but did not modify the test. How is the new functionality tested?

src/Frontend/state/helpers/load-action-helpers.ts
```diff
@@ -0,0 +1,64 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import objectHash from 'object-hash';
+import {
+  Attributions,
+  AttributionsToHashes,
+  PackageInfo,
+} from '../../../shared/shared-types';
+
+const FIELDS_TO_IGNORE_WHEN_HASHING_ATTRIBUTIONS = new Set<string>([
```
benedikt-richter, line 13: `string` could be `KeysOfPackageInfo`.

benedikt-richter (approved):

src/Frontend/state/helpers/load-action-helpers.ts
lennartclaas: This leads to typing problems since the `excludeKeys` function in `hashOptions` requires a string type.

vasily-pozdnyakov (approved):
Please do not forget to squash the commits.

PR #1544, opened 2023-04-05
DRAFT! Deduplicate Signals: save ids

lennartclaas:


src/Frontend/Components/AggregatedAttributionsPanel/accordion-panel-helpers.ts
```diff
@@ -85,6 +89,9 @@ function getMergedAttributionWithCountFromAttributions(
   if (originIds.length > 0) {
     attributionToShow.originIds = originIds;
   }
+  if (originalAttributionIds.length > 1) {
```
MarkusObendrauf, line 92: why do we have this condition?

lennartclaas: If the length is 1, it is not a merged package but an ordinary one. In this case, `originalAttributionIds` should not be set.

vasily-pozdnyakov: Should we also test the case?

MarkusObendrauf: this fucntion returns type `MergedAttributionWithCount` which is defined as
```
export interface MergedAttributionWithCount extends AttributionIdWithCount {
  attribution: MergedPackageInfo;
}
```
therefore `attributionToShow` has type `MergedPackageInfo`. there is no reason to expect a `MergedPackageInfo` to be non-merged. to me it makes more sense to use `originalAttributionIds` with a list of length 1

lennartclaas: Yeah. I also think this design is a bit awkward. Having `originalAttributionIds` set or not is currently the only way to distinguish 'real' merged signals form not merged signals. We should talk to Leslie about that.

src/shared/shared-types.ts
```diff
@@ -54,6 +54,7 @@ export interface PackageInfo extends PackageInfoCore {
 export interface MergedPackageInfo extends PackageInfo {
   type: 'MergedPackageInfo';
   comments?: Array<string>;
+  originalAttributionIds?: Array<string>;
```
benedikt-richter, line 57: this one should be non-optional

src/Frontend/Components/AggregatedAttributionsPanel/accordion-panel-helpers.ts
lennartclaas: Changed it according to our discussion.

src/shared/shared-types.ts
lennartclaas: Changed it according to our discussion.

vasily-pozdnyakov (approved):

benedikt-richter (approved):

PR #1550, opened 2023-04-06
Rename MergedPackageInfo to DisplayPackageInfo

leslielazzarino:

The name was confusing.

### DIFF
--- a/src/Frontend/Components/AggregatedAttributionsPanel/AccordionPanel.tsx
+++ b/src/Frontend/Components/AggregatedAttributionsPanel/AccordionPanel.tsx
@@ -11,7 +11,7 @@ import React, { ReactElement, useMemo, useState } from 'react';
 import { PackagePanel } from '../PackagePanel/PackagePanel';
 import {
   AttributionIdWithCount,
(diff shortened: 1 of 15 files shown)


src/Frontend/Components/AggregatedAttributionsPanel/AccordionPanel.tsx
```diff
@@ -11,7 +11,7 @@ import React, { ReactElement, useMemo, useState } from 'react';
 import { PackagePanel } from '../PackagePanel/PackagePanel';
 import {
   AttributionIdWithCount,
-  MergedAttributionWithCount,
+  DisplayAttributionWithCount,
```
lennartclaas, line 14: Why `DisplayAttribution...` and `DisplayPackage...` and not `Displayed...`? Is it meant like a noun 'display-package'?

leslielazzarino: We mostly use display instead of displayed in the code, and in the end it means the same thing and is shorter.

benedikt-richter (approved):

PR #1551, opened 2023-04-06
Show resources for merged signals

MarkusObendrauf:
### Summary of changes

The resource path popup shows resources for all attributions of a merged package.

Currently, it is possible that there are many identical signals that only differ by the comment. This should be fixed. In this PR, the resource path popup is updated to display resource paths for all attributions of a merged package, instead of just one.

### DIFF
--- a/src/Frontend/Components/PackageCard/PackageCard.tsx
+++ b/src/Frontend/Components/PackageCard/PackageCard.tsx
@@ -56,7 +56,11 @@ import {
   getRightIcons,
 } from './package-card-helpers';
 import OpenInBrowserIcon from '@mui/icons-material/OpenInBrowser';
-import { PackageInfo } from '../../../shared/shared-types';
+import {
+  isDisplayPackageInfo,
+  DisplayPackageInfo,
+  PackageInfo,
+} from '../../../shared/shared-types';
 import MuiBox from '@mui/material/Box';
 import { openAttributionWizardPopup } from '../../state/actions/popup-actions/popup-actions';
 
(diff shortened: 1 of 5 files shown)


src/Frontend/Components/ResourcePathPopup/ResourcePathPopup.tsx
```diff
@@ -33,10 +34,12 @@ export function ResourcePathPopup(props: ResourcePathPopupProps): ReactElement {
   const manualAttributionsToResources = useAppSelector(
     getManualAttributionsToResources
   );
-  const allResourceIds = props.isExternalAttribution
-    ? externalAttributionsToResources[props.attributionId]
-    : manualAttributionsToResources[props.attributionId];
-
+  const allResourcePaths = getAllResourcePathsForAttributions(
+    props.attributionIds,
+    props.isExternalAttribution
```
lennartclaas, line 39: 👍

src/Frontend/Components/PackageCard/PackageCard.tsx
```diff
@@ -396,13 +396,19 @@ export function PackageCard(props: PackageCardProps): ReactElement | null {
     />
   ) : undefined;
 
+  const attributionIdsForResourcePathPopup = isMergedPackageInfo(
```
lennartclaas, line 399: In `PackageCardProps`, only `PackageInfo` is allowed as a type for `props.packageInfo`. `isMergedPackageInfo` should therefore only yield `false`.

lennartclaas (changes requested):

src/Frontend/Components/PackageCard/PackageCard.tsx
lennartclaas: But is works anyway. I'm confused.

MarkusObendrauf: `MergedPackageInfo` is a subclass of `PackageInfo`, so the props allowed it. changed the type signature of `props.packageInfo` to `PackageInfo | MergedPackageInfo` to make it explicit.

PR #1558, opened 2023-04-11
Deduplicate Signals: Provide Hide Functionality.

lennartclaas:
### Summary of changes

Enabled hide feature for merged signals in PackageCard context menu as well as in AttributionColumn.

Is part of the story 'Deduplicate reused signals that are identical'.

With new tests in `audit-view.test.tsx` as well as`context-menu-audit-view.test.tsx` and with the test file `opossum-input.json` (merged signal attached to `menu.ts`).

### DIFF
--- a/src/Frontend/Components/AttributionColumn/AttributionColumn.tsx
+++ b/src/Frontend/Components/AttributionColumn/AttributionColumn.tsx
@@ -6,7 +6,10 @@
 
 import React, { ChangeEvent, ReactElement } from 'react';
 
-import { PackageInfo } from '../../../shared/shared-types';
+import {
+  PackageInfo,
+  isDisplayPackageInfo,
+} from '../../../shared/shared-types';
 import { setTemporaryPackageInfo } from '../../state/actions/resource-actions/all-views-simple-actions';
 import {
   getAttributionIdMarkedForReplacement,
@@ -31,7 +34,7 @@ import {
   getFollowUpChangeHandler,
   getMergeButtonsDisplayState,
   getResolvedToggleHandler,
-  selectedPackageIsResolved,
+  selectedPackagesAreResolved,
   usePurl,
   useRows,
 } from './attribution-column-helpers';
@@ -85,9 +88,7 @@ export function AttributionColumn(props: AttributionColumnProps): ReactElement {
   const resolvedExternalAttributions = useAppSelector(
     getResolvedExternalAttributions
   );
-  const temporaryPackageInfo: PackageInfo = useAppSelector(
-    getTemporaryPackageInfo
-  );
+  const temporaryPackageInfo = useAppSelector(getTemporaryPackageInfo);
   const packageInfoWereModified = useAppSelector(
     wereTemporaryPackageInfoModified
   );
@@ -247,6 +248,12 @@ export function AttributionColumn(props: AttributionColumnProps): ReactElement {
     !props.displayPackageInfo.firstParty &&
     !props.displayPackageInfo.excludeFromNotice;
 
+  const attributionIdsToResolveOrUnresolve = isDisplayPackageInfo(
+    temporaryPackageInfo
+  )
+    ? temporaryPackageInfo.attributionIds
+    : []; // TODO: remove when refactoring
+
   return (
     <MuiBox sx={classes.root}>
       <PackageSubPanel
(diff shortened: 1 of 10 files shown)


src/Frontend/Components/AttributionColumn/AttributionColumn.tsx
```diff
@@ -247,6 +250,12 @@ export function AttributionColumn(props: AttributionColumnProps): ReactElement {
     !props.displayPackageInfo.firstParty &&
     !props.displayPackageInfo.excludeFromNotice;
 
+  const attributionIdsToResolveOrUnresolve = isDisplayPackageInfo(
+    temporaryPackageInfo
+  )
+    ? temporaryPackageInfo.attributionIds
+    : [selectedPackageId];
```
benedikt-richter, line 257: Why do we need the second case (`[selectedPackageId]`)? Attributions cannot be hidden.

lennartclaas: Changed it to an empty list with the comment to consider this when refactoring. Using `const attributionIdsToResolveOrUnresolve = temporaryPackageInfo?.attributionIds ?? []` was not possible because the type checker is complaining in this case.

MarkusObendrauf (changes requested):

src/Frontend/Components/AttributionColumn/attribution-column-helpers.ts
```diff
@@ -119,29 +119,35 @@ export function getFirstPartyChangeHandler(
...
+        for (const attributionId of attributionIds) {
+          dispatch(addResolvedExternalAttribution(attributionId));
+        }
       }
       dispatch(saveManualAndResolvedAttributionsToFile());
     }
   };
 }
 
 export function selectedPackageIsResolved(
```
benedikt-richter, line 144: The name seems outdated

src/Frontend/Components/PackageCard/PackageCard.tsx
```diff
@@ -152,6 +152,10 @@ export function PackageCard(props: PackageCardProps): ReactElement | null {
       ? getPackageCardHighlighting(manualAttributions[attributionId])
       : undefined;
 
+  const displayAttributionIds = isDisplayPackageInfo(props.packageInfo)
+    ? props.packageInfo.attributionIds
+    : [attributionId];
```
benedikt-richter, line 157: Same issue as above

benedikt-richter (changes requested):

src/Frontend/Components/AttributionColumn/attribution-column-helpers.ts
```diff
@@ -119,29 +119,35 @@ export function getFirstPartyChangeHandler(
...
-  attributionId: string,
+  attributionIds: Array<string>,
   resolvedExternalAttributions: Set<string>,
   dispatch: AppThunkDispatch
 ): () => void {
   return (): void => {
-    if (attributionId) {
-      if (resolvedExternalAttributions.has(attributionId)) {
-        dispatch(removeResolvedExternalAttribution(attributionId));
+    if (attributionIds) {
```
benedikt-richter, line 127: This check should be removed, as it cannot occur.

src/Frontend/Components/AttributionColumn/__tests__/attribution-column-helpers.test.ts
```diff
@@ -29,19 +29,31 @@ describe('The AttributionColumn helpers', () => {
...
   });
 
   it('selectedPackageIsResolved returns false if empty attributionId', () => {
-    expect(selectedPackageIsResolved('', new Set<string>().add('123'))).toEqual(
-      false
-    );
+    expect(
+      selectedPackagesAreResolved([''], new Set<string>().add('123'))
+    ).toEqual(false);
+  });
```
MarkusObendrauf, line 40: do we need this test? is it expected that `attributionIds` can contain empty strings? i would expect the array to be empty instead

MarkusObendrauf (approved):

src/Frontend/Components/AttributionColumn/__tests__/attribution-column-helpers.test.ts
lennartclaas: Only if the input file is flawed or due to some other bug. I think it should be kept. Runtime is negligible.

PR #1579, opened 2023-04-14
Deprecate originId field in input files

MarkusObendrauf:
### Summary of changes

Deprecate the originId filed in input files

In #1561, identical signals are merged in OpossumUI. This means that one signal may have multiple origins. In order to support this, the field `originIds` was introduced, and now `originId` is deprecated.

### DIFF
--- a/src/ElectronBackend/input/OpossumInputFileSchema.json
+++ b/src/ElectronBackend/input/OpossumInputFileSchema.json
@@ -140,7 +140,8 @@
           },
           "originId": {
             "type": "string",
-            "description": "Can be set to track a signal from the tooling that generated the input file. Will be copied into the output when the signal is converted into an attribution."
+            "description": "Deprecated - originIds should be used instead. Can be set to track a signal from the tooling that generated the input file. Will be copied into the output when the signal is converted into an attribution.",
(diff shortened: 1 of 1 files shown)


src/ElectronBackend/input/OpossumInputFileSchema.json
```diff
@@ -140,7 +140,8 @@
           },
           "originId": {
             "type": "string",
-            "description": "Can be set to track a signal from the tooling that generated the input file. Will be copied into the output when the signal is converted into an attribution."
+            "description": "Can be set to track a signal from the tooling that generated the input file. Will be copied into the output when the signal is converted into an attribution.",
```
benedikt-richter, line 143: We could add the information that the new field `originIds` should be used instead of the deprecated one.

benedikt-richter (changes requested):

src/ElectronBackend/input/OpossumInputFileSchema.json
MarkusObendrauf: Added `Deprecated - originIds should be used instead. ` to the front of the description

benedikt-richter (approved):

PR #1583, opened 2023-04-17
Use index for withAttributedChildren to save memory.

leslielazzarino:
### Summary of changes

ResourcesWithAttributedChildren type has been changed from: 
````
export interface ResourcesWithAttributedChildren {
  [path: string]: Set<string>;
}
````
to:
````
export interface ResourcesWithAttributedChildren {
  paths: Array<string>;
  pathToIndex: { [path: string]: number };
  attributedChildren: { [index: number]: Set<number> };
}
````

This change reduces a lot the memory use, increasing the size of input files that can be handled without crashes.

### DIFF
--- a/src/Frontend/Components/ResourceBrowser/get-resource-browser-tree-item-label.tsx
+++ b/src/Frontend/Components/ResourceBrowser/get-resource-browser-tree-item-label.tsx
@@ -149,9 +149,13 @@ function containsExternalAttribution(
   nodeId: string,
   resourcesWithExternalAttributedChildren: ResourcesWithAttributedChildren
 ): boolean {
+  const nodeIndex =
+    resourcesWithExternalAttributedChildren.pathsToIndices[nodeId];
+
   return (
-    resourcesWithExternalAttributedChildren &&
-    nodeId in resourcesWithExternalAttributedChildren
+    nodeIndex !== undefined &&
+    resourcesWithExternalAttributedChildren.attributedChildren[nodeIndex] !==
+      undefined
   );
 }
 
@@ -161,7 +165,7 @@ function containsManualAttribution(
 ): boolean {
   return (
     resourcesWithManualAttributedChildren &&
-    nodeId in resourcesWithManualAttributedChildren
+    nodeId in resourcesWithManualAttributedChildren.pathsToIndices
   );
 }
 
--- a/src/Frontend/shared-constants.ts
+++ b/src/Frontend/shared-constants.ts
@@ -13,7 +13,11 @@ export const EMPTY_ATTRIBUTION_DATA: AttributionData = {
   attributions: {},
   resourcesToAttributions: {},
   attributionsToResources: {},
-  resourcesWithAttributedChildren: {},
+  resourcesWithAttributedChildren: {
+    paths: [],
+    pathsToIndices: {},
+    attributedChildren: {},
+  },
 };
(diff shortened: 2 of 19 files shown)

### CODE REVIEW

File: src/Frontend/shared-constants.ts [hunk 1]
```diff
@@ -13,7 +13,11 @@ export const EMPTY_ATTRIBUTION_DATA: AttributionData = {
   attributions: {},
   resourcesToAttributions: {},
   attributionsToResources: {},
-  resourcesWithAttributedChildren: {},
+  resourcesWithAttributedChildren: {
+    paths: [],
+    pathToIndex: {},
```
benedikt-richter, line 18: suggestion: `pathsToIndex`

src/Frontend/state/helpers/action-and-reducer-helpers.ts
```diff
@@ -49,15 +53,43 @@ export function _addParentsToResourcesWithAttributedChildren(
...
     }
 
-    childrenWithAttributions[parent].add(attributedPath);
+    childrenWithAttributions.attributedChildren[parentIndex].add(
+      attributedPathIndex
+    );
   });
 }
 
+export function addPathToIndexesIfMissingInResourcesWithAttributedChildren(
```
benedikt-richter, line 81: seems to be untested

benedikt-richter (changes requested):
Nice refactoring that really reduces memory usage 👍

src/Frontend/Components/ResourceBrowser/get-resource-browser-tree-item-label.tsx
```diff
@@ -149,9 +149,15 @@ function containsExternalAttribution(
   nodeId: string,
   resourcesWithExternalAttributedChildren: ResourcesWithAttributedChildren
 ): boolean {
+  if (!resourcesWithExternalAttributedChildren) {
+    return false;
+  }
```
antonbauhofer, line 154: Doesn't seem to be necessary, since the type is not optional.

src/Frontend/shared-constants.ts
antonbauhofer: actually, I disagree. I think `pathsToIndex` would imply that multiple paths are mapped to one index.

src/Frontend/state/helpers/action-and-reducer-helpers.ts
leslielazzarino: It is just a small, trivial element that has been extracted. I would not write an unit test for it.

src/Frontend/state/helpers/action-and-reducer-helpers.ts
```diff
@@ -49,15 +53,43 @@ export function _addParentsToResourcesWithAttributedChildren(
   attributedPath: string,
```
antonbauhofer, line 53: I would rename the function to
_addAttributedPathAndParentsToResourcesWithAttributedChildren

antonbauhofer (approved):
Looks good to me.
Just some minor comments.

src/Frontend/shared-constants.ts
antonbauhofer: oh, I didn't see that the comment was updated. I agree with `pathsToIndices`.

antonbauhofer (approved):

PR #1598, opened 2023-04-20
Do not zip mac release

benedikt-richter:
### Summary of changes

The zipping operation is removed. 

The release for mac a has size of approximately 293MB and 285 when zipped. The advantage for the user is that no unzipping is necessary.

Run `yarn ship-mac` and test release. The Github action is hard to test.


maxhbr (approved):
LGTM

lennartclaas (approved):
👍

Issue #1604, opened 2023-04-24
Add missing package counts.


lennartclaas:
During the deduplicate signals story a bug slipped in: The counts on package cards disappeared.

The goal of this ticket is to fix this bug in order to display the signal counts again.

Current (buggy):
![Screenshot 2023-04-24 111637](https://user-images.githubusercontent.com/83081698/233959453-db81d62b-4319-4a51-8c11-379d6c3b421d.png)

Goal:
![Screenshot 2023-04-24 111432](https://user-images.githubusercontent.com/83081698/233959497-7b3b346e-9867-41fc-9e4a-a0c00dd12ce8.png)

PR #1605, opened 2023-04-24
Bugfix: Add missing signal counts in package panels.

lennartclaas:
### Summary of changes

The deuplicate signals story led to a bug that removed the counts on package cards in package panels. Enable the counts again.

Counts on package cards are valuable information for the user.

Open test files and check package panels on a folder near root.

### Further comments

Integration test is added in an upcoming pull request to facilitate fast release of a version without this bug.

See #1604 for screenshots.

Fix: #1604


src/Frontend/Components/AggregatedAttributionsPanel/accordion-panel-helpers.ts
```diff
@@ -75,18 +77,33 @@ export function getDisplayAttributionsWithCount(
...
+  const counts: Array<number> = attributionsWithIdsAndCounts.reduce(
+    (filteredCounts, attributionWithIdAndCount) => {
+      const count = attributionWithIdAndCount[2];
+      if (count !== undefined) {
+        filteredCounts.push(count);
+      }
+      return filteredCounts;
+    },
+    Array<number>()
+  );
```
benedikt-richter, line 98: This can simply be:

```
const counts = attributionsWithIdsAndCounts.filter((element) =>  count !== undefined)
```

benedikt-richter (changes requested):

src/Frontend/Components/AggregatedAttributionsPanel/accordion-panel-helpers.ts
lennartclaas: I don't think so. `element` is an array [attributionId, attribution, count]. Filtering would lead to an array of arrays with type `DisplayAttributionWithCount` and not `number`.

benedikt-richter: True. Seems I've been confused.

vasily-pozdnyakov (approved):

PR #1629, opened 2023-04-28
Refactor DisplayPackageInfo: AttributionView

lennartclaas:
### Summary of changes

Apply the refactoring regarding DisplayPackageInfo to AttributionView as well as ReplaceAttributionPopup.

Separate business and display logic by using DisplayPackageInfo in displaying components.


src/Frontend/Components/AttributionList/__tests__/AttributionList.test.tsx
```diff
@@ -218,4 +232,82 @@ describe('The AttributionList', () => {
       true
     );
   });
+
+  // it('shows correct replace attribution buttons in the context menu', () => {
```
benedikt-richter, line 236: why is this one commented out?

src/Frontend/Components/ReplaceAttributionPopup/ReplaceAttributionPopup.tsx
```diff
@@ -59,11 +62,22 @@ export function ReplaceAttributionPopup(): ReactElement {
     dispatch(closePopup());
   }
 
+  const displayAttributions: Array<DisplayAttributionWithCount> =
+    Object.entries(attributions).map(([attributionId, packageInfo]) => {
```
benedikt-richter, line 66: Why was this popup modified?
Before, we just converted the attributions that we actually displayed, while now all attributions are converted and then we still only show two of these. Seems to be unnecessary compute.

src/Frontend/util/get-attribution-from-display-attribution-with-count.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  if (
+    attributionIdOrDisplayAttributionWithCount &&
+    isDisplayAttributionWithCount(attributionIdOrDisplayAttributionWithCount)
+  ) {
+    return attributionIdOrDisplayAttributionWithCount.attribution;
+  } else {
+    return {
+      attributionIds: [],
+    } as DisplayPackageInfo;
```
benedikt-richter, line 31: We should avoid type casts.

src/Frontend/util/get-attribution-from-display-attribution-with-count.ts
```diff
@@ -0,0 +1,33 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+
+import { DisplayPackageInfo } from '../../shared/shared-types';
+import {
+  DisplayAttributionWithCount,
+  isDisplayAttributionWithCount,
+} from '../types/types';
+
+export function getAttributionFromDisplayAttributionWithCount(
```
benedikt-richter, line 12: This is a duplicate of `getAttributionFromDisplayAttributionsWithCount`. Probably a relict of rebasing.

src/Frontend/Components/AttributionList/__tests__/AttributionList.test.tsx
lennartclaas: I just commented it out for reference and forgot to delete it. The apropriate test is above.

src/Frontend/Components/ReplaceAttributionPopup/ReplaceAttributionPopup.tsx
lennartclaas: Good point, I reverted the changes.

src/Frontend/util/get-attribution-from-display-attribution-with-count.ts
lennartclaas: Removed it as it does not seem to be needed anymore.

src/Frontend/util/get-attribution-from-display-attribution-with-count.ts
lennartclaas: That's correct. I removed it.

benedikt-richter (approved):

PR #1638, opened 2023-05-02
Add checkbox for needsReview

antonbauhofer:
### Summary of changes

This adds a checkbox next to the buttons on the attribution view to signal that an attribution needs further review.

- Run the tests in `AttributionColumn.test.tsx`.
- Open a file in OpossumUI and verify that the new checkbox shows up and works.

### DIFF
--- a/src/Frontend/Components/AttributionColumn/ButtonRow.tsx
+++ b/src/Frontend/Components/AttributionColumn/ButtonRow.tsx
@@ -7,16 +7,22 @@ import React, { ReactElement } from 'react';
 import { ToggleButton } from '../ToggleButton/ToggleButton';
 import { ButtonGroup, MainButtonConfig } from '../ButtonGroup/ButtonGroup';
 import MuiTypography from '@mui/material/Typography';
-import { ButtonText } from '../../enums/enums';
+import { ButtonText, CheckboxLabel } from '../../enums/enums';
 import { ContextMenuItem } from '../ContextMenu/ContextMenu';
 import MuiBox from '@mui/material/Box';
+import { Checkbox } from '../Checkbox/Checkbox';
+import { checkboxClass } from '../../shared-styles';
+import { DisplayPackageInfo, PackageInfo } from '../../../shared/shared-types';
 
 const classes = {
+  ...checkboxClass,
   root: {
     marginLeft: '10px',
     marginTop: '5px',
   },
   buttonRow: {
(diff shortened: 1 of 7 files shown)


lennartclaas (changes requested):
The implementations look nice. However, the styling in the `EditAttributionPopup` should be considered as well:

Screenshot 2023-05-02 161859

antonbauhofer:
> The implementations look nice. However, the styling in the `EditAttributionPopup` should be considered as well:
> 

Nice catch!
I fixed it in `ButtonRow.tsx`. 
Screenshot 2023-05-02 at 18 47 01

src/Frontend/Components/AttributionColumn/ButtonRow.tsx
```diff
@@ -28,6 +34,10 @@ const classes = {
     marginTop: '0px',
     marginRight: '0px',
   },
+  extraPadding: {
```
lennartclaas, line 37: I would choose a more meaningful name. First, only margins are set and not padding. Second, one could name it according to for which component and why it is used, e.g., 'checkboxWithHiddenButtons'. But that's only a suggestion.

src/Frontend/Components/AttributionColumn/ButtonRow.tsx
```diff
@@ -38,9 +48,16 @@ interface ButtonRowProps {
   displayTexts: Array<string>;
   mainButtonConfigs: Array<MainButtonConfig>;
   hamburgerMenuButtonConfigs?: Array<ContextMenuItem>;
+  isEditable: boolean;
+  displayPackageInfo: PackageInfo | DisplayPackageInfo;
+  needsReviewChangeHandler(event: React.ChangeEvent<HTMLInputElement>): void;
 }
 
 export function ButtonRow(props: ButtonRowProps): ReactElement {
+  const largerPaddingForNeedsReviewCheckbox = props.areButtonsHidden
```
lennartclaas, line 57: Naming as above.

lennartclaas (changes requested):
Only some nits I have overlooked in the first review. Sorry for that. We should discuss whether the position of the checkbox is suitable in the case of attributions in the `Global` tab or the `ContainedManualPackages` panel. I looks a bit lost as no other buttons are displayed.

Screenshot 2023-05-03 073919

lennartclaas (approved):
lgtm

PR #1644, opened 2023-05-03
Needs review icon

antonbauhofer:
### Summary of changes

Adds a new icon to signal that an attribution needs review.
Note that in the cards in attribution view, there should be a maximum of 4 icons shown at the same time, since pre-selected attributions cannot have the needs-preview attribute.
Also, the color of the follow-up icon was adjusted in the attribution view to make it consistent with the report view.

- Run the tests in `Icons.test.tsx` and `ReportTableItem.test.tsx`
- Open a file in OpossumUI and verify that the new icon shows up in the report- and attribution view (twice).

### DIFF
--- a/src/Frontend/Components/ProjectStatisticsPopup/__tests__/project-statistics-popup-helpers.test.tsx
+++ b/src/Frontend/Components/ProjectStatisticsPopup/__tests__/project-statistics-popup-helpers.test.tsx
@@ -34,6 +34,7 @@ const testAttributions_1: Attributions = {
     criticality: Criticality.Medium,
     licenseName: 'Apache License Version 2.0',
     firstParty: true,
+    needsReview: true,
   },
   uuid2: {
     source: {
@@ -50,6 +51,7 @@ const testAttributions_1: Attributions = {
     },
     licenseName: ' Apache license version-2.0 ',
     followUp: FollowUp,
+    needsReview: true,
   },
   uuid4: {
     source: {
@@ -202,6 +204,7 @@ describe('aggregateAttributionPropertiesFromAttributions', () => {
     const expectedAttributionPropertyCounts: {
       [attributionPropertyOrTotal: string]: number;
     } = {
+      needsReview: 2,
       followUp: 1,
       firstParty: 2,
       incomplete: 4,
@@ -227,6 +230,7 @@ describe('aggregateAttributionPropertiesFromAttributions', () => {
     const expectedAttributionPropertyCounts: {
       [attributionPropertyOrTotal: string]: number;
     } = {
+      needsReview: 0,
       followUp: 0,
       firstParty: 0,
       incomplete: 5,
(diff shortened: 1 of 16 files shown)


src/e2e-tests/__tests__/e2e-deprecated-file-format.test.ts
```diff
@@ -36,18 +36,6 @@ test.describe('Open outdated .json file via command line', () => {
     }
   });
 
-  test('should open FileSupportPopup when .json file is provided as command line arg', async () => {
```
vasily-pozdnyakov, line 39: why?

src/e2e-tests/test-helpers/test-helpers.ts
```diff
@@ -19,7 +19,7 @@ import {
 
 const ELECTRON_LAUNCH_TEST_TIMEOUT = 75000;
 export const E2E_TEST_TIMEOUT = 120000;
-export const EXPECT_TIMEOUT = 30000;
```
vasily-pozdnyakov, line 22: why?

src/Frontend/Components/ProjectStatisticsPopup/__tests__/project-statistics-popup-helpers.test.tsx
```diff
@@ -202,6 +203,7 @@ describe('aggregateAttributionPropertiesFromAttributions', () => {
     const expectedAttributionPropertyCounts: {
       [attributionPropertyOrTotal: string]: number;
     } = {
+      needsReview: 1,
```
vasily-pozdnyakov, line 206: is it the only test where needsReview>0? Would be nice to have more, e.g. with more items which needReview

vasily-pozdnyakov (changes requested):

src/e2e-tests/__tests__/e2e-deprecated-file-format.test.ts
antonbauhofer: This test was redundant with the test below which implicitly also tests that the popup is shown (with `await getButtonWithName(window, 'Keep');`)

src/e2e-tests/test-helpers/test-helpers.ts
antonbauhofer: The CI repeatedly ran into this timeout. So, we decided to increase it. The value of `30000` was kind of arbitrarily chosen anyway.

src/Frontend/Components/ProjectStatisticsPopup/__tests__/project-statistics-popup-helpers.test.tsx
antonbauhofer: I added the property to another one of the test attributions. I would not add a separate test for that, since the logic is trivial and the same as for follow-ups.
Is that ok for you?

vasily-pozdnyakov: fine, thx

antonbauhofer:
> well done. The failing CI - not good.

We ran the CI on a separate branch and locally. In both cases it was green. So, we decided to ignore the failing tests and still land this diff (hoping that the CI will be green when we make a PR into the main branch).

PR #1645, opened 2023-05-03
Added NeedsReview filter

vasily-pozdnyakov:
### Summary of changes

Added filter item "Needs Review" to Attribution and Report Views.

use-filters.test.tsx

Mark some attributions "needs review" and see filters in Attribution and Report Views.

image


lennartclaas (approved):
👍

PR #1650, opened 2023-05-04
Update user guide

antonbauhofer:
### Summary of changes

Add description of the needs review checkbox to the OpossumUI user guide

Check the formatting of the user guide in an IDE.

### DIFF
--- a/USER_GUIDE.md
+++ b/USER_GUIDE.md
@@ -266,18 +266,19 @@ attribution information and signals cannot be edited.
 
 IMPORTANT: Some fields in the column have special meanings/behaviors:
 
-- _PURL_: if provided, package name and version are extracted from it, and the corresponding fields are not editable. A
+- _PURL_: If provided, package name and version are extracted from it, and the corresponding fields are not editable. A
   basic validity check is done on the purl: if the purl text is red it means it is invalid and saving is prevented.
-- _License Text_: it will appear in the attributions document. It will be automatically filled in for licenses suggested
+- _License Text_: It will appear in the attributions document. It will be automatically filled in for licenses suggested
   in the license name dropdown.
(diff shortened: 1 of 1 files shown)


USER_GUIDE.md
```diff
@@ -266,18 +266,19 @@ attribution information and signals cannot be edited.
...
-  In the case of first party code the respective flag should be preferred.
+- _Exclude From Notice Checkbox_: If checked, the relative attribution will not be shown in the notice document.
+  In the case of first party code, the respective flag should be preferred.
   _Exclude From Notice Checkbox_ should be used only if:
   - the content of the attribution does not need attribution or
   - the attribution isn't an actual attribution or
   - it was globally decided that this attribution does not need attribution (e.g. it is proprietary but bought for the
     whole company).
 - _Comment / Comments_: In the case of an ordinary signal or an attribution, the comment textbox is displaying a single comment. In the case of a merged signal, the comment textbox is displaying multiple comments according to the comments of the individual merged signals.
+- _Needs Review Checkbox_: Checking this checkbox does not have any effect on the overall workflow. The state of the checkbox is however persisted when saving the attribution.
```
lennartclaas, line 281: I would be more specific and explain how this checkbox can be used in a workflow with multiple users. In this sense, it, in deed, has an effect on the overall workflow.

lennartclaas (approved):
I do not have a strong opinion in this case, feel free to just merge if you like.

USER_GUIDE.md
antonbauhofer: I had the same thought and in fact, I had more text in the beginning. However, I noticed that for the other points we don't describe the actual usage, but the function / effect of the features, so I wanted to keep it consistent and not tell the user how to use the app, but just describe it's functions.

PR #1651, opened 2023-05-04
Needs review checkbox

antonbauhofer:
### Summary of changes

In this branch we are adding a feature to signal that an attribution needs review.

In order to improve the workflow of a typical QA process, we add this feature to provide a way for an OpossumUI user to signal to QA that an attribution needs special attention.

Open a file in OpossumUI and use the checkbox for a manual attribution. Check also that the statistics popup is updated accordingly.

### DIFF
--- a/USER_GUIDE.md
+++ b/USER_GUIDE.md
@@ -278,7 +278,8 @@ IMPORTANT: Some fields in the column have special meanings/behaviors:
   - it was globally decided that this attribution does not need attribution (e.g. it is proprietary but bought for the
     whole company).
 - _Comment / Comments_: In the case of an ordinary signal or an attribution, the comment textbox is displaying a single comment. In the case of a merged signal, the comment textbox is displaying multiple comments according to the comments of the individual merged signals.
-- _Needs Review Checkbox_: Checking this checkbox does not have any effect on the overall workflow. The state of the checkbox is however persisted when saving the attribution.
+- _Needs Review Checkbox_: This checkbox can be used to signal to another OpossumUI user that an attribution needs further review.
+  The state of the checkbox is persisted when saving the attribution, so it can e.g. be used for a typical QA workflow.
 
 The `Attribution Details Column`, if editable, shows the following buttons:
 


USER_GUIDE.md
```diff
@@ -266,18 +266,19 @@ attribution information and signals cannot be edited.
...
-  In the case of first party code the respective flag should be preferred.
+- _Exclude From Notice Checkbox_: If checked, the relative attribution will not be shown in the notice document.
+  In the case of first party code, the respective flag should be preferred.
   _Exclude From Notice Checkbox_ should be used only if:
   - the content of the attribution does not need attribution or
   - the attribution isn't an actual attribution or
   - it was globally decided that this attribution does not need attribution (e.g. it is proprietary but bought for the
     whole company).
 - _Comment / Comments_: In the case of an ordinary signal or an attribution, the comment textbox is displaying a single comment. In the case of a merged signal, the comment textbox is displaying multiple comments according to the comments of the individual merged signals.
+- _Needs Review Checkbox_: Checking this checkbox does not have any effect on the overall workflow. The state of the checkbox is however persisted when saving the attribution.
```
benedikt-richter, line 281: I do disagree here. It should have an effect on the overall workflow (otherwise why do we introduce it?). Please, be more specific here.

src/Frontend/Components/PackageCard/package-card-helpers.tsx
```diff
@@ -21,9 +22,12 @@ export function getKey(prefix: string, cardId: string): string {
 }
 
 const classes = {
-  followUpIcon: {
+  needsReviewIcon: {
     color: OpossumColors.orange,
   },
+  followUpIcon: {
```
benedikt-richter, line 28: Not sure that we should change this color. The user might be used to it. Therefore, just give a different meaning to the color might be confusing.

benedikt-richter (changes requested):
Looks good and works.
Just two minor points.

src/Frontend/Components/PackageCard/package-card-helpers.tsx
antonbauhofer: The reason why we changed it here was that this icon was actually used with two different colors in the attribution- and the report view.
I would strongly suggest that we change it in one of the two locations to have one color assigned per icon. I have no strong opinion, though, on which color to use. So, I would be ok with leaving it here as is and changing it in report view. What do you think?

USER_GUIDE.md
antonbauhofer: Ok, I'll give it a try and re-write it. As discussed in the original PR for this, my intention was not to use the explanation to suggest to the user how to use the feature, but rather to describe its function. But I see your point and I'll try to improve it.

src/Frontend/Components/PackageCard/package-card-helpers.tsx
benedikt-richter: That's a good point! I was not aware of that. I'm definitely in favor of making it consistent. No strong opinion which one to choose.

benedikt-richter: We can leave it as it is.

benedikt-richter (approved):

PR #1659, opened 2023-05-08
Refactoring DisplayPackageInfo

lennartclaas:
### Summary of changes

Replace `PackageInfo` with `DisplayPackageInfo` in all components that display attributions and adapt their parents accordingly. Furthermore, make `temporaryPackageInfo` a `DisplayPackageInfo`.

Improve separation between business and display logic. The 'deduplicate resuser signals' story introduced merged signals which made these changes necessary in order to keep the code clean.


benedikt-richter (approved):
Tested this collection of refactorings and it works. Also performance is good. Reviewed individual commits before.

PR #1769, opened 2023-06-29
Fix vulnerability

benedikt-richter:
### Summary of changes

This commit fixes vulnerabilities pointed out by "yarn audit".

[](https://github.com/opossum-tool/OpossumUI/security/dependabot/37)

Run `yarn audit` and observe that semver vuln is gone.

### DIFF
--- a/package.json
+++ b/package.json
@@ -175,7 +175,7 @@
     ]
   },
   "resolutions": {
-    "minimatch": "^3.0.5"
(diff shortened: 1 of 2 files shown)


lennartclaas (approved):
I still see 2 vulnerabilities concerning word-wrap but these do not have patches. So, I assume this is expected. It seems the fix removes almost 50 vulnerabilities. Nice!

Issue #1711, opened 2023-05-31
Open project statistics popup on opening new file


benedikt-richter:
As an OpossumUI user, I want to get a quick overview of the project when opening a new file.
In order to achieve this the project statistics popup should automatically open when a new file is opened.

PR #1774, opened 2023-06-30
Show project statistics popup after new file was loaded. Fix: 1711

FlorianSchepers:
### Summary of changes

In `src\Frontend\Components\BackendCommunication\BackendCommunication.tsx` in `fileLoadedListener()`  the `ProjectStatisticsPopup` is now dispatched right after the `parsedFileContent`. Thus, the `ProjectStatisticsPopup` will open once a new file is loaded.
Some tests had to be addapted.

As an OpossumUI user, I want to get a quick overview of the project when opening a new file.
In order to achieve this the project statistics popup should automatically open when a new file is opened.

Fix: #1711 

Start OpossumUI and open a non empty opossum file. 

### DIFF
--- a/setupTests.ts
+++ b/setupTests.ts
@@ -7,12 +7,18 @@
 import '@testing-library/jest-dom';
 
 const TEST_TIMEOUT = 15000;
+//We suppress webworker errors that are due to jest not supporting these.
 const SUBSTRINGS_TO_SUPPRESS_IN_CONSOLE_INFO = [
   'Web worker error in workers context',
   'Error in rendering folder progress bar',
   'Error in rendering top progress bar',
   'Error in ResourceDetailsTab',
 ];
+//We supress recharts warning that is due to our mocking in tests.
+const SUBSTRINGS_TO_SUPPRESS_IN_CONSOLE_WARN = [
+  'The width(0) and height(0) of chart should be greater than 0,',
+];
+
 // this is a quick fix for #938
 const SUBSTRINGS_TO_SUPPRESS_IN_CONSOLE_ERROR = [
   'should be wrapped into act(...)',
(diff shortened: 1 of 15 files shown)


setupTests.ts
```diff
@@ -74,8 +94,16 @@ function getMockConsoleImplementation(
...
-      ? originalConsoleError.call(console, args)
-      : originalConsoleInfo.call(console, args);
+    switch (type) {
+      case 'error': {
+        return originalConsoleError.call(console, args);
+      }
+      case 'warn': {
+        return originalConsoleWarn.call(console, args);
+      }
+      default: {
```
benedikt-richter, line 104: instead of handling this in default, it would be better to have an explicit case.

setupTests.ts
```diff
@@ -13,6 +13,10 @@ const SUBSTRINGS_TO_SUPPRESS_IN_CONSOLE_INFO = [
   'Error in rendering top progress bar',
   'Error in ResourceDetailsTab',
 ];
+const SUBSTRINGS_TO_SUPPRESS_IN_CONSOLE_WARN = [
```
benedikt-richter, line 16: we could add a comment here explaining, why we do suppress these warnings

setupTests.ts
```diff
@@ -23,9 +27,19 @@ jest.setTimeout(TEST_TIMEOUT);
 jest.mock('./src/Frontend/web-workers/get-new-accordion-workers');
 jest.mock('./src/Frontend/web-workers/get-new-progress-bar-workers');
 
+const originalResizeObserver = window.ResizeObserver;
 const originalConsoleError = console.error;
+const originalConsoleWarn = console.warn;
 const originalConsoleInfo = console.info;
 
+function mockResizeObserver(): void {
```
benedikt-richter, line 35: We should remove the old definition of that function that we had before.

src/Frontend/Components/EditAttributionPopup/__tests__/EditAttributionPopup.test.tsx
```diff
@@ -142,6 +139,9 @@ describe('The EditAttributionPopup', () => {
         })
       )
     );
+    store.dispatch(
```
benedikt-richter, line 142: why was this test changed?

benedikt-richter (changes requested):

src/Frontend/Components/EditAttributionPopup/__tests__/EditAttributionPopup.test.tsx
FlorianSchepers: Must have been a mistake. I removed it.

setupTests.ts
benedikt-richter: This comment could read something like "We supress recharts warning that is due to our mocking in tests."

benedikt-richter (approved):
lgtm

Just consider adding the comment

benedikt-richter (approved):
Times for CI vary a lot. This probably also depends on the actual machine utilized. Should be ok to merge and to improve if necessary.

PR #1792, opened 2023-07-10
Resource icon is grey after deletion of all manual attribution

### LINKED ISSUE #1685: Resources icon remains green after deleting attributions
## Context
User can create and delete attributions for resources. Unattributed resources have grey icons, attributed - dark green. 

## Issue
After creating and deleting all attributions for a resource the icon remains light green.
image

The light green color is actually used in Progress Bar for 'Files with only pre-selected attributions'. This is confusing.
One should check if it is an intended behaviour and fix it if not.

vasily-pozdnyakov:
### Summary of changes

resourcesWithManualAttributedChildren.pathToIndices is not updated once attribution is deleted. Paths is the correct list which is updated.

Fix #1685

Start the app, find any resource without attributions (resource with grey icon), create a new attribution for the resource, delete it and check the color of resource icon (it has to be in the same state as it was at the beginning - grey)

### DIFF
--- a/src/Frontend/Components/ResourceBrowser/__tests__/get-resource-browser-tree-item-label.test.ts
+++ b/src/Frontend/Components/ResourceBrowser/__tests__/get-resource-browser-tree-item-label.test.ts
@@ -3,8 +3,14 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
-import { Criticality } from '../../../../shared/shared-types';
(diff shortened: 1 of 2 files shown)


maxhbr (approved):
LGTM

vasily-pozdnyakov:
Added tests and realized that it is not python ('in' does not work for arrays in js) :)

benedikt-richter (approved):

Issue #1821, opened 2023-07-20
Fetch data button tooltip shows wrong text for signals


benedikt-richter:
When hovering the "fetch data" button in the URL field on audit view for a signal with e.g. the URL "https://github.com/reactchartjs/react-chartjs-2/tree/d8cbcb7dc050d749771c6bb1347e22ec63bdddf9" the button is disabled (as expected) but shows the text "... please enter a URL with one of the following domains ... github.com ...". However, the URL actually has this domain (see screenshot).
Instead the button should show an appropriate tooltip text, like: "Fetching data not possible for signals as these are read-only" or similar.

<img width="940" alt="Screenshot 2023-07-20 at 14 47 54" src="https://github.com/opossum-tool/OpossumUI/assets/46576389/a4bbf0a8-d6ff-400d-a239-3fd1deead060">

PR #1825, opened 2023-07-21
Fix tooltip

benedikt-richter:
### Summary of changes

Introduce new tooltip for fetch data button in order to inform the user that fetching data is not possible for signals.

Fix: #1821

Observe correct tooltips in OpossumUI.


meretp (approved):
LGTM

PR #1872, opened 2023-08-11
Replace jszip with fflate async

leslielazzarino:
The result is better speed and possibly better reliability.

### Summary of changes

Replace jszip with fflate async in the functions writing the .opossum files.

Users have complained that sporadically .opossum files get corrupted after saving.

Speed: Compare with the Chrome performance profiler the speed with tha latest release and this brunch (after building it). Reliability: unless we notice something before, we can gather feedback only after releasing it, as the previous error could not be reproduced by me in a short ammount of time.


benedikt-richter (changes requested):
When opening a corrupt zip file, the loading spinner is shown forever with no user facing message displayed to inform what is going on.

src/ElectronBackend/output/writeJsonToOpossumFile.ts
```diff
@@ -3,57 +3,84 @@
...
 // SPDX-License-Identifier: Apache-2.0
 
-import JSZip from 'jszip';
+import * as fflate from 'fflate';
 import fs from 'fs';
-import log from 'electron-log';
 import { OPOSSUM_FILE_COMPRESSION_LEVEL } from '../shared-constants';
+import { Unzipped } from 'fflate';
 
+// just used in tests
```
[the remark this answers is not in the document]
leslielazzarino: Actually, it isn't used in production any longer.

src/ElectronBackend/input/parseFile.ts
```diff
@@ -188,3 +185,11 @@ function parseAndValidateJson(
   jsonSchemaValidator.validate(jsonContent, schema, validationOptions);
   return jsonContent;
 }
+
+function convertUint8ArrayToString(fileData: Uint8Array): string {
```
leslielazzarino, line 189: Test that this function works with unicode.

leslielazzarino, line 189: write unit test for unicode chars

src/ElectronBackend/output/writeJsonToOpossumFile.ts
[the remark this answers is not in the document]
leslielazzarino: I will move it to test helpers.

benedikt-richter (changes requested):
The app is stuck when attempting to open broken zip files. There should be an error instead displayed to the user. Currently, there is a never ending spinner.

leslielazzarino:
> The app is stuck when attempting to open broken zip files. There should be an error instead displayed to the user. Currently, there is a never ending spinner.

It isn't when further testing.

src/ElectronBackend/input/__tests__/parseFile.test.ts
```diff
@@ -327,7 +330,8 @@ describe('parseOutputJsonFile', () => {
     );
 
     expect(() => parseOutputJsonFile(attributionPath)).toThrow(
-      `Error: ${attributionPath} is not a valid attribution file.`,
+      `Error: ${attributionPath} contains an invalid output file.
+instance requires property \"metadata\"`,
```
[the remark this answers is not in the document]
leslielazzarino: Improved.

src/ElectronBackend/input/parseFile.ts
```diff
@@ -34,57 +34,38 @@ export async function parseOpossumFile(
...
-                  })
-              : new Promise<void>((resolve) => resolve());
-          await Promise.all([promiseInputJson, promiseOutputJson]);
-          resolve();
-        }
-      });
-    });
-  });
+  const zip: fflate.Unzipped = await readZipAsync(opossumFilePath);
+  if (!zip['input.json']) {
```
benedikt-richter, line 38: Here, the shared constants should be used

src/ElectronBackend/input/parseFile.ts
```diff
@@ -34,57 +34,38 @@ export async function parseOpossumFile(
...
+    try {
+      parsedInputData = parseAndValidateJson(inputJson, OpossumInputFileSchema);
+    } catch (err) {
+      jsonParsingError = {
+        message: `Error: ${opossumFilePath} does not contain a valid input file.\n Original error message: ${err}`,
+        type: 'jsonParsingError',
+      };
+    }
+
+    if (zip['output.json']) {
```
benedikt-richter, line 57: Here, the shared constants should be used

benedikt-richter:
> > The app is stuck when attempting to open broken zip files. There should be an error instead displayed to the user. Currently, there is a never ending spinner.
> 
> It isn't when further testing.

Could reproduce. Must have had an older version. It is now working. 

Not sure only showing "OK" button with no other choice is perfect, as clicking the button leads the user to infinite loading. However, better as before and can be improved later.

benedikt-richter (changes requested):
While working with the app, I get the zip warnings repeatedly.

In particular, when adding signals to resources in quick succession, I get the warning popup.

benedikt-richter (commented):
It seems the library `jszip` is still used in some places

src/ElectronBackend/types/types.ts
```diff
@@ -37,6 +37,7 @@ export interface GlobalBackendState {
   projectId?: string;
   inputContainsCriticalExternalAttributions?: boolean;
   inputFileChecksum?: string;
+  inputFileUint8Array?: Uint8Array;
```
benedikt-richter, line 40: I would not name the variable after its type. We typically do not do that and the type is there if anyone wonders about its type.

benedikt-richter (commented):
Retested it. The error popup does no longer show up.

leslielazzarino:
> jszip

True, in some scripts and in the backwards-compatibility code. As we are not entirely sure which library we will be using in the future (see https://github.com/opossum-tool/OpossumUI/issues/1893), and it takes quite some work to remove it, I would postpone it.

benedikt-richter:
> [> jszip](https://github.com/opossum-tool/OpossumUI/pull/1872#pullrequestreview-1591340297)
> 
> True, in some scripts and in the backwards-compatibility code. As we are not entirely sure which library we will be using in the future (see #1893), and it takes quite some work to remove it, I would postpone it.

In this case, we should create an issue for it.

benedikt-richter (approved):

PR #1873, opened 2023-08-11
[DO NOT MERGE] Replace jszip with fflate: use promise generators

leslielazzarino:
### Summary of changes

An attempt at refactoring that does not work, maybe someone has an idea why. ___Just look at the last commit.___

### DIFF
--- a/src/ElectronBackend/utils/returnPromise.ts
+++ b/src/ElectronBackend/utils/returnPromise.ts
@@ -1,0 +1,30 @@
+// SPDX-FileCopyrightText: Meta Platforms, Inc. and its affiliates
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+
+export async function returnPromiseOfData<T>(
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  asyncFunction: (...args: Array<any>) => void,
+  ...args: Array<unknown>
+): Promise<T> {
+  return new Promise((resolve): void => {
+    asyncFunction(...args, (err: unknown, data: T): void => {
+      if (err) throw err;
+      resolve(data);
+    });
+  });
+}
+
+export async function returnPromiseOfVoid(
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  asyncFunction: (...args: Array<any>) => void,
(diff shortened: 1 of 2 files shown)


benedikt-richter:
Is this one still needed? Which issue does it relate to?

leslielazzarino:
> Is this one still needed? Which issue does it relate to?

It is just a small refactoring, that for some reason does not work. Maybe someone has an idea how to fix it?

src/ElectronBackend/utils/returnPromise.ts
```diff
@@ -0,0 +1,30 @@
+// SPDX-FileCopyrightText: Meta Platforms, Inc. and its affiliates
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+
+export async function returnPromiseOfData<T>(
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  asyncFunction: (...args: Array<any>) => void,
+  ...args: Array<unknown>
```
mstykow, line 9: with this approach, the args are no longer typed, i.e., no longer connected to the function you pass in as first argument. maybe you can improve the typing of your helper functions to fix this issue. but i'm also not convinced that this PR is worth putting any more time into since
a) these abstractions don't get used a lot
b) they don't actually work and produce errors instead

src/ElectronBackend/utils/returnPromise.ts
```diff
@@ -0,0 +1,30 @@
+// SPDX-FileCopyrightText: Meta Platforms, Inc. and its affiliates
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+
+export async function returnPromiseOfData<T>(
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  asyncFunction: (...args: Array<any>) => void,
```
mstykow, line 8: naming is misleading since this is not actually an async function (notice that the typing does not involve any promises)

src/ElectronBackend/utils/returnPromise.ts
leslielazzarino: As you also don't know why we get the errors, I will just close this PR.

PR #1883, opened 2023-08-16
Fix computation of "files with only external attribution count"

antonbauhofer:
### Summary of changes

Fix computation of "files with only external attribution count"

When calculating the number of "Files with only signals" for the tooltip of the progress bar, we want to sum up all files that only have a signal and all files that only have a parent which only has a signal. Before this fix, the logic was such that only files with a direct parent with signal were counted, but not files with a parent higher up the tree, that has a signal.

For one example Opossum input file, we calculated the number that should be shown in the metrics by hand and compared it to the displayed number. We also used a console log to verify that the correct paths were counted.

### DIFF
--- a/src/Frontend/state/helpers/__tests__/progress-bar-data-helpers.test.ts
+++ b/src/Frontend/state/helpers/__tests__/progress-bar-data-helpers.test.ts
@@ -21,6 +21,11 @@ describe('The getUpdatedProgressBarData function', () => {
   it('gets updated progress data', () => {
     const testResources: Resources = {
       thirdParty: {
+        package: {
+          contents: {
+            'readme.txt': 1,
+          },
+        },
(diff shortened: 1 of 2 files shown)


benedikt-richter:
Would be good to have a test that catches the bug if it re-occurs.

antonbauhofer:
> Would be good to have a test that catches the bug if it re-occurs.

I extended one of our tests to cover the fix, so it would fail if we have the same error again.

benedikt-richter (approved):
👍

antonbauhofer:
> Btw the commit message should contain the issue number, like `Fix: #number `
> One might also consider adding some more information/context in the commit message.

Yes, makes sense.
I added the ticket number and a description to the commit message.

PR #1884, opened 2023-08-18
Bump dependencies

benedikt-richter:
### Summary of changes

Bump eslint-plugin-testing-library from 5.11.1 to 6.0.0 and @testing-library/jest-dom from 5.17.0 to 6.0.0. Also remove test that only tested implementation detail.

Dependabot PRs are broken. The current PR replaces  #1879 and #1876


src/Frontend/Components/FileSearchPopup/__tests__/FileSearchPopup.test.tsx
```diff
@@ -116,33 +116,4 @@ describe('FileSearch popup ', () => {
 
     expect(screen.queryByText('/', { exact: false })).not.toBeInTheDocument();
   });
-
-  it('has search with state', () => {
```
benedikt-richter, line 120: This test is artificial and includes "dirty" unmounting of the component. All it tests is that we store the search term in state. This is an implementation detail.

leslielazzarino (approved):

Issue #1939, opened 2023-09-08
[1935] Introduce new filter state


benedikt-richter:
Introduce new filter state in resource state:

```
locatePopup: {
 selectedCriticality: Criticality | "any"   (introduce new type for this one),
 selectedLicenses: Set<string>;
},

```

--
Part of  #1935

benedikt-richter:
Guess, it is closed by https://github.com/opossum-tool/OpossumUI/pull/1948

PR #1948, opened 2023-09-08
Introduce new locatePopup filter state

MarkusObendrauf:
Issue: #1939

### Summary of changes

Introduce a new filter state in resource state:

```
locatePopup: {
 selectedCriticality: Criticality | "any"   (introduce new type for this one),
 selectedLicenses: Set<string>;
}
```

as well as getters and setters.

This state is intended to be used by a new locate popup, that will make is possible to search for attributions by criticality and/or by licenses.

See unit tests

### DIFF
--- a/src/Frontend/state/reducers/resource-reducer.ts
+++ b/src/Frontend/state/reducers/resource-reducer.ts
@@ -15,6 +15,7 @@ import {
 } from '../../../shared/shared-types';
 import { PackagePanelTitle } from '../../enums/enums';
 import {
+  SelectedCriticality,
   PackageAttributeIds,
   PackageAttributes,
   PanelPackage,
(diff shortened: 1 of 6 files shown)


src/Frontend/state/reducers/resource-reducer.ts
```diff
@@ -136,6 +139,10 @@ export const initialResourceState: ResourceState = {
     },
     totalAttributionCount: null,
   },
+  locatePopup: {
+    selectedCriticality: 'any',
+    selectedLicenses: new Set(),
```
lennartclaas, line 144:
```suggestion
    selectedLicenses: new Set<string>(),
```

lennartclaas (approved):
lgtm

src/Frontend/state/reducers/resource-reducer.ts
MarkusObendrauf: tbh I considered this one too. however, there are 3 other places in this function where we initialize an empty set like this, so ill keep it this way for consistency

Issue #1938, opened 2023-09-08
[1935] Create popup skeleton 


benedikt-richter:
Create popup skeleton 
* Just empty new popup with apply and clear buttons (non-functional)
* Also add new enum for popups
* Add to global popup
* Has functional cancel button

--
Part of  #1935

meretp:
fixed with #1949

PR #1949, opened 2023-09-08
Add LocatorPopup to find signals

meretp:
### Summary of changes

This diff adds the skeleton for a locator popup. 

We want to introduce a new feature which enables the user to search for signals based on criticality and licenses.

See the unit tests. We have a different ticket to make the popup appear with a shortcut, once this is implemented one can see the actual popup. 

Issue: #1938


MarkusObendrauf (approved):
lgtm

image

Issue #1935, opened 2023-09-07
Initial Locate Popup


benedikt-richter:
This issue has the goal to implement a first version of the locate popup that is functional but does not yet contain all features.

**Description**
* The user should be able to open the popup via "Ctrl + L"
* The user should be able to select a criticality 
* The user should be able to select a license in an auto-complete single-select
  * these two filters are independent
* The user should be able to press "apply" with the following effects
  * The popup closes
  * The resource tree is highlighted accordingly
  * There is an icon indicating that the location feature is active and there are filter selected. Clicking that icon opens the popup.
* The user should be able to press "cancel" with the following effects
  * The popup closes

**Mock:**
![Untitled scene(1)](https://github.com/opossum-tool/OpossumUI/assets/46576389/10c48d75-8665-46a8-a32b-b8f2dd0026e7)

benedikt-richter:
Part of release: https://github.com/opossum-tool/OpossumUI/releases/tag/OpossumUI-2023-09-27

Issue #1934, opened 2023-09-07
Introduce "Locate Popup"


benedikt-richter:
**Goal**
Introduce a new popup that enables the user to locate signals in the resource tree.
For this reason, we want to introduce a new popup that allows to configure criteria for signals. When the user configures a set of cirteria and clicks "apply", the resources that have attached signals that match these criteria are highlighted in the resource tree.

The visuals as well as some additional details are described in the following mock:

![Untitled scene](https://github.com/opossum-tool/OpossumUI/assets/46576389/79793609-f242-4721-86fa-7198ac19ff12)

A first implementation is provided by #1935

mstykow:
This is completed.

PR #1960, opened 2023-09-12
Introduce icon indicating that there are filters selected

MarkusObendrauf:
### Summary of changes

Display a filter icon on the top bar whenever locator popup filters are active. Clicking on it opens the locator popup.
image

This allows a user to see that they have filters active, and gives another way to open the popup.

Open the popup with Ctrl+L, then select some filters.


MarkusObendrauf:
Just saw the mock-up and realized I put the icon in the wrong place. These issues are incorrectly linked to #1935 when it should be #1934 - #1934 contains all the relevant info and images.

PR #1961, opened 2023-09-12
Add license selector to popup

### LINKED ISSUE #1941: [1935] Add license selector to popup
Add license selector to popup (functional; sets value in state) 

--
Part of  #1935

meretp:
### Summary of changes

This commit adds an auto-completion input field with a dropdown to the locator popup. The dropdown contains all licenses found in any external attribution and does not aggregate licenses. If a license is selected and the apply button is pressed the corresponding state is set. If the clear button is pressed the state will be reset. Fixes #1941.
We want to imlement a new feature to be able to locate signals based on criticality and certain licenses.
Check out the commit, open an .opossum file and open the popup. Observe the license field and the dropdown when clicking into the field, the dropdown should contain all license names of all external attributions (see the screenshot below).
If you start to type anything the dropdown will filter for those licenses matching your text. Click apply to set the state for this selected license. If you close the popup and reopen it your selected license will be shown in the corresponding field.
If you click the clean button the field will be cleared and the state is reset.

Screenshot from 2023-09-12 16-09-44

### DIFF
--- a/src/Frontend/Components/LocatorPopup/LocatorPopup.tsx
+++ b/src/Frontend/Components/LocatorPopup/LocatorPopup.tsx
@@ -9,14 +9,26 @@ import { closePopup } from '../../state/actions/view-actions/view-actions';
 import { ButtonText, CriticalityTypes } from '../../enums/enums';
 import { useAppDispatch, useAppSelector } from '../../state/hooks';
 import { Dropdown, menuItem } from '../InputElements/Dropdown';
-import { getLocatePopupSelectedCriticality } from '../../state/selectors/locate-popup-selectors';
-import { setLocatePopupSelectedCriticality } from '../../state/actions/resource-actions/locate-popup-actions';
-import { SelectedCriticality } from '../../../shared/shared-types';
+import {
+  getLocatePopupSelectedCriticality,
+  getLocatePopupSelectedLicenses,
+} from '../../state/selectors/locate-popup-selectors';
+import {
+  setLocatePopupSelectedCriticality,
+  setLocatePopupSelectedLicenses,
+} from '../../state/actions/resource-actions/locate-popup-actions';
+import {
+  Attributions,
+  SelectedCriticality,
+} from '../../../shared/shared-types';
+import { getExternalAttributions } from '../../state/selectors/all-views-resource-selectors';
+import { AutoComplete } from '../InputElements/AutoComplete';
 
 const classes = {
   dropdown: {
     marginTop: '8px',
   },
+  autocomplete: { marginTop: '8px' },
 };
 
 const criticalityMenuItems: Array<menuItem> = [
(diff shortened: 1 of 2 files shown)


src/Frontend/Components/LocatorPopup/LocatorPopup.tsx
```diff
@@ -3,19 +3,57 @@
...
 import { ButtonText } from '../../enums/enums';
-import { useAppDispatch } from '../../state/hooks';
+import { useAppDispatch, useAppSelector } from '../../state/hooks';
+import { getFirstLocatePopupSelectedLicense } from '../../state/selectors/locate-popup-selectors';
+import { AutoComplete } from '../InputElements/AutoComplete';
+import { setLocatePopupSelectedLicenses } from '../../state/actions/resource-actions/locate-popup-actions';
+import { getExternalAttributions } from '../../state/selectors/all-views-resource-selectors';
+import { Attributions } from '../../../shared/shared-types';
+
+export function getLicenseNamesFromExternalAttributions(
```
benedikt-richter, line 17: If it is exported only for testing, please, add a respetice comment.

src/Frontend/Components/LocatorPopup/LocatorPopup.tsx
```diff
@@ -3,19 +3,57 @@
...
+  const selectedLicense = useAppSelector(getFirstLocatePopupSelectedLicense);
+  const [searchedLicense, setSearchedLicense] = useState(selectedLicense);
   const dispatch = useAppDispatch();
   function close(): void {
     dispatch(closePopup());
   }
-
-  const content = <></>;
+  const externalAttributions = useAppSelector(getExternalAttributions);
+  const licenseNamesInSignals =
```
benedikt-richter, line 38: Is the performance sufficient to do it on every render? Did you try with a realistic example?

src/Frontend/Components/LocatorPopup/LocatorPopup.tsx
```diff
@@ -3,19 +3,57 @@
...
+    getLicenseNamesFromExternalAttributions(externalAttributions);
+  const content = (
+    <AutoComplete
+      isEditable={true}
+      sx={{ padding: 2 }}
```
benedikt-richter, line 43: Please, avoid inline styling

src/Frontend/Components/LocatorPopup/LocatorPopup.tsx
```diff
@@ -3,19 +3,57 @@
...
+      title={'License'}
+      handleChange={(event: ChangeEvent<HTMLInputElement>): void => {
+        setSearchedLicense(event.target.value);
+      }}
+      isHighlighted={false}
+      options={Array.from(licenseNamesInSignals.values())}
```
benedikt-richter, line 49: Why do we transform the return value of `getLicenseNamesFromExternalAttributions` instead of returning a different value. I.e. moving the logic into `getLicenseNamesFromExternalAttributions`

src/Frontend/Components/LocatorPopup/LocatorPopup.tsx
```diff
@@ -3,19 +3,57 @@
...
+      inputValue={searchedLicense}
+      showTextBold={false}
+      formatOptionForDisplay={(option: string): string => {
+        return option;
+      }}
```
benedikt-richter, line 54: could be 
```
  formatOptionForDisplay={(option: string): string => option}
```

src/Frontend/state/selectors/locate-popup-selectors.ts
```diff
@@ -14,3 +14,14 @@ export function getLocatePopupSelectedCriticality(
 export function getLocatePopupSelectedLicenses(state: State): Set<string> {
   return state.resourceState.locatePopup.selectedLicenses;
 }
+
+export function getFirstLocatePopupSelectedLicense(state: State): string {
```
benedikt-richter, line 18: I'd suggest to still use `getLocatePopupSelectedLicenses` and to handle the return (i.e. taking first value) in the component.
The danger of the current approach is that e.g. the case size = 0 is handled separately such that if replacing `getFirstLocatePopupSelectedLicense` in the component in the future by `getLocatePopupSelectedLicenses` one might forget to also include such handling in`getLocatePopupSelectedLicenses`.

src/Frontend/Components/LocatorPopup/LocatorPopup.tsx
benedikt-richter: btw we should never use `signals` in variable names. We always refer to these as `externalAttributions`. The variable might just be called `licenseNameOptions`

src/Frontend/Components/LocatorPopup/LocatorPopup.tsx
benedikt-richter: The function might be called `getLicenseNames`. The fact that it uses "ExternalAttributions" is not required in the name. From the sginature it should be clear that attributions are used. (btw it would also work for manual attributions)

src/Frontend/Components/LocatorPopup/__tests__/LocatorPopup.test.tsx
```diff
@@ -16,5 +31,85 @@ describe('Locator popup ', () => {
...
+    const licenseSet = new Set(['MIT']);
+
+    renderComponentWithStore(<LocatorPopup />, { store: testStore });
+
+    expectElementsInAutoCompleteAndSelectFirst(screen, ['MIT']);
+    fireEvent.click(screen.getByRole('button', { name: 'Apply' }) as Element);
+    expect(getLocatePopupSelectedLicenses(testStore.getState())).toEqual(
+      licenseSet,
+    );
+  });
```
benedikt-richter, line 70: typically, we have empty lines between tests

src/Frontend/Components/LocatorPopup/__tests__/LocatorPopup.test.tsx
```diff
@@ -16,5 +31,85 @@ describe('Locator popup ', () => {
...
+    testStore.dispatch(setLocatePopupSelectedLicenses(licenseSet));
+
+    renderComponentWithStore(<LocatorPopup />, { store: testStore });
+
+    fireEvent.click(screen.getByRole('button', { name: 'Clear' }) as Element);
+    expect(getLocatePopupSelectedLicenses(testStore.getState())).toEqual(
+      new Set(),
+    );
+  });
+  it('shows license if selected beforehand', () => {
```
benedikt-richter, line 84: typically, we have empty lines between tests

src/Frontend/Components/LocatorPopup/__tests__/LocatorPopup.test.tsx
```diff
@@ -16,5 +31,85 @@ describe('Locator popup ', () => {
...
+    expect(screen.getByDisplayValue('MIT')).toBeInTheDocument();
+  });
+});
+
+describe('getLicenseNamesFromExternalAttributions', () => {
+  it('collects the correct license names', () => {
+    const testExternalAttribution: PackageInfo = {
+      licenseName: 'MIT',
+    };
+    const testExternalAttribution_2: PackageInfo = {
```
benedikt-richter, line 100: If possible avoid `_2` or similar. For readability, it would be nice to have e.g. `testExternalApacheAttribution` and `testExternalMITAttribution` or similar

benedikt-richter (changes requested):

src/Frontend/Components/LocatorPopup/LocatorPopup.tsx
meretp: I tried with the provided example file from ort and didn't observe any performance issues.

src/Frontend/Components/LocatorPopup/LocatorPopup.tsx
```diff
@@ -3,19 +3,59 @@
...
 export function LocatorPopup(): ReactElement {
+  const selectedLicenses = useAppSelector(getLocatePopupSelectedLicenses);
+  // currently we only support sets with one element
+  // once we support multiple elements we will have to adapt the logic to not take one arbitrary element of the set
+  const selectedLicense =
+    selectedLicenses.size == 0 ? '' : selectedLicenses.values().next().value;
+  const [searchedLicense, setSearchedLicense] = useState(selectedLicense);
   const dispatch = useAppDispatch();
+
+  const classes = { autocomplete: { padding: 2 } };
```
benedikt-richter, line 38: typically, we do define these `classes` right after the imports.

benedikt-richter (approved):
Looks good 👍 

just two minor comments.

meretp:
After the rebase the popup now looks like this:  
Screenshot from 2023-09-14 08-03-56

PR #1975, opened 2023-09-14
Disable deps optimization

benedikt-richter:
### Summary of changes

For the e2e tests deps optimization is disabled.

For the e2e tests deps optimization is disabled as it seems to occasionally lead to timeouts in e2e tests if node modules are cached.

`yarn test:e2e`


benedikt-richter:
seems not to solve the issue

PR #1978, opened 2023-09-14
Add logic to calculate located resources

### LINKED ISSUE #1944: [1935] Calculate tree nodes that are highlighted
Calculate tree nodes that are highlighted:

* Input: criticality, license name (in future first party,...)
* Loop over attributions to filter for relevant ones
  * If the searched license matches either the short name or full name of a frequent license, both names should be found 
* Use attributionToResources to find relevant resources
* Use resources to calculate ResourcesWithAttributedChildren
  * We already calculate ResourcesWithAttributedChildren in quite some other places
* Update state on hitting apply or reset

--
Part of  #1935

meretp:
### Summary of changes

This commit adds new logic to calculate all resources that should be highlighted when we filter for signals with certain licenses or criticality. Fixes #1944.

We want to imlement a new feature to be able to locate signals based on criticality and certain licenses. 

Checkout the commit, open a file and select a criticality and or license name. Observe the new state of resourcesWithLocatedAttributions in the redux dev tools browser. Make sure that there exists at least one resource with a signal of the selected type.

### DIFF
--- a/src/Frontend/Components/LocatorPopup/LocatorPopup.tsx
+++ b/src/Frontend/Components/LocatorPopup/LocatorPopup.tsx
@@ -23,6 +23,8 @@ import {
 } from '../../../shared/shared-types';
 import { getExternalAttributions } from '../../state/selectors/all-views-resource-selectors';
 import { AutoComplete } from '../InputElements/AutoComplete';
+import { locateResourcesByCriticalityAndLicense } from '../../state/helpers/action-and-reducer-helpers';
+import { setResourcesWithLocatedAttributions } from '../../state/actions/resource-actions/all-views-simple-actions';
 
 const classes = {
   dropdown: {
@@ -80,6 +82,12 @@ export function LocatorPopup(): ReactElement {
   function handleApplyClick(): void {
     dispatch(setLocatePopupSelectedCriticality(criticalityDropDownChoice));
     dispatch(setLocatePopupSelectedLicenses(new Set([searchedLicense])));
+    dispatch(
+      locateResourcesByCriticalityAndLicense(
+        criticalityDropDownChoice,
+        new Set([searchedLicense]),
+      ),
+    );
   }
 
   function handleClearClick(): void {
@@ -87,6 +95,16 @@ export function LocatorPopup(): ReactElement {
     dispatch(setLocatePopupSelectedCriticality(SelectedCriticality.Any));
     setSearchedLicense('');
     dispatch(setLocatePopupSelectedLicenses(new Set()));
+    dispatch(
+      setResourcesWithLocatedAttributions(
+        {
+          paths: [],
+          pathsToIndices: {},
+          attributedChildren: {},
+        },
+        new Set(),
+      ),
+    );
   }
   function close(): void {
     dispatch(closePopup());
--- a/src/Frontend/state/helpers/__tests__/action-and-reducer-helpers.test.ts
+++ b/src/Frontend/state/helpers/__tests__/action-and-reducer-helpers.test.ts
@@ -6,9 +6,11 @@
 import {
   AttributionData,
   Attributions,
+  Criticality,
   PackageInfo,
   Resources,
   ResourcesToAttributions,
+  SelectedCriticality,
 } from '../../../../shared/shared-types';
 import { createTestAppStore } from '../../../test-helpers/render-component-with-store';
 import { getParsedInputFileEnrichedWithTestData } from '../../../test-helpers/general-test-helpers';
@@ -20,12 +22,19 @@ import {
   createExternalAttributionsToHashes,
   getAttributionIdOfFirstPackageCardInManualPackagePanel,
   getIndexOfAttributionInManualPackagePanel,
+  locateResourcesByCriticalityAndLicense,
 } from '../action-and-reducer-helpers';
 import {
   initialResourceState,
   ResourceState,
 } from '../../reducers/resource-reducer';
 import { EMPTY_ATTRIBUTION_DATA } from '../../../shared-constants';
+import {
+  setLocatePopupSelectedCriticality,
+  setLocatePopupSelectedLicenses,
+} from '../../actions/resource-actions/locate-popup-actions';
+import { getResourcesWithLocatedAttributions } from '../../selectors/all-views-resource-selectors';
+import { setExternalData } from '../../actions/resource-actions/all-views-simple-actions';
 
 describe('The attributionForTemporaryDisplayPackageInfoExists function', () => {
   it('checks if manual attributions exist', () => {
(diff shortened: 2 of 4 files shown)


src/Frontend/state/helpers/action-and-reducer-helpers.ts
```diff
@@ -254,3 +263,76 @@ export function getIndexOfAttributionInManualPackagePanel(
...
+  return (dispatch: AppThunkDispatch, getState: () => State): void => {
+    const locatedResources = new Set<string>();
+    const attributions = getExternalAttributions(getState());
+    const attributionsToResources = getExternalAttributionsToResources(
+      getState(),
+    );
+    const resourcesToExternalAttributions = getResourcesToExternalAttributions(
+      getState(),
+    );
+    // if license Name check frequent licenses for either short name or full name?
```
meretp, line 280: In the ticket description we also added the requirement
"If the searched license matches either the short name or full name of a frequent license, both names should be found", this part is still missing.

src/Frontend/state/helpers/action-and-reducer-helpers.ts
```diff
@@ -254,3 +263,76 @@ export function getIndexOfAttributionInManualPackagePanel(
...
+      attribution: PackageInfo,
+    ): boolean {
+      return (
+        criticality == SelectedCriticality.Any &&
+        licenseNames.size > 0 &&
+        attribution.licenseName !== undefined &&
+        licenseNames.has(attribution.licenseName)
+      );
+    }
+    function CriticalityMatchesAndLicenseNAmesIsEmpty(
```
antonbauhofer, line 301: Nit: NAmes -> Names

src/Frontend/state/helpers/action-and-reducer-helpers.ts
```diff
@@ -254,3 +263,76 @@ export function getIndexOfAttributionInManualPackagePanel(
 
   return packageCardIndex !== -1 ? packageCardIndex : null;
 }
+
+export function locateResourcesByCriticalityAndLicense(
```
antonbauhofer, line 267: I would have expected to have this logic in LocatorPopup.tsx, but here is good as well.

src/Frontend/state/helpers/action-and-reducer-helpers.ts
antonbauhofer: We decided to do this part in a separate issue.

src/Frontend/state/helpers/action-and-reducer-helpers.ts
```diff
@@ -254,3 +263,76 @@ export function getIndexOfAttributionInManualPackagePanel(
...
+    const locatedResources = new Set<string>();
+    const attributions = getExternalAttributions(getState());
+    const attributionsToResources = getExternalAttributionsToResources(
+      getState(),
+    );
+    const resourcesToExternalAttributions = getResourcesToExternalAttributions(
+      getState(),
+    );
+    // if license Name check frequent licenses for either short name or full name?
+    function CriticalityAndLicenseMatches(attribution: PackageInfo): boolean {
```
antonbauhofer, line 281: We don't usually use nested functions. But here, I think it is ok.

src/Frontend/state/helpers/action-and-reducer-helpers.ts
```diff
@@ -254,3 +263,76 @@ export function getIndexOfAttributionInManualPackagePanel(
...
+        CriticalityAndLicenseMatches(attribution) ||
+        CriticalityIsAnyAndLicenseMatches(attribution) ||
+        CriticalityMatchesAndLicenseNAmesIsEmpty(attribution)
+      ) {
+        attributionsToResources[attributionId].forEach((resource) => {
+          locatedResources.add(resource);
+        });
+      }
+    }
+    const resourcesToAttributions: ResourcesToAttributions = {};
```
antonbauhofer, line 322: This name should be more specific.

src/Frontend/state/helpers/action-and-reducer-helpers.ts
```diff
@@ -254,3 +263,76 @@ export function getIndexOfAttributionInManualPackagePanel(
...
+    for (const resourcePath in resourcesToExternalAttributions) {
+      if (locatedResources.has(resourcePath)) {
+        resourcesToAttributions[resourcePath] =
+          resourcesToExternalAttributions[resourcePath];
+      }
+    }
+    const resourcesWithLocatedChildren: ResourcesWithAttributedChildren =
+      computeChildrenWithAttributions(resourcesToAttributions);
```
antonbauhofer, line 330: Instead of doing this, would it not be easier / more efficient to calculate the parents of the located resources and then just use those as resourcesWithLocatedChildren?

src/Frontend/state/helpers/action-and-reducer-helpers.ts
```diff
@@ -254,3 +263,76 @@ export function getIndexOfAttributionInManualPackagePanel(
...
+        licenseNames.size == 0
+      );
+    }
+    for (const attributionId in attributions) {
+      const attribution = attributions[attributionId];
+      if (
+        CriticalityAndLicenseMatches(attribution) ||
+        CriticalityIsAnyAndLicenseMatches(attribution) ||
+        CriticalityMatchesAndLicenseNAmesIsEmpty(attribution)
+      ) {
```
antonbauhofer, line 316: For now, this logic works well. However, if we add more criteria to the locate popup in the future, this approach would lead to n! functions (with n being the number of criteria) if I am not mistaken. Maybe we should redesign it from the beginning to be more scalable.

antonbauhofer (changes requested):

src/Frontend/Components/LocatorPopup/LocatorPopup.tsx
```diff
@@ -80,13 +82,29 @@ export function LocatorPopup(): ReactElement {
...
     dispatch(setLocatePopupSelectedCriticality(SelectedCriticality.Any));
     setSearchedLicense('');
     dispatch(setLocatePopupSelectedLicenses(new Set()));
+    dispatch(
+      setResourcesWithLocatedAttributions(
+        {
+          paths: [],
+          pathsToIndices: {},
+          attributedChildren: {},
+        },
```
antonbauhofer, line 104: Do we actually need all this? I think we could just save a list of paths.

src/Frontend/state/helpers/__tests__/action-and-reducer-helpers.test.ts
```diff
@@ -344,3 +353,117 @@ describe('getIndexOfAttributionInManualPackagePanel', () => {
...
+
+    testStore.dispatch(
+      locateResourcesByCriticalityAndLicense(criticality, licenseNames),
+    );
+    const expectedLocatedResources = {
+      resourcesWithLocatedChildren: {
+        attributedChildren: { 1: new Set([0]) },
+        paths: ['pathToMITHigh/pathToMITMedium', 'pathToMITHigh/'],
+        pathsToIndices: {
+          'pathToMITHigh/pathToMITMedium': 0,
```
antonbauhofer, line 401: This path should not be in resourcesWithLocatedChildren, since it is located itself.
The same applies for the tests below.

src/Frontend/state/helpers/__tests__/action-and-reducer-helpers.test.ts
```diff
@@ -344,3 +353,117 @@ describe('getIndexOfAttributionInManualPackagePanel', () => {
...
+        SelectedCriticality.Any,
+        licenseNames,
+      ),
+    );
+    const expectedLocatedResources = {
+      resourcesWithLocatedChildren: {
+        attributedChildren: { 0: new Set([1]) },
+        paths: ['pathToMITHigh/', 'pathToMITHigh/pathToMITMedium'],
+        pathsToIndices: {
+          'pathToMITHigh/pathToMITMedium': 1,
```
antonbauhofer, line 426: see above.

src/Frontend/state/helpers/__tests__/action-and-reducer-helpers.test.ts
```diff
@@ -344,3 +353,117 @@ describe('getIndexOfAttributionInManualPackagePanel', () => {
...
+    testStore.dispatch(
+      locateResourcesByCriticalityAndLicense(criticality, new Set()),
+    );
+    const expectedLocatedResources = {
+      resourcesWithLocatedChildren: {
+        attributedChildren: { 1: new Set([0]) },
+        paths: [
+          'pathToMITHigh/pathToMITMedium',
+          'pathToMITHigh/',
+          'pathToApacheMedium',
```
antonbauhofer, line 452: see above.

src/Frontend/state/helpers/action-and-reducer-helpers.ts
```diff
@@ -254,3 +263,76 @@ export function getIndexOfAttributionInManualPackagePanel(
...
+    const resourcesToExternalAttributions = getResourcesToExternalAttributions(
+      getState(),
+    );
+    // if license Name check frequent licenses for either short name or full name?
+    function CriticalityAndLicenseMatches(attribution: PackageInfo): boolean {
+      return (
+        criticality != SelectedCriticality.Any &&
+        attribution.criticality == criticality &&
+        attribution.licenseName !== undefined &&
+        licenseNames.size > 0 &&
```
antonbauhofer, line 286: Actually, this condition does not work as expected in the current state, since initiate the set with an empty string entry if it is not specified by the user. Our tests should have caught that.

antonbauhofer:
To address the issues pointed out, we made a separate PR #2005.

PR #1996, opened 2023-09-18
Highlight located resources

MarkusObendrauf:
### Summary of changes

Display an arrow on the virtualized tree pointing to located resources.
image
This will highlight resources satisfying one of the following conditions:
1. The resource is located
2. The resource contains located children, and the resource is not expanded.

This allows a user to find located resources, which satisfy the filters chosen in the locator popup (Ctrl+L).

See unit tests.
The locator popup does not yet set any state, so testing it normally is not possible (See #1944). To try it in the app, I hardcoded some located resources in `ResourceBrowser.tsx`:
```
  // const locatedResources = locateSignalActive
  //   ? resourcesWithLocatedAttributions.locatedResources
  //   : undefined;
  // const resourcesWithLocatedChildren = locateSignalActive
  //   ? resourcesWithLocatedAttributions.resourcesWithLocatedChildren.paths
  //   : undefined;
  const locatedResources = new Set<string>(['/Frontend/shared-styles.ts', '/index.tsx']);
  const resourcesWithLocatedChildren = ['/', '/Frontend/'];
```

### DIFF
--- a/src/Frontend/Components/ResourceBrowser/ResourceBrowser.tsx
+++ b/src/Frontend/Components/ResourceBrowser/ResourceBrowser.tsx
@@ -3,6 +3,7 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
+import WestRoundedIcon from '@mui/icons-material/WestRounded';
 import MyLocationIcon from '@mui/icons-material/MyLocation';
 import remove from 'lodash/remove';
 import React, { ReactElement } from 'react';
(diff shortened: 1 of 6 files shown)


FlorianSchepers (approved):

MarkusObendrauf:
> As discussed earlier with Benedikt I realy do not like the solution with arrows. However, I am fine with fixing this in a follow up ticket.

I was only trying to match the PoC! lol
We have a finalize styling ticket, so lets leave it for that ticket.

Issue #1984, opened 2023-09-18
[1952] Add icons to "critical licenses" table


benedikt-richter:
There should be icons (proposal: locate icon) next to the elements in the "critical licenses" table in project statistics popup.

part of #1952

PR #1997, opened 2023-09-18
Add search icons to `CriticalLicensesTable` in `ProjectStatisticsPopup`

lennartclaas:
### Summary of changes
This commit adds icons for searching licenses to the `CriticalLicensesTable` in `ProjectStatisticsPopup`. The logic is implemented in an upcoming PR.

We want the user to be able to search for resources with specific licenses via the `ProjectStatisticsPopup`. The new search icons allow for searching licenses in the `ResourceBrowser` according to the functionality of the `LocatorPopup`.

**Automatic testing:**
Run `ProjectStatisticsPopup.test.tsx`.
**Manual testing:**
Load test file, open `ProjectStatisticsPopup`, and hover over the icons in the third column of the `CriticalLicensesTable`.

Fix: #1984

### DIFF
--- a/src/Frontend/Components/CriticalLicensesTable/CriticalLicensesTable.tsx
+++ b/src/Frontend/Components/CriticalLicensesTable/CriticalLicensesTable.tsx
@@ -7,6 +7,9 @@ import React, { ReactElement } from 'react';
 import { Criticality } from '../../../shared/shared-types';
 import { ProjectLicensesTable } from '../ProjectLicensesTable/ProjectLicensesTable';
 import { LicenseNamesWithCriticality } from '../../types/types';
+import { IconButton } from '../IconButton/IconButton';
+import { LocateAttributionsIcon } from '../Icons/Icons';
+import { clickableIcon } from '../../shared-styles';
 
 const LICENSE_COLUMN_NAME_IN_TABLE = 'License name';
 const COUNT_COLUMN_NAME_IN_TABLE = 'Count';
@@ -21,6 +24,10 @@ const classes = {
     maxHeight: '400px',
     maxWidth: '500px',
     marginBottom: '3px',
+  },
+  clickableIcon,
+  iconButton: {
+    marginLeft: '8px',
   },
 };
 
@@ -70,6 +77,12 @@ export function CriticalLicensesTable(
       containerStyle={classes.container}
       columnHeaders={TABLE_COLUMN_NAMES}
       columnNames={TABLE_COLUMN_NAMES}
+      firstColumnIconButtons={Object.fromEntries(
+        criticalLicensesTotalAttributions.map(({ licenseName }) => [
+          licenseName,
+          getLocateAttributionIconButton(licenseName),
+        ]),
+      )}
       rowNames={criticalLicensesTotalAttributions.map(
         (attribution) => attribution.licenseName,
       )}
@@ -77,7 +90,9 @@ export function CriticalLicensesTable(
         criticalLicensesTotalAttributions.map(
           ({ licenseName, totalNumberOfAttributions }) => [
             licenseName,
-            { [COUNT_COLUMN_NAME_IN_TABLE]: totalNumberOfAttributions },
+            {
+              [COUNT_COLUMN_NAME_IN_TABLE]: totalNumberOfAttributions,
+            },
           ],
         ),
       )}
@@ -135,4 +150,19 @@ function getTotalNumberOfAttributions(
         licenseNameAndTotalAttributions.totalNumberOfAttributions,
     )
     .reduce((total, value) => total + value, 0);
+}
+
+function getLocateAttributionIconButton(licenseName: string): ReactElement {
+  const onLocateAttributionButtonClick = function (): void {
+    // TODO: dispatch license locator actions in next ticket
+  };
+  return (
+    <IconButton
+      tooltipTitle={`locate attributions with "${licenseName}"`}
+      tooltipPlacement="right"
+      onClick={onLocateAttributionButtonClick}
+      iconSx={classes.iconButton}
+      icon={<LocateAttributionsIcon sx={classes.clickableIcon} />}
+    />
+  );
 }
--- a/src/Frontend/Components/Icons/Icons.tsx
+++ b/src/Frontend/Components/Icons/Icons.tsx
@@ -17,6 +17,7 @@ import LocalParkingIcon from '@mui/icons-material/LocalParking';
 import SearchIcon from '@mui/icons-material/Search';
 import AutoAwesomeIcon from '@mui/icons-material/AutoAwesome';
 import QuestionMarkIcon from '@mui/icons-material/QuestionMark';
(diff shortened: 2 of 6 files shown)


src/Frontend/Components/CriticalLicensesTable/CriticalLicensesTable.tsx
```diff
@@ -136,3 +146,25 @@ function getTotalNumberOfAttributions(
...
+function getSearchLicenseIconButton(
+  licenseName: string,
+  totalNumberOfAttributions: number,
+): ReactElement {
+  const onSearchLicenseButtonClick = function (): void {
+    // TODO: dispatch license locator actions in next ticket
+  };
+  const buttonIsEnabled = Boolean(totalNumberOfAttributions);
+  const tooltipTitle = buttonIsEnabled
+    ? `search for "${licenseName}" in resource browser`
```
benedikt-richter, line 159: It's not a "search" but a "locate" in our terms

benedikt-richter:
Maybe it would be more consistent to use the "locate" icon instead of the search icon

src/Frontend/Components/CriticalLicensesTable/CriticalLicensesTable.tsx
```diff
@@ -77,7 +81,13 @@ export function CriticalLicensesTable(
         criticalLicensesTotalAttributions.map(
           ({ licenseName, totalNumberOfAttributions }) => [
             licenseName,
-            { [COUNT_COLUMN_NAME_IN_TABLE]: totalNumberOfAttributions },
+            {
+              [COUNT_COLUMN_NAME_IN_TABLE]: totalNumberOfAttributions,
+              [SEARCH_COLUMN_NAME_IN_TABLE]: getSearchLicenseIconButton(
```
vasily-pozdnyakov, line 86: I do not think that it should be a new column, maybe just place it next to the license names.

src/Frontend/Components/CriticalLicensesTable/CriticalLicensesTable.tsx
```diff
@@ -136,3 +146,25 @@ function getTotalNumberOfAttributions(
...
 }
+
+function getSearchLicenseIconButton(
+  licenseName: string,
+  totalNumberOfAttributions: number,
+): ReactElement {
+  const onSearchLicenseButtonClick = function (): void {
+    // TODO: dispatch license locator actions in next ticket
+  };
+  const buttonIsEnabled = Boolean(totalNumberOfAttributions);
```
vasily-pozdnyakov, line 157: why do we need this?

src/Frontend/Components/CriticalLicensesTable/CriticalLicensesTable.tsx
vasily-pozdnyakov: and I am not sure that every user knows what resource browser is. Propose something like: "locate resources with 'some license' "

vasily-pozdnyakov (changes requested):
Honestly, I do not like the appearance (not sure that new column is required, also Total looks not good in this case)
image

src/Frontend/Components/Icons/Icons.tsx
```diff
@@ -73,6 +74,10 @@ interface LabelDetailIconProps extends IconProps {
   disabled?: boolean;
 }
 
+interface SearchLicenseIconProps extends IconProps {
```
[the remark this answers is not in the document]
lennartclaas: I renamed everything according to `LocatorPopup`. In particular, it is searching for `attributions` (see button label). I guess to keep everything as generic as possible (one could think about searching for attributions via licenses and other properties).

src/Frontend/Components/CriticalLicensesTable/CriticalLicensesTable.tsx
lennartclaas: Just for security. I wasn't sure whether situations are possible in which a license points to zero attributions (maybe some edge case). But if that's not possible I can remove it and expect the buttons to be always enabled.

src/Frontend/Components/CriticalLicensesTable/CriticalLicensesTable.tsx
lennartclaas: I found that having the button after the license name looks quite oddly when the lenghts of licenses names vary a lot. Before the names looks kind of confusing. However, I have no strong opinion. I will place the icons after the license name.

vasily-pozdnyakov:
image
or
image

lennartclaas:
I decided to place the icons after the license name because I want to preserve text alignment on the left side of the first column.

Screenshot 2023-09-18 181741

src/Frontend/Components/ProjectLicensesTable/ProjectLicensesTable.tsx
```diff
@@ -19,6 +19,12 @@ import { LicenseNamesWithCriticality } from '../../types/types';
 
 const PLACEHOLDER_ATTRIBUTION_COUNT = '-';
 
+const classes = {
+  rowNameWithIconButton: {
+    marginRight: '8px',
```
benedikt-richter, line 24: I'd attach the margin to the button instead. The butten itself should take care of it.

benedikt-richter (approved):
just minor comment

src/Frontend/Components/CriticalLicensesTable/CriticalLicensesTable.tsx
```diff
@@ -136,3 +148,17 @@ function getTotalNumberOfAttributions(
...
     .reduce((total, value) => total + value, 0);
 }
+
+function getLocateAttributionIconButton(licenseName: string): ReactElement {
+  const onLocateAttributionButtonClick = function (): void {
+    // TODO: dispatch license locator actions in next ticket
+  };
+  return (
+    <IconButton
+      tooltipTitle={`locate attributions with "${licenseName}"`}
```
vasily-pozdnyakov, line 158: why attributions? should it be resources?

vasily-pozdnyakov (approved):
Just a minor thing about naming:
do we locate attributions or resources?

src/Frontend/Components/CriticalLicensesTable/CriticalLicensesTable.tsx
benedikt-richter: I, personally, would argue that it is signals that we are locating.

lennartclaas: To my understanding, we locate attributions (license is a property of an attribution). As attributions are attached to resources, we locate resources at the same time. It's correct that the visual hints (arrows) point to resources, but I see that as some kind of incompleteness of the location approach we employ. Ideally, we would mark the located attributions, not only the resources. But that would be overkill as the user can easily identify the attribution of interest given the correct resource. Regarding the tooltip, a correct title would be something like 'locate resources that have attributions with <licenseName> attached'. But I don't like that.

Issue #1986, opened 2023-09-18
Criticality bar should show number of resources having at least one highly critical signal as highly critical 


benedikt-richter:
**Context:**

Criticality bar should show number of resources having at least one highly critical signal as highly critical instead of showing the number of resources with only highly critical signals. The same is true for medium critical signals.

**This issue::**

Improved tooltip texts:

* Header line: “Files with no attributions and”
* “with signals” 
  * Count: All files with signals and no attributions
* “with non-critical signals” 
  * Count: All files with non-critical signals and no attribution
*  “with medium-critical signals” 
  * Count: All files with at least one medium-critical signals, but not highly-critical signal and no attribution
* “with highly-critical signals”
  * Count: All files with at least one highly-critical signal and no attribution

**Definition of done:**

* The criticality bar shows the improve tooltip texts
* The counting of the criticality bar was adapted to the descriptions above
* Iterating through the signals by clicking on the criticality bar should be consistent with the signals counted in the tooltip
* Update the OpossumUI user guide

FlorianSchepers:
This logic is already in place. Therefore, we only need to update the tooltip text.

PR #1998, opened 2023-09-18
Update Tooltip text of the criticality bar.

FlorianSchepers:
### Summary of changes

- Updated tooltip text of criticality bar.

OpossumUI_CriticallityBarToolTip

The previous tooltip text was missleading as it was not clear what the "only" related to, e.g., "Files with only highly critical signals". Therefore, the tooltip was updated. The new headline makes it clear, that only resources without attributions but with signals only are considered.
The logic used to calculate the numbers did not have to be changed, as it already implements the behaviour requested in the Issue #1986.

- Checkout the commit
- run `yarn start`
- open the example file `opossum_ui_scan_input.opossum`
- use the toggle on the right hand side of the progress bar in order to switch to the criticallity bar
- Move cursor over the criticallity bar in order to see the updated tooltip text

Issue: #1986
Fix: #1986

### DIFF
--- a/src/Frontend/Components/ProgressBar/progress-bar-helpers.tsx
+++ b/src/Frontend/Components/ProgressBar/progress-bar-helpers.tsx
@@ -58,17 +58,16 @@ export function getCriticalityBarTooltipText(
     progressBarData.filesWithMediumCriticalExternalAttributionsCount;
   return (
     <MuiBox>
-      Files with only signals:{' '}
-      {progressBarData.filesWithOnlyExternalAttributionCount}
+      Files without attributions but <br />
+      with signals: {progressBarData.filesWithOnlyExternalAttributionCount}
       <br />
-      Files with only highly critical signals:{' '}
+      with highly-critical signals:{' '}
       {progressBarData.filesWithHighlyCriticalExternalAttributionsCount}
       <br />
-      Files with only medium critical signals:{' '}
+      with medium-critical signals:{' '}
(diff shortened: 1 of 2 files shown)


src/Frontend/Components/ProgressBar/progress-bar-helpers.tsx
```diff
@@ -58,17 +58,16 @@ export function getCriticalityBarTooltipText(
     progressBarData.filesWithMediumCriticalExternalAttributionsCount;
   return (
     <MuiBox>
-      Files with only signals:{' '}
-      {progressBarData.filesWithOnlyExternalAttributionCount}
+      Files without attributions but <br />
+      with signals: {progressBarData.filesWithOnlyExternalAttributionCount}
       <br />
-      Files with only highly critical signals:{' '}
+      with highly-critical signals:{' '}
```
benedikt-richter, line 64: We have `{' '}` for two of the lines. Do we actually need these? In any case it would be good to use these consistently.

benedikt-richter (approved):
Looks good.

Just a minor comment

src/Frontend/Components/ProgressBar/progress-bar-helpers.tsx
FlorianSchepers: They are automatically created and removed by the formatter depending on wether ther is linebreak or not. So I can not change this.

PR #1999, opened 2023-09-18
Extend packageinfo and schema for preferred attributions

### LINKED ISSUE #1987: Introduce preferred attributions
The goal of this issue is to allow the user to mark attributions as "preferred". 

The .opossum file should reflect that user decision and it also should expose which signals an attribution is preferred over.

### LINKED ISSUE #1988: [1987] Extend PackageInfo and schema
Extend `PackageInfo` and schema by adding the following fields (here example data is shown):

```
"preferredOverOriginIds": [
 "1eb75f3b-fc84-e781-ad9d-d46d6d61a668"
],
"preferred": true,

```

--
Part of #1987

davidmundelius:
### Summary of changes

We add two new fields "preferred" and "prefferedOverOriginIds" to the PackageInfo and the manual attributions in the schema for the output file.

We want to support preferred attributions marked as such by the user. For this, the two new fields are necessary.

Run the updated unit tests.

Fixes #1988

### DIFF
--- a/src/ElectronBackend/output/writeCsvToFile.ts
+++ b/src/ElectronBackend/output/writeCsvToFile.ts
@@ -91,6 +91,8 @@ export function getHeadersFromColumns(columns: Array<KeysOfAttributionInfo>): {
     excludeFromNotice: 'exclude-from-notice',
     criticality: 'criticality',
     needsReview: 'needs-review',
+    preferred: 'preferred',
+    preferredOverOriginIds: 'preferred-over-origin-ids',
   };
 
   const headers: { [key: string]: string } = {};
--- a/src/ElectronBackend/input/OpossumInputFileSchema.json
+++ b/src/ElectronBackend/input/OpossumInputFileSchema.json
@@ -149,6 +149,17 @@
             "items": {
               "type": "string"
             }
+          },
+          "preferred":{
+            "type": "boolean",
+            "description": "Indicates that the attribution has been marked as preferred by a user."
+          },
+          "preferredOverOriginIds": {
+            "type": "array",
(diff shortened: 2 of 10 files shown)


benedikt-richter:
Seems the DCO check fails. You have to sign your commits: https://github.com/opossum-tool/OpossumUI/pull/1999/checks?check_run_id=16892129931

src/ElectronBackend/output/writeCsvToFile.ts
```diff
@@ -91,6 +91,8 @@ export function getHeadersFromColumns(columns: Array<KeysOfAttributionInfo>): {
     excludeFromNotice: 'exclude-from-notice',
     criticality: 'criticality',
     needsReview: 'needs-review',
+    preferred: 'preferred',
+    preferredOverOriginIds: 'preferred-over-origin-ids',
```
benedikt-richter, line 95: I think, we do not have to expose this - very technical - data in this csv export.

benedikt-richter (changes requested):
The input schema should also be adapted accordingly. Currently, the app fails, when adding the new fields to the inpit (see screenshot)


Screenshot 2023-09-18 at 17 17 16

davidmundelius:
Fixed the DCO and input file schema. The csv part is fine as we choose columns explicitly when creating them.

src/ElectronBackend/output/writeCsvToFile.ts
benedikt-richter: Noted that we have an additional config that specifies the columns going into csvs. Therefore it is fine to keep

benedikt-richter (approved):
lgtm

PR #2005, opened 2023-09-18
Calculate located resources

antonbauhofer:
### Summary of changes

Calculate located resources and resources with located children
- Introduce conditions to check the entries in the locate popup
- Adjust the state to only store paths for resources with located children
- Consider short- and full names of frequent licenses
- Fix and add tests
- code style improvements

We are introducing a feature to "locate" resources that have or contain attributions with criteria that can be set in a new popup.

Run the tests

### DIFF
--- a/src/Frontend/Components/LocatorPopup/__tests__/LocatorPopup.test.tsx
+++ b/src/Frontend/Components/LocatorPopup/__tests__/LocatorPopup.test.tsx
@@ -12,17 +12,25 @@ import React from 'react';
 import { getLocatePopupSelectedCriticality } from '../../../state/selectors/locate-popup-selectors';
 import { clickOnButton } from '../../../test-helpers/general-test-helpers';
 import { setLocatePopupSelectedCriticality } from '../../../state/actions/resource-actions/locate-popup-actions';
-import { SelectedCriticality } from '../../../../shared/shared-types';
+import {
+  Criticality,
+  FrequentLicenses,
+  SelectedCriticality,
+} from '../../../../shared/shared-types';
 import { getLicenseNames, LocatorPopup } from '../LocatorPopup';
 import { getLocatePopupSelectedLicenses } from '../../../state/selectors/locate-popup-selectors';
 import { expectElementsInAutoCompleteAndSelectFirst } from '../../../test-helpers/general-test-helpers';
-import { setExternalData } from '../../../state/actions/resource-actions/all-views-simple-actions';
+import {
+  setExternalData,
+  setFrequentLicenses,
+} from '../../../state/actions/resource-actions/all-views-simple-actions';
 import {
   Attributions,
   PackageInfo,
   ResourcesToAttributions,
 } from '../../../../shared/shared-types';
 import { setLocatePopupSelectedLicenses } from '../../../state/actions/resource-actions/locate-popup-actions';
+import { getResourcesWithLocatedAttributions } from '../../../state/selectors/all-views-resource-selectors';
 
 describe('Locator popup ', () => {
   jest.useFakeTimers();
@@ -103,6 +111,10 @@ describe('Locator popup ', () => {
       ),
     );
     const licenseSet = new Set(['MIT']);
+    const expectedLocatedResources = {
+      resourcesWithLocatedChildren: new Set(['/']),
+      locatedResources: new Set(['/root/']),
+    };
 
     renderComponentWithStore(<LocatorPopup />, { store: testStore });
 
@@ -110,6 +122,9 @@ describe('Locator popup ', () => {
     fireEvent.click(screen.getByRole('button', { name: 'Apply' }) as Element);
     expect(getLocatePopupSelectedLicenses(testStore.getState())).toEqual(
       licenseSet,
+    );
+    expect(getResourcesWithLocatedAttributions(testStore.getState())).toEqual(
+      expectedLocatedResources,
     );
   });
 
@@ -125,6 +140,10 @@ describe('Locator popup ', () => {
     expect(getLocatePopupSelectedLicenses(testStore.getState())).toEqual(
       new Set(),
     );
+    expect(getResourcesWithLocatedAttributions(testStore.getState())).toEqual({
+      resourcesWithLocatedChildren: new Set(),
+      locatedResources: new Set(),
+    });
   });
 
   it('shows license if selected beforehand', () => {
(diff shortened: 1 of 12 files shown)


src/Frontend/state/helpers/action-and-reducer-helpers.ts
```diff
@@ -23,6 +25,14 @@ import { ResourceState } from '../reducers/resource-reducer';
...
 import { getClosestParentAttributionIds } from '../../util/get-closest-parent-attributions';
 import { getAttributionBreakpointCheckForResourceState } from '../../util/is-attribution-breakpoint';
+import { AppThunkAction, AppThunkDispatch } from '../types';
+import { setResourcesWithLocatedAttributions } from '../actions/resource-actions/all-views-simple-actions';
+import {
+  getExternalAttributions,
+  getExternalAttributionsToResources,
+  getFrequentLicensesNameOrder,
+} from '../selectors/all-views-resource-selectors';
+import { State } from '../../types/types';
```
MarkusObendrauf, line 35: I really dont think we should be importing this stuff here.

src/Frontend/state/helpers/action-and-reducer-helpers.ts
```diff
@@ -254,3 +264,96 @@ export function getIndexOfAttributionInManualPackagePanel(
 
   return packageCardIndex !== -1 ? packageCardIndex : null;
 }
+
+export function locateResourcesByCriticalityAndLicense(
```
MarkusObendrauf, line 268: Not sure about this structure. Here's how I understand these helpers:

In `resource-reducer.ts` we have a bunch of 'setters' and 'updaters'. These are all mostly a couple of lines - they take an input and write it to the state. If some processing is needed to get the data into the right form, that is done in a helper function here, so that the `resource-reducer.ts` function remains concise.

With that in mind, I think the structure should be the following:
* In `resource-reducers.ts`, replace the function `setResourcesWithLocatedAttributions` with a new function `updateResourcesWithLocatedAttributions`.
* The new function takes as input criticality and licenseNames
* Inside action-and-reducer-helpers, we have a function with signature `(criticality, licensenames, state) -> (resourcesWithLocatedChildren, locatedResources)`. It does pretty much what this function does but doesnt update state. (not 100% sure about passing state here, but it seems reasonable an is done elsewhere.)
* Then `updateResourcesWithLocatedAttributions` just calls the helper function and updates the state with the return value.

MarkusObendrauf: more generally, imo, when your code does `read -> calculate -> write` then its nice (and very unit-testable) to have the calculations independent of the reading & writing.

src/Frontend/state/helpers/action-and-reducer-helpers.ts
```diff
@@ -254,3 +264,96 @@ export function getIndexOfAttributionInManualPackagePanel(
...
+      } else if (licenseNames.has(frequentLicense.fullName)) {
+        licenseNames.add(frequentLicense.shortName);
+      }
+    }
+
+    for (const attributionId in attributions) {
+      const attribution = attributions[attributionId];
+      if (
+        (criticalityMatches(criticality, attribution) &&
+          !licenseIsDifferent(licenseNames, attribution)) ||
```
MarkusObendrauf, line 294: i dont like double negatives, and i  dont understand what is happening here.

src/Frontend/state/helpers/action-and-reducer-helpers.ts
```diff
@@ -254,3 +264,96 @@ export function getIndexOfAttributionInManualPackagePanel(
...
+    );
+  };
+}
+
+function criticalityMatches(
+  criticality: SelectedCriticality,
+  attribution: PackageInfo,
+): boolean {
+  return (
+    criticality != SelectedCriticality.Any &&
```
MarkusObendrauf, line 324: might be clearer if written as
`criticality != initialResourceState.locatePopup.selectedCriticality`

MarkusObendrauf (changes requested):
.

src/Frontend/state/helpers/action-and-reducer-helpers.ts
antonbauhofer: I was playing with the same ideas, but actually, the logic you are suggesting is not correct, since it would pass if neither the license, nor the criticality is set.

The point is that there are three possible states for each variable: Matching, not matching and not set. This will also be true when we introduce additional criteria. We want to evaluate as true, if any one of the criteria is matching and none of the other is not matching. With that in mind, I still think the way it is currently written is good.
But I can add a comment to clarify things.

src/Frontend/state/helpers/action-and-reducer-helpers.ts
antonbauhofer: Yes, I agree with you. Actually, I had kind of a similar comment on the original diff. I will change the structure to avoid this.

src/Frontend/state/helpers/action-and-reducer-helpers.ts
antonbauhofer: Yes, I am with you here. I had the same thought, but just didn't fix it yet. I will change the structure as you suggested.

src/Frontend/state/helpers/action-and-reducer-helpers.ts
antonbauhofer: Not sure if I agree. We could also have `SelectedCriticality.Any` as a constant with a meaningful name like `unspecifiedCriticality` or similar.

src/Frontend/Components/LocatorPopup/__tests__/LocatorPopup.test.tsx
```diff
@@ -157,3 +176,147 @@ describe('getLicenseNamesFromExternalAttributions', () => {
...
+    const criticality = SelectedCriticality.Medium;
+    const licenseNames = new Set(['MIT']);
+    testStore.dispatch(setLocatePopupSelectedCriticality(criticality));
+    testStore.dispatch(setLocatePopupSelectedLicenses(licenseNames));
+
+    renderComponentWithStore(<LocatorPopup />, { store: testStore });
+    clickOnButton(screen, 'Apply');
+
+    const expectedLocatedResources = {
+      resourcesWithLocatedChildren: new Set(['pathToMITHigh/']),
```
MarkusObendrauf, line 239: in the real world, all paths begin with a `/`, and we expect the path `'/'` to be in `resourceWithLocatedChildren`. could you add this to the test?

src/Frontend/Components/LocatorPopup/__tests__/LocatorPopup.test.tsx
```diff
@@ -157,3 +176,147 @@ describe('getLicenseNamesFromExternalAttributions', () => {
...
+    },
+    GPLMediumAttribution: {
+      licenseName: 'General Public License',
+      criticality: Criticality.Medium,
+    },
+  };
+  const testResourcesToAttributions: ResourcesToAttributions = {
+    'pathToMITHigh/': ['MITHighAttribution'],
+    'pathToMITHigh/pathToMITMedium': ['MITMediumAttribution'],
+    pathToApacheHigh: ['ApacheHighAttribution'],
```
MarkusObendrauf, line 206: can we add quotation marks around all of these?

MarkusObendrauf (changes requested):

MarkusObendrauf (approved):
lgtm

Issue #1985, opened 2023-09-18
[1952] Make icon in "critical licenses" table clickable


benedikt-richter:
Clicking the icon:

* It opens Audit View
* It will perform the same filtering action as the [locator popup](https://github.com/opossum-tool/OpossumUI/issues/1952)
  * As in the project statistics table multiple license variants are counted as one. For the filter the license variant with the most entries should be taken.
*  It directly navigates to a (random?) representative resource with this license

part of #1952
Follow up of  #1984.

PR #2006, opened 2023-09-19
Add logic to license locator icons in `ProjectStatisticsPopup`

lennartclaas:
### Summary of changes

Clicking on an icon initiates the location process. All attributions that contain the underlying license are highlighted in the resource browser. The view changes to Audit View and the resource browser navigates to a resource with a located attribution.

The new search icons allow for searching licenses in the `ResourceBrowser` according to the functionality of the `LocatorPopup`. 

Manually by using a test file.

### Comment
Meanwhile, this PR has accumulated some refactorings.

Fix: #1985

### DIFF
--- a/src/Frontend/Components/CriticalLicensesTable/CriticalLicensesTable.tsx
+++ b/src/Frontend/Components/CriticalLicensesTable/CriticalLicensesTable.tsx
@@ -8,8 +8,11 @@ import { Criticality } from '../../../shared/shared-types';
 import { ProjectLicensesTable } from '../ProjectLicensesTable/ProjectLicensesTable';
 import { LicenseNamesWithCriticality } from '../../types/types';
 import { IconButton } from '../IconButton/IconButton';
-import { LocateAttributionsIcon } from '../Icons/Icons';
+import { LocateSignalsIcon } from '../Icons/Icons';
 import { clickableIcon } from '../../shared-styles';
+import { useAppDispatch } from '../../state/hooks';
+import { AppThunkDispatch } from '../../state/types';
+import { locateSignalsFromProjectStatisticsPopup } from '../../state/actions/popup-actions/popup-actions';
 
 const LICENSE_COLUMN_NAME_IN_TABLE = 'License name';
 const COUNT_COLUMN_NAME_IN_TABLE = 'Count';
@@ -40,6 +43,7 @@ interface CriticalLicensesTableProps {
 export function CriticalLicensesTable(
   props: CriticalLicensesTableProps,
 ): ReactElement {
+  const dispatch = useAppDispatch();
   const allLicensesWithCriticality = Object.entries(
     props.licenseNamesWithCriticality,
   ).map((licenseNameAndCriticality) => {
@@ -80,7 +84,7 @@ export function CriticalLicensesTable(
       firstColumnIconButtons={Object.fromEntries(
         criticalLicensesTotalAttributions.map(({ licenseName }) => [
           licenseName,
-          getLocateAttributionIconButton(licenseName),
+          getLocateSignalsIconButton(licenseName, dispatch),
         ]),
       )}
       rowNames={criticalLicensesTotalAttributions.map(
@@ -152,17 +156,20 @@ function getTotalNumberOfAttributions(
     .reduce((total, value) => total + value, 0);
 }
 
-function getLocateAttributionIconButton(licenseName: string): ReactElement {
-  const onLocateAttributionButtonClick = function (): void {
-    // TODO: dispatch license locator actions in next ticket
+function getLocateSignalsIconButton(
+  licenseName: string,
+  dispatch: AppThunkDispatch,
+): ReactElement {
+  const onLocateSignalButtonClick = function (): void {
+    dispatch(locateSignalsFromProjectStatisticsPopup(licenseName));
   };
   return (
     <IconButton
-      tooltipTitle={`locate attributions with "${licenseName}"`}
+      tooltipTitle={`locate signals with "${licenseName}"`}
       tooltipPlacement="right"
-      onClick={onLocateAttributionButtonClick}
+      onClick={onLocateSignalButtonClick}
       iconSx={classes.iconButton}
-      icon={<LocateAttributionsIcon sx={classes.clickableIcon} />}
+      icon={<LocateSignalsIcon sx={classes.clickableIcon} />}
     />
   );
 }
--- a/src/Frontend/Components/LocatorPopup/LocatorPopup.tsx
+++ b/src/Frontend/Components/LocatorPopup/LocatorPopup.tsx
@@ -5,33 +5,28 @@
 
 import React, { ChangeEvent, ReactElement, useState } from 'react';
 import { NotificationPopup } from '../NotificationPopup/NotificationPopup';
-import { closePopup } from '../../state/actions/view-actions/view-actions';
+import {
+  closePopup,
+  setShowNoSignalsLocatedMessage,
+} from '../../state/actions/view-actions/view-actions';
 import { ButtonText, CriticalityTypes } from '../../enums/enums';
 import { useAppDispatch, useAppSelector } from '../../state/hooks';
 import { Dropdown, menuItem } from '../InputElements/Dropdown';
 import {
-  getLocatePopupSelectedCriticality,
-  getLocatePopupSelectedLicenses,
+  getLocatePopupFilters,
+  getShowNoSignalsLocatedMessage,
 } from '../../state/selectors/locate-popup-selectors';
-import {
-  setLocatePopupSelectedCriticality,
-  setLocatePopupSelectedLicenses,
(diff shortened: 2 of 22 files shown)


src/Frontend/Components/CriticalLicensesTable/CriticalLicensesTable.tsx
```diff
@@ -152,9 +195,40 @@ function getTotalNumberOfAttributions(
...
+    criticality,
+    new Set<string>([licenseName]),
+    externalAttributions,
+    externalAttributionsToResources,
+    frequentLicensesNameOrder,
+  );
+  const resourcesWithLocatedChildren =
+    getResourcesWithLocatedChildren(locatedResources);
+
   const onLocateAttributionButtonClick = function (): void {
```
leslielazzarino, line 218: This is not how you want to do this. The best pattern for selectors and actions in the components is as follows:
- get from the state only what is relevant for visualisation: the component will then re-render only if something it is displaying (or a derivate thereof) changes
- get everything else needed for an action in the action itself: this will not cause re-renders and better separate action and component dependencies.

leslielazzarino (changes requested):

leslielazzarino:
I will have a deeper look after the selectors have been migrated.

src/Frontend/Components/CriticalLicensesTable/CriticalLicensesTable.tsx
lennartclaas: I created a new action that handles all this stuff. Thanks for pointing that out.

src/Frontend/state/actions/popup-actions/popup-actions.ts
```diff
@@ -346,3 +364,40 @@ export function closeAttributionWizardPopup(): AppThunkAction {
...
+      new Set<string>([licenseName]),
+      externalAttributions,
+      externalAttributionsToResources,
+      frequentLicenseNames,
+    );
+    const resourcesWithLocatedChildren =
+      getResourcesWithLocatedChildren(locatedResources);
+
+    dispatch(setLocatePopupSelectedCriticality(criticality));
+    dispatch(setLocatePopupSelectedLicenses(new Set([licenseName])));
```
leslielazzarino, line 390: Why are these two separate actions? It will potentially cause unwanted re-renders, as the dispatches in a thunk action are applied one after the other, causing re-renders (and calculations) after each. I would prefer to have a new extra action "setLocatePopupFilters" that allows to set all filters at once.

leslielazzarino (changes requested):

src/Frontend/Components/ProjectStatisticsPopup/__tests__/ProjectStatisticsPopup.test.tsx
```diff
@@ -83,6 +100,46 @@ describe('The ProjectStatisticsPopup', () => {
...
+    store.dispatch(openPopup(PopupType.ProjectStatisticsPopup));
+    renderComponentWithStore(<ProjectStatisticsPopup />, { store });
+
+    expect(getSelectedView(store.getState())).toBe(View.Attribution);
+    expect(getSelectedResourceId(store.getState())).toBe('');
+    expect(getOpenPopup(store.getState())).toBe(
+      PopupType.ProjectStatisticsPopup,
+    );
+    const iconButtonMIT = screen.getByRole('button', {
+      name: 'locate attributions with "MIT"',
```
MarkusObendrauf, line 133: not sure this is the right place, but shouldn't we call it 'signals' instead of 'attributions'?

lennartclaas:
I refactored the two individual actions for setting and getting `selectedCriticality` and `selectedLicenses`. Now, both are always set and got at once via `locatePopupFilters`. This prevents unnecessary re-renderings.

src/Frontend/util/locate-attributions.ts
```diff
@@ -0,0 +1,72 @@
...
+
+import {
+  Attributions,
+  AttributionsToResources,
+  FrequentLicenseName,
+  SelectedCriticality,
+} from '../../shared/shared-types';
+import { getParents } from '../state/helpers/get-parents';
+
+export function calculateResourcesWithLocatedAttributions(
```
benedikt-richter, line 14: Did not look into it in detail. However, this is very specific for the locate feature and should not be in the `util` folder. Probably, move to component helpers (that's what we typically do to extract specific helpers.

lennartclaas: This stuff is used in `LocatorPopup` and `popup-actions`. To me it seems a little bit unconventional to import component helpers to an action.

benedikt-richter: We discussed to actually move it to the reducer.

vasily-pozdnyakov:
I found an issue: statistics popup contains invalid pairs of license name (the most frequent one is taken) and criticality (there is some heuristics) and it fails when no results are found. In this case the locator popup informs user that there are no signals.
I do not know what is the best way to handle this at statistics popup.

lennartclaas:
@vasily-pozdnyakov concerning your [comment](https://github.com/opossum-tool/OpossumUI/pull/2006#issuecomment-1733982024). Nice catch! To really fix this properly, I would argue that substantial changes in the logic of the statistics popup are required. I would therefore create a new ticket for that. For now, I would say that this issue is quite an edge case and does not introduce major obstacles for the users.

src/Frontend/Components/LocatorPopup/LocatorPopup.tsx
```diff
@@ -95,107 +84,53 @@ export function LocatorPopup(): ReactElement {
...
+  if (applyWasClicked) {
     const noSignalsAreFound =
       locatedResources.size === 0 && resourcesWithLocatedChildren.size === 0;
     const allFiltersAreEmpty =
       criticalityDropDownChoice === SelectedCriticality.Any &&
-      searchedLicenses.size === 0;
+      searchedLicense.length === 0;
     if (noSignalsAreFound && !allFiltersAreEmpty) {
       setShowNoSignalsLocatedMessage(true);
+      setApplyWasClicked(false);
```
lennartclaas, line 97: I don't appreciate this approach, but I could not think of any other way to set `showNoSignalsLocatedMessage` without writing it to the redux state (which I would also consider bad practice).

src/Frontend/state/helpers/action-and-reducer-helpers.ts
```diff
@@ -258,3 +260,63 @@ export function getIndexOfAttributionInManualPackagePanel(
...
+  licenseNames: Set<string>,
+  externalAttributions: Attributions,
+  externalAttributionsToResources: AttributionsToResources,
+  frequentLicenseNames: Array<FrequentLicenseName>,
+): Set<string> {
+  completeFrequentLicenseNamesIfPresent(licenseNames, frequentLicenseNames);
+  const locatedResources = new Set<string>();
+
+  const licenseIsSet = licenseNames.size > 0;
+  const criticalityIsSet = selectedCriticality != SelectedCriticality.Any;
```
benedikt-richter, line 275: Do not use `!=` and use `!==` instead as we want to avoid type coercion

src/Frontend/state/actions/popup-actions/popup-actions.ts
```diff
@@ -346,3 +352,54 @@ export function closeAttributionWizardPopup(): AppThunkAction {
...
+      locatedResources.size === 0 && resourcesWithLocatedChildren.size === 0;
+    const allFiltersAreEmpty =
+      criticality === SelectedCriticality.Any && licenseNames.size === 0;
+    const showNoSignalsLocatedMessage =
+      noSignalsAreFound && !allFiltersAreEmpty;
+
+    dispatch(setShowNoSignalsLocatedMessage(showNoSignalsLocatedMessage));
+
+    if (!showNoSignalsLocatedMessage) {
+      dispatch(closePopup());
```
benedikt-richter, line 380: Why don't we include 
```
    dispatch(
      openResourceInResourceBrowser(locatedResources.values().next().value),
    );
```

here?

benedikt-richter:
Thanks for your effort to refactor the code base! The code is much improved now 👍

src/Frontend/state/helpers/action-and-reducer-helpers.ts
```diff
@@ -258,3 +260,63 @@ export function getIndexOfAttributionInManualPackagePanel(
...
 }
+
+export function calculateResourcesWithLocatedAttributions(
+  selectedCriticality: SelectedCriticality,
+  licenseNames: Set<string>,
+  externalAttributions: Attributions,
+  externalAttributionsToResources: AttributionsToResources,
+  frequentLicenseNames: Array<FrequentLicenseName>,
+): Set<string> {
+  completeFrequentLicenseNamesIfPresent(licenseNames, frequentLicenseNames);
```
leslielazzarino, line 271: do we really whant to mutate such a small set? We are mutating the input of the parent function, it is error-prone

leslielazzarino (changes requested):

PR #2008, opened 2023-09-19
Add button to mark attribution as preferred

MarkusObendrauf:
### Summary of changes

Add a button to the ... menu in the audit view, to mark an attribution as preferred.
image
This option is not present in the attribution view, because on that view, you don't see which attributions you prefer this one over. The icon is also added to the report view.
image

We are implementing a new feature to mark attributions as 'preferred'. Preferred attributions give new information which we can use in internal tools to produce higher quality input files.

Open your favorite input file, and mark an attribution as preferred.

### DIFF
--- a/src/Frontend/Components/AttributionColumn/AttributionColumn.tsx
+++ b/src/Frontend/Components/AttributionColumn/AttributionColumn.tsx
@@ -19,6 +19,7 @@ import { getSelectedView } from '../../state/selectors/view-selector';
 import {
   getDisplayedPackage,
   getResolvedExternalAttributions,
+  getSelectedResourceId,
 } from '../../state/selectors/audit-view-resource-selectors';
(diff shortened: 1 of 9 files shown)


davidmundelius (commented):
When we click on Mark as preferred with no attribution selected, the existing attribution is replaced by an empty one in the UI.

MarkusObendrauf:
> When we click on Mark as preferred with no attribution selected, the existing attribution is replaced by an empty one in the UI.

good catch. in fact the menu shouldn't even be openable, and it seems to be slightly buggy on main. created an issue: #2016 
working on a fix for this pr

davidmundelius (approved):
Looks good to me now.

JakobSchubert (approved):

MarkusObendrauf:
rebased & fixed merge conflicts

Issue #1990, opened 2023-09-18
[1987] Add filter for preferred attributions


benedikt-richter:
Add respective filter option in attribution and report view to the existing filter.

--
Part of #1987

davidmundelius:
I will work on this now but cannot assign myself.

PR #2014, opened 2023-09-20
Add filter for preferred attributions

davidmundelius:
### Summary of changes

We add a new filter to attribution view and audit view to filter for preferred attributions.

We want to make it possible for users to mark attributiions as preferred. Therefore we also want a filter for these attributions.

Run the new unit test in src/Frontend/util/__tests__/use-filters.test.tsx.
Open the updated file opossum_input.json from the example-files folder in OpossumUI. Go to attribution view and select the new filter. There should be exactly one attribution in the filtered list.

Fix: #1990


antonbauhofer (commented):
The implemented changes look good to me. The testplan works as expected.

antonbauhofer (approved):

PR #2017, opened 2023-09-20
Add `isRelevantForPreferred` field to external attribution sources

MarkusObendrauf:
### Summary of changes

Add new `isRelevantForPreferred` field to external attribution sources. This field has type optional boolean. If set, it indicates that signals from this source may be 'preferred over' using the preference feature.

If no sources have this field set, then the preference feature is disabled.

We would like for the preference feature to only apply to signals from certain sources.

See unit tests

I have updated the example files `example.opossum` & `opossum_input.json` to include relevant sources. The preferred feature should be enabled on those files. It should be disabled on every other random file you have.


src/Frontend/Components/AttributionColumn/attribution-column-helpers.ts
```diff
@@ -234,10 +235,12 @@ export function getMergeButtonsDisplayState(currentState: {
         currentState.packageInfoWereModified) ||
       currentState.targetAttributionIsPreSelected,
     hideMarkAsPreferredButton:
+      !currentState.isPreferenceFeatureEnabled ||
```
FlorianSchepers, line 238: Optional: A integration test that checks if the button is hidden if the preference Feature is enables, i.e., by using get by text.

FlorianSchepers (approved):
lgtm 

- see one comment in code
- squash commits
- add proper commit message

src/Frontend/Components/AttributionColumn/attribution-column-helpers.ts
MarkusObendrauf: good idea! added one

MarkusObendrauf:
squashed and rebased

Issue #2007, opened 2023-09-19
[1935] Handle pre-selected preferred attributions


benedikt-richter:
When confirming a pre-selected preferred attribution the preferred flag should be dropped. The same is true for confirm globally.

--
Part of https://github.com/opossum-tool/OpossumUI/issues/1935

PR #2018, opened 2023-09-20
Preferred flag is dropped for pre-selected attributions

FlorianSchepers:
### Summary of changes

When creating a manual attribution from a pre-selected external attribution the preferred flag and the list of preferredOverOriginIds are deleted. 

We do not want that new preferred attributions are created automatically via pre-selection.

Run tests in 'importFromFile.test.ts'.

Issue: #2007
Fix: #2007


benedikt-richter (approved):
lgtm

PR #2022, opened 2023-09-20
Cache PanelAttributionData in the manual progress bar worker

MarkusObendrauf:
### Summary of changes

Cache PanelAttributionData in the manual progress bar worker. Then, instead of re-sending it every time we click on a resource, we only re-send it when we save.

Sending this data takes up a significant amount of time when clicking on a folder. By caching the data, navigation speed is improved, at the cost of slower loads.

Open a large file, and measure how long it takes to load a large directory.

We tried it on the biggest file we have and got the following results:
Before: 7 seconds
After: 4 seconds

### DIFF
--- a/src/Frontend/web-workers/contained-manual-attributions-accordion-worker.ts
+++ b/src/Frontend/web-workers/contained-manual-attributions-accordion-worker.ts
@@ -4,23 +4,38 @@
 
 import { DisplayPackageInfosWithCountAndResourceId } from '../types/types';
 import { getContainedManualDisplayPackageInfosWithCount } from '../Components/AggregatedAttributionsPanel/accordion-panel-helpers';
+import { PanelAttributionData } from '../util/get-contained-packages';
+
+let cachedManualData: PanelAttributionData | null = null;
 
 self.onmessage = ({
   data: { selectedResourceId, manualData, panelTitle },
 }): void => {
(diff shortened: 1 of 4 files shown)


src/Frontend/web-workers/contained-manual-attributions-accordion-worker.ts
```diff
@@ -4,23 +4,32 @@
...
-    displayPackageInfosWithCount,
-  };
+  if (manualData !== undefined) {
+    cachedManualData = manualData;
+  }
 
-  self.postMessage({
-    output,
-  });
+  if (selectedResourceId && cachedManualData) {
```
antonbauhofer, line 18: As discussed, I think we should split this up and 

      self.postMessage({
        output: null,
      });

if only `selectedResourceId` is truthy.

antonbauhofer (commented):
Looks good to me. When evaluating the performance on a large app, I found that the loading time was reduced by almost half.
Just consider my comment in the worker.

src/Frontend/web-workers/contained-manual-attributions-accordion-worker.ts
MarkusObendrauf: done

MarkusObendrauf:
squashed and rebased

Issue #2021, opened 2023-09-20
Error for some licenses


vasily-pozdnyakov:
Open example.opossum and try to locate the license "MIT " (the last one in the list, with empty space at the end) with the new Locate Popup
![image](https://github.com/opossum-tool/OpossumUI/assets/85183359/c3709d13-a150-49e4-8dd6-5ecbb096701f)

Not sure what is the problem: example file is corrupted or there is an actual bug in the code.

Related to https://github.com/opossum-tool/OpossumUI/issues/1935

benedikt-richter:
Suggestion: 
* do not write signals without resources into the redux state.
This would solve the issue and avoid running into the same issue at other places.

PR #2026, opened 2023-09-21
Locator popup does not fail when a signal is not linked to any resource

leslielazzarino:
### Summary of changes

When loading an input file into the state, external attributions without linked resources are pruned. The goas is, as they are unreachable, thei should not be present in the state.  The pruning is done in the FE as attributionsToResources are needed to do the search. The attribution without linked resources has been kept in the input file, as a test for this feature.

See linked issue.

Test that the bug does not exist any longer: build the app and try to reproduce the bug, using the opossum_input.json file.

### DIFF
--- a/src/Frontend/Components/PackageCard/__tests__/PackageCard.test.tsx
+++ b/src/Frontend/Components/PackageCard/__tests__/PackageCard.test.tsx
@@ -33,21 +33,28 @@ import {
 } from '../../../state/actions/resource-actions/all-views-simple-actions';
 import { setSelectedResourceId } from '../../../state/actions/resource-actions/audit-view-simple-actions';
 
-const testResources: Resources = {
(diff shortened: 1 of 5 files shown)


vasily-pozdnyakov (approved):

benedikt-richter:
I personally find it strange to argue that it cannot happen but still modify the code to handle it... If we have that code why do we modify the example file?

benedikt-richter:
Don't we ave the issue that licenses from signals that are not linked to resources still show up in the auto-complete of locate popup?

benedikt-richter (changes requested):
see https://github.com/opossum-tool/OpossumUI/issues/2021#issuecomment-1733119479

vasily-pozdnyakov (approved):

PR #2027, opened 2023-09-21
Fix incorrect count of merged hhc signals

vasily-pozdnyakov:
### Summary of changes

Instead of summarizing of all the similar hhc signals, now we just take the max value. I decided to go easy way (not adding a lot of extra logic of deduplication of resources) as the metrics is not the most important and accurate anyway.

For merged hhc signals (where only comments are different ) there is incorrect count of signals (just a sum of all of them) which is not realistic, e.g.
Screenshot 2023-09-21 at 14 09 53

Run tests.
Open any project with merged signals and see that the number is more real now, e.g.
Screenshot 2023-09-21 at 14 07 41


antonbauhofer (approved):
Looks good to me.
Nice catch!

vasily-pozdnyakov:
simplified the logic

Issue #1995, opened 2023-09-18
[1987] Add integration test


benedikt-richter:
Preferred attributions that we introduced in https://github.com/opossum-tool/OpossumUI/issues/1987 should be integration tested.
A test covering a typical flow of user actions should be added.

--
Part of #1987

leslielazzarino:
Already present

PR #2029, opened 2023-09-21
[merge after 2033] Save list of 'preferred over' origin IDs

MarkusObendrauf:
### Summary of changes

When marking an attribution as preferred, save a list of signals that it is preferred over. This list contains all signals in subfolders from sources relevant for preference. As an exception, resources with different manual attributions act as breakpoints.

We are implementing a new feature to mark attributions as 'preferred'. Preferred attributions give new information which we can use in internal tools to produce higher quality input files.

First, make sure you have a sensible choice for "signals relevant for preference" in your input file. Mark an attribution as preferred. Check that the corresponding attribution in the input file is updated.

### DIFF
--- a/src/Frontend/Components/AttributionColumn/AttributionColumn.tsx
+++ b/src/Frontend/Components/AttributionColumn/AttributionColumn.tsx
@@ -53,7 +53,8 @@ import { ContextMenuItem } from '../ContextMenu/ContextMenu';
 import { useAppDispatch, useAppSelector } from '../../state/hooks';
 import MuiBox from '@mui/material/Box';
 import { EMPTY_DISPLAY_PACKAGE_INFO } from '../../shared-constants';
-import { cloneDeep, isEqual } from 'lodash';
+import { isEqual } from 'lodash';
+import { toggleIsSelectedPackagePreferred } from '../../state/actions/resource-actions/preference-actions';
 
 const classes = {
   root: {
@@ -156,15 +157,6 @@ export function AttributionColumn(props: AttributionColumnProps): ReactElement {
 
   const mainButtonConfigs: Array<MainButtonConfig> = [];
 
-  function toggleIsSelectedPackagePreferred(): void {
-    const newTemporaryDisplayPackageInfo = cloneDeep(
-      temporaryDisplayPackageInfo,
-    );
-    newTemporaryDisplayPackageInfo.preferred =
-      !newTemporaryDisplayPackageInfo.preferred;
-    dispatch(setTemporaryDisplayPackageInfo(newTemporaryDisplayPackageInfo));
-  }
-
   if (props.onSaveButtonClick) {
     mainButtonConfigs.push({
       buttonText: temporaryDisplayPackageInfo.preSelected
@@ -239,7 +231,9 @@ export function AttributionColumn(props: AttributionColumnProps): ReactElement {
       buttonText: ButtonText.MarkAsPreferred,
       onClick: (): void => {
         if (selectedPackage) {
-          toggleIsSelectedPackagePreferred();
+          dispatch(
+            toggleIsSelectedPackagePreferred(temporaryDisplayPackageInfo),
+          );
         }
       },
       hidden: mergeButtonDisplayState.hideMarkAsPreferredButton,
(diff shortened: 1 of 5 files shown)


src/Frontend/Components/AttributionColumn/attribution-column-helpers.ts
```diff
@@ -377,3 +383,100 @@ export function getSelectedManualAttributionIdForAuditView(
...
+  for (const [childResourceName, childResource] of Object.entries(
+    rootResource,
+  )) {
+    const pathToChildResource = pathToRootResource + childResourceName;
+    if (isBreakpoint(pathToChildResource)) {
+      continue;
+    }
+    resources.push(pathToChildResource);
+    if (childResource === 1) {
+      resources.push(pathToChildResource);
```
lennartclaas, line 451: Why is `pathToChildResource` added twice?

davidmundelius, line 451: Why do we need this? It seems like you did this already two lines before.

lennartclaas (changes requested):
Would it make sense to add an e2e-test that checks whether `output.json` really contains all required data?

src/Frontend/Components/AttributionColumn/attribution-column-helpers.ts
MarkusObendrauf: good point! removed the first call. it turns out that later deduplicating meant that nothing was broken, but its nicer to have this function return exactly what we need.

MarkusObendrauf:
> Would it make sense to add an e2e-test that checks whether `output.json` really contains all required data?

seperate issue: https://github.com/opossum-tool/OpossumUI/issues/1995

src/Frontend/Components/AttributionColumn/attribution-column-helpers.ts
MarkusObendrauf: fixed

src/Frontend/Components/AttributionColumn/AttributionColumn.tsx
```diff
@@ -112,6 +118,17 @@ export function AttributionColumn(props: AttributionColumnProps): ReactElement {
...
+    getResourcesToManualAttributions,
+  );
+  const resourcesToExternalAttributions = useAppSelector(
+    getResourcesToExternalAttributions,
+  );
+  const resources = useAppSelector(getResources);
+  const externalAttributions = useAppSelector(getExternalAttributions);
+  const externalAttributionsSources = useAppSelector(
+    getExternalAttributionSources,
+  );
```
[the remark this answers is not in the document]
MarkusObendrauf: ok, moved this function to a custom action

src/Frontend/state/actions/resource-actions/preference-actions.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    if (newTemporaryDisplayPackageInfo.preferred) {
+      newTemporaryDisplayPackageInfo.preferredOverOriginIds =
+        _getOriginIdsToPreferOver(
+          resourceId,
+          allViewsState.resources ?? {},
+          allViewsState.externalData.resourcesToAttributions,
+          allViewsState.manualData.attributionsToResources,
+          allViewsState.externalData.attributions,
+          allViewsState.externalAttributionSources,
```
lennartclaas, line 40: So far, I haven't seen this approach of getting state data. Seems a bit like bad practice to me. Usually, we call already existing selectors like
```
const state = getState();
const resourceId = getSelectedResourceId(state);
```

lennartclaas (changes requested):

src/Frontend/state/actions/resource-actions/preference-actions.ts
MarkusObendrauf: good point, thanks. it should now look like the other actions.

lennartclaas (approved):
lgtm

MarkusObendrauf:
ive rebased onto https://github.com/opossum-tool/OpossumUI/pull/2033 to incorporate those changed. i'd like to wait for that to be finalized before merging anything

benedikt-richter:
Would it be safer to convert this one to a draft to avoid it being merged accidentaly, as it is based on a draft?

MarkusObendrauf:
rebased

PR #2033, opened 2023-09-22
Update temporary package info instead of saving when prefering or unprefering an attribution on Audit View

leslielazzarino:
Some tests are still missing

### Summary of changes

Update temporary package info instead of saving when prefering or unprefering an attribution on Audit View. Also the global saving button is disabled if an attribution has been changed to preferred.

That prevents collisions with unsaved changes in the attribution.

Set-unset the preferred attribute.


src/Frontend/Components/ResourceDetailsAttributionColumn/ResourceDetailsAttributionColumn.tsx
```diff
@@ -116,6 +119,17 @@ export function ResourceDetailsAttributionColumn(
...
+      dispatch(
+        unlinkAttributionAndSavePackageInfoIfSavingIsNotDisabled(
+          selectedResourceId,
+          attributionIdOfSelectedPackageInManualPanel as string, // showSaveGloballyButton is true, so attributionIdOfSelectedPackageInManualPanel is not null
+          temporaryDisplayPackageInfo,
+        ),
+      );
+      return;
+    }
+
```
MarkusObendrauf, line 132: i dont understand this change. why is it needed now, and not before?

MarkusObendrauf (changes requested):

src/Frontend/state/actions/resource-actions/__tests__/all-views-simple-actions.test.ts
```diff
@@ -273,11 +275,29 @@ describe('The load and navigation simple actions', () => {
     );
   });
 
-  it('sets and gets isSavingDisabled', () => {
+  // replace following 3 tests with a single parametric one
```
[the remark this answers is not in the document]
leslielazzarino: I think in this case it displays nicely what is expected when in an easy-to-read table

src/Frontend/Components/ResourceDetailsAttributionColumn/ResourceDetailsAttributionColumn.tsx
leslielazzarino: Before ctrl+S was always triggering a global save. Now it triggers a local save, if the global save is disabled.

src/Frontend/state/actions/resource-actions/save-actions.ts
```diff
@@ -57,8 +61,8 @@ import { openPopup } from '../view-actions/view-actions';
 import { getMultiSelectSelectedAttributionIds } from '../../selectors/attribution-view-resource-selectors';
 import { setMultiSelectSelectedAttributionIds } from './attribution-view-simple-actions';
 
-export function setIsSavingDisabled(
-  isSavingDisabled: boolean,
+export function setAllowedSaveOperations(
+  isSavingDisabled: AllowedSaveOperations,
```
MarkusObendrauf, line 65: rename this variable

src/Frontend/state/actions/resource-actions/types.ts
```diff
@@ -221,7 +222,7 @@ export interface SetTargetSelectedAttributionIdAction {
 
 export interface SetIsSavingDisabled {
   type: typeof ACTION_SET_IS_SAVING_DISABLED;
```
MarkusObendrauf, line 224: rename function and action name

MarkusObendrauf (changes requested):
i think the refactoring tools missed some things

src/Frontend/state/actions/resource-actions/types.ts
```diff
@@ -39,7 +40,8 @@ export const ACTION_SET_SELECTED_RESOURCE_ID =
 export const ACTION_SET_EXPANDED_IDS = 'ACTION_SET_EXPANDED_IDS';
 export const ACTION_SET_TARGET_SELECTED_RESOURCE_ID =
   'ACTION_SET_TARGET_SELECTED_RESOURCE_ID';
-export const ACTION_SET_IS_SAVING_DISABLED = 'ACTION_SET_IS_SAVING_DISABLED';
+export const ACTION_SET_ALLOWED_SAVE_OPERATIONS =
+  'ACTION_SET_IS_SAVING_DISABLED';
```
MarkusObendrauf, line 44: this still needs to be changed

Issue #2030, opened 2023-09-21
[1987] Update 'preferred over' list when an attribution is changed in the subtree


MarkusObendrauf:
When an attribution is added or removed in the subtree under a preferred attribution, the list of origin ids has to be updated

--
Part of https://github.com/opossum-tool/OpossumUI/issues/1987

PR #2039, opened 2023-09-25
Unit test linking and unlinking attributions

MarkusObendrauf:
### Summary of changes

Unit test linking and unlinking resources.
Also, improve the test for deleting resources, which currently doesn't assert any changes between the output and input `childrenFromAttributedResources`.

I plan to change these functions soon as part of #2030, so having unit tests is nice.

`yarn jest src/Frontend/state/helpers/__tests__/save-action-helpers.test.ts`

### DIFF
--- a/src/Frontend/state/helpers/__tests__/save-action-helpers.test.ts
+++ b/src/Frontend/state/helpers/__tests__/save-action-helpers.test.ts
@@ -19,6 +19,8 @@ import {
   createManualAttribution,
   deleteManualAttribution,
   updateManualAttribution,
+  unlinkResourceFromAttributionId,
+  linkToAttributionManualData,
 } from '../save-action-helpers';
 
 const testUuid: string = uuidNil;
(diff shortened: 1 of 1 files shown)


src/Frontend/state/helpers/__tests__/save-action-helpers.test.ts
```diff
@@ -193,6 +182,190 @@ describe('The updateManualAttribution function', () => {
...
+      attributedChildren: {},
+      pathsToIndices: {
+        '/': 0,
+        '/first/': 1,
+      },
+      paths: ['/', '/first/'],
+    });
+  });
+
+  it('correctly maintains childrenFromAttributedResources', () => {
```
davidmundelius, line 325: Does this item test anything which is not covered by the previous one? It seems that the only difference between the two is that the first one contains more expect statements.

davidmundelius (commented):
Aside from the one point I mentioned above, this looks good to me.

src/Frontend/state/helpers/__tests__/save-action-helpers.test.ts
MarkusObendrauf: good point. now that im looking at it, theres really no need to have two tests for linking or for unlinking. and i should revert my changes to the deletion test, since i only made it match the first test.

davidmundelius (approved):
Looks good to me now.

PR #2041, opened 2023-09-25
Update preferences when an attribution is created, deleted, linked or unlinked

MarkusObendrauf:
### Summary of changes

Whenever an attribution is created, deleted, linked or unlinked, recalculate the preferences of its parents.

Manual attributions act as break points for preference calculations. Therefore, any time we change an attribution placement in the resource tree, we need to update the preference calculations of its parents. Only the parents are relevant, because the 'prefered over' calculation travels down the resource tree only.

We have 6 save operations for attributions:

- Create
- Replace
- Update
- Delete
- Link
- Unlink

Replace & update don't require any recalculations, because they don't change any breakpoint locations. The other four operations do.

Recalculating preferences requires some external attribution data from the state. I wasn't totally sure how best to pass this around. I didn't want to pass the entire state to the helper functions. Passing the params needed for recalculations was possible too, but passing the function around instead made more sense to me.

See unit tests.
Play around with it in OpossumUI with the output json open, and check that it is updated as expected.

### DIFF
--- a/src/Frontend/state/helpers/__tests__/save-action-helpers.test.ts
+++ b/src/Frontend/state/helpers/__tests__/save-action-helpers.test.ts
@@ -12,6 +12,11 @@ import {
   ResourcesToAttributions,
 } from '../../../../shared/shared-types';
 import { EMPTY_ATTRIBUTION_DATA } from '../../../shared-constants';
+import { getParsedInputFileEnrichedWithTestData } from '../../../test-helpers/general-test-helpers';
+import { createTestAppStore } from '../../../test-helpers/render-component-with-store';
+import { loadFromFile } from '../../actions/resource-actions/load-actions';
+import { initialResourceState } from '../../reducers/resource-reducer';
+import { getManualData } from '../../selectors/all-views-resource-selectors';
 import {
   _getIdsOfResourcesThatMightHaveChildrenWithTheSameAttributions,
   _removeAttributionsFromChildrenAndParents,
@@ -21,6 +26,7 @@ import {
   updateManualAttribution,
   unlinkResourceFromAttributionId,
   linkToAttributionManualData,
+  getCalculatePreferredOverOriginIds,
 } from '../save-action-helpers';
 
 const testUuid: string = uuidNil;
@@ -37,6 +43,7 @@ describe('The createManualAttribution function', () => {
       testManualData,
       testSelectedResourceId,
       testTemporaryDisplayPackageInfo,
+      getCalculatePreferredOverOriginIds(initialResourceState),
     );
     expect(newManualData.attributions[newAttributionId]).toEqual(
       testTemporaryDisplayPackageInfo,
@@ -44,6 +51,61 @@ describe('The createManualAttribution function', () => {
     expect(newManualData.attributionsToResources[newAttributionId]).toEqual([
       '/something.js',
     ]);
+  });
+
+  it('correctly updates preferences', () => {
+    const testSelectedResourceId = '/child';
+    const testTemporaryDisplayPackageInfo: PackageInfo = {
+      packageName: 'React',
+    };
+
+    const testStore = createTestAppStore();
+    testStore.dispatch(
+      loadFromFile(
+        getParsedInputFileEnrichedWithTestData({
+          resources: { child: 1 },
+          resourcesToExternalAttributions: {
+            '/child': ['externalUuid'],
+          },
+          externalAttributions: {
+            externalUuid: {
+              source: { name: 'testSource', documentConfidence: 0 },
+              originIds: ['originId'],
+            },
+          },
+          externalAttributionSources: {
+            testSource: {
+              name: 'Test source',
+              priority: 0,
+              isRelevantForPreferred: true,
+            },
+          },
+          manualAttributions: {
+            manualUuid1: {
+              preferred: true,
+              preferredOverOriginIds: ['originId'],
+            },
+          },
+          resourcesToManualAttributions: {
+            '/': ['manualUuid1'],
+          },
+        }),
+      ),
+    );
+    const resourceState = testStore.getState().resourceState;
+    const testManualData = getManualData(testStore.getState());
+
+    const { newManualData } = createManualAttribution(
+      testManualData,
+      testSelectedResourceId,
+      testTemporaryDisplayPackageInfo,
+      getCalculatePreferredOverOriginIds(resourceState),
+    );
+
+    expect(newManualData.attributions['manualUuid1']).toEqual({
+      preferred: true,
+      preferredOverOriginIds: [],
+    });
   });
 });
 
@@ -78,6 +140,7 @@ describe('The deleteManualAttribution function', () => {
       testManualData,
       testUuid,
       () => false,
+      getCalculatePreferredOverOriginIds(initialResourceState),
     );
     expect(isEmpty(newManualData.attributions)).toBe(true);
     expect(isEmpty(newManualData.resourcesToAttributions)).toBe(true);
@@ -127,6 +190,7 @@ describe('The deleteManualAttribution function', () => {
       testManualData,
       testUuid,
       () => false,
+      getCalculatePreferredOverOriginIds(initialResourceState),
     );
     expect(newManualData.attributions).toEqual({
       '000': { packageName: 'another testpackage' },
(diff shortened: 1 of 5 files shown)


src/Frontend/state/helpers/__tests__/save-action-helpers.test.ts
```diff
@@ -45,6 +48,60 @@ describe('The createManualAttribution function', () => {
...
+        originIds: ['originId'],
+      },
+    };
+    testResourceState.allViews.externalAttributionSources = {
+      testSource: {
+        name: 'Test source',
+        priority: 0,
+        isRelevantForPreferred: true,
+      },
+    };
```
leslielazzarino, line 91: Crating manually all of this will lead to a likely unrealistic state. You should do:
const testStore = createTestAppStore();
testStore.dispatch(
  loadFromFile(
    getParsedInputFileEnrichedWithTestData({
      resources: testResources,
      manualAttributions: testManualAttributions,
      resourcesToManualAttributions: testResourcesToManualAttributions,
    }),
  ),
);

leslielazzarino: Also I cannot see uses of testResourceState, you are just modifying initialResourceState by reference.

src/Frontend/state/helpers/__tests__/save-action-helpers.test.ts
```diff
@@ -148,6 +207,59 @@ describe('The deleteManualAttribution function', () => {
...
```
leslielazzarino, line 249: See other comments.

src/Frontend/state/helpers/__tests__/save-action-helpers.test.ts
```diff
@@ -148,6 +207,59 @@ describe('The deleteManualAttribution function', () => {
...
+        manualUuid2: ['/child/'],
+      },
+      resourcesWithAttributedChildren: {
+        attributedChildren: {},
+        pathsToIndices: {},
+        paths: [],
+      },
+    };
+
+    const testResourceState = initialResourceState;
```
leslielazzarino, line 232: This is not a deep copy, all of you tests will be modifying and using the same object, which is to avoid.

src/Frontend/state/helpers/__tests__/save-action-helpers.test.ts
```diff
@@ -242,6 +355,62 @@ describe('The linkToAttributionManualData function', () => {
...
+        originIds: ['originId'],
+      },
+    };
+    testResourceState.allViews.externalAttributionSources = {
+      testSource: {
+        name: 'Test source',
+        priority: 0,
+        isRelevantForPreferred: true,
+      },
+    };
```
leslielazzarino, line 399: See comments above.

src/Frontend/state/helpers/__tests__/save-action-helpers.test.ts
```diff
@@ -290,6 +460,63 @@ describe('The unlinkResourceFromAttributionId function', () => {
...
+
+    const newManualData = unlinkResourceFromAttributionId(
+      testManualData,
+      testSelectedResourceId,
+      'childAttributionUuid',
+      getCalculatePreferredOverOriginIds(testResourceState),
+    );
```
leslielazzarino, line 513: See comments above.

src/Frontend/state/helpers/save-action-helpers.ts
```diff
@@ -326,6 +363,44 @@ function linkAttributionAndResource(
...
+  calculatePreferredOverOriginIds: CalculatePreferredOverOriginIds,
+): void {
+  getParents(pathToChangedResource).forEach((pathToParent) => {
+    const manualAttributionsIds =
+      newManualData.resourcesToAttributions[pathToParent] ?? [];
+
+    manualAttributionsIds.forEach((manualAttributionId) => {
+      const packageInfo = newManualData.attributions[manualAttributionId];
+      if (packageInfo.preferred) {
+        packageInfo.preferredOverOriginIds = calculatePreferredOverOriginIds(
```
leslielazzarino, line 395: Here you are potentially calling the same function multiple time, if a resource has multiple attributions. Please use a for loop that you break once you find a preferred attribution.

leslielazzarino (changes requested):

src/Frontend/state/helpers/__tests__/save-action-helpers.test.ts
MarkusObendrauf: done

src/Frontend/state/helpers/save-action-helpers.ts
MarkusObendrauf: for the inner loop, in case there are multiple preferred attributions attached to the same resource, we need to update all of them. we can break out of the outer loop though.

MarkusObendrauf: added an early exit

src/Frontend/state/helpers/__tests__/save-action-helpers.test.ts
```diff
@@ -242,6 +372,69 @@ describe('The linkToAttributionManualData function', () => {
...
+
+  it('correctly updates preferences', () => {
+    const testSelectedResourceId = '/child';
+    const testManualData: AttributionData = {
+      attributions: {
+        parentAttriubtionUuid: {
+          preferred: true,
+          preferredOverOriginIds: ['originId'],
+        },
+        childAttributionUuid: { preferred: false },
```
leslielazzarino, line 384: This attribution should alredy be linked somewhere else, otherwise the example isn't realistic.

src/Frontend/state/helpers/__tests__/save-action-helpers.test.ts
```diff
@@ -45,6 +51,67 @@ describe('The createManualAttribution function', () => {
...
+      packageName: 'React',
+    };
+    const testManualData: AttributionData = {
+      attributions: {
+        manualUuid1: { preferred: true, preferredOverOriginIds: ['originId'] },
+      },
+      resourcesToAttributions: {
+        '/': ['manualUuid1'],
+      },
+      attributionsToResources: {
```
leslielazzarino, line 67: For all the tests: if you load manual attributions and resourcesToAttributions with the external attributions, and then get them from the state, attributionsToResourses and WithAttributedChildren come for free.

src/Frontend/state/helpers/save-action-helpers.ts
```diff
@@ -326,6 +363,51 @@ function linkAttributionAndResource(
...
+  for (const pathToParent of getParents(pathToChangedResource)) {
+    let wasPreferredParentFound = false;
+
+    const manualAttributionsIds =
+      newManualData.resourcesToAttributions[pathToParent] ?? [];
+
+    manualAttributionsIds.forEach((manualAttributionId) => {
+      const packageInfo = newManualData.attributions[manualAttributionId];
+      if (packageInfo.preferred) {
+        wasPreferredParentFound = true;
```
leslielazzarino, line 397: can't you remove this variable and just break after calling calculate...? I think it would be as clear and more concise.

src/Frontend/state/reducers/resource-reducer.ts
```diff
@@ -661,10 +665,12 @@ export const resourceState = (
       };
     case ACTION_UNLINK_RESOURCE_FROM_ATTRIBUTION:
       const manualDataAfterUnlinking: AttributionData =
+        // TODO figure out how this is called & where to test
```
leslielazzarino, line 668: This comment should be removed.

leslielazzarino (changes requested):

src/Frontend/state/helpers/save-action-helpers.ts
MarkusObendrauf: see this comment:
https://github.com/opossum-tool/OpossumUI/pull/2041#discussion_r1337025339

src/Frontend/state/reducers/resource-reducer.ts
MarkusObendrauf: lol

src/Frontend/state/helpers/__tests__/save-action-helpers.test.ts
MarkusObendrauf: nice, done

leslielazzarino (approved):

PR #2049, opened 2023-09-26
Remove preferred icon from signals

vasily-pozdnyakov:
### Summary of changes

Only manual attributions get a preferred star icon.

For now, signals should not be marked with a star as preferred.
How it looks like now:
image

Open opossum_input file and check signals


benedikt-richter (changes requested):
I feel that we want to show icons for signals that have `isPreferred` set. If one does not want these icons, the respective field in packageInfo should not be set.

PR #2056, opened 2023-09-27
Add feature flag

benedikt-richter:
### Summary of changes

We add a feature flag in order to deactivate the preferred attributions feature.

Deactivate the preferred attributions feature for a new release, as it is not yet finalized.


FlorianSchepers (approved):
lgtm

Issue #2020, opened 2023-09-20
Consider unsaved changes when locating attributions via ProjectStatisticsPopup


lennartclaas:
It is possible to open the `ProjectStatisticsPopup` while having unsaved changes of an attribution. If the user then locates attributions via the icon buttons in the `CriticalLicensesTable`, the `UnsavedChangesPopup` should appear. After the user has decided to save or undo the changes (or has cancelled), the `ProjectStatisticsPopup` should be shown again,

The same logic has to be applied to the `LocatorPopup`.

PR #2057, opened 2023-09-27
Handle unsaved changes when locating attributions

lennartclaas:
### Summary of changes

It is possible to open the `ProjectStatisticsPopup` or the `LocatorPopup` while having unsaved changes of an attribution. If the user then locates attributions, the `NotSavedPopup` appears. After the user has decided to save or undo the changes (or cancelled), the `ProjectStatisticsPopup` or `LocatorPopup` is shown again. From here, the user can again initialize the location process.

We introduced signal location that can currently be initiated via the ' LocatorPopup'  or the `ProjectStatisticsPopup`. However, so far, we do not consider the possibility of unsaved changes of the `temporaryDisplayPackageInfo` when starting the location process.

1. Open an example file
2. Select a manual attribution (either in audit or attribution view)
3. Change the content of a textbox in the attribution column
4. Open  ' LocatorPopup'  or  `ProjectStatisticsPopup` (via main menu)
5. Start the location process
6. Choose to undo or save the changes
7. See that the   ' LocatorPopup'  or  `ProjectStatisticsPopup` appears again
8. Start the location process

Fix: #2020

### DIFF
--- a/src/Frontend/state/actions/popup-actions/__tests__/popup-actions.test.ts
+++ b/src/Frontend/state/actions/popup-actions/__tests__/popup-actions.test.ts
@@ -104,61 +104,6 @@ describe('The actions checking for unsaved changes', () => {
       ]);
       expect(getSelectedView(testStore.getState())).toBe(View.Audit);
     });
-
-    it('closes an open popup if dispatched from an open popup', () => {
-      const testStore = createTestAppStore();
-      testStore.dispatch(openPopup(PopupType.ProjectStatisticsPopup));
-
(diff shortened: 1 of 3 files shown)


leslielazzarino (changes requested):

leslielazzarino:
I would expect, when getting the usaved canges popup, that clicking on cancel would not bring me back to the locate popup but to the unsaved attribution. I would also expect that clicking on undo after clicking on apply would close the locate popup.

leslielazzarino (approved):

PR #2074, opened 2023-09-29
Add wasPreferred field to package info and schemas

davidmundelius:
### Summary of changes

We add a field wasPreferred to the schemas for input and output files and the packageInfo for attributions which had previously been marked as preferred.

We recently added the possibility to mark attributions as preferred. We also want to be able to highlight attributions which had been marked as preferred before, but not in the current vetting process.

Run the unit tests in src\ElectronBackend\input\__tests__\importFromFile.test.ts.
Open the file opossum_input.json and check in the DevTools that in the state the external attribution with ID bb61b094-b5ba-4bd0-b218-8ec96b7c5ae1 has the property "wasPreferred: true".


lennartclaas (approved):
lgtm

PR #2086, opened 2023-10-02
Handle 'wasPreferred' for merged signals

FlorianSchepers:
### Summary of changes

- Added `wasPreferred` to the list of keys to be excluded when calculating the hash for the an attribution
- Sets `wasPreferred` to true for the `DisplayPackageInfo` if at least for one of the origin attributions the field `wasPreferred` was set to true
- Added tests

Recently we have introduced the `wasPreferred` field  that allowes to highlight external Attributions that were previously preferred. 

In this diff we make sure that the `wasPreferred` filed is handled correctly during merging of multiple attributions.

- Checkout commit
- run `yarn start`
- delete `opossum_input.opossum` in `example-files` if pressent
- open `opossum_input.json`
- create a new file with a ".opossum" extension during opening process
- In audit view navigate to `/index.ts`
- Check that the merged signal with the license `Apache License Version 2.0` for `TypeScript 3.9.6` is marked by the gray star for `wasPreferred`


example-files/opossum_input.json
```diff
@@ -167,8 +167,7 @@
...
-      "originId": "1eb75f3b-fc84-e781-ad9d-d46d6d61abf2",
-      "wasPreferred": true
+      "originId": "1eb75f3b-fc84-e781-ad9d-d46d6d61abf2"
```
FlorianSchepers, line 170: Before the changes introduced by this diff the 'wasPreferred' state dependet on the first attribution in the list of attributions to merge. If for this attribution 'wasPreferred' was set to true it was set to true for the mered attribution as well. This was also the case here in the example file. 
Therefore, I switched the 'wasPreferred' state to the second attribution in the list in order to reliably test the changes.

PR #2089, opened 2023-10-04
Extend license single select to multi select

davidmundelius:
### Summary of changes

We adapt the AutoComplete field for license selection in the locate popup so that multiple licenses can be selected.

We want to extend the locate popup to make it possible to search for multiple licenses at once. Here we change the UI component. The logic behind it will be changed in the next step.

Open the locate popup and check that multiple licenses can be entered.

### Note

In order to make this work, I needed to skip one unit test in LocatorPopup.test.tsx. This one should be reenabled when the new logic for the license search is added.

### DIFF
--- a/src/Frontend/Components/InputElements/AutoComplete.tsx
+++ b/src/Frontend/Components/InputElements/AutoComplete.tsx
@@ -15,6 +15,7 @@ interface AutoCompleteProps extends InputElementProps {
   endAdornmentText?: string;
   inputValue: string;
   showTextBold?: boolean;
+  multiple?: boolean;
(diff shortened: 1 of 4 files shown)


leslielazzarino (changes requested):
1. You should test the multiselect.
2. The x to remove a license looks broken, as it overlaps with the license name. That should be fixed.

davidmundelius:
There is nothing yet that can be tested, since the logic for the multiselect will only be added in the next step. I fixed the positioning of the x.

leslielazzarino (approved):

PR #2093, opened 2023-10-04
Implement signal search

MarkusObendrauf:
### Summary of changes

Adds signal search in the locate signals popup. It uses the already existing search function to search for packages that match the search term.

Gives the user another option for locating signals.

Open the popup with ctrl+L and play around. (For example, try searching for '4.1.1' in the example file - it should give a result where a package has this version.)

### DIFF
--- a/src/Frontend/state/selectors/__tests__/locate-popup-selectors.test.ts
+++ b/src/Frontend/state/selectors/__tests__/locate-popup-selectors.test.ts
@@ -11,7 +11,7 @@ import { isLocateSignalActive } from '../locate-popup-selectors';
 describe('isLocateSignalActive', () => {
   it('returns false in the default state', () => {
     const testStore = createTestAppStore();
-    expect(!isLocateSignalActive(testStore.getState()));
+    expect(!isLocateSignalActive(testStore.getState())).toEqual(true);
   });
 
   it('returns true if the selected criticality is not the default', () => {
@@ -25,7 +25,7 @@ describe('isLocateSignalActive', () => {
       }),
     );
 
-    expect(isLocateSignalActive(testStore.getState()));
+    expect(isLocateSignalActive(testStore.getState())).toEqual(true);
   });
 
   it('returns true if there are selected licenses', () => {
@@ -38,7 +38,8 @@ describe('isLocateSignalActive', () => {
         searchOnlyInLicenseField: false,
       }),
     );
-    expect(isLocateSignalActive(testStore.getState()));
+
+    expect(isLocateSignalActive(testStore.getState())).toEqual(true);
   });
 
   it('returns true if the search term is set', () => {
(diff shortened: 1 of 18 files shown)


src/Frontend/state/selectors/__tests__/locate-popup-selectors.test.ts
```diff
@@ -32,6 +34,34 @@ describe('isLocateSignalActive', () => {
       setLocatePopupFilters({
         selectedCriticality: SelectedCriticality.Any,
         selectedLicenses: new Set<string>(['testLicenseId']),
+        searchTerm: '',
+        searchOnlyInLicenseField: false,
+      }),
+    );
+    expect(isLocateSignalActive(testStore.getState()));
```
leslielazzarino, line 41: This is not the best pattern. Please add to all of these assertions .toEqual(true) or equivalent.

src/Frontend/util/__tests__/search-package-info.test.ts
```diff
@@ -0,0 +1,81 @@
...
+
+describe('packageInfoContainsSearchTerm', () => {
+  it('searches by package name', () => {
+    const testPackageInfo = {
+      packageName: 'Search_term package',
+      attributionIds: ['uuid1'],
+    };
+
+    expect(packageInfoContainsSearchTerm(testPackageInfo, 'SeArCh_TeRm')).toBe(
+      true,
```
leslielazzarino, line 19: We had the convention to alsways use toEqual instead of toBe, as it is strictly better. Please replace it in all tests.

src/Frontend/util/__tests__/search-package-info.test.ts
```diff
@@ -0,0 +1,81 @@
...
+});
+
+describe('licenseContainsSearchTerm', () => {
+  it('searches by license name', () => {
+    const testPackageInfo = {
+      licenseName: 'Search_term licence',
+      attributionIds: ['uuid3'],
+    };
+
+    expect(licenseContainsSearchTerm(testPackageInfo, 'SeArCh_TeRm')).toBe(
```
leslielazzarino, line 77: I find this confusing for the user. We have no license field, but a license name and a license text. I believe we have two alternatives:
- either we rename everything here to license name (from license field)
- or we raplace field with fields and also search in the licanse text.

leslielazzarino (changes requested):

src/Frontend/util/__tests__/search-package-info.test.ts
MarkusObendrauf: done

src/Frontend/util/__tests__/search-package-info.test.ts
MarkusObendrauf: renamed it to `searchOnlyLicenseName` etc. everywhere

leslielazzarino (approved):

PR #2100, opened 2023-10-05
Introduce state for QA mode

meretp:
### Summary of changes

This commit adds a new state for the QA mode. If the corresponding menu item is clicked the mode will change, either from QA to normal mode or normal to QA mode.

~~Note: The branch is based on #2090 and needs to be rebased before merging once the other PR is merged.~~ Done

Selecting preferred attributions should only be possible in the new QA mode. This PR adds the respective state and integrates it with the previous implemented menu item. 

Checkout the changes and open OpossumUI. Open the Redux tab in the developer tools and see the new field `qaMode` as part of the `viewState`. This one is set to false.  Click the menu item `QA Mode` in the `View` submenu and observe the change of the state to be `true`. Click the menu item again and the state will change to `false`. The state should correspond to the respective icon: if the `qaMode` is `true` the icon should be a checked checkbox, if `false` unchecked.

### DIFF
--- a/src/Frontend/Components/BackendCommunication/BackendCommunication.tsx
+++ b/src/Frontend/Components/BackendCommunication/BackendCommunication.tsx
@@ -18,6 +18,7 @@ import {
   FileSupportPopupArgs,
   IsLoadingArgs,
   ParsedFileContent,
+  QAModeArgs,
 } from '../../../shared/shared-types';
 import { PopupType } from '../../enums/enums';
 import {
@@ -44,6 +45,7 @@ import { getFileWithChildrenCheck } from '../../util/is-file-with-children';
 import {
   openPopup,
   setIsLoading,
+  setQAMode,
 } from '../../state/actions/view-actions/view-actions';
 
 export function BackendCommunication(): ReactElement | null {
@@ -280,6 +282,15 @@ export function BackendCommunication(): ReactElement | null {
     }
   }
 
+  function setQAModeListener(
+    event: IpcRendererEvent,
+    qaModeArgs: QAModeArgs,
+  ): void {
+    if (qaModeArgs) {
+      dispatch(setQAMode(qaModeArgs.qaMode));
+    }
+  }
+
   useIpcRenderer(AllowedFrontendChannels.FileLoaded, fileLoadedListener, [
     dispatch,
   ]);
@@ -342,6 +353,9 @@ export function BackendCommunication(): ReactElement | null {
     showUpdateAppPopupListener,
     [dispatch],
   );
+  useIpcRenderer(AllowedFrontendChannels.SetQAMode, setQAModeListener, [
+    dispatch,
+  ]);
 
   return null;
 }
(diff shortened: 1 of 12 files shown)


src/Frontend/state/actions/view-actions/view-actions.ts
```diff
@@ -110,3 +112,8 @@ export function setShowNoSignalsLocatedMessage(
     payload: showMessage,
   };
 }
+
+export function setQAMode(qaMode: boolean): SetQAMode {
+  console.log('ViewAction' + qaMode);
```
FlorianSchepers, line 117: remove logging

src/Frontend/Components/BackendCommunication/BackendCommunication.tsx
```diff
@@ -280,6 +282,16 @@ export function BackendCommunication(): ReactElement | null {
     }
   }
 
+  function setQAModeListener(
+    event: IpcRendererEvent,
+    qaModeArgs: QAModeArgs,
+  ): void {
+    if (qaModeArgs) {
+      console.log('Listener: ' + qaModeArgs);
```
FlorianSchepers, line 290: remove logging

src/Frontend/state/reducers/view-reducer.ts
```diff
@@ -83,6 +86,12 @@ export function viewState(
         ...state,
         showNoSignalsLocatedMessage: action.payload,
       };
+    case ACTION_SET_QA_MODE:
+      console.log(action.payload);
```
FlorianSchepers, line 90: remove logging

src/Frontend/state/actions/view-actions/types.ts
```diff
@@ -24,7 +26,8 @@ export type ViewAction =
   | OpenPopupAction
   | UpdateActiveFilters
   | SetIsLoadingAction
-  | SetShowNoSignalsLocatedMessage;
+  | SetShowNoSignalsLocatedMessage
+  | SetQAMode;
```
FlorianSchepers, line 30: I am a bit supprised that the naming here is so inconsistent with some actions having "Action" as suffix and some not. Is there any system behind this?

src/Frontend/util/use-ipc-renderer.ts
```diff
@@ -47,14 +48,20 @@ type ShowFileSupportPopupListener = (
...
 type Listener =
   | ResetStateListener
   | SetStateListener
   | LoggingListener
   | ExportFileRequestListener
   | SetBaseURLForRootListener
   | IsLoadingListener
-  | ShowFileSupportPopupListener;
+  | ShowFileSupportPopupListener
+  | SetQAModeListener;
```
FlorianSchepers, line 64: Here it is more consisten compared to the actions above.

FlorianSchepers (changes requested):

src/Frontend/state/actions/view-actions/types.ts
meretp: I don't think that there is a system behind it, but I will add the suffix because I think it makes more sense as it really is an action but I think refactoring the other ones is out of scope of this ticket.

FlorianSchepers (approved):

PR #2103, opened 2023-10-05
Prefer attribution only in QA mode

meretp:
### Summary of changes

With these changes the user will only be able to prefer an attribution if OpossumUI is in QA mode. If OpossumUI is not in QA mode, the button to prefer (or undo prefer if it was preferred before) is hidden.

Note: This PR is based on #2100 and therefore includes two commits, only the second one is relevant for this PR. This PR needs to be rebased once #2100 is merged. 

Preferring an attribution should only be possible in QA mode. 

Besides the unit and integration tests: 
Check out the commit and open `opossum_input.json` from the example-files. Click on the `index.tsx` file in the resource tree and see one attribution with a grey star. If you open the context menu `...` for this attribution you only see the options `Mark for replacement` and `Delete`. Enable QA mode by clicking the menu item below `View` and observe that there is an addtional option `Mark as preferred`. If you prefer this attribution and save it, the option will change to `Unmark as preferred`. If you once again change the mode by clicking the respective menu item, the option will disappear.

### DIFF
--- a/src/Frontend/Components/AttributionColumn/AttributionColumn.tsx
+++ b/src/Frontend/Components/AttributionColumn/AttributionColumn.tsx
@@ -17,7 +17,10 @@ import {
   getTemporaryDisplayPackageInfo,
   wereTemporaryDisplayPackageInfoModified,
 } from '../../state/selectors/all-views-resource-selectors';
-import { getSelectedView } from '../../state/selectors/view-selector';
+import {
+  getQAMode,
+  getSelectedView,
+} from '../../state/selectors/view-selector';
 import {
   getDisplayedPackage,
   getResolvedExternalAttributions,
@@ -114,6 +117,7 @@ export function AttributionColumn(props: AttributionColumnProps): ReactElement {
   const wasPreferredFieldChanged: boolean =
     initialManualDisplayPackageInfo.preferred !==
     temporaryDisplayPackageInfo.preferred;
+  const qaMode = useAppSelector(getQAMode);
 
   const {
     isLicenseTextShown,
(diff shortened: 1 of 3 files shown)


src/Frontend/Components/BackendCommunication/BackendCommunication.tsx
```diff
@@ -280,6 +282,16 @@ export function BackendCommunication(): ReactElement | null {
     }
   }
 
+  function setQAModeListener(
+    event: IpcRendererEvent,
+    qaModeArgs: QAModeArgs,
+  ): void {
+    if (qaModeArgs) {
+      console.log('Listener: ' + qaModeArgs);
```
benedikt-richter, line 290: Before merging into `main` we should remove the `console.log`

src/Frontend/state/actions/view-actions/view-actions.ts
```diff
@@ -110,3 +112,8 @@ export function setShowNoSignalsLocatedMessage(
     payload: showMessage,
   };
 }
+
+export function setQAMode(qaMode: boolean): SetQAMode {
+  console.log('ViewAction' + qaMode);
```
benedikt-richter, line 117: as above

src/Frontend/Components/AttributionColumn/AttributionColumn.tsx
```diff
@@ -153,6 +157,7 @@ export function AttributionColumn(props: AttributionColumnProps): ReactElement {
     isPreferenceFeatureEnabled,
     attributionIsPreferred: temporaryDisplayPackageInfo.preferred ?? false,
     view,
+    qaMode,
```
benedikt-richter, line 160: Instead of adding a new argument for `getMergeButtonsDisplayState`, we could modify line 157 to be `isPreferenceFeatureEnabled && qaMode`

benedikt-richter (approved):
lgtm

Just consider the comments

src/Frontend/Components/BackendCommunication/BackendCommunication.tsx
meretp: Yes, this was part of the previous PR, but I deleted the logging there and rebased this one.

src/Frontend/Components/AttributionColumn/AttributionColumn.tsx
meretp: Nice suggestion, changed it. With this I only need to do small adaptions to the tests which makes the changes much clearer.

PR #2113, opened 2023-10-09
Improve styling in Locator popup

meretp:
### Summary of changes

This diffs adds a border around the search and the associated checkbox to clarify where the checkbox belongs to. The corresponding screenshot in the documentation is also updated.

On this branch we add the search functionality to the Locator Popup, the style changes should clarify where the introduced checkbox belongs to.

Only style changes.


benedikt-richter (approved):
Nice 👍

PR #2124, opened 2023-10-10
Improve typecheck

mstykow:
### Summary of changes

- use TypeScript instead of ESLint to handle unused vars
- prevent implicit overrides and returns
- rename script "compile-all" to "typecheck" to avoid confusion

By tuning the TypeScript config, we can get more mileage out of type checking than we currently do.
One major improvement I have not yet done is to turn on this setting: https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess
However, this would mean needing to fix hundreds of errors.

Run the type checker with `yarn typecheck`.

I recommend commit-wise review. Note that the first commit includes removing unused `React` imports in hundreds of files.


benedikt-richter (approved):
Thanks for cleaning that up!

PR #2126, opened 2023-10-10
Fix bug regarding preferred attributions

davidmundelius:
### Summary of changes

We improve the preformance of collecting the resource ids in a subtree to avoid an error for large files.

When one tries to mark an attribution for a large folder as preferred, previously we got an error "Maximum call stack size exceeded". These improvements avoid this error.

Open a large input file and try to mark an attribution of the root folder as preferred. Then click "save".


meretp (approved):
LGTM

PR #2130, opened 2023-10-10
Improve styling V1: using TSS

mstykow:
### Summary of changes

- add tss-react dependency
- convert App and Spinner to use TSS
- use .styles files to contain style definitions

This is the first version of styling components consistently using [TSS](https://docs.tss-react.dev/).
See the other alternative here: https://github.com/opossum-tool/OpossumUI/pull/2131

Main disadvantage: extra dependency.
Main advantage: very similar to defining `classes` the way we currently do, but with the benefit of being performant, fully typed, and having access to props.

MUI is not meant to be consistently styled using the `sx` prop.

The way we're currently styling  our components using `classes` as an object where we define all CSS classes and then reference them via `sx` has the following disadvantages:
a) negative performance implications
b) untyped
c) we have no access to props inside the FCs

Read more here: https://mui.com/system/getting-started/usage/#when-to-use-mui-system

Do a UI check that `App` and `Spinner` still look correct.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -63,7 +63,7 @@ module.exports = {
     'object-shorthand': 2,
     'react/prop-types': 'off',
     'testing-library/no-node-access': 'off', // TODO: remove this line and fix warnings
-    'filenames/match-regex': [2, '^[a-zA-Z0-9\\-]+(.d|.test)?$', true],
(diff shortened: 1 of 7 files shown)


mstykow:
I can't get this to start up without error. Closing since we have another alternative.

PR #2131, opened 2023-10-10
Improve styling V2: using styled components

mstykow:
### Summary of changes

- convert App and Spinner to use styled components
- use .styles files to contain style definitions

This is the second version of styling components consistently using styled components.
See the other alternative here: https://github.com/opossum-tool/OpossumUI/pull/2130

Main disadvantage: more code by needing to define wrappers around all components.
Main advantage: native styling (no extra dependencies) recommended by MUI with high performance, full typing, access to props.

MUI is not meant to be consistently styled using the `sx` prop.

The way we're currently styling  our components using `classes` as an object where we define all CSS classes and then reference them via `sx` has the following disadvantages:
a) negative performance implications
b) untyped
c) we have no access to props inside the FCs

Read more here: https://mui.com/system/getting-started/usage/#when-to-use-mui-system

Do a UI check that `App` and `Spinner` still look correct.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -63,7 +63,7 @@ module.exports = {
     'object-shorthand': 2,
     'react/prop-types': 'off',
     'testing-library/no-node-access': 'off', // TODO: remove this line and fix warnings
-    'filenames/match-regex': [2, '^[a-zA-Z0-9\\-]+(.d|.test)?$', true],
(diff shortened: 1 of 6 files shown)


leslielazzarino:
This PR looks promising. How long did it take to migrate the two components? I would guess not long at all.

mstykow:
I'll close this until actual work on this may start again.

PR #2133, opened 2023-10-11
Fix console warnings caused by unnecessary rerenders

### LINKED ISSUE #1259: Replace improper selectors with custom hooks
There are selectors, e.g. getAttributionsOfSelectedResourceClosestParent, that create a new object or array every time they are called, causing unnecessary re-renderings. Replace all of them with custom hooks, in order to prevent the performance loss.

mstykow:
### Summary of changes

- memoize selectors to prevent rerenders using [proxy-memoize](https://github.com/dai-shi/proxy-memoize)
- simplify some selectors

Selectors returning mutable data types such as arrays and objects should be memoized in react-redux. Otherwise they cause unnecessary rerenders. Read more here: https://react-redux.js.org/api/hooks#equality-comparisons-and-updates

I also tried using `lodash.memoize` as memoizer, or using `lodash.isEqual`/`react-redux.shallowEqual` as second argument to `useSelector`. Both approaches resulted in more rerender cycles than using `proxy-memoize`.

Closes #1259.

Observe the console while using the app in development mode (`yarn start`) and compare with main. You'll see that the warnings have disappeared.


src/Frontend/util/get-filtered-attributions-by-id.ts
```diff
@@ -1,23 +0,0 @@
...
-  function reducer(
-    filteredAttributions: Attributions,
-    id: string,
-  ): Attributions {
-    filteredAttributions[id] = attributions[id];
-
-    return filteredAttributions;
-  }
-
-  return reduce(ids, reducer, {});
```
mstykow, line 22: all the logic that happened in this file using custom logic can be replaced by lodash's `pick`.

src/Frontend/state/hooks.ts
```diff
@@ -3,10 +3,17 @@
...
+import { useDispatch, useSelector } from 'react-redux';
 import { State } from '../types/types';
 import { AppThunkDispatch } from './types';
 
 export const useAppDispatch = (): AppThunkDispatch =>
   useDispatch<AppThunkDispatch>();
-export const useAppSelector: TypedUseSelectorHook<State> = useSelector;
+export const useAppSelector = <T>(
+  fn: (state: State) => T,
+  deps: unknown[] = [],
```
benedikt-richter, line 16: We typically use the notiation `Array` instead of `[]`. Therefore, I'd prefer `Array<unknown>`

benedikt-richter (approved):
lgtm.

Just a minor comment on conventions.

src/Frontend/state/hooks.ts
mstykow: I'll do this. We should consider adding more linter rules (which can cover what you're describing) so that comments like the one you just made become unnecessary.

PR #2134, opened 2023-10-11
Fix: Context menu does not check for disabled

### LINKED ISSUE #2016: Disabled triple dots in audit view are clickable
In Audit view, when no attribution is selected, the triple dots in the corner are greyed out and should be disabled. However, it is possible to click on them, and this opens up a context menu containing a disabled Undo button.

image
image

mstykow:
### Summary of changes

- add disabled check to ContextMenu
- add new unit tests and fix some integration tests that were failing as a result

Closes #2016.

See task description: you won't be able to reproduce the bug anymore.

### DIFF
--- a/src/Frontend/Components/ContextMenu/__tests__/ContextMenu.test.tsx
+++ b/src/Frontend/Components/ContextMenu/__tests__/ContextMenu.test.tsx
@@ -4,6 +4,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { fireEvent, render, screen } from '@testing-library/react';
+import { noop } from 'lodash';
 import { ButtonText } from '../../../enums/enums';
 import { doNothing } from '../../../util/do-nothing';
 import { ContextMenu, ContextMenuItem } from '../ContextMenu';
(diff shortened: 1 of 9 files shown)


src/Frontend/test-helpers/attribution-column-test-helpers.ts
```diff
@@ -54,10 +54,6 @@ export function expectButtonInHamburgerMenuIsNotShown(
   expect(
     screen.queryByRole('button', { name: buttonLabel }),
   ).not.toBeInTheDocument();
-
-  if (screen.queryByRole('presentation')) {
```
mstykow, line 58: a method named `expectButtonInHamburgerMenuIsNotShown` should not contain unrelated side-effects. these side-effects were causing other tests to fail.

src/Frontend/integration-tests/audit-view-tests/__tests__/audit-view.test.tsx
```diff
@@ -174,6 +174,7 @@ describe('The App in Audit View', () => {
     clickOnValueInManualPackagePanelForParentAttribution(screen, 'React');
     expectValueInTextBox(screen, 'Name', 'React');
 
+    fireEvent.click(screen.getByRole('presentation').firstChild as Element);
```
mstykow, line 177: only place where the side effect from `expectButtonInHamburgerMenuIsNotShown` was actually needed to clear the context menu.

src/Frontend/Components/ContextMenu/__tests__/ContextMenu.test.tsx
```diff
@@ -145,4 +146,58 @@ describe('The ContextMenu', () => {
     fireEvent.click(screen.getByText(ButtonText.Save));
     expect(onCloseMock).toHaveBeenCalledTimes(1);
   });
+
+  it('is disabled when all menu items are disabled', () => {
+    // given
```
leslielazzarino, line 151: As a convention, we do not write given/when/then. Exceptions are possible, but in this case I do no think that they make the tests easier to understand, as they are pretty short and simple already.

src/Frontend/Components/ContextMenu/__tests__/ContextMenu.test.tsx
```diff
@@ -145,4 +146,58 @@ describe('The ContextMenu', () => {
...
+    const menuItem: ContextMenuItem = {
+      buttonText: ButtonText.Undo,
+      disabled: true,
+      onClick: noop,
+      hidden: false,
+    };
+    const trigger = 'Test Element';
```
leslielazzarino, line 158: Why are you calling this "trigger"?

src/Frontend/integration-tests/audit-view-tests/__tests__/audit-view.test.tsx
leslielazzarino: I find this line quite obscure. Can we pack it in a function, whose name explains what we are doing?

leslielazzarino (changes requested):

src/Frontend/integration-tests/audit-view-tests/__tests__/audit-view.test.tsx
mstykow: done

src/Frontend/Components/ContextMenu/__tests__/ContextMenu.test.tsx
mstykow: removed since such annotations are not commonly current practice

src/Frontend/Components/ContextMenu/__tests__/ContextMenu.test.tsx
mstykow: because this is the button label that, when clicked, triggers opening the menu. i find this more helpful as a variable name than "testElementText" which is what the other tests are using.

src/Frontend/test-helpers/context-menu-test-helpers.ts
```diff
@@ -363,6 +363,10 @@ export function clickOnButtonInPackageContextMenu(
   fireEvent.click(button);
 }
 
+export function clearPopover(screen: Screen): void {
```
leslielazzarino, line 366: nice

leslielazzarino (approved):

PR #2145, opened 2023-10-12
Improve search bar

### LINKED ISSUE #1773: Cursor unchanged when hovering "clear search" button
Typically, the curser changes when hovering a button. However, this is currently, not the case for the "clear search" button (the "x" icon that appears after entering a search term):

Screenshot 2023-06-30 at 12 43 48

### LINKED ISSUE #2076: Remove second scroll bar at All Attributions
Open Audit view and open tab "Global" where signals are. This tab has only one category "All Attributions", but when the list is quite long there are two scroll bars, which is quite confusing.

mstykow:
### Summary of changes

- remove default clear icon that comes with input fields of type "search"
- in search textfield, move search icon before cursor and place clear icon with custom hover state at the rear
- always display search bar in resource details tabs
- make tabs in resource details full width 
- move scroll bar to below search bar to make search bar sticky

image
image
image
image

The default clear icon that webkit browsers add by default to text inputs of type "search" is not good enough for us.

About always showing the search bar inside the resource details tabs:
- there is enough space to always show the search
- better UX (fewer clicks)
- makes the code simpler
- avoids the need to treat this search bar differently from others where the spy glass is always shown

Closes #1773.
Closes #2076.

Check out a few places where the search bar is used: resource details tabs, package search popup, file search.

### DIFF
--- a/src/Frontend/Components/ResourceDetailsTabs/ResourceDetailsTabs.tsx
+++ b/src/Frontend/Components/ResourceDetailsTabs/ResourceDetailsTabs.tsx
@@ -3,39 +3,40 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
-import { ReactElement, useEffect, useMemo, useState } from 'react';
-import MuiTabs from '@mui/material/Tabs';
+import MuiBox from '@mui/material/Box';
 import MuiTab from '@mui/material/Tab';
-import { getManualData } from '../../state/selectors/all-views-resource-selectors';
-import { AggregatedAttributionsPanel } from '../AggregatedAttributionsPanel/AggregatedAttributionsPanel';
-import { AllAttributionsPanel } from '../AllAttributionsPanel/AllAttributionsPanel';
+import MuiTabs from '@mui/material/Tabs';
 import { remove } from 'lodash';
+import { ReactElement, useEffect, useState } from 'react';
+import { Attributions } from '../../../shared/shared-types';
+import { PackagePanelTitle } from '../../enums/enums';
+import { OpossumColors } from '../../shared-styles';
+import { setPackageSearchTerm } from '../../state/actions/resource-actions/audit-view-simple-actions';
+import { useAppDispatch, useAppSelector } from '../../state/hooks';
+import { getManualData } from '../../state/selectors/all-views-resource-selectors';
 import {
   getAttributionIdsOfSelectedResource,
   getDisplayedPackage,
-  getIsAccordionSearchFieldDisplayed,
   getPackageSearchTerm,
   getSelectedResourceId,
 } from '../../state/selectors/audit-view-resource-selectors';
-import { OpossumColors } from '../../shared-styles';
-import { useAppDispatch, useAppSelector } from '../../state/hooks';
-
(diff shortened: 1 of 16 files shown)


src/Frontend/Components/SearchTextField/SearchTextField.tsx
```diff
@@ -3,33 +3,42 @@
...
-import { Search } from '@mui/icons-material';
 import { ReactElement } from 'react';
 import { OpossumColors } from '../../shared-styles';
 import { getSxFromPropsAndClasses } from '../../util/get-sx-from-props-and-classes';
 import { inputElementClasses } from '../InputElements/shared';
 
 const classes = {
   searchField: {
-    marginTop: '4px',
+    marginTop: '8px',
```
mstykow, line 17: prevents cutting off the "Search" label, e.g., in package search

src/Frontend/Components/SearchTextField/SearchTextField.tsx
```diff
@@ -3,33 +3,42 @@
...
     marginBottom: '8px',
     '& div': {
       borderRadius: '0px',
     },
+    '& input[type=search]::-webkit-search-cancel-button': { display: 'none' },
```
mstykow, line 22: this removes the built-in "x" icon of input field of type "search"

src/Frontend/Components/ResourceDetailsTabs/ResourceDetailsTabs.tsx
```diff
@@ -113,38 +97,26 @@ export function ResourceDetailsTabs(
...
-    dispatch(setPackageSearchTerm(''));
-  }
-
   function onSearchInputChange(input: string): void {
     dispatch(setPackageSearchTerm(input));
   }
 
   return (
-    <div style={{ position: 'relative' }}>
+    <Box sx={classes.container}>
```
leslielazzarino, line 111: We prefix MUI components with "Mui", to better distinguish them from our custom components.

src/Frontend/Components/SearchTextField/SearchTextField.tsx
leslielazzarino: It would be nice to remove the label all together, as it is already clear from the icon that this is a search field. It would also be nice if the search icon would disapper when the user clicks in the search field or there is text, as the user would gain some space.

leslielazzarino (changes requested):
I believe it looks much better now.

src/Frontend/Components/SearchTextField/SearchTextField.tsx
mstykow: I agree with removing the label. I think the spyglass appearing/disappearing is a matter of taste since either way there is plenty of space. I'd like to keep as is for now because I'm not convinced of the benefit of dynamically displaying this icon while adding this functionality would increase code and UX complexity (I may even be bothered by the jumping cursor that results from the dynamic icon).

leslielazzarino: ok

leslielazzarino (approved):

PR #2183, opened 2023-10-23
Mark located signals in package card

### LINKED ISSUE #2148: Mark located signals in package card
**Context**
When locating signals, we display arrows in the resource browser. This helps the user to find the resource that has signals that are located.
Screenshot 2023-10-13 at 08 36 35

**Problem**
For the user it is, in general, difficult to find the signals that triggered the location (i.e. the signals that e.g. contain the licenses the user searches for).

**Solution**
Add the "locate icon" to the package card of signals that trigger the location to give visual guidance to the user.

related to #1934

meretp:
### Summary of changes

We already implemented the functionality to locate resources which contain signals based on criticality and license names. This commit adds an icon to each signal which matches the selected filter.

Fixes #2148

There are new unit tests, so the CI should be green.
Additionally you can checkout the PR, open the "opossum_input.json" file with OpossumUI and click on the locate Icon next to the Apache license in the project statistics. Click on the resources "index.tsx" which is marked with an arrow and observe the locate icons in all suggested signals like below: 
image

If you open the LocatorPopup again and clear the search, all icons will be gone.

### DIFF
--- a/src/Frontend/Components/Icons/__tests__/Icons.test.tsx
+++ b/src/Frontend/Components/Icons/__tests__/Icons.test.tsx
@@ -14,6 +14,7 @@ import {
   FirstPartyIcon,
   FollowUpIcon,
   IncompleteAttributionsIcon,
+  LocateSignalsIconWithTooltip,
   NeedsReviewIcon,
   PreSelectedIcon,
   SearchPackagesIcon,
@@ -91,6 +92,12 @@ describe('The Icons', () => {
     render(<SearchPackagesIcon />);
 
     expect(screen.getByLabelText('Search packages icon'));
+  });
+
+  it('renders LocateSignalsIconWithTooltip', () => {
+    render(<LocateSignalsIconWithTooltip />);
+
+    expect(screen.getByLabelText('locate signals icon'));
   });
 });
 
--- a/src/Frontend/state/helpers/action-and-reducer-helpers.ts
+++ b/src/Frontend/state/helpers/action-and-reducer-helpers.ts
@@ -29,6 +29,7 @@ import {
   licenseNameContainsSearchTerm,
   packageInfoContainsSearchTerm,
 } from '../../util/search-package-info';
+import { LocatePopupFilters } from '../../types/types';
 
 export function getMatchingAttributionId(
   packageInfoToMatch: PackageInfo,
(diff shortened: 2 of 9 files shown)


src/Frontend/Components/Icons/__tests__/Icons.test.tsx
```diff
@@ -92,6 +93,12 @@ describe('The Icons', () => {
 
     expect(screen.getByLabelText('Search packages icon'));
   });
+
+  it('renders LocateSignalsIconWithTooltip', () => {
+    render(<LocateSignalsIconWithTooltip />);
+
+    expect(screen.getByLabelText('locate signals icon'));
+  });
```
meretp, line 101: Not sure if this test makes sense as it doesn't really test any logic but I wanted to keep it consistent.

src/Frontend/state/helpers/action-and-reducer-helpers.ts
```diff
@@ -331,6 +321,42 @@ export function calculateResourcesWithLocatedAttributions(
...
+  const licenseIsSet = locatePopupFilters.selectedLicenses.size > 0;
+  const criticalityIsSet =
+    locatePopupFilters.selectedCriticality != SelectedCriticality.Any;
+  const searchTermIsSet: boolean = locatePopupFilters.searchTerm != '';
+  return licenseIsSet || criticalityIsSet || searchTermIsSet;
+}
+export function attributionMatchesLocateFilter(
+  attribution: PackageInfo,
+  locatePopupFilter: LocatePopupFilters,
+): boolean {
```
vasily-pozdnyakov, line 336: these two functions have some repeating functionality, can they be somehow refactored to reuse?

vasily-pozdnyakov (changes requested):
well done, just a small refactoring question.

src/Frontend/state/helpers/action-and-reducer-helpers.ts
meretp: I am not sure or at least I couldn't think of a good refactoring.
The first function is used during the calculation of resources with located attributions to early return if no filter is set. If a filter is set we loop through all attributions and compare if they match the filter. 
As we have different filter and we don't know which filter is set we also need these variables `licenseIsSet` and `criticalityIsSet` in the second function. 
The only refactoring I could think of is to implement a function that based on the `LocatePopupFilters` returns some kind of mapping which filter is set. This could then be used in the calculation - if no filter is set -> early return and in the match function as we would only need to check the filter which is set. But I am not sure if this would produce more overhead and lead to less readability of the code.

vasily-pozdnyakov: how about the following logic for the second function:
if no filter is set -> return true
if some filter is set -> check if the attribution matches the filter and return the corresponding value.

meretp: I am not sure if I get your suggestion right, but to check if the attribution matches the filter we would again need to check if the filter is set. The filter for the criticality is set to `SelectedCriticality.ANY` if "not set", so comparing the criticality would lead to `false` although the filter is just not set.

vasily-pozdnyakov: I see, ok. Then keep it as it is.
Please test anyLocateFilterIsSet()

vasily-pozdnyakov (approved):

PR #2194, opened 2023-10-25
Fix marking of located signals

meretp:
### Summary of changes

Mark located signals if the selected license matches either the short name or full name of a frequent license.

#2183 introduced the marking of signals with an icon if they match the filters from the LocatorPopup. If we have `frequentLicenses` in the file and search for a short name of a `frequentLicense` we also want to locate and mark the signals that have the corresponding full name as license name or the other way around. We marked the resources correctly for these cases but didn't mark the corresponding signal, this is fixed with this PR.

See the new test case and the passing CI. Additionally you can checkout this PR, open the "opossum_input.json" file and locate signals with the license name "MIT License" as in the screenshot below:
image

Observe that the corrresponding signal in "index.tsx" is located and marked with an icon eventhough the license name is just "MIT". 
image


FlorianSchepers (approved):
lgtm

PR #2195, opened 2023-10-25
fix: search request sent without search term

mstykow:
### Summary of changes

- the package search request is enabled based on the current search term, however, it is the debounced search term that's sent as a query param
- fix enabled flag to also be based on debounced search term

Fix this console error:

image

Open the console to see that there is no longer a 400 BAD REQUEST when first searching for a package.


alexzurbonsen (approved):
Looks good to me. Thanks for cleaning up the tests.

PR #2204, opened 2023-10-26
feat: eliminate List duplication

mstykow:
### Summary of changes

- merge the two List components into a single component
- delete redundant types

Remove unnecessary code duplication.

Check that all places where virtualized lists are used still look and work correctly. Mostly this affects the resource browser, attribution view and package list.


src/Frontend/Components/List/List.tsx
```diff
@@ -9,54 +9,66 @@ import { Height, NumberOfDisplayedItems } from '../../types/types';
...
-
-export function List(props: ListProps): ReactElement {
-  const defaultCardHeight = 24;
-  const cardHeight = props.cardVerticalDistance || defaultCardHeight;
-  const maxHeight = maxHeightWasGiven(props.max)
-    ? props.max.height
-    : props.max.numberOfDisplayedItems * cardHeight;
-  const currentHeight = props.length * cardHeight;
-  const listHeight = props.alwaysShowHorizontalScrollBar
+export const List: React.FC<Props> = ({
```
FlorianSchepers, line 33: Why did you switch to an arrow function here? We typically use regular functions for components.

If we should decide that arrow function have some advantage over "regular" functions we should discuss if we want to adapt it everywhere. Otherwise i would preferre to stay consistent.

FlorianSchepers (changes requested):

src/Frontend/Components/List/List.tsx
mstykow: By using arrow functions for components, I'm able to use the `React.FC` type which results in

```const Foo: React.FC<Props> = (props) => {```

This is more concise than writing

```function Foo(props: Props): React.ReactElement {```

I personally don't feel like there is a need to force one way or the other as it's just personal preference.

leslielazzarino: So far we have enforced consistency there, as it makes life easier for people new to React and TS. We can discuss it with the team, personally I would still enforce consistency, possibly switching to arrow functions if they turn out to be better. But the improved conciseness isn't much, and I havent checked recently but arrow functions used to have a much worse stack trace.

FlorianSchepers (approved):
I checked and the virtualized lists still work as expected also for a bigger project.

lgtm

PR #2206, opened 2023-10-26
fix: package card padding missing

mstykow:
### Summary of changes

- add a small left padding when there are no icons to the left of the text on a package card

BEFORE:
image

AFTER:
image

WITH SECOND LINE TEXT:
image

UI bugfix.

Check package cards with and without icons to the left.

### DIFF
--- a/src/Frontend/Components/ListCard/ListCard.tsx
+++ b/src/Frontend/Components/ListCard/ListCard.tsx
@@ -164,7 +164,7 @@ interface ListCardProps {
 }
 
 export function ListCard(props: ListCardProps): ReactElement | null {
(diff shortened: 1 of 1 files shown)


lucavb (approved):

meretp:
Looks good, but I think we should also include a padding if we have an additional copyright to make it consistent, see the example below:
image

mstykow:
> Looks good, but I think we should also include a padding if we have an additional copyright to make it consistent, see the example below: image

Thanks for the catch! Fixed.

meretp (approved):
Nice improvement!

Issue #971, opened 2022-09-12
Enable keyboard navigation in package lists


benedikt-richter:
**Is your feature request related to a problem? Please describe.**
Currently, navigating the package lists, a user has to click each package individually in order to traverse the list.
For the user that implies the need to click a lot.

Navigation:
- simple up-down navigation with arrows
- experiment if it's better to select the item with space or if an item should be immediately selected after the navigation event

**Describe the solution you'd like**
As a user navigating a package list, I want to be able to use keyboard navigation.

someshkhandelia:
As per verbal discussion, it is assumed that the user first clicks (selects) atleast one item in the given tree / list and then there is the possibility to navigate using keyboard.

someshkhandelia:
So far, could find two ways to do this:
1. Revamp entire architecture and use an actual Tree instead of a VirtualizedList, https://mui.com/material-ui/react-tree-view/
2. Create a custom hook for keyboard input for the existing VirtualizedList, https://whereisthemouse.com/create-a-list-component-with-keyboard-navigation-in-react

mstykow:
As of #2208 we can implement this feature pretty directly by following these instructions: https://virtuoso.dev/keyboard-navigation/

PR #2208, opened 2023-10-27
fix: issues with multiple vertical scroll bars

### LINKED ISSUE #2076: Remove second scroll bar at All Attributions

mstykow:
### Summary of changes

- replace react-window with virtuoso
- virtuoso comes with built-in full-height sizing which obviates the need for complicated height calculations based on window height and paddings in multiple components
- removing the height calculations and instead using flex-box design with flex grow factor removes the vertical scroll bar issues

Closes #2076.
Also preparation for #971 since virtuoso supports keyboard navigation natively: https://virtuoso.dev/keyboard-navigation/

These videos demo that first of all no existing functionality has been broken, and second, that small UI bugs such as the doubled vertical scroll bars have been fixed.

BEFORE is on the left, AFTER is on the right of the screen.

[Untitled_ Oct 30 2023 11_17 AM.webm](https://github.com/opossum-tool/OpossumUI/assets/46091730/384a119c-a978-4f5b-a02a-7847979c691b)

[Untitled_ Oct 30 2023 11_21 AM.webm](https://github.com/opossum-tool/OpossumUI/assets/46091730/107d43c5-e48e-4d48-8c36-6e51e2d2b39a)


src/Frontend/Components/ReportView/ReportView.tsx
```diff
@@ -62,12 +62,12 @@ export function ReportView(): ReactElement {
         isFileWithChildren={isFileWithChildren}
         onIconClick={getOnIconClick()}
         topElement={
-          <>
+          <div>
```
mstykow, line 65: `div` now necessary because of the parent's flex-box behavior.

src/Frontend/Components/SearchTextField/SearchTextField.tsx
```diff
@@ -13,7 +13,7 @@ import { inputElementClasses } from '../InputElements/shared';
 
 const classes = {
   searchField: {
-    marginTop: '4px',
+    marginTop: '8px',
```
mstykow, line 16: necessary to prevent cutting off the label that moves to above the input field when user is entering text.

src/Frontend/util/has-attribution-multiple-resources.ts
```diff
@@ -10,7 +10,7 @@ export function hasAttributionMultipleResources(
   attributionsToResources: AttributionsToResources,
 ): boolean {
   if (attributionId) {
-    return attributionsToResources[attributionId].length > 1;
+    return (attributionsToResources[attributionId] || []).length > 1;
```
mstykow, line 13: notice that index access can result in `undefined` so that accessing `length` may result in a null pointer exception. there is a [typescript rule](https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess) to prevent this error but turning it on right now would result in too many places to fix.

src/Frontend/Components/FileSearchPopup/FileSearchPopup.tsx
```diff
@@ -9,17 +9,11 @@ import { closePopup } from '../../state/actions/view-actions/view-actions';
...
 import { useAppDispatch } from '../../state/hooks';
 
 export function FileSearchPopup(): ReactElement {
   const dispatch = useAppDispatch();
   const [filteredPaths, setFilteredPaths] = useState<Array<string>>([]);
-  const fileSearchPopupOffset = 260;
-  const maxPossibleHeightInPx = useWindowHeight() - fileSearchPopupOffset;
-  const resourceListMaxHeightInPx =
-    // eslint-disable-next-line @typescript-eslint/no-magic-numbers
-    500 > maxPossibleHeightInPx ? maxPossibleHeightInPx : 500;
```
mstykow, line 22: example of complicated height calculations that are no longer necessary.

src/Frontend/extracted/VirtualisedTree/VirtualizedTree.tsx
```diff
@@ -66,53 +52,38 @@ export function VirtualizedTree(
...
   );
 
-  const maxListLength: NumberOfDisplayedItems | Height = props.maxHeight
-    ? { height: props.maxHeight }
-    : {
-        numberOfDisplayedItems: min([
-          treeNodeProps.length,
-          DEFAULT_MAX_TREE_DISPLAYED_NODES,
-        ]) as number,
-      };
```
mstykow, line 82: another example of a complicated height calculation that is no longer necessary.

src/Frontend/Components/ResourceDetailsTabs/ResourceDetailsTabs.tsx
```diff
@@ -120,14 +121,6 @@ export function ResourceDetailsTabs(
 
   const isAddToPackageEnabled: boolean =
     props.isGlobalTabEnabled && props.isAddToPackageEnabled;
-  const aggregatedAttributionsPanel = useMemo(
-    () => (
-      <AggregatedAttributionsPanel
-        isAddToPackageEnabled={isAddToPackageEnabled}
```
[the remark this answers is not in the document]
leslielazzarino: Unless they changed something recently, if a component is re-rendered in React its children are re-rendered too. That was the purpose of this useMemo. We need to double-check that it is not needed any longer.

src/Frontend/extracted/VirtualisedTree/VirtualizedTree.tsx
leslielazzarino: cool

src/Frontend/util/has-attribution-multiple-resources.ts
leslielazzarino: This place in the code was discussed a month ago or so. It was decided not to add this fix as it would only be needed in the case of a broken input file. Personally I'm not against it, but it should be redicussed before changing it.

leslielazzarino (changes requested):

src/Frontend/Components/ResourceDetailsTabs/ResourceDetailsTabs.tsx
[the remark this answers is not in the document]
mstykow: Thanks for your comment @leslielazzarino. I was wrong and you're right that the child components will always re-render when the parent changes, even if the prop value is still the same. I have two points about this though:
1. For the purpose of memoizing components, React introduced `React.memo`. `React.useMemo` is meant for constants (objects, arrays, ...). So I suggest that we change to `React.memo` if performance is an issue here. Which brings me to the second point:
2. Have you tested that this memoization really leads to a performance optimization? I didn't notice any issues when I removed it. I mean re-rendering is a normal part of React and we could literally go as far as wrapping all components, constants, functions in memoization hooks. See here for example: https://attardi.org/why-we-memo-all-the-things/ In my opinion, doing so would at least make the code generally harder to read.

src/Frontend/util/has-attribution-multiple-resources.ts
mstykow: Discussed with Leslie: we agreed to keep this change because without it some tests are failing. In addition, it may be a good idea in general to turn on the rule referenced above in a separate PR.

src/Frontend/Components/ResourceDetailsTabs/ResourceDetailsTabs.tsx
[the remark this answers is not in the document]
mstykow: I added `React.memo` to the aggregated attributions panel. We could check in a separate PR if other components could benefit from this, too.

src/Frontend/Components/AttributionView/AttributionView.tsx
```diff
@@ -62,25 +63,19 @@ export function AttributionView(): ReactElement {
     );
 
   function onCardClick(packageCardId: string): void {
-    // In AttributionView, attribtionIds still serve as packageCardIds
+    // In AttributionView, attributionIds still serve as packageCardIds
```
MarkusObendrauf, line 66: lol nice

src/Frontend/Components/NotificationPopup/NotificationPopup.tsx
```diff
@@ -64,21 +58,22 @@ export function NotificationPopup(props: NotificationPopupProps): ReactElement {
...
         {props.header}
       </MuiDialogTitle>
       <MuiDialogContent
-        sx={getSxFromPropsAndClasses({
-          sxProps: props.contentSx,
-          styleClass: classes.dialogContent,
-        })}
+        sx={{
+          ...props.contentSx,
+          ...(props.fullHeight && { display: 'flex', flexDirection: 'column' }),
```
MarkusObendrauf, line 75: we usually dont inline styles, could you extract it to something like a `fullHeightDialogContent` class?

MarkusObendrauf (changes requested):
Nice! Im definitely a fan of using Virtuoso here, it simplifies a lot of our styling code. couldnt spot any bugs when i tried it.

i just have a few comments. im especially not sure about how we are mixing two different styling conventions in this PR.

src/Frontend/integration-tests/all-views-tests/__tests-ci__/context-menu-all-views.test.tsx
```diff
@@ -363,6 +364,7 @@ describe('The ContextMenu', () => {
 
     expectPackageInPackagePanel(screen, 'JQuery, 16.5.0', 'Signals');
 
+    closePopup(screen);
```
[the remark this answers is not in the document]
mstykow: this test was failing after my changes. tbh, i'm not sure how the test succeeded before because as far as i understand the opened "resources popup" should be blocking the UI. but i suspect that through react-testing-library it is possible to access nodes in the DOM which are visually obscured by a modal with a higher z-index. either way, it seems correct to me that the popup should be closed after interactions/assertions with it and apparently this also fixes the tests.

from my conversations with other maintainers, it seems that the existing so-called integration tests have various issues that rear their heads when refactorings are performed. in my opinion, these integration tests are one of the biggest tech debts we currently have in the project:
- they are slow (in total they run more than 20 mins here on GitHub)
- they are fragile (subject to easily be broken by internal refactorings)
- they are hard to understand because of numerous abstractions layered on top of react-testing-library which, again in my opinion, is an anti-pattern

src/Frontend/Components/ResizableBox/ResizableBox.tsx
```diff
@@ -3,13 +3,11 @@
...
 // SPDX-License-Identifier: Apache-2.0
 
-import MuiBox from '@mui/material/Box';
-import { SxProps } from '@mui/system';
 import { Resizable, ResizableProps } from 're-resizable';
 
 interface Props extends ResizableProps {
   children: React.ReactNode;
-  sx?: SxProps;
+  style?: React.CSSProperties;
```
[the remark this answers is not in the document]
mstykow: i experimented a bit more and it turns out that `Resizable` already has another `sx` prop with a typing that's different from the one MUI would expect. also, using this prop does not produce the desired CSS effects. so rather than complicating things, i suggest we live with the fact that non-MUI components may use a styling system upon which the current `sx` styling system cannot be forced.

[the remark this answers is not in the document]
mstykow: i found a way to revert everything back to using `sx` by casting the type in the `ResizableBox` component. although i don't think this is the best approach, it's at least consistent with how styling in this app is done in general.

src/Frontend/Components/NotificationPopup/NotificationPopup.tsx
mstykow: done

src/Frontend/Components/PackageList/AttributionsViewPackageList.tsx
```diff
@@ -38,16 +40,16 @@ export function AttributionsViewPackageList(
...
+    <MuiBox sx={classes.container}>
       <SearchTextField onInputChange={setSearch} search={search} />
       <List
         getListItem={(index: number): ReactElement | null =>
           props.getAttributionCard(filteredAndSortedPackageCardIds[index])
         }
-        max={props.max}
         length={filteredAndSortedPackageCardIds.length}
-        cardVerticalDistance={CARD_VERTICAL_DISTANCE}
+        cardHeight={41}
```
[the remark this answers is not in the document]
mstykow: done

src/Frontend/Components/ResizableBox/ResizableBox.tsx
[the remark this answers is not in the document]
MarkusObendrauf: nice! i understand that the approach isnt the best, but for now i think this is the right choice.

MarkusObendrauf (approved):
lgtm 👍

Issue #2213, opened 2023-10-31
Improve PURL quality


mstykow:
**Motivation:**
Typically the data quality of attributions is far from perfect. This, in particular, also applies to the PURL that is an identifier for a package.

Actual: pkg:generic/boost
Expected: pkg:github/boostorg/boost@1.83.0

**Goal:**
Improve the average quality and completeness of PURL data produced by OpossumUI.

**Ideas:**
- Provide validated PURLs in OpossumUI: create database containing data of valid PURLs and, using this database, enable OpossumUI help the user choose valid and complete PURLs
- Highlight PURLs that are missing namespaces (currently there is only generic syntax validation)
- Improve the way users add or edit attributions (currently, individual input fields are disabled when a PURL is present which complicates the UX since we then require the user to understand the PURL syntax)
- Improve the package search (currently poor results), possibly by extending improving the "attribution wizard"

mstykow:
Closing in favor of #2333 #2332 #2334 .

PR #2214, opened 2023-10-31
feat: highlight missing namespaces

mstykow:
### Summary of changes

- explicitly display package type and namespace in attribution details view
- highlight PURLs whose namespace is required but missing
- create dedicated file for storing user-facing texts and extract package sub-panel labels there
- create hook to update temporarily displayed package info to avoid prop-drilling redux actions through several levels of components

image

Progress on #2213.

This is only the first step in a line work that should lead to better UX and higher PURL quality.

### DIFF
--- a/src/Frontend/Components/AttributionColumn/PackageSubPanel.tsx
+++ b/src/Frontend/Components/AttributionColumn/PackageSubPanel.tsx
@@ -6,13 +6,18 @@
 import OpenInNewIcon from '@mui/icons-material/OpenInNew';
 import MuiBox from '@mui/material/Box';
 import MuiPaper from '@mui/material/Paper';
-import { ChangeEvent, ReactElement } from 'react';
+import { ReactElement } from 'react';
 
 import { DisplayPackageInfo } from '../../../shared/shared-types';
+import { text } from '../../../shared/text';
 import { HighlightingColor } from '../../enums/enums';
 import { clickableIcon, disabledIcon } from '../../shared-styles';
-import { isImportantAttributionInformationMissing } from '../../util/is-important-attribution-information-missing';
+import {
+  isImportantAttributionInformationMissing,
+  isNamespaceRequiredButMissing,
+} from '../../util/is-important-attribution-information-missing';
 import { openUrl } from '../../util/open-url';
+import { usePackageInfoChangeHandler } from '../../util/use-package-info-change-handler';
 import { FetchLicenseInformationButton } from '../FetchLicenseInformationButton/FetchLicenseInformationButton';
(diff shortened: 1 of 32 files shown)


src/Frontend/util/is-important-attribution-information-missing.ts
```diff
@@ -38,6 +38,7 @@ export function isImportantAttributionInformationMissing(
     case 'copyright':
     case 'licenseName':
     case 'packageName':
+    case 'packageType':
```
leslielazzarino, line 41: Are we sure about this? There will be a ton more orange in the app with this change.

src/shared/text.ts
```diff
@@ -0,0 +1,20 @@
...
+      packageName: 'Package Name',
+      packageVersion: 'Package Version',
+      purl: 'PURL',
+      repositoryUrl: 'Repository URL',
+      openLinkInBrowser: 'Open link in browser',
+      noLinkToOpen: 'No link to open. Please enter a URL.',
+      searchForPackage: 'Search for package information',
+    },
+  },
+};
```
leslielazzarino, line 20: cool

leslielazzarino (changes requested):
A second and last point: I'm not sure about the usability of the feature. Sure, we are not exposing the appendix, but the purl is much easier to edit manually through the part fields (e.g. to update the version). Disabling them might not be the best for the users.

src/Frontend/Components/AttributionColumn/PackageSubPanel.tsx
```diff
@@ -36,75 +38,131 @@ interface PackageSubPanelProps {
...
+      <MuiBox sx={attributionColumnClasses.displayRow}>
+        {renderPackageName()}
+        {renderPackageVersion()}
+      </MuiBox>
+      {renderPurl()}
+      {renderRepositoryUrl()}
+    </MuiPaper>
+  );
+
+  function renderPackageType(): React.ReactElement {
```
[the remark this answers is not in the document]
mstykow: yes, i always do it like this because then i can read the component like a book from top to bottom. instead of down, up, and then down again.

src/Frontend/util/is-important-attribution-information-missing.ts
mstykow: can you explain more what you mean by "orange"?

leslielazzarino: Incomplete attributions are marked with an orange wedge. After checking a real case scenario, it turn out that your change would not make much of a difference here.

leslielazzarino (approved):

PR #2215, opened 2023-10-31
feat: improve/speed up staged file linting

mstykow:
### Summary of changes

- run linting on staged files in parallel
- improve scoping to only perform tasks that relate to staged files
- perform `reuse lint` on all staged files, not just ts/tsx

The previous configuration led to a slower than necessary pre-commit hook.

Change some files and measure the time it takes to commit your changes.

In my tests, I was able to **reduce the pre-commit hook time from 13s to 5s** when comparing this branch with main.


.lintstagedrc.js
```diff
@@ -0,0 +1,12 @@
+// SPDX-FileCopyrightText: Meta Platforms, Inc. and its affiliates
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+
+module.exports = {
+  '*': [() => 'yarn copyright-lint-check', 'prettier --write --ignore-unknown'],
+  '*.{ts,tsx}': 'eslint -c .eslintrc.js',
```
alexzurbonsen, line 8: I noticed that the `--fix` is not there anymore. Couldn't find it in the config either. Is that on purpose?

mstykow: one can argue about this, but i would prefer if we fix linter errors manually rather than magically.
also, since prettier already makes edits to files concurrently to running the linter, there is a risk of a race condition if we were to allow the linter to also edit files.
all in all, i think it's quite alright to only allow the formatter to make file edits during the hook.

alexzurbonsen: Thanks for explaining. Sounds reasonable to me.

alexzurbonsen (approved):

PR #2222, opened 2023-10-31
feat: re-enable E2E tests

### LINKED ISSUE #1977: Re-enable E2E tests in CI
We disabled E2E tests in the CI as these became flaky. The PR that disabled these is  #1976.

An example of a red CI job is https://github.com/opossum-tool/OpossumUI/actions/runs/6182398024/job/16782051056.

The goal of this issue is to fix the underlying cause and to re-enable the E2E tests in the CI.

mstykow:
### Summary of changes

- create proper playwright config
- introduce fixtures to simplify test setup
- rewrite tests with modern syntax
- simplify package.json script to run E2E tests
- consolidate vite configs using conditional modes
- handle devtools using command line arg instead of env var
- reenable E2E tests in pipeline
- allow workflow dispatch for pipelines

Closes #1977 and improves setup and code quality of E2E tests.

Run `yarn test:e2e` locally and also watch the pipeline output in CI/CD.


src/ElectronBackend/main/main.ts
```diff
@@ -69,12 +60,6 @@ export async function main(): Promise<void> {
     ipcMain.handle(IpcChannel.ExportFile, getExportFileListener(mainWindow));
     ipcMain.handle(IpcChannel.OpenLink, getOpenLinkListener());
 
-    app.on('activate', function () {
-      if (BrowserWindow.getAllWindows().length === 0) {
```
mstykow, line 73: not needed since we always shut down the app as soon as there are 0 windows.

src/ElectronBackend/app.ts
```diff
@@ -3,11 +3,25 @@
...
 // SPDX-License-Identifier: Apache-2.0
 import { app } from 'electron';
+import installExtension, {
+  REACT_DEVELOPER_TOOLS,
+  REDUX_DEVTOOLS,
+} from 'electron-devtools-installer';
 
 import { main } from './main/main';
 
-app.whenReady().then(main);
```
mstykow, line 9: `.whenReady` returns a promise. so technically we'd need to await it or void it. using `.on('ready')` is equivalent but  is an event handler.

package.json
```diff
@@ -98,22 +99,15 @@
...
     "test:local": "jest --testPathIgnorePatterns=src/e2e-tests --testMatch=[ \"**/__tests__/**/*.ts?(x)\" ]",
     "test:changed": "jest --onlyChanged",
     "test:all": "jest --testPathIgnorePatterns=src/e2e-tests --testMatch=[ \"**/__(tests|tests-ci)__/**/*.ts?(x)\", \"**/?(*.)+(test).ts?(x)\" ] && yarn test:e2e",
     "test:integration-ci": "jest src/Frontend/integration-tests --testMatch=[ \"**/__(tests|tests-ci)__/**/*.ts?(x)\", \"**/?(*.)+(test).ts?(x)\" ]",
-    "test:e2e": "run-script-os",
-    "test:e2e:darwin:linux": "start-server-and-test \"yarn build:dev; BROWSER=none vite --config src/e2e-tests/vite-e2e-config.ts\" \"http://localhost:5173/\" \"DISPLAY=\"${DISPLAY:-:99}\" RUNNING_IN_E2E_TEST=true npx playwright test src/e2e-tests\"",
-    "test:e2e:win32": "start-server-and-test \"yarn build:dev & cross-env BROWSER=none vite --config src/e2e-tests/vite-e2e-config.ts\" \"http://localhost:5173/\" \"cross-env RUNNING_IN_E2E_TEST=true DISPLAY=:99 npx playwright test src/e2e-tests\"",
+    "test:e2e": "start-server-and-test \"tsc -p src/ElectronBackend && vite -m e2e\" \"http://localhost:5173/\" \"yarn playwright test -c src/e2e-tests/playwright.config.ts\"",
```
mstykow, line 110: notice that we don't need to build the application using `yarn vite`. doing so was unnecessarily slowing down the test startup time.

src/ElectronBackend/main/__tests__/mainErrorCase.test.ts
```diff
@@ -44,9 +40,6 @@ jest.mock('electron', () => ({
 }));
 describe('The App backend', () => {
   it('handles errors', async () => {
-    // we expect warnings that we do not want to see
-    jest.spyOn(console, 'log');
```
mstykow, line 48: just spying on `console` does nothing at all.

.eslintrc.js
```diff
@@ -64,6 +65,7 @@ module.exports = {
     'object-shorthand': 2,
     'react/prop-types': 'off',
     'testing-library/no-node-access': 'off', // TODO: remove this line and fix warnings
+    'testing-library/prefer-screen-queries': 'off',
```
mstykow, line 68: playwright uses RTL-like syntax which triggers this linter rule but cannot be fixed.
also, i think there is nothing wrong with using non-screen queries, even in RTL.
i've usually turned this rule off in all my other projects as well.

leslielazzarino (changes requested):
Great PR, I would just avoid the lint rule change. Btw, do you know why playwright hangs 20s (on my Windows sistems) after all tests are finished, before exiting?

.eslintrc.js
mstykow: i'll keep the rule on for now and use an override instead, as this discussion is not material to my PR.

leslielazzarino (approved):

PR #2233, opened 2023-11-02
Modify statistics popup

### LINKED ISSUE #2146: Incorrect data in "critical signals" pie chart in statistics popup
**Context**

It seems that there is some discrepancy between the number of highly critical licenses shown in the "Critical Signals" piechart in Statistics Popup and in the “Criticality Bar” shown in the top bar.

**How to reproduce**

* Open `output_opossum.json` example file
* checkout the number of highly critical signals in the "Critical Signals" piechart in the Statistics Popup (0)
* navigate to audit view and switch to the criticality bar
* checkout the number of highly critical signals in the criticality bar (311)

In this example it can be seen that there are no highly critical signals in the piechart while there are highly critical signals in the criticality bar:

Screenshot 2023-10-11 at 18 12 51
Screenshot 2023-10-11 at 18 13 54

**Definition of Done**

The cause for this difference is identified and if necessary fixed.

vasily-pozdnyakov:
### Summary of changes

* Merged signals should be counted as one item at the project statistics popup
* Use "Signals Count" instead of "Count"
* Rename "Critical signals" to "Signals by criticality" and the pie chart legend
* Modify the heuristics to take the highest criticality at the statistics when there are license variants with different criticalities (here we take the risk of false-critical signals). At the end, colors at the "Critical signals" pie chart should agree with colors at Criticality Progress bar

Fix #2146

See unit tests.
Open any example file and see the changes at Project Statistics popup. (see opossum_input file below)

**Before**
image

**After**
image

### DIFF
--- a/src/Frontend/Components/AccordionWithPieChart/__tests__/accordion-with-pie-chart-helpers.test.tsx
+++ b/src/Frontend/Components/AccordionWithPieChart/__tests__/accordion-with-pie-chart-helpers.test.tsx
@@ -2,7 +2,10 @@
 // SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
 //
 // SPDX-License-Identifier: Apache-2.0
-import { ProjectStatisticsPopupTitle } from '../../../enums/enums';
+import {
+  PieChartCriticalityNames,
+  ProjectStatisticsPopupTitle,
+} from '../../../enums/enums';
 import { OpossumColors } from '../../../shared-styles';
 import { PieChartData } from '../../../types/types';
 import { getColorsForPieChart } from '../AccordionWithPieChart';
@@ -16,15 +19,15 @@ describe('getColorsForPieChart', () => {
     ];
     const criticalSignalsCount: Array<PieChartData> = [
       {
-        name: 'High',
+        name: PieChartCriticalityNames.HighCriticality,
         count: 3,
       },
       {
-        name: 'Medium',
+        name: PieChartCriticalityNames.MediumCriticality,
         count: 4,
       },
       {
-        name: 'Not critical',
+        name: PieChartCriticalityNames.NoCriticality,
         count: 2,
       },
     ];
--- a/src/Frontend/Components/CriticalLicensesTable/CriticalLicensesTable.tsx
+++ b/src/Frontend/Components/CriticalLicensesTable/CriticalLicensesTable.tsx
@@ -16,7 +16,7 @@ import { LocateSignalsIcon } from '../Icons/Icons';
 import { ProjectLicensesTable } from '../ProjectLicensesTable/ProjectLicensesTable';
 
 const LICENSE_COLUMN_NAME_IN_TABLE = 'License name';
-const COUNT_COLUMN_NAME_IN_TABLE = 'Count';
+const COUNT_COLUMN_NAME_IN_TABLE = 'Signals Count';
 const FOOTER_TITLE = 'Total';
 const TABLE_COLUMN_NAMES = [
   LICENSE_COLUMN_NAME_IN_TABLE,
--- a/src/Frontend/Components/ProjectStatisticsPopup/__tests__/ProjectStatisticsPopup.test.tsx
+++ b/src/Frontend/Components/ProjectStatisticsPopup/__tests__/ProjectStatisticsPopup.test.tsx
@@ -200,9 +200,21 @@ describe('The ProjectStatisticsPopup', () => {
 
     renderComponentWithStore(<ProjectStatisticsPopup />, { store });
 
-    expect(screen.getByText('Most Frequent Licenses')).toBeInTheDocument();
-    expect(screen.getByText('Critical Signals')).toBeInTheDocument();
-    expect(screen.getAllByText('Incomplete Attributions')).toHaveLength(2);
+    expect(
+      screen.getByText(
+        ProjectStatisticsPopupTitle.MostFrequentLicenseCountPieChart,
+      ),
+    ).toBeInTheDocument();
+    expect(
+      screen.getByText(
+        ProjectStatisticsPopupTitle.CriticalSignalsCountPieChart,
+      ),
+    ).toBeInTheDocument();
+    expect(
+      screen.getAllByText(
+        ProjectStatisticsPopupTitle.IncompleteLicensesPieChart,
+      ),
+    ).toHaveLength(2);
   });
 
   it('does not render pie charts when there are no attributions', () => {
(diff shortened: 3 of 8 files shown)


src/Frontend/Components/CriticalLicensesTable/CriticalLicensesTable.tsx
```diff
@@ -17,11 +17,16 @@ import { ProjectLicensesTable } from '../ProjectLicensesTable/ProjectLicensesTab
 
 const LICENSE_COLUMN_NAME_IN_TABLE = 'License name';
 const COUNT_COLUMN_NAME_IN_TABLE = 'Count';
```
mstykow, line 19: what is the reason we need to keep the old "Count" column name? it seems that if you also adjust line 113 to use the new value, there is no need to introduce two different names for the same column.

src/Frontend/Components/AccordionWithPieChart/__tests__/accordion-with-pie-chart-helpers.test.tsx
```diff
@@ -16,11 +16,11 @@ describe('getColorsForPieChart', () => {
     ];
     const criticalSignalsCount: Array<PieChartData> = [
       {
-        name: 'High',
+        name: 'Highly critical signals',
```
mstykow, line 19: please don't repeat user facing text in tests. instead, reference the enum you created.

src/Frontend/Components/ProjectStatisticsPopup/__tests__/ProjectStatisticsPopup.test.tsx
```diff
@@ -201,7 +201,7 @@ describe('The ProjectStatisticsPopup', () => {
     renderComponentWithStore(<ProjectStatisticsPopup />, { store });
 
     expect(screen.getByText('Most Frequent Licenses')).toBeInTheDocument();
-    expect(screen.getByText('Critical Signals')).toBeInTheDocument();
+    expect(screen.getByText('Signals by Criticality')).toBeInTheDocument();
```
mstykow, line 204: here too you can use the enum

src/Frontend/Components/ProjectStatisticsPopup/__tests__/project-statistics-popup-helpers.test.tsx
```diff
@@ -198,6 +213,16 @@ describe('getLicenseCriticality', () => {
 
     expect(licenseCriticality).toEqual(expectedLicenseCriticality);
   });
+
+  it('obtains the highest high criticality', () => {
```
mstykow, line 217: i don't see how this test is any different from the one that's called "obtains high criticality". i don't really see you testing something new and so i suggest you revert adding this test.

src/Frontend/Components/ProjectStatisticsPopup/project-statistics-popup-helpers.ts
```diff
@@ -173,28 +174,38 @@ export function getLicenseCriticality(licenseCriticalityCounts: {
...
       if (!uniqueLicenseNameToAttributions[strippedLicenseName]) {
         uniqueLicenseNameToAttributions[strippedLicenseName] = [];
       }
-      uniqueLicenseNameToAttributions[strippedLicenseName].push(attributionId);
+
+      if (!usedHashes.has(hash)) {
+        uniqueLicenseNameToAttributions[strippedLicenseName].push(
+          attributionId,
+        );
+        if (hash) {
```
mstykow, line 205: typescript says that `hash` is always a string because we haven't turned on the rule [noUncheckedIndexAccess](https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess) (this is a tech debt IMHO).

i think you're performing this check not because you're concerned about the empty string but because you know that `hash` could be undefined, correct?

but then it seems you should perform this null check before line 201 where you use `useHashes.has(hash)` which even has a risk of a null-pointer exception because `has` requires `hash` to always be a string.

mstykow (changes requested):

src/Frontend/Components/CriticalLicensesTable/CriticalLicensesTable.tsx
vasily-pozdnyakov: good point, you are right.

src/Frontend/Components/ProjectStatisticsPopup/__tests__/project-statistics-popup-helpers.test.tsx
vasily-pozdnyakov: It was slightly different, but I get your point. Will delete it and slightly change "obtains high criticality" test.

src/Frontend/Components/ProjectStatisticsPopup/project-statistics-popup-helpers.ts
vasily-pozdnyakov: I made a refactoring here.

src/Frontend/Components/ProjectStatisticsPopup/project-statistics-popup-helpers.ts
```diff
@@ -173,27 +174,36 @@ export function getLicenseCriticality(licenseCriticalityCounts: {
...
-
     if (licenseName) {
       const strippedLicenseName = getStrippedLicenseName(licenseName);
-
       if (!uniqueLicenseNameToAttributions[strippedLicenseName]) {
         uniqueLicenseNameToAttributions[strippedLicenseName] = [];
       }
+
+      // We do not take into account duplicating (a.k.a merged) signals
+      if (attributionId in externalAttributionsToHashes) {
```
mstykow, line 199: this seen-check can and probably should be moved to the beginning of the for-loop.

src/Frontend/Components/ProjectStatisticsPopup/project-statistics-popup-helpers.ts
```diff
@@ -173,27 +174,36 @@ export function getLicenseCriticality(licenseCriticalityCounts: {
...
+    ? Criticality.Medium
     : undefined;
 }
 
 export function getUniqueLicenseNameToAttribution(
   attributions: Attributions,
+  externalAttributionsToHashes: AttributionsToHashes,
 ): UniqueLicenseNameToAttributions {
   const uniqueLicenseNameToAttributions: UniqueLicenseNameToAttributions = {};
+  const usedHashes: Set<string> = new Set<string>();
```
mstykow, line 189: unnecessary type declaration

mstykow (approved):

PR #2247, opened 2023-11-08
Updated docs

vasily-pozdnyakov:
### Summary of changes

Updated contributors list and contribution guide.

See the changed files

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -108,10 +108,9 @@ The following software is required for working on the repository:
 ### Building the app
 
 To build for a single OS run either `yarn ship-linux`, `yarn ship-mac` or `yarn ship-win`. To build for all three
-systems run `yarn ship`.
-**Important:** wine must be installed to build a Windows installer on a non-Windows system.
+systems run `yarn ship`. The built release(s) can be found under _/release_.
 
-The built release(s) can be found under _/release_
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -108,10 +108,9 @@ The following software is required for working on the repository:
 ### Building the app
 
 To build for a single OS run either `yarn ship-linux`, `yarn ship-mac` or `yarn ship-win`. To build for all three
-systems run `yarn ship`.
-**Important:** wine must be installed to build a Windows installer on a non-Windows system.
+systems run `yarn ship`. The built release(s) can be found under _/release_.
 
-The built release(s) can be found under _/release_
+**Important:** `wine` must be installed to build a Windows installer on a non-Windows system.
```
mstykow, line 113: this does not seem to be correct. i'm on a mac and did not have to install wine. also, if you reference an actual library it's probably better to link it instead of just putting it into backticks.

mstykow (changes requested):

CONTRIBUTING.md
vasily-pozdnyakov: good point, we tested it on all the systems and adjusted the guideline accordingly.

mstykow (approved):

Issue #2253, opened 2023-11-08
Replace "audit view" integration with E2E tests


mstykow:
As part of the ongoing effort to replace all integration with E2E tests, the goal of this task is to replace all tests in the folder "audit-view-tests".

PR #2257, opened 2023-11-08
feat: replace audit view tests part 1

mstykow:
### Summary of changes

- replace first half of audit view integration tests with E2E tests
- bugfix discovered while testing: properly omit the plus-icon on a package card when the signal it represents is hidden (previously the plus-icon only had hidden visibility but was still clickable)

Part of #2253.

Run locally and see CI/CD.


src/Frontend/Components/PackageCard/PackageCard.tsx
```diff
@@ -427,30 +424,30 @@ export function PackageCard(props: PackageCardProps): ReactElement | null {
...
-          : 'add'
-      }
-      tooltipPlacement="left"
-      disabled={props.displayPackageInfo.preferred}
-      onClick={props.onIconClick}
-      key={getKey('add-icon', props.cardId)}
-      icon={
-        <PlusIcon
-          sx={{
-            ...(props.cardConfig.isResolved ? classes.hiddenIcon : {}),
```
mstykow, line 444: this is a bug: all we're doing here is made the visibility of the plus-icon hidden but it's still there and thus still clickable. you can see that in the UI when hovering over the empty spot at the beginning of the package card.
thus: i'm now not rendering the plus-icon at all instead to fix this bug.

src/e2e-tests/__tests__/using-app-without-a-file.test.ts
```diff
@@ -0,0 +1,55 @@
...
+  },
+});
+
+test('provides expected functionality when no file is open and then opens file', async ({
+  menuBar,
+  resourceBrowser,
+  topBar,
+  projectStatisticsPopup,
+}) => {
+  await menuBar.assert.showsExpectedTitle();
```
MarkusObendrauf, line 38: I would prefer to have `showsTitle("OpossumUI")` or similar here

MarkusObendrauf (changes requested):
Thanks for this, I love the new style of tests.

Btw, could you try splitting the pr more in the future? I think review time grows quadratically with diff size. To review this I spent a lot of time just trying to figure out which tests were converted to what. Please try to split the next part more or i might just reject it right away

src/e2e-tests/utils/fixtures.ts
```diff
@@ -38,6 +40,7 @@ export const test = base.extend<{
   /** Run this function at any point in a test to abort the test at that point and inspect the opossum file. */
```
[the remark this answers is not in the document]
mstykow: because by default the test is executed without any data at all, i.e., no opossum file

src/e2e-tests/utils/opossum-files.ts
```diff
@@ -15,6 +15,7 @@ export interface OpossumData {
   inputData: ParsedOpossumInputFile;
   outputData?: ParsedOpossumOutputFile;
   decompress?: boolean;
+  openManually?: boolean;
```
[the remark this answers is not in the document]
mstykow: to me reading `opossumData.openManually` makes sense. but now i'll be deleting this anyway since opening a file is too flaky.

src/e2e-tests/__tests__/using-app-without-a-file.test.ts
mstykow: done

src/e2e-tests/__tests__/browsing-resources.test.ts
```diff
@@ -7,29 +7,83 @@ import { faker, test } from '../utils';
...
+        [faker.opossum.filePath(resourceName5, resourceName6, resourceName7)]: [
+          attributionId3,
+        ],
       }),
     }),
   },
 });
 
-test('shows and hides resources as user browses through resources', async ({
+test('shows expected resources as user browses through resources or clicks on progress bar', async ({
```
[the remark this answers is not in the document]
mstykow: i've split it now

src/Frontend/integration-tests/audit-view-tests/__tests__/audit-view.test.tsx
```diff
@@ -15,641 +15,36 @@ import {
...
-    );
-    clickOnButton(screen, ButtonText.Save);
-    expectValueInTextBox(
-      screen,
-      text.attributionColumn.packageSubPanel.packageName,
-      'Angular',
-    );
-  });
-
-  it('show confidence correctly', () => {
```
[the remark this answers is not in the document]
mstykow: i'm already testing most of this as part of other tests. for example, the assertion that the attribution details are empty/blank also asserts that the confidence is high.
reading through this test one more time just now, there is only one thing i haven't already tested: that the confidence changes to high when you convert a signal to an attribution on a resource. i have made myself a note that i will add this assertion to the tests about adding attributions in general (next PR).

src/e2e-tests/playwright.config.ts
```diff
@@ -12,6 +12,7 @@ const config: PlaywrightTestConfig = {
   preserveOutput: process.env.CI ? 'failures-only' : 'always',
   quiet: !!process.env.CI,
   reportSlowTests: null,
+  retries: process.env.CI ? 1 : 0,
```
MarkusObendrauf, line 15: what does this do?

MarkusObendrauf (approved):
cool, lgtm

src/e2e-tests/playwright.config.ts
mstykow: it allows one retry per test in the CI pipeline to help combat flakiness.

PR #2263, opened 2023-11-09
feat: add popup to warn user if attribution is saved globally that is marked as preferred

alexzurbonsen:
### Summary of changes

This PR introduces a new warning popup that is warning the user if changes to an attribution are saved globally and these changes encompass that the attribution is marked as preferred. 

This PR is part of a stack and just adds the popup without wiring it to the code base. Opening the popup and sufficient integration testing follows in the next PRs.

We want to give the user the possibility to globally mark attributions as preferred. But since that should be handled with caution a user should be alerted and asked for confirmation if he saves changes to an attribution globally that include that the attribution is marked as preferred.

This PR introduces the popup. In the next PR saving globally will be enabled and in a subsequent PR the warning popup will be opened. 

Can't be tested at this stage.

### DIFF
--- a/src/Frontend/Components/PreferGloballyPopup/PreferGloballyPopup.tsx
+++ b/src/Frontend/Components/PreferGloballyPopup/PreferGloballyPopup.tsx
@@ -1,0 +1,44 @@
+// SPDX-FileCopyrightText: Meta Platforms, Inc. and its affiliates
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+import { ReactElement } from 'react';
+
+import { text } from '../../../shared/text';
+import { ButtonText } from '../../enums/enums';
+import {
+  closePopupAndUnsetTargets,
+  saveTemporaryDisplayPackageInfoAndNavigateToTargetViewIfSavingIsNotDisabled,
+} from '../../state/actions/popup-actions/popup-actions';
+import { useAppDispatch } from '../../state/hooks';
+import { NotificationPopup } from '../NotificationPopup/NotificationPopup';
+
+export function PreferGloballyPopup(): ReactElement {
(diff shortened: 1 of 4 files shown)


src/Frontend/Components/PreferGloballyPopup/PreferGloballyPopup.tsx
```diff
@@ -0,0 +1,45 @@
...
+    dispatch(
+      saveTemporaryDisplayPackageInfoAndNavigateToTargetViewIfSavingIsNotDisabled(),
+    );
+  }
+
+  function handleCancelClick(): void {
+    dispatch(closePopupAndUnsetTargets());
+  }
+
+  const content = 'Do you really want to prefer the attribution globally?';
```
mstykow, line 28: let's put user-facing text into `text.ts` (i started doing this a few PRs ago).

mstykow (changes requested):
I'm not sure if we want to go the way of having diffs so small that they cannot even be tested. Even with stacked diffs, I would still prefer that the smallest diff is not something that just results in unreachable code.

src/Frontend/Components/PreferGloballyPopup/PreferGloballyPopup.tsx
alexzurbonsen: sounds like a good idea.

alexzurbonsen:
I made a small change: renamed the `Apply` button to `Ok` - that was what the ticket actually said.

alexzurbonsen:
> I'm not sure if we want to go the way of having diffs so small that they cannot even be tested. Even with stacked diffs, I would still prefer that the smallest diff is not something that just results in unreachable code.

Yes, that is a problem. I thought that it might make things potential more reviewable, but without proper tests that expectation doesn't hold. I can also keep adding to this diff, if it is more convenient for you?

alexzurbonsen:
reopened directly on opossum upstream #2270

PR #2264, opened 2023-11-09
feat: replacing audit view tests part 2

mstykow:
### Summary of changes

- replace more int by E2E tests
- bugfix discovered while testing: the save globally button became and then remained disabled when a QA user toggled the was-preferred attribute twice although this is an idempotent operation

Part of #2253.

Run locally and see CI/CD.


src/Frontend/Components/AttributionColumn/AttributionColumn.tsx
```diff
@@ -107,9 +107,10 @@ export function AttributionColumn(props: AttributionColumnProps): ReactElement {
   const isPreferenceFeatureEnabled = useAppSelector(
     getIsPreferenceFeatureEnabled,
   );
-  const wasPreferredFieldChanged: boolean =
-    initialManualDisplayPackageInfo.preferred !==
-    temporaryDisplayPackageInfo.preferred;
```
mstykow, line 112: there was a bug here: initially we may be comparing `undefined !== undefined` but after toggling the was-preferred attribute twice this would become `undefined !== false` thus disabling the global save button. but a double-toggle is idempotent and the user should thus be returned to the initial state.

src/shared/Faker.ts
```diff
@@ -149,6 +151,26 @@ class OpossumModule {
...
+      ...props,
+    };
+  }
+
+  public static externalAttributionSources(
+    props?: ExternalAttributionSources,
+  ): ExternalAttributionSources {
+    return {
+      ...(props || {
+        [faker.string.alpha({ length: 2 })]: this.externalAttributionSource(),
```
MarkusObendrauf, line 169: the key here is the essentially an id so it needs to be unique, please add this constraint.

src/e2e-tests/__tests__/saving-attributions.test.ts
```diff
@@ -0,0 +1,93 @@
...
+  await notSavedPopup.saveGloballyButton.click();
+  await modifyWasPreferredAttributionPopup.assert.isVisible();
+
+  await modifyWasPreferredAttributionPopup.saveGloballyButton.click();
+  await attributionList.attributionCard.assert.wasPreferredIconIsHidden(
+    packageInfo2,
+  );
+  await attributionList.attributionCard.assert.wasPreferredIconIsHidden(
+    packageInfo1,
+  );
```
MarkusObendrauf, line 92: in `audit-view.test.tsx` line 269, the asssertion was `expectAttributionIsMarkedAsWasPreferred(screen);`. why are we now asserting the opposite?

MarkusObendrauf (changes requested):
It's always a good sign when the tests find new bugs!

src/e2e-tests/__tests__/preferring-attributions.test.ts
```diff
@@ -0,0 +1,90 @@
...
+  await attributionDetails.assert.buttonInHamburgerMenuIsVisible(
+    'unmarkAsPreferred',
+  );
+  await attributionDetails.assert.buttonInHamburgerMenuIsHidden(
+    'markAsPreferred',
+  );
+
+  await attributionDetails.hamburgerMenu.unmarkAsPreferred.click();
+  await attributionDetails.assert.saveButtonIsEnabled();
+  await attributionDetails.assert.saveGloballyButtonIsEnabled();
```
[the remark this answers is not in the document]
mstykow: done

src/shared/Faker.ts
mstykow: since this fallback object only contains a single key uniqueness is guaranteed. but for consistency, i've changed it now to be the name of the source, as also the app would do it like this.

src/e2e-tests/__tests__/saving-attributions.test.ts
mstykow: i've rewritten, and in my opinion improved, the test slightly. if you check the previous test, it just asserted generally that some grey star is still present at the end. i don't find this assertion as useful as asserting one-by-one that grey stars are disappearing so that at the end of my test all grey stars are gone.

src/shared/Faker.ts
MarkusObendrauf: oh youre right, diregard that

MarkusObendrauf (approved):
lgtm

PR #2265, opened 2023-11-09
feat: replacing audit view tests part 3

mstykow:
### Summary of changes

- replace more int by E2E tests

Part of #2253.

Run locally and see CI/CD.


src/e2e-tests/__tests__/confirming-preselected-attributions.test.ts
```diff
@@ -64,7 +64,8 @@ test('updates progress bar and confidence when user confirms preselected attribu
     'confirmGloballyButton',
   );
 
-  await resourceDetails.attributionCard.contextMenu.confirmButton.click();
+  await resourceDetails.attributionCard.closeContextMenu();
+  await attributionDetails.confirmButton.click();
```
mstykow, line 68: i'm making this change so that i'm testing both context menu and button in attribution details column in one test rather than having to write two separate tests for essentially the same thing.

src/e2e-tests/__tests__/saving-attributions.test.ts
```diff
@@ -33,61 +37,105 @@ test.use({
...
+  await attributionDetails.name.fill(newPackageInfo.packageName!);
+  await attributionDetails.version.fill(newPackageInfo.packageVersion!);
+  await attributionDetails.url.fill(newPackageInfo.url!);
+  await attributionDetails.copyright.fill(newPackageInfo.copyright!);
+  await attributionDetails.licenseName.click();
+  await attributionDetails.selectLicense(license1);
+  await attributionDetails.assert.matchPackageInfo(newPackageInfo);
   await attributionDetails.saveButton.click();
-  await modifyWasPreferredAttributionPopup.assert.isVisible();
+});
```
MarkusObendrauf, line 76: i think there should be some more assertions at the end of this test. right now we click save but dont really check that anything was saved. for example, maybe click away, click back, and check that the package info matches `newPackageInfo`

MarkusObendrauf (approved):
cool, just a couple of small things

src/shared/Faker.ts
```diff
@@ -62,7 +63,10 @@ class OpossumModule {
         max: DiscreteConfidence.High - 1,
       }),
       copyright: faker.lorem.sentences(),
-      licenseName: faker.commerce.productName(),
+      licenseName: faker.commerce
+        .productName()
+        .match(/\b([A-Z])/g)!
+        .join(''),
```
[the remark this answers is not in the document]
mstykow: fair point

src/e2e-tests/__tests__/saving-attributions.test.ts
mstykow: good point. should be better now.

PR #2268, opened 2023-11-10
feat: enable global marking of attributions as preferred 

alexzurbonsen:


src/Frontend/Components/AttributionColumn/attribution-column-helpers.ts
```diff
@@ -275,8 +274,6 @@ export function usePurl(
 
   const savingStatus = isAllSavingDisabled
     ? AllowedSaveOperations.None
-    : wasPreferredFieldChanged
-    ? AllowedSaveOperations.Local
```
alexzurbonsen, line 279: Currently, I do not understand why the savingStatus is determined in this function. Also, there is no test for it. But I haven't understood enough of it to write one.

PR #2270, opened 2023-11-10
feat: enable preferring attribution globally and add warning popup

### LINKED ISSUE #2290: Save an attribution globally that is marked as preferred.
We want to enable the option to save a preferred attribution globaly.

Definition of done
- When an attribution is preferred it can be saved both, locally and globally.
- In case an newly preferred attribution is saved globally a popup warning is shown asking “Do you really want to prefer globally?”
  - “Ok”
  - “Cancel”

After confirming the warning the attribution is set as preferred for all its resources

### LINKED ISSUE #2291: Save an attribution globally that is unmarked as preferred
Goal of this ticket is to enable the option to unmark attributions globally in a similar fashion as the global marking of an attribution as preferred, see #2290.

Definition of done
- When an attribution is unmarked as preferred it can be saved both, locally and globally.
- In case an unmarked attribution is saved globally a popup warning is shown asking “Do you really want to unmark this attribution globally?”
  - “Ok”
  - “Cancel”

After confirming the warning the attribution is unmarked as preferred for all its resources.

alexzurbonsen:
### Summary of changes

This PR enables preferring an attribution globally, as well as removing the preferred status again. The user is shown a warning message whenever he is about to save such a global modification of the preferred status.

The commits:
1) create the warning popup
2) enable save globally if attribution is marked or unmarked as preferred
3) open warning popup on save. There are three ways to save and that lead into the warning popup:
 - save globally button
 - via save globally button on unsaved changes warning
 - via menu and hotkey
 4) add integration tests

The warning popup gives the user the choise to proceed with the modification of the preferred status or to undo it.
Undoing it reverts the changes to the preferred status but keeps remaining changes, if there are any.

Currently, modifying the preferred status is only possible in Audit view. I.e. all this changes only take effect in audit view.

Screenshot 2023-11-21 at 14 57 27
Screenshot 2023-11-21 at 14 57 54

We want to give the user the possibility to mark and unmark attributions as preferred globally. But since that should be handled with caution a user should be alerted and asked for confirmation if he globally saves changes that include marking or unmarking the attribution as preferred.

Closes #2290
Closes #2291 

Added two integration tests.
Manual testing.

### DIFF
--- a/src/Frontend/Components/AttributionColumn/attribution-column-helpers.ts
+++ b/src/Frontend/Components/AttributionColumn/attribution-column-helpers.ts
@@ -251,7 +251,6 @@ export function getMergeButtonsDisplayState(currentState: {
 export function usePurl(
   dispatch: AppThunkDispatch,
   packageInfoWereModified: boolean,
-  wasPreferredFieldChanged: boolean,
   temporaryDisplayPackageInfo: DisplayPackageInfo,
   selectedPackage: PanelPackage | null,
   selectedAttributionId: string | null,
@@ -275,9 +274,7 @@ export function usePurl(
 
   const savingStatus = isAllSavingDisabled
     ? AllowedSaveOperations.None
-    : wasPreferredFieldChanged
-      ? AllowedSaveOperations.Local
-      : AllowedSaveOperations.All;
+    : AllowedSaveOperations.All;
 
   useEffect(() => {
     dispatch(setAllowedSaveOperations(savingStatus));
--- a/src/Frontend/Components/ChangePreferredStatusGloballyPopup/ChangePreferredStatusGloballyPopup.tsx
+++ b/src/Frontend/Components/ChangePreferredStatusGloballyPopup/ChangePreferredStatusGloballyPopup.tsx
@@ -48,6 +48,7 @@ export function ChangePreferredStatusGloballyPopup(): ReactElement {
         buttonText: ButtonText.Cancel,
       }}
       isOpen={true}
+      aria-label={'change preferred status globally popup'}
     />
   );
 }
--- a/src/Frontend/Components/ChangePreferredStatusGloballyPopup/ChangePreferredStatusGloballyPopup.tsx
+++ b/src/Frontend/Components/ChangePreferredStatusGloballyPopup/ChangePreferredStatusGloballyPopup.tsx
@@ -1,0 +1,53 @@
+// SPDX-FileCopyrightText: Meta Platforms, Inc. and its affiliates
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+import { ReactElement } from 'react';
+
+import { text } from '../../../shared/text';
+import { ButtonText } from '../../enums/enums';
+import {
+  closePopupAndUnsetTargets,
+  saveTemporaryDisplayPackageInfoAndNavigateToTargetViewIfSavingIsNotDisabled,
+} from '../../state/actions/popup-actions/popup-actions';
+import { useAppDispatch, useAppSelector } from '../../state/hooks';
+import { getTemporaryDisplayPackageInfo } from '../../state/selectors/all-views-resource-selectors';
+import { NotificationPopup } from '../NotificationPopup/NotificationPopup';
+
+export function ChangePreferredStatusGloballyPopup(): ReactElement {
+  const dispatch = useAppDispatch();
+  const temporaryDisplayPackageInfo = useAppSelector(
+    getTemporaryDisplayPackageInfo,
+  );
+
+  function handleOkClick(): void {
+    dispatch(
+      saveTemporaryDisplayPackageInfoAndNavigateToTargetViewIfSavingIsNotDisabled(),
+    );
+  }
+
+  function handleCancelClick(): void {
+    dispatch(closePopupAndUnsetTargets());
+  }
+
+  const content = temporaryDisplayPackageInfo.preferred
+    ? text.changePreferredStatusGloballyPopup.markAsPreferred
+    : text.changePreferredStatusGloballyPopup.unmarkAsPreferred;
+
+  return (
+    <NotificationPopup
+      content={content}
+      header={'Warning'}
+      centerRightButtonConfig={{
+        onClick: handleOkClick,
+        buttonText: ButtonText.Ok,
+        isDark: true,
+      }}
+      rightButtonConfig={{
+        onClick: handleCancelClick,
+        buttonText: ButtonText.Cancel,
+      }}
+      isOpen={true}
+    />
+  );
+}
--- a/src/Frontend/enums/enums.ts
+++ b/src/Frontend/enums/enums.ts
@@ -30,6 +30,7 @@ export enum PopupType {
   UpdateAppPopup = 'UpdateAppPopup',
   LocatorPopup = 'LocatorPopup',
   ModifyWasPreferredAttributionPopup = 'ModifyWasPreferredAttributionPopup',
+  ChangePreferredStatusGloballyPopup = 'ChangePreferredStatusGloballyPopup',
 }
 
 export enum SavePackageInfoOperation {
@@ -83,6 +84,7 @@ export enum ButtonText {
   OpenDotOpossumFile = 'Open ".opossum" file',
   MarkAsPreferred = 'Mark as preferred',
   UnmarkAsPreferred = 'Unmark as preferred',
+  Ok = 'Ok',
 }
 
 export enum FilterType {
--- a/src/e2e-tests/__tests__/saving-preferred-attributions-globally.test.ts
+++ b/src/e2e-tests/__tests__/saving-preferred-attributions-globally.test.ts
@@ -1,0 +1,276 @@
+// SPDX-FileCopyrightText: Meta Platforms, Inc. and its affiliates
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+import { faker, test } from '../utils';
+
+const [resourceName1, resourceName2, resourceName3, resourceName4] =
+  faker.opossum.resourceNames({ count: 4 });
+const [attributionId1, packageInfo1] = faker.opossum.manualAttribution();
+const [attributionId2, packageInfo2] = faker.opossum.manualAttribution();
+const externalAttributionSource = faker.opossum.externalAttributionSource({
+  isRelevantForPreferred: true,
+});
+const [attributionId3, packageInfo3] = faker.opossum.externalAttribution({
+  source: faker.opossum.source({ name: externalAttributionSource.name }),
+});
+
+test.use({
+  data: {
+    inputData: faker.opossum.inputData({
+      resources: faker.opossum.resources({
+        [resourceName1]: 1,
+        [resourceName2]: 1,
+        [resourceName3]: 1,
+        [resourceName4]: 1,
+      }),
+      externalAttributionSources: faker.opossum.externalAttributionSources({
+        externalAttributionSource,
+      }),
+      externalAttributions: faker.opossum.externalAttributions({
+        [attributionId3]: packageInfo3,
+      }),
+      resourcesToAttributions: faker.opossum.resourcesToAttributions({
+        [faker.opossum.filePath(resourceName1)]: [attributionId3],
+        [faker.opossum.filePath(resourceName2)]: [attributionId3],
+        [faker.opossum.filePath(resourceName3)]: [attributionId3],
+        [faker.opossum.filePath(resourceName4)]: [attributionId3],
+      }),
+    }),
+    outputData: faker.opossum.outputData({
+      manualAttributions: faker.opossum.manualAttributions({
+        [attributionId1]: packageInfo1,
+        [attributionId2]: packageInfo2,
+      }),
+      resourcesToAttributions: faker.opossum.resourcesToAttributions({
+        [faker.opossum.filePath(resourceName1)]: [attributionId1],
+        [faker.opossum.filePath(resourceName2)]: [attributionId1],
+        [faker.opossum.filePath(resourceName3)]: [attributionId1],
+        [faker.opossum.filePath(resourceName4)]: [attributionId2],
+      }),
+    }),
+  },
+});
+
+test('marks and unmarks an attribution as preferred globally if user saves globally via button', async ({
+  attributionDetails,
+  changePreferredStatusGloballyPopup,
+  projectStatisticsPopup,
+  resourceBrowser,
+  resourceDetails,
+  menuBar,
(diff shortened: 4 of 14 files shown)


alexzurbonsen:
originated from #2263
This version doesn't use the fork workflow to enable stacking if necessary.

alexzurbonsen:
The ticket originally said to name the buttons of the warning popup "Ok" and "Cancel". But while implementing it, I changed it to "Yes" instead of "Ok" because the was the more in accordance with the question asked in the popup. And I changed "Cancel" to "Undo" because the current implementation undoes changes, which seemed to be the expected effect to me.

src/Frontend/Components/ChangePreferredStatusGloballyPopup/ChangePreferredStatusGloballyPopup.tsx
```diff
@@ -0,0 +1,74 @@
...
+    EMPTY_DISPLAY_PACKAGE_INFO;
+
+  function handleYesClick(): void {
+    dispatch(
+      saveTemporaryDisplayPackageInfoAndNavigateToTargetViewIfSavingIsNotDisabled(),
+    );
+  }
+
+  function undoChangeOfPreferredStatus(): void {
+    // TODO I am not sure the deep clone is needed, but might be safer...
```
alexzurbonsen, line 39: still need to figure this out.

src/Frontend/integration-tests/audit-view-tests/__tests__/audit-view.test.tsx
```diff
@@ -268,4 +226,229 @@ describe('The App in Audit View', () => {
...
+      },
+    },
+    externalAttributionSources: {
+      SC: {
+        name: 'ScanCode',
+        priority: 1,
+        isRelevantForPreferred: true,
+      },
+    },
+    // TODO do I really need the next two fields? test is green without them
```
alexzurbonsen, line 264: clarification needed

src/Frontend/Components/AttributionColumn/attribution-column-helpers.ts
```diff
@@ -275,8 +274,6 @@ export function usePurl(
 
   const savingStatus = isAllSavingDisabled
     ? AllowedSaveOperations.None
-    : wasPreferredFieldChanged
-    ? AllowedSaveOperations.Local
```
alexzurbonsen, line 279: I am not sure why the `savingStatus` was computed within this function. It is not tested and I didn't understand it sufficiently to write a test.

src/Frontend/Components/AttributionColumn/attribution-column-helpers.ts
```diff
@@ -275,9 +274,7 @@ export function usePurl(
 
   const savingStatus = isAllSavingDisabled
     ? AllowedSaveOperations.None
-    : wasPreferredFieldChanged
-      ? AllowedSaveOperations.Local
-      : AllowedSaveOperations.All;
```
alexzurbonsen, line 280: I don't know why `savingStatus` was defined here. There is no test for the function and I didn't understand it sufficiently to write one.

src/e2e-tests/__tests__/saving-preferred-attributions-globally.test.ts
```diff
@@ -0,0 +1,268 @@
...
+        [faker.opossum.filePath(resourceName1)]: [attributionId1],
+        [faker.opossum.filePath(resourceName2)]: [attributionId1],
+        [faker.opossum.filePath(resourceName3)]: [attributionId1],
+        [faker.opossum.filePath(resourceName4)]: [attributionId2],
+      }),
+    }),
+  },
+});
+
+test('marks and unmarks an attribution as preferred globally if user saves globally via button', async ({
```
alexzurbonsen, line 57: With this structure the tests take between 10s-14s for me locally. Seems more suitable.

src/Frontend/enums/enums.ts
```diff
@@ -83,6 +84,7 @@ export enum ButtonText {
   OpenDotOpossumFile = 'Open ".opossum" file',
   MarkAsPreferred = 'Mark as preferred',
   UnmarkAsPreferred = 'Unmark as preferred',
+  Ok = 'Ok',
```
mstykow, line 87: we could avoid introducing this new enum value and just use the existing "confirm". together with my other comment, that means the two buttons in the dialog are simply: "confirm" or "cancel". what do you think?

src/Frontend/Components/ChangePreferredStatusGloballyPopup/ChangePreferredStatusGloballyPopup.tsx
```diff
@@ -0,0 +1,75 @@
...
+      content={content}
+      header={'Warning'}
+      centerRightButtonConfig={{
+        onClick: handleYesClick,
+        buttonText: ButtonText.Ok,
+        isDark: true,
+      }}
+      rightButtonConfig={{
+        onClick: handleUndoClick,
+        buttonText: ButtonText.Undo,
```
mstykow, line 69: just now testing your PR, i was suprised that this button is not just a simple "cancel" button.

the dialog here opens, when i press "save globally" so my first thought would have been to either have the option to confirm saving my change globally, or to simply cancel my attempt.

that this button here not only cancels my attempt to save globally but also undoes my change to the preferred state, seems too much to me.

so my suggestion would be to simplify the logic here and just have a simple cancel button. if i as a user really want to undo my change to the preferred attribute, i can do so myself.

src/e2e-tests/__tests__/saving-preferred-attributions-globally.test.ts
```diff
@@ -0,0 +1,268 @@
...
+const resourceName2 = 'resource2';
+const resourceName3 = 'resource3';
+const resourceName4 = 'resource4';
+const [attributionId1, packageInfo1] = faker.opossum.manualAttribution();
+const [attributionId2, packageInfo2] = faker.opossum.manualAttribution();
+const externalAttributionSource = faker.opossum.externalAttributionSource({
+  isRelevantForPreferred: true,
+});
+const [attributionId3, packageInfo3] = faker.opossum.externalAttribution({
+  source: { name: externalAttributionSource.name, documentConfidence: 100 },
```
mstykow, line 17: please use `faker.opossum.source` to build the source. this avoid you needing to hard-code values such as 100 and avoids putting undue importance on specific attributes. right now reading 100, i would wonder if this is essential to the test being green.

src/e2e-tests/__tests__/saving-preferred-attributions-globally.test.ts
mstykow: yes, for me only 5sec.

src/e2e-tests/__tests__/saving-preferred-attributions-globally.test.ts
```diff
@@ -0,0 +1,268 @@
+// SPDX-FileCopyrightText: Meta Platforms, Inc. and its affiliates
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+import { faker, test } from '../utils';
+
+const resourceName1 = 'resource1';
+const resourceName2 = 'resource2';
+const resourceName3 = 'resource3';
+const resourceName4 = 'resource4';
```
mstykow, line 10: here you can use this new function: https://github.com/opossum-tool/OpossumUI/pull/2272/files#diff-3df6fa26c7089e828b9b5129bf858d652e646ab294bf10ba6e30b0cb201ebefdR51

src/e2e-tests/__tests__/saving-preferred-attributions-globally.test.ts
```diff
@@ -0,0 +1,268 @@
...
+    await changePreferredStatusGloballyPopup.assert.showsUnmarkAsPreferredWarning();
+
+    await changePreferredStatusGloballyPopup.okButton.click();
+    await resourceDetails.attributionCard.assert.preferredIconIsHidden(
+      packageInfo1,
+    );
+  });
+});
+
+test('popup is only shown if necessary and undo reverts preferrence status only', async ({
```
mstykow, line 218:
```suggestion
test('popup is only shown if necessary and undo reverts preference status only', async ({
```

...but this test will change if we no longer actually undo the operation.

src/e2e-tests/page-objects/ChangePreferredStatusGloballyPopup.ts
```diff
@@ -0,0 +1,54 @@
...
+    },
+    isHidden: async (): Promise<void> => {
+      await expect(this.node).toBeHidden();
+    },
+    showsMarkAsPreferredWarning: async (): Promise<void> => {
+      const regex = new RegExp(
+        text.changePreferredStatusGloballyPopup.markAsPreferred,
+      );
+      const innerText = await this.node.textContent();
+      expect(innerText).toMatch(regex);
```
mstykow, line 40: using a regex seems unnecessarily complicated to me. does this not work?
```
await expect(
        this.node.getByText(
          text.changePreferredStatusGloballyPopup.markAsPreferred,
        ),
      ).toBeVisible();
```

src/e2e-tests/page-objects/PackageCard.ts
```diff
@@ -89,6 +89,10 @@ export class PackageCard {
     return this.node(packageInfo).getByRole('button').getByLabel('add');
   }
 
+  public preferredIcon(packageInfo: PackageInfo): Locator {
+    return this.node(packageInfo).getByLabel('preferred icon');
```
mstykow, line 93: here we need to use `exact: true` lest it also matches the "was preferred icon" (see below).

src/shared/text.ts
```diff
@@ -17,4 +17,9 @@ export const text = {
       searchForPackage: 'Search for package information',
     },
   },
+  changePreferredStatusGloballyPopup: {
```
mstykow, line 20: 👍

mstykow (changes requested):

src/Frontend/enums/enums.ts
alexzurbonsen: I discussed the "confirm" option with Benedikt last week. His argument was that "confirm" is used to confirm attributions, i.e. it has a clear meaning that he didn't want to overload. Which seemed reasonable to me.

src/Frontend/Components/ChangePreferredStatusGloballyPopup/ChangePreferredStatusGloballyPopup.tsx
alexzurbonsen: Also discussed that with Benedikt. I initially started off with a cancel button. It can lead to confusion of the user because there is no visual feedback on the preferred status before it was saved.

If you press cancel, afterwards you are in a state where the package card doesn't show any golden star. But the hamburger menu only gives you the option to unmark the attribution as preferred (and the other way around for preferred attributions).

We decided to create the visual feedback first, and then add a cancel button to the popup.

When I played around with it, I also found it convenient to undo the mark when cancelling the save action. Otherwise you have to do two extra clicks to do something you probably want to do anyways.

alexzurbonsen: Sorry, I should have mentioned that in the PR somewhere!

src/e2e-tests/__tests__/saving-preferred-attributions-globally.test.ts
alexzurbonsen: Nice!

src/e2e-tests/page-objects/ChangePreferredStatusGloballyPopup.ts
alexzurbonsen: I didn't work for some reason if I remeber correctly. But I think there would be a better way: Make the aria label depend on the content line of the pop up as well. Then I can simply assert on that?!

src/e2e-tests/page-objects/PackageCard.ts
alexzurbonsen: Oh, right...

src/e2e-tests/__tests__/saving-preferred-attributions-globally.test.ts
alexzurbonsen: that's quite a difference :D

src/e2e-tests/page-objects/ChangePreferredStatusGloballyPopup.ts
alexzurbonsen: I checked again. It seems that it didn't work because I forgot the await. (And I didn't forget it in the regex solution.)
I am changing it to what you proposed.

alexzurbonsen:
another rebase to resolve conflicts.

src/e2e-tests/page-objects/ChangePreferredStatusGloballyPopup.ts
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  public assert = {
+    isVisible: async (): Promise<void> => {
+      await expect(this.node).toBeVisible();
+    },
+    isHidden: async (): Promise<void> => {
+      await expect(this.node).toBeHidden();
+    },
+    showsMarkAsPreferredWarning: async (): Promise<void> => {
```
mstykow, line 35: let's stick with the pattern:
```suggestion
    markAsPreferredWarningIsVisible: async (): Promise<void> => {
```
and similar for the hidden state.

src/Frontend/Components/ChangePreferredStatusGloballyPopup/ChangePreferredStatusGloballyPopup.tsx
alexzurbonsen: Thanks for the discussion. It now is a cancel button that just cancels.

src/Frontend/Components/ChangePreferredStatusGloballyPopup/ChangePreferredStatusGloballyPopup.tsx
```diff
@@ -0,0 +1,54 @@
...
+    getTemporaryDisplayPackageInfo,
+  );
+
+  function handleOkClick(): void {
+    dispatch(
+      saveTemporaryDisplayPackageInfoAndNavigateToTargetViewIfSavingIsNotDisabled(),
+    );
+  }
+
+  function handleCancelClick(): void {
```
mstykow, line 29: could you please create the task here in opossum to display/hide the golden star as the user changes the preferred status via the hamburger/context menu?

mstykow (approved):

src/Frontend/Components/ChangePreferredStatusGloballyPopup/ChangePreferredStatusGloballyPopup.tsx
alexzurbonsen: Sure,  #2330.

PR #2272, opened 2023-11-10
feat: improve E2E developer experience

mstykow:
- extend developer guide about E2E debugging
- run tests with single worker by default (for slower machines) but allow increasing the number of workers via environment variable
- introduce package.json script to create an app release for the current OS
- remove unnecessary "increase watches" in pipeline
- rename pipeline jobs to better describe what they are doing
- remove yarn4 "enableHardenedMode" to speed up install step
- create faker method to generate given number of resource names in alphabetical order


CONTRIBUTING.md
```diff
@@ -105,6 +105,51 @@ The following software is required for working on the repository:
 - [reuse/tool](https://git.fsfe.org/reuse/tool#install) (to check that copyright information is provided, for more context see https://reuse.software/),
 - [wine](https://www.winehq.org/) (only to build the Windows version).
 
+### Running the end-to-end tests
+
+As mentioned above, the end-to-end tests can be run using `yarn test:e2e`. This runs the tests against the dev server. To run the tests against a built version of the app, which is also what the CI/CD pipeline does, run `yarn test:e2e:ci`. See below for how to build the app for your OS.
+
+Additionally, we recommend you install the [Playwright Test for VSCode extension](https://marketplace.visualstudio.com/items?itemName=ms-playwright.playwright). This will allow you running individual tests from within VSCode, either by clicking the green play button next to a test, or by going to the "testing" tab in the sidebar.
+
+Note that the extension cannot start the dev server for you, so you will have to run the tests against a built version of the app. In addition, you must add the following environment variable in your `settings.json`:
```
FlorianSchepers, line 114: This means we need to run the ci option for the e2e tests right? May be add something like "... against a build version of the app **(i.e. `yarn test:e2e:ci`)**.".

CONTRIBUTING.md
```diff
@@ -105,6 +105,51 @@ The following software is required for working on the repository:
...
+
+```json
+"playwright.env": {
+   "CI": true
+}
+```
+
+### Debugging the end-to-end tests
+
+Each executed end-to-end test creates artifacts in the folder `src/e2e-tests/artifacts`. The artifacts contain the auto-generated .opossum file that the test was run against and, in case the test failed, a Playwright trace file.
```
FlorianSchepers, line 124: The auto-generated .opossum file is great. However, it will contain all the changes performed by the test, e.g., delted attributions will be deleted. Would there be an option to also provide the initial .opossum file? This would allow the user to manually reproduce each step of the test.

CONTRIBUTING.md
mstykow: No, that's not what I mean. VSCode's green play button will run the tests. So there is no need for you to do anything on the CLI. However, you do need a built version of the app in your `release` folder. So at some point you must have run `yarn ship-X`. I'll see if I can formulate that more clearly.

CONTRIBUTING.md
mstykow: I think it's correct that the .opossum file represents the state at which the test failed. If you need to debug to an earlier point, you would use the `debug` util described below.

alexzurbonsen (approved):
I tried to figure it out with Intellij, but didn't work so far. Will try to get it going and add the instructions.

PR #2280, opened 2023-11-10
feat: replace "all views" integration tests with E2E tests

### LINKED ISSUE #2254: Replace "all views" integration tests with E2E tests
As part of the ongoing effort to replace all integration with E2E tests, the goal of this task is to replace all tests in the folder "all-views-tests".

mstykow:
### Summary of changes

- replace remaining "all views" int by E2E tests
- delete any reference to integration tests
- delete any unused test helpers

Closes #2254.

Run locally and see CI/CD.
Note that there are no more integration tests left now. So `yarn test:integration` no longer exists.


src/Frontend/integration-tests/all-views-tests/__tests-ci__/open-file-in-external-link.test.tsx
```diff
@@ -1,64 +0,0 @@
...
-    clickGoToLinkIcon(screen, 'link to open');
-    expect(window.electronAPI.openLink).toHaveBeenCalledTimes(1);
-    expect(window.electronAPI.openLink).toHaveBeenCalledWith(
-      expectedLinkForParent,
-    );
-
-    clickOnElementInResourceBrowser(screen, 'something_else.js');
-    expectGoToLinkIconIsVisible(screen);
-    clickGoToLinkIcon(screen, 'link to open');
-    expect(window.electronAPI.openLink).toHaveBeenCalledTimes(2);
```
mstykow, line 59: i did not find a way to mock the `openLink` API when running E2E tests. thus, i cannot write an E2E test that opens a valid link because such a test would open a browser tab every time it runs.

we do, however, already have other (unit) tests where we mock this API and ensure that the functionality to open a valid link is working as expected.

src/Frontend/integration-tests/all-views-tests/__tests-ci__/other-popups.test.tsx
```diff
@@ -1,430 +0,0 @@
...
-
-    mockSaveFileRequestChannel();
-    insertValueIntoTextBox(
-      screen,
-      text.attributionColumn.packageSubPanel.purl,
-      testInvalidPurl,
-    );
-
-    // Trigger sending external attribution in IpcChannel without the clicking on save button
-    clickOnElementInResourceBrowser(screen, 'firstResource.js');
```
mstykow, line 365: such mocks go against the philosophy of integration/e2e tests: if we cannot run the test via interactions with the UI but need to artificially produce a backend error, it's probably not the best candidate for an E2E test.

src/Frontend/integration-tests/all-views-tests/__tests-ci__/not-saved-popup.test.tsx
```diff
@@ -1,467 +0,0 @@
...
-        resourcesToAttributions: {
-          '/root/src/file_2': ['uuid_1'],
-        },
-      },
-    };
-
-    mockElectronBackendOpenFile(mockChannelReturn);
-  }
-
-  it('switches between views', () => {
```
mstykow, line 82: in this file there were various tests that were either unrelated to the "not saved" popup, or were duplicates of tests elsewhere.

src/Frontend/integration-tests/all-views-tests/__tests-ci__/other-popups.test.tsx
leslielazzarino: What happened to the other tests in this file? I cannot find the new ones in this diff.

leslielazzarino (changes requested):
I cannot find the new version of any test in src/Frontend/integration-tests/all-views-tests/__tests-ci__/other-popups.test.tsx

src/Frontend/integration-tests/all-views-tests/__tests-ci__/other-popups.test.tsx
mstykow: each test in this file has already been accounted for:
- some were tests of the "unsaved changes" popup and thus just duplicating those tests. those tests started with "warning popup".
- some were tests of the "resource path" popup which is already tested in `showing-resources-belonging-to-attributions.test.ts`. those tests started with "working popup".
- the remaining tests, starting with "error popup", were not real integration/e2e tests as described above

leslielazzarino (approved):

PR #2283, opened 2023-11-13
feat: add path breadcrumbs

### LINKED ISSUE #1501: Add breadcrumbs instead of path
Should allow navigation. An icon next to them should still help by copying the full path.

mstykow:
### Summary of changes

- convert path bar to a component that renders clickable breadcrumbs instead of a static path
- add button to copy path to clipboard which works on all OS
- keep path in Attribution Wizard unchanged

image

image

image

image

Closes #1501.

File with resources which are nested 10 levels deep:
[input.json](https://github.com/opossum-tool/OpossumUI/files/13333743/input.json)

### DIFF
--- a/src/Frontend/Components/AttributionWizardPopup/AttributionWizardPopup.tsx
+++ b/src/Frontend/Components/AttributionWizardPopup/AttributionWizardPopup.tsx
@@ -2,13 +2,16 @@
 // SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
 //
 // SPDX-License-Identifier: Apache-2.0
+import MuiNavigateNextIcon from '@mui/icons-material/NavigateNext';
 import MuiBox from '@mui/material/Box';
+import MuiTooltip from '@mui/material/Tooltip';
+import MuiTypography from '@mui/material/Typography';
 import { ReactElement, useState } from 'react';
 import { v4 as uuid4 } from 'uuid';
 
 import { DisplayPackageInfo } from '../../../shared/shared-types';
 import { ButtonText } from '../../enums/enums';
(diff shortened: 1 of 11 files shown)


mstykow:
> Great PR, just one request: a way to collapse the path again, without reopening the file.

MUI does not support re-collapsing. The question would probably be: where should such a button to re-collapse be located and why should it even be there? In my opinion, we're also fine without this.

leslielazzarino:
> > Great PR, just one request: a way to collapse the path again, without reopening the file.
> 
> MUI does not support re-collapsing. The question would probably be: where should such a button to re-collapse be located and why should it even be there? In my opinion, we're also fine without this.

I would argue, that if it worth collapsing in the first place, it is worth being able to go back to that default condition. Having said that, I have no genius idea. There is no obvious opposite of the three points, and an explicit collapse button would probably look bad. We could just reset it when moving to a different resource. What do you think about that?

mstykow:
> > > Great PR, just one request: a way to collapse the path again, without reopening the file.
> > 
> > 
> > MUI does not support re-collapsing. The question would probably be: where should such a button to re-collapse be located and why should it even be there? In my opinion, we're also fine without this.
> 
> I would argue, that if it worth collapsing in the first place, it is worth being able to go back to that default condition. Having said that, I have no genius idea. There is no obvious opposite of the three points, and an explicit collapse button would probably look bad. We could just reset it when moving to a different resource. What do you think about that?

Yes, that I can do: https://github.com/opossum-tool/OpossumUI/pull/2283/files#diff-c3f47559928cee00d819d0ab68bd8a3510c5c8c436e8e9d074c3ea0f488aaa87R60

leslielazzarino (approved):

PR #2301, opened 2023-11-15
feat: add browser history

### LINKED ISSUE #2299: Introduce resource browser history
As a user, I want to be able to navigate back and forth through my resources the same way I would navigate through web pages.

There should be a back and a forward button part of the path bar in audit view through which the user can go back and forth through the selected-resource history.
There should also be the usual shortcuts (see Chrome) for going back/forth via keyboard.

mstykow:
### Summary of changes

- add buttons to navigate back and forth through resource browser history
- enable navigation via keyboard shortcuts
- enable/disable path bar buttons and show appropriate tooltips instead of hiding them completely
- add tooltip delay to icon buttons to make them feel less jumpy
- scroll to resource when navigating via browser or path bar
- expand resource browser to selected resource if a resource is navigated to but hidden in the browser due to file/folder collapse

image

image

Closes #2299.

Use mouse and keyboard to navigate through the browser history. Pay special attention to edge cases such as being at the beginning or end of the history stack.

### DIFF
--- a/src/Frontend/Components/IconButton/IconButton.tsx
+++ b/src/Frontend/Components/IconButton/IconButton.tsx
@@ -34,6 +34,7 @@ export function IconButton(props: IconButtonProps): ReactElement {
       disableInteractive
       title={props.hidden ? '' : props.tooltipTitle}
       placement={props.tooltipPlacement}
+      enterDelay={1500}
     >
       <MuiBox component="span" sx={props.containerSx}>
         {
--- a/src/Frontend/Components/PathBar/PathBar.tsx
+++ b/src/Frontend/Components/PathBar/PathBar.tsx
@@ -2,15 +2,24 @@
 // SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
 //
 // SPDX-License-Identifier: Apache-2.0
+import ArrowBackIcon from '@mui/icons-material/ArrowBack';
+import ArrowForwardIcon from '@mui/icons-material/ArrowForward';
 import ContentCopyIcon from '@mui/icons-material/ContentCopy';
 import MuiBox from '@mui/material/Box';
-import { compact } from 'lodash';
-import { ReactElement } from 'react';
+import { compact, uniq } from 'lodash';
+import { ReactElement, useCallback, useEffect, useMemo, useState } from 'react';
+import { useHotkeys } from 'react-hotkeys-hook';
 
-import { clickableIcon } from '../../shared-styles';
+import { clickableIcon, disabledIcon } from '../../shared-styles';
 import { setSelectedResourceIdOrOpenUnsavedPopup } from '../../state/actions/popup-actions/popup-actions';
+import { setExpandedIds } from '../../state/actions/resource-actions/audit-view-simple-actions';
+import { getParents } from '../../state/helpers/get-parents';
 import { useAppDispatch, useAppSelector } from '../../state/hooks';
-import { getSelectedResourceId } from '../../state/selectors/audit-view-resource-selectors';
+import {
+  getExpandedIds,
+  getSelectedResourceId,
+} from '../../state/selectors/audit-view-resource-selectors';
+import { usePrevious } from '../../util/use-previous';
 import { Breadcrumbs } from '../Breadcrumbs/Breadcrumbs';
 import { GoToLinkButton } from '../GoToLinkButton/GoToLinkButton';
 import { IconButton } from '../IconButton/IconButton';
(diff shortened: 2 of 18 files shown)


src/Frontend/Components/IconButton/IconButton.tsx
```diff
@@ -34,6 +34,7 @@ export function IconButton(props: IconButtonProps): ReactElement {
       disableInteractive
       title={props.hidden ? '' : props.tooltipTitle}
       placement={props.tooltipPlacement}
+      enterDelay={1500}
```
MarkusObendrauf, line 37: if i first hover somewhere else, eg the top bar, then the hint appears without delay.
also, 1500ms is a bit long imo (maybe that's just personal preference)

src/Frontend/Components/PathBar/PathBar.tsx
```diff
@@ -20,35 +23,123 @@ const classes = {
...
-  };
+  const cmdOrCtrl = useMemo(
+    () => (window.navigator.platform.match(/^Mac/) ? '⌘' : 'Ctrl'),
+    [],
+  );
 
-  return pathElements.length > 0 ? (
+  const handleGoBack = useCallback((): void => {
+    if (activeIndex > 0) {
+      setActiveIndex((prev) => prev - 1);
```
MarkusObendrauf, line 46: what does `prev` mean here? previous?

src/Frontend/Components/PathBar/PathBar.tsx
```diff
@@ -20,35 +23,123 @@ const classes = {
...
+          tooltipPlacement={'left'}
+          tooltipTitle={`Go back (${cmdOrCtrl} + Left arrow)`}
+          disabled={activeIndex === 0}
+        />
+        <IconButton
+          icon={
+            <ArrowForwardIcon
+              aria-label={'go forward'}
+              sx={
+                activeIndex === history.length - 1
```
MarkusObendrauf, line 108: im reading this condition a lot - maybe `isLastHistoryItemActive()` and `isFirstHistoryItemActive()` is nicer?

MarkusObendrauf (changes requested):
Right now, undo could jump to a resource which is not shown in the resource browser. If that happens, it's very hard to find your place. Imo that doesn't need to be fixed in this PR.

src/Frontend/Components/IconButton/IconButton.tsx
mstykow: yes, that's intentional: once any tooltip is open, the next one should not wait again.
i picked the duration from what google chrome does on its back/forward buttons.

src/Frontend/Components/PathBar/PathBar.tsx
mstykow: yes

mstykow:
> Right now, undo could jump to a resource which is not shown in the resource browser. If that happens, it's very hard to find your place. Imo that doesn't need to be fixed in this PR.

Sorry, I'm not understanding this comment. What do you mean by "undo"? Do you mean "go back"? And how can I reproduce what you're saying here: in particular, how can it be that there is a resource that's not part of the resource browser? Do you mean part of the viewport (in case of long lists of resources)?

src/Frontend/Components/PathBar/PathBar.tsx
mstykow: good point. done.

mstykow:
> > Right now, undo could jump to a resource which is not shown in the resource browser. If that happens, it's very hard to find your place. Imo that doesn't need to be fixed in this PR.
> 
> Sorry, I'm not understanding this comment. What do you mean by "undo"? Do you mean "go back"? And how can I reproduce what you're saying here: in particular, how can it be that there is a resource that's not part of the resource browser? Do you mean part of the viewport (in case of long lists of resources)?

@MarkusObendrauf I think you've discovered a very valid point and I think I created a fix now. Please try again. :)

src/Frontend/Components/IconButton/IconButton.tsx
MarkusObendrauf: TIL chrome has tooltips. but even if we're using chrome as a reference: in chrome, the tooltips for different sections of the UI are not connected.
what makes this really wierd in opossumui is the fact that the progress bar's tooltip appears instantly. so if you move your cursor over the progress bar and then the arrows, the tooltip on the arrow shows instantly.

src/Frontend/Components/PathBar/PathBar.tsx
MarkusObendrauf: i don't understand why it's `prev` and not `current`

mstykow:
> There are a couple of strange things regarding collapsing folders:
> 
> 1. If you select a resource in a folder, and then collapse that folder, the resource browser scrolls to the top of the screen.
> 2. It's quite difficult to find a resource contained in a collapsed folder. Imo, navigating in history or clicking a breadcrumb should expand the resource tree so that the current resource is shown.

Both issues should be fixed now. Please check again.

src/Frontend/Components/PathBar/PathBar.tsx
mstykow: it's a convention to call this state "previous" or "pending": https://react.dev/reference/react/useState#updating-state-based-on-the-previous-state

i guess one could argue either way. so better just stick to the naming convention they recommend.

src/Frontend/Components/IconButton/IconButton.tsx
mstykow: the instant response you're observing happens, because you move quickly from the progress bar to the icon buttons. if you first pause your mouse somewhere else and then move to the icon buttons, you'll see that the usual 1.5 secs kick in.

i do like that the progress bar tooltip shows up quicker because it actually contains essential information which is not the case for the helper text on the icon buttons. there, a fast tooltip just creates noise because in most cases i already know what that icon button does and don't need to be reminded every time i want to click on it.

i don't think this point is worth discussing much more.

MarkusObendrauf: ah i see, theres no problem then :)

MarkusObendrauf (changes requested):
Nice! I'm loving the new scrolling and file tree opening. However, if you do both in a single operation, the scroll doesn't work.

src/Frontend/util/use-previous.ts
```diff
@@ -0,0 +1,15 @@
+// SPDX-FileCopyrightText: Meta Platforms, Inc. and its affiliates
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+import { useEffect, useRef } from 'react';
+
+export function usePrevious<T>(value: T, fallback: T): T;
+export function usePrevious<T>(value: T, fallback?: T): T | undefined;
+export function usePrevious<T>(value: T, fallback?: T): T | undefined {
```
MarkusObendrauf, line 9: something went wrong here

MarkusObendrauf (changes requested):

src/Frontend/util/use-previous.ts
mstykow: why do you say this? this is how function overloading in typescript looks like.

MarkusObendrauf: thx for the explanation

MarkusObendrauf (approved):
lgtm

PR #2315, opened 2023-11-17
feat: hide search clear button when no search term

### LINKED ISSUE #2281: Hide "clear" icon (x) when search field is empty
**Describe the bug**
"Clear" icon (x) at search field is always shown (even when the field is empty). AT Attribution view and Search popup.

**To Reproduce**
1. Go to Attribution View
2. See the empty search field with shown X icon

**Expected behavior**
When the field is empty, the icon should be hidden.

**Screenshots**
image

**Desktop (please complete the following information):**

- OS: Windows

mstykow:
### Summary of changes

- conditionally hide search clear button when no search term is present
- improve unit tests by using @testing-library/user-event as recommended by RTL maintainers

Closes #2281.

Check out places in the app that use the search bar.

### DIFF
--- a/src/Frontend/Components/SearchTextField/__tests__/SearchTextField.test.tsx
+++ b/src/Frontend/Components/SearchTextField/__tests__/SearchTextField.test.tsx
@@ -2,20 +2,44 @@
 // SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
 //
 // SPDX-License-Identifier: Apache-2.0
-import { fireEvent, screen } from '@testing-library/react';
+import { screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 
+import { faker } from '../../../../shared/Faker';
 import { renderComponentWithStore } from '../../../test-helpers/render-component-with-store';
 import { SearchTextField } from '../SearchTextField';
 
 describe('The SearchTextField', () => {
-  it('has search functionality', () => {
-    const onInputchange = jest.fn();
+  it('clears and updates the search term', async () => {
+    // given
+    const onInputChange = jest.fn();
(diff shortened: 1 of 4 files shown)


src/Frontend/Components/SearchTextField/__tests__/SearchTextField.test.tsx
```diff
@@ -2,20 +2,44 @@
 // SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
 //
 // SPDX-License-Identifier: Apache-2.0
-import { fireEvent, screen } from '@testing-library/react';
+import { screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
```
vasily-pozdnyakov, line 6: Why do we need this? We usually use
`import { fireEvent } from '@testing-library/react';`
`fireEvent.click(...)`
It seems to have the same functionality.

vasily-pozdnyakov (changes requested):

src/Frontend/Components/SearchTextField/__tests__/SearchTextField.test.tsx
mstykow: it's not the same.
1. `user-event` simulates actual user interaction while `fireEvent` dispatches DOM events.
2. `user-event` has a lot of convenient APIs that `fireEvent` is missing.
3.  even RTL encourages the use of `user-event` over `fireEvent`: https://testing-library.com/docs/user-event/intro#differences-from-fireevent

i'll talk about this in the next COP meeting.

vasily-pozdnyakov: ok

vasily-pozdnyakov (approved):

PR #2325, opened 2023-11-21
feat: improve file processing state

### LINKED ISSUE #2323: Improve file processing state
As a user, I would like to see more context while an file is being opened/exported instead of a simple loading spinner.

Display a dialog describing concrete steps that are performed by the backend when a file is processed.
Show which steps are completed and which steps are still in progress.
If possible, show a linear progress indicator.

mstykow:
### Summary of changes

- replace loading spinner with new process popup that displays detailed information about the file processing process
- introduce `dayjs` to format dates and times
- create logger class which writes to stdout in the backend but also sends the log to the frontend
- delay follow-up export handler because this resource intensive task blocks the UI to such a degree that the process popup is not displayed in time
- fix `ipcRenderer` subscription mechanism which was not correctly unsubscribing from backend event listeners (the attempt to fix issue #825 in #828 was not correct as it only partly mitigated the problem but didn't fix the underlying cause that `removeListener` was not called on the correct listener)

image

image

Closes #2323.

- Open some files that take longer or shorter to load.
- Also test that exporting a file for follow-up does not create multiple exports. This is regression for #825.


src/Frontend/Components/OpenFilePopup/OpenFilePopup.style.ts
mstykow, line 1: I'm using the styled components pattern here as per #2131

src/ElectronBackend/preload.ts
```diff
@@ -7,29 +7,14 @@ import { contextBridge, ipcRenderer } from 'electron';
...
-    ipcRenderer.on(channel, listener);
-  }
-}
-
-function removeListener(channel: AllowedFrontendChannels): void {
-  if (Object.values(AllowedFrontendChannels).includes(channel)) {
-    // we remove all listeners to fix a bug we had with using
-    // ipcRenderer.removeListener in combination with the
-    // context bridge.
-    ipcRenderer.removeAllListeners(channel);
```
mstykow, line 28: we should not remove all listeners of a particular channel because then we cause race conditions between subscribers: only the last subscriber wins the subscription.
so the "fix" introduced here is actually causing other bugs. the proper fix is to call `.removeListener` on the correct listener. see line 35 below.

src/ElectronBackend/preload.ts
```diff
@@ -45,8 +30,10 @@ const electronAPI: IElectronAPI = {
     ipcRenderer.invoke(IpcChannel.ExportFile, args),
   saveFile: (saveFileArgs: SaveFileArgs) =>
     ipcRenderer.invoke(IpcChannel.SaveFile, saveFileArgs),
-  on,
-  removeListener,
```
mstykow, line 49: having a `removeListener` separate from the subscriber (`on`) is the cause of the bug from #825.

setupTests.ts
```diff
@@ -5,15 +5,16 @@
...
+
+// We suppress webworker errors that are due to jest not supporting these.
 const SUBSTRINGS_TO_SUPPRESS_IN_CONSOLE_INFO = [
   'Web worker error in workers context',
   'Error in rendering folder progress bar',
   'Error in rendering top progress bar',
   'Error in ResourceDetailsTab',
 ];
-//We supress recharts warning that is due to our mocking in tests.
+// We suppress recharts warning that is due to our mocking in tests.
```
leslielazzarino, line 17: ...the recharts...

src/ElectronBackend/input/importFromFile.ts
```diff
@@ -199,10 +192,7 @@ async function createOutputInOpossumFile(
   resourcesToExternalAttributions: AttributionsToResources,
   projectId: string,
 ): Promise<ParsedOpossumOutputFile> {
-  log.info(
-    `Starting to create output in .opossum file, project ID is ${projectId}`,
-  );
-
+  logger.info('Preparing output file');
```
leslielazzarino, line 195: As there is only one .opossum file, the new messages are a bit confusing. Something like"Writing output to file" and "Reading output from file" would be clearer.

src/ElectronBackend/input/importFromFile.ts
```diff
@@ -232,20 +219,18 @@ async function parseOrCreateOutputJsonFile(
...
       resourcesToExternalAttributions,
       projectId,
       inputFileMD5Checksum,
     );
     await writeFile({ path: filePath, content: attributionJSON });
-    log.info('... Successfully created output file.');
   }
 
-  log.info(`Starting to parse output file ${filePath} ...`);
+  logger.info('Parsing output file');
```
leslielazzarino, line 232: Same for this function

src/ElectronBackend/main/listeners.ts
```diff
@@ -213,16 +212,17 @@ export function getDeleteAndCreateNewAttributionFileListener(
     const globalBackendState = getGlobalBackendState();
     const resourceFilePath = globalBackendState.resourceFilePath as string;
 
-    log.info(
-      'Deleting attribution file and opening input file: ',
-      resourceFilePath,
+    logger.info(
+      `Deleting attribution file and opening input file ${resourceFilePath.replace(
+        os.homedir(),
+        '~',
```
leslielazzarino, line 218: This is confusing for a Windows user.

src/ElectronBackend/input/importFromFile.ts
```diff
@@ -133,17 +131,17 @@ export async function loadInputAndOutputFromFilePath(
...
   );
 
-  log.info('Parsing frequent licenses from input');
+  logger.info('Parsing frequent licenses from input');
   const frequentLicenses = parseFrequentLicenses(
     parsedInputData.frequentLicenses,
   );
 
-  log.info('Sanitizing external resources to attributions');
+  logger.info('Sanitizing external resources to attributions');
```
leslielazzarino, line 139: This one was already wrong before, it should have been resources to external attributions. Anyway, the user knows external attributions as signals, so now that it is surfaced, it should read "Sanitizing resources to signals".

leslielazzarino (changes requested):
Cool popup.

src/ElectronBackend/input/importFromFile.ts
mstykow: I agree that speaking of an output file when there is only a .opossum file may be confusing.

I've changed it to just say "preparing output". Reason: at this point we're not writing any output to a file but we're just setting up the output.json inside the .opossum archive.

leslielazzarino (approved):

PR #2339, opened 2023-11-22
fix: preserve symlinks when zipping macOS release

mstykow:
### Summary of changes

- when zipping the macOS release, we were not correctly preserving symlinks thus duplicating the Electron framework three times
- reduces the zipped release size from ~300 to ~100MB and consequently also the unpacked app size in a similar fashion

Run `yarn ship-mac` and notice that the zipped app file size has been reduced by a factor of three compared to the main branch.


MarkusObendrauf (approved):
Nice!

PR #2342, opened 2023-11-23
Add original source field

vasily-pozdnyakov:
### Summary of changes

Original source name is shown in case if the additionalName is defined.

We would like to see original source for suggested (=MERGER) signals

See unit tests.
Add the field to any source of any input.json file and check it in the UI.

### DIFF
--- a/example-files/opossum_input.json
+++ b/example-files/opossum_input.json
@@ -156,7 +156,8 @@
     "bb61b094-b5ba-4bd0-b218-8ec96b7c5ae0": {
       "source": {
         "name": "MERGER",
-        "documentConfidence": 50.0
+        "documentConfidence": 50.0,
+        "additionalName": "ScanCode"
       },
       "attributionConfidence": 100,
       "comment": "TypeScript. \nhttps://www.npmjs.com/package/typescript",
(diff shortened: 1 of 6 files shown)


example-files/opossum_input.json
```diff
@@ -172,7 +173,8 @@
     "bb61b094-b5ba-4bd0-b218-8ec96b7c5af0": {
       "source": {
         "name": "MERGER",
-        "documentConfidence": 50.0
+        "documentConfidence": 50.0,
+        "additionalName": "1TEST Name"
```
mstykow, line 177: it seems to me that these example files try to use fairly realistic mock data. so perhaps you can find a better name of what one would expect as as additional name. something like "ScanCode"?

mstykow (approved):

src/Frontend/Components/AttributionColumn/__tests__/AttributionColumn.test.tsx
```diff
@@ -311,6 +311,41 @@ describe('The AttributionColumn', () => {
...
+    act(() => {
+      store.dispatch(setSelectedResourceId('test_id'));
+      store.dispatch(
+        setTemporaryDisplayPackageInfo(testTemporaryDisplayPackageInfo),
+      );
+    });
+
+    expect(
+      screen.getByDisplayValue(
+        (testTemporaryDisplayPackageInfo.source as Source)
```
mstykow, line 343:
```suggestion
        (testTemporaryDisplayPackageInfo.source!)
```

mstykow (changes requested):

example-files/opossum_input.json
vasily-pozdnyakov: actually, that should have not been committed. But ok, will rename it and keep.

mstykow (approved):

PR #2352, opened 2023-11-23
feat: make project statistics popup optional on startup

### LINKED ISSUE #2351: Make project statistics popup optional
As a user, I want to be able to choose if the project statistics popup opens every time I open a file or not.

Add a checkbox to the project statistics popup: "Show project statistics on startup"

**Additional context**
Persist user settings in backend, e.g. by using [this library](https://github.com/nathanbuchar/electron-settings).

mstykow:
### Summary of changes
- introduce electron app user settings
- allow user to toggle preference of showing project statistic on startup
- create state and hook for using a redux-persisted variable
- create hook for working with user settings
- create CLI arg to start app without project statistics and use it to simplify E2E tests

image

Closes #2351 .

Change the new checkbox inside the project statistics popup and observe the effect when opening a file.


src/ElectronBackend/main/__tests__/main.test.ts
```diff
@@ -1,117 +0,0 @@
...
-  dialog: {
-    showMessageBox: jest.fn(),
-  },
-  systemPreferences: {
-    setUserDefault: jest.fn(),
-  },
-}));
-
-describe('The App backend', () => {
-  it('calls ipc handler', async () => {
```
mstykow, line 54: this test did nothing more than assert that a certain line of code existed. in my opinion, this kind of test does not add to my confidence that the functionality of the app is as expected and merely makes it harder for me to work with/refactor the code.

it is also rather uncommon to unit test the main entry point to an electron app. probably better done using E2E tests.

src/Frontend/Components/ProjectStatisticsPopup/ProjectStatisticsPopup.tsx
```diff
@@ -141,6 +147,14 @@ export function ProjectStatisticsPopup(): ReactElement {
       onBackdropClick={close}
       onEscapeKeyDown={close}
       aria-label={'project statistics'}
+      customAction={
+        <Checkbox
+          checked={showProjectStatistics}
+          onChange={(event) => setShowProjectStatistics(event.target.checked)}
+          disabled={!hydrated}
```
MarkusObendrauf, line 154: what's this?

src/Frontend/test-helpers/render-component-with-store.tsx
```diff
@@ -1,12 +1,13 @@
+/* eslint-disable @typescript-eslint/explicit-function-return-type */
```
MarkusObendrauf, line 1: is this intentional?

src/Frontend/test-helpers/render-component-with-store.tsx
```diff
@@ -15,32 +16,49 @@ export interface EnhancedTestStore extends Store {
...
   return {
+    ...nativeRenderHook(callback, {
+      wrapper: ({ children }) => <Provider store={store}>{children}</Provider>,
+      initialProps,
+    }),
     store,
-    ...render(component, { wrapper: Wrapper, ...renderOptions }),
   };
-};
+}
```
MarkusObendrauf, line 64: i dont understand what this function does, could you explain it for me?

src/Frontend/util/__tests__/use-user-setting.test.ts
```diff
@@ -0,0 +1,82 @@
...
+    window.electronAPI = electronAPI as ElectronAPI;
+  });
+
+  it('returns default value when no stored value exist', async () => {
+    const defaultValue = faker.datatype.boolean();
+    const { result } = renderHook(() =>
+      useUserSetting({ key: 'showProjectStatistics', defaultValue }),
+    );
+
+    await waitFor(() => expect(result.current[2]).toBe(true));
```
MarkusObendrauf, line 31: what's `result.current[2]`? I don't understand at all what result is

MarkusObendrauf (approved):
Looks good, I just have some questions

src/Frontend/Components/ProjectStatisticsPopup/ProjectStatisticsPopup.tsx
mstykow: retrieving user settings from the file system is an asynchronous process. the `hydrated` flag indicates that the process has completed because other processes can/should not be blocked by the disk read operation. but until hydration is complete we also don't want to allow the user to make a change.

src/Frontend/test-helpers/render-component-with-store.tsx
mstykow: yes it was, but since the linter changes no longer necessary so it's gone now.

src/Frontend/test-helpers/render-component-with-store.tsx
mstykow: yes: react-testing-library offers a tool for testing hooks without needing to render them inside a component. but we often need a context around the hook, such as the redux store. so this function here combines the RTL `renderHook` tool with our custom context which ATM is just the redux provider. we can extend this function in the future if need be.

src/Frontend/util/__tests__/use-user-setting.test.ts
mstykow: the `result` is the return value of the hook we're testing. `result.current` is this value at the time that `current` is called. since the hook returns `[value, setValue, hydrated]`, `result.current[2]` is the value of `hydrated` at the time this line of code is executed.

MarkusObendrauf: I really don't like that, please make it more cleat in the code

MarkusObendrauf (changes requested):

MarkusObendrauf (approved):

PR #2356, opened 2023-11-24
Add original source field to manual attributions

MarkusObendrauf:
### Summary of changes

The original source name is kept when promoting a signal to an attribution.

We would like to see original source for suggested (=MERGER) signals

See updated unit & e2e test.
Take a file with an origin source, and promote it to an attribution. Check that the source is saved in the file, and that the file with the new fields can be loaded.


MarkusObendrauf:
Closed because we decided to change our approach.

We decided that we should not write the original source in the attribution. We want to avoid doing this because:

1. This would be duplicaetd info
2. It is not backwards compatible.

Instead, we take the following approach:
* Create a map originId → signal
* Create a function to look up the original source of an attribution by looking up the corresponding signal using the originId

Issue #2267, opened 2023-11-10
When a pre-selected attribution from a scan file has a 20% confidence, it does not change to 80% when confirmed


mstykow:
**Describe the bug**
The value of 20 is ignored during the save operation and not converted to 80 because it is one of only two values the user can choose manually. However, when the original input file has a 20-percent confidence for some attribution and there was no user choice involved, this 20 percent should also be converted to 80 upon saving/confirming the (pre-selected) attribution.

**To Reproduce**

1. Generate an opossum file with a pre-selected attribution that has a confidence of 20%.
2. Open file in opossum and confirm this attribution.
3. Notice that confidence has remained at 20%.

**Expected behavior**
Confidence should jump to 80%.

mstykow:
Fixed in #2358 by no longer overriding the confidence when an attribution is saved.

Issue #1572, opened 2023-04-13
Revise comment stack


lennartclaas:
The goal of this ticket is twofold:
- remove whitespace before and after a comment (signals only, not for (pre-selected) attributions)
- adjust the styling such that wrapping a long line does not introduce a second scroll bar

Screenshot that shows the issues that should be fixed:
![Screenshot 2023-04-13 092242](https://user-images.githubusercontent.com/83081698/231685300-8a111386-31cf-434a-821f-ee765d3361d8.png)

It is reasonable to fix this issue together with #1559.

mstykow:
#2357

mstykow:
#2357

Issue #1559, opened 2023-04-11
Comments not reaching the bottom of the UI


MarkusObendrauf:
The comment box does not reach the bottom of the window. Instead, there is some empty space between the box and the bottom of the UI.

![image](https://user-images.githubusercontent.com/107913663/231192686-5f5a6e82-ccfb-47cb-bac3-642a8feb5601.png)
![image](https://user-images.githubusercontent.com/107913663/231192790-610339c2-fb42-4894-a16a-894ad58c6957.png)

Expected behaviour: The comments box reaches the bottom of the screen, like the other text boxes:
![image](https://user-images.githubusercontent.com/107913663/231193036-f8be5e10-0e22-4974-a316-130fdad2e717.png)

mstykow:
#2357

PR #2358, opened 2023-11-24
feat: improve package info display

### LINKED ISSUE #2357: Improve package info display
There are several issues with the section displaying package info (either attribution or signal):
- cannot scroll vertically
- action buttons in bottom-right corner may overlap input fields
- text displaying additional info such as "marked for replacement" or "was pre-selected" floats seemingly without concept in mid-air between input fields and button row at the bottom
- #1559
- #1572
- copyright and license text fields influence each other (change size) when one or the other is opened

Clean up these design/concept/implementation issues.

mstykow:
### Summary of changes

- split attribution details into package coordinates and legal information
- redesign action buttons at bottom of attribution column to prevent them from overlapping content while allowing vertical scrolling in the attribution details
- make attribution column actions, especially replacing attributions, clearer
- introduce auditing options and labels at the top of the attribution column to highlight particular states of a signal or attribution
- allow user to set it to an arbitrary value on a scale from 1 to 5 for confidence
- rearrange package coordinate input fields to display most important fields first
- make first-party checkbox a toggle button between first and third party
- simplify multi-line text areas to not affect each other's display state

image

image

image

image

image

Closes #2357, #1559, #1572, #2267.

- Check the attribution column in audit view, attribution view, and the edit attribution popup in report view.
- Create first and third party attributions.
- Assign (multiple) auditing options, including modification of confidence.
- Replace attributions.


leslielazzarino (changes requested):
In general I like this PR. It needs some adaptations though.
1. The package name, and version potentially, are usually the most important part of an attribution, as far as identifying a piece of code goes. Now they are lost somewhere in the middle of the page. Imo they need to be the top text fields.
2. Moving the checkboxes to the top and making the comment field smaller leads, on my 1440p screen, to a cluttered look and half of the page being empty. Imo at the very least the check boxes and confidence+source need to go back where they were, and extra vertical space between some components should be used to group them semantically.

mstykow:
> In general I like this PR. It needs some adaptations though.
> 
> 1. The package name, and version potentially, are usually the most important part of an attribution, as far as identifying a piece of code goes. Now they are lost somewhere in the middle of the page. Imo they need to be the top text fields.
> 2. Moving the checkboxes to the top and making the comment field smaller leads, on my 1440p screen, to a cluttered look and half of the page being empty. Imo at the very least the check boxes and confidence+source need to go back where they were, and extra vertical space between some components should be used to group them semantically.

I've reworked the layout a bit and now there are no more checkboxes. Please have a look again. Probably best to also have a call instead of writing here back and forth.

leslielazzarino:
Some comments so far (I still need to look at the code):
1. I like the first/third party toggle, but I find the gray background looks too much like disabled. Can you use a shade of blue?
2. Personally I don't like the red delete button, in a blue color scheme. What about e.g. black?
3. I still think that two lines of permanent toggle-like chips (btw, I love the confidence one), preferred/was preferred/needs QA at the top, follow up/exclude from notice/confidence below, would look neater. But this could be postponed to "maybe later".

mstykow:
> See my previous comments.

1. I played around with a few color schemes but couldn't come up with anything that's not too intense but fits into the scheme of existing blue shades. Also bear in mind that the background is also already blue. I think the way it is right now is OK. I think any user will very quickly understand how this toggle button works and if there is any confusion as to whether the button is disabled or not they'll see it immediately when hovering over it.
2. I think the red button is fine. I used the color that's part of the OpossumUI color scheme/palette. Let's not waste any more time on this right now. What I really need is a code review.
3. This was already mentioned in a larger round/call where you were present. Nobody jumped onto the idea so let's keep it as is for now.

Concerning the usability of the chip menu: here too I would vote for keeping this as-is for now as we've already spent a lot of time discussing the pros and cons. By the way, GitHub labels actually work exactly the same way so I don't think that we're asking for something unpragmatic of the user.

image

So, please, let's focus on the code review now.

leslielazzarino (approved):

PR #2207, opened 2023-10-26
Add feature to directly open a file via ENV variable

meretp:
### Summary of changes

In addition to opening a file from a cli argument this commit adds the function to open a file by setting an environment variable. If the ENV variable OPOSSUM_FILE is set, the app will try to open this file directly when starting. To simplify the usage of this feature and make it independent of the operating system the yarn start command now accepts an optional argument and sets the environment variable when starting the app.
### Context and reason for change

With these changes the app can be started with `yarn start` (using vite) and directly open an opossum input file.
### How can the changes be tested

See the additional tests. 
Alternatively, run `yarn start <input_file>` where `<input_file>` is the absolute path to the file which should be opened. The app is now started and directly opens the provided file.

PR #2360, opened 2023-11-28
Add bash script to check out a PR and directly open a file in OpossumUI

meretp:
### Summary of changes

This PR adds a script to make development easier. 
```
Usage: ./tools/checkout_PR_and_start_opossum_ui.sh [OPTIONS]
Options:
 -p, Set the number of the PR to use
 -c, Set the commit hash to use
 -n, Set the name of the remote, default: origin
 -f, Set absolute path to file which should be opened
 -o, Checkout only, don't open the app
Note: either use -c or -p
Note: The script assumes to be run in the root directory of a local clone of the original repo

```

This script should help developers to test changes on large files,

Play around with the script.
Note that you need to store a duplicate of the script (for example change the name and don't add it to Git afterwards) as the script will check out a different PR where this script no  longer exists. Run e.g.
`./tools/checkout_PR_and_start_opossum_ui.sh -p 2359 -f ./example-files/opossum_input.json` . See that the respective PR is checked out and the OpossumUI is opened with the provided file.


tools/checkout_PR_and_start_opossum_ui.sh
meretp, line 1: I decided to put the script in a tools directory as this seems to be best practice (see https://medium.com/code-factory-berlin/github-repository-structure-best-practices-248e6effc405 and https://backrightup.com/blog/how-to-organize-github-repositories/)

mstykow (changes requested):

tools/checkout_PR_and_start_opossum_ui.sh
meretp: I agree "yet another folder" containing only one script is not the best approach. Concerning the naming I feel like we have two clashing opinions here (see https://github.com/opossum-tool/OpossumUI/pull/2207#discussion_r1395544931) but imho the best solution with respect to the content of the blogposts is to put the script in the yarn_scripts folder and name it tools.

mstykow (approved):

PR #2361, opened 2023-11-28
feat: make PURL read-only

### LINKED ISSUE #2346: PURL field should be read only
The PURL field should be read only and be calculated from its parts.
Having a PURL should not disable other input fields.

mstykow:
### Summary of changes

- calculate PURL from individual package attributes and make it read-only
- add buttons to paste/copy a PURL

image

Closes #2346.

- copy & paste PURLs


mstykow:
> The changes look good to me. Just one thing I noticed: it is possible to enter a space in type and name which results in a purl as in the screenshot. This is probably more related to validation of the input and not necessarily in the scope of this ticket but I think we shouldn't allow this. Actually, I checked that this is currently allowed as well, maybe we can create a follow-up ticket to add some validation for the PURL, what do you think? image

Good catch! Fixed.

meretp:
It is still possible to only enter a space in package name and package version and save this "empty" attribution. As said above, I am not sure if it's in the scope of this ticket but we should probably validate against empty output and not save empty (despite spaces) attributions.

PR #2362, opened 2023-11-28
feat: handle saving of first-party attributions

### LINKED ISSUE #2347: Handle first party
When a user marks a resource as 1st party, all inputs determining the PURL should be cleared and disabled (but keep the inputs until user actually hits "save" so that toggling does not lose input data).

mstykow:
### Summary of changes

- when saving a first-party attribution, delete third-party attributes
- simplify stripping functions

Closes #2347.

Modify a third-party attribution to become first-party and notice that the third-party attributes are removed after saving. That is, if after saving you switch back to third-party, you won't see the deleted values any longer. Previously, a save operation would not remove these values.


src/Frontend/util/get-stripped-package-info.ts
```diff
@@ -2,46 +2,90 @@
...
-  rawDisplayPackageInfo: DisplayPackageInfo,
-): DisplayPackageInfo {
-  const displayPackageInfoKeys = getDisplayPackageInfoKeys();
-  return pick(rawDisplayPackageInfo, displayPackageInfoKeys);
-}
+const strippedPackageInfoTemplate: {
+  [P in keyof Required<PackageInfo>]: boolean;
+} = {
+  ...strippedPackageInfoCoreTemplate,
+  comment: true,
```
meretp, line 78: Why is it comment here not comments?

src/Frontend/util/get-stripped-package-info.ts
```diff
@@ -2,46 +2,90 @@
...
+};
+
+const strippedDisplayPackageInfoTemplate: {
+  [P in keyof Required<DisplayPackageInfo>]: boolean;
+} = {
+  ...strippedPackageInfoCoreTemplate,
+  attributionIds: true,
+  comments: true,
```
meretp, line 86: Not sure if I'm mixing it up, but I think here it should be comment

meretp (changes requested):
Looks good to me, just one minor comment and a question which should be clarified before merging.

src/Frontend/util/get-stripped-package-info.ts
meretp: makes sense :D

PR #2372, opened 2023-12-04
fix: include all resources attached to an attribution when preferring globally

### LINKED ISSUE #2370: Preferring an attribution globally: Preferred over ids are not set correctly
**Describe the bug**
When preferring an attribution globally, the preferred over ids should contain the relevant attribution ids from all resources, that the attribution is attached to. Currently, only the resource that is selected when preferring the attribution globally, is taken into account.

**To Reproduce**
This is not a bug that is visible in the front end. The list of preferred over ids is potentially too short / the data is incomplete.

**Expected behavior**
All resources attached to an attribution should be taken into account when computing preferred over ids.

alexzurbonsen:
### Summary of changes

Use all resources that an attribution is attached to when computing the preferred over ids for that attribution if it gets preferred globally.

Fixes #2370 

Added unit test.


alexzurbonsen:
Closing in favour of #2374 to enable stacking.

Issue #2332, opened 2023-11-21
Improve experience of adding or editing attribution details


mstykow:
**Is your feature request related to a problem? Please describe.**
The current attribution wizard's value is not clear, it is hidden behind a context menu, and it is not intuitive to use.

The goal of this task is to replace it with an alternative of assigning an attribution that is less rigid than picking a signal directly but also less free than manually entering unrelated free text into individual input fields.

**Describe the solution you'd like**
1. Attribution wizard is replaced by autocomplete input fields in the attribution details.
2. Each autocomplete input (used for the fields making up the PURL) should offer:
    - custom data entry
    - a list of likely values to choose from (collected from the resource's signals, if any exist)
    - each offered suggestion should show additional context to see where it came from (name, version, source, license, ...)
    - suggestions coming from a grey star should be at the top of the list
4. The PURL input field is filled deterministically, becomes read-only and no longer disables individual input fields. We may even chose to delete it entirely (TBD).

**Additional requirements and context**
- Since the PURL is determined by the individual text inputs such as type, namespace, name, etc., we can generate it from the individual input fields which are more user friendly to fill and don't require the user to have PURL syntax knowledge. Thus, disabling individual inputs because of the presence of a PURL is exactly the wrong way around, from a UX/UI point-of-view.
- A PURL may contain additional information as part of qualifiers such as OS, architecture, distro, subpaths. Is this important for users of OpossumUI? And if so: how do we offer the user to enter such information, again without requiring the user to understand PURL syntax?
- The "Package Type" input field should not allow free-text but only offer a list of known package ecosystems with those from signals ranking first.
- When a user marks a resource as 1st party, all inputs determining the PURL should be cleared and disabled (but keep the inputs until user actually hits "save" so that toggling does not lose input data).
- If the feature to paste a PURL and thus generate an attribution that way is still desired, we can think of a solution for that too, but in a follow-up task.

See here for PURL syntax: https://github.com/package-url/purl-spec#purl

**Subtasks:**
- [x] #2343
- [x] #2344
- [x] #2345
- [ ] #2346
- [ ] #2347
- [x] #2348

maxhbr:
Not sure if that still accomplishes the goal to give an overview of the contained signals. The Attribution Wizard was build as a move to more package  oriented approach where directories are resolved. This proposal moves away from that again.

mstykow:
[...]

PR #2373, opened 2023-12-04
feat: use Opossum colors in theme palette

mstykow:
### Summary of changes

- to properly leverage the "color" prop on many MUI components, we set the theme palette to the values used by OpossumUI
- this also means that MUI correctly and automatically calculates the required states for `light`, `dark`, `contrastText` instead of us having to supply them: https://mui.com/material-ui/customization/palette/#customization

Following MUI best practices and improving button states. Also required as part of #2332.

Check buttons throughout OpossumUI that they still look as before but have improved hover/active states (calculated by MUI now).


mstykow:
> Member

Cleaning up the various button state styles is the main reason for this PR. Yes, this comes with some changes.

Previously, there were situations in which the contrast in particular states was not according to basic design principles, or where some states such disabled or hovered did not match the default state.

By giving the MUI palette a main color tone and then letting MUI calculate the derivatives, we can be sure that all tones match with each other harmoniously.

In the absence of a proper designer this is the best we can hope for and will hopefully avoid future discussions among developers about what color is best.

Thank you for the catch with the hamburger menu button. I aligned it now with the other buttons.

MarkusObendrauf (approved):
Ah, so grey is the new highlight color? got it.

Im not sure i prefer the new colors, but the code is better this way.

PR #2374, opened 2023-12-04
 fix: include all resources attached to an attribution when preferring globally

### LINKED ISSUE #2370: Preferring an attribution globally: Preferred over ids are not set correctly

alexzurbonsen:
### Summary of changes

Use all resources that an attribution is attached to when computing the preferred over ids for that attribution if it gets preferred globally.

Fixes #2370 

Added unit test.

### DIFF
--- a/src/Frontend/state/actions/resource-actions/__tests__/preference-actions.test.ts
+++ b/src/Frontend/state/actions/resource-actions/__tests__/preference-actions.test.ts
@@ -2,6 +2,7 @@
 // SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
 //
 // SPDX-License-Identifier: Apache-2.0
+import { faker } from '../../../../../shared/Faker';
 import { Resources } from '../../../../../shared/shared-types';
 import { getOriginIdsToPreferOver } from '../preference-actions';
 
@@ -154,5 +155,53 @@ describe('getOriginIdsToPreferOver', () => {
 
     const expectedOriginUuids = ['originUuid1'];
     expect(actualOriginUuids).toEqual(expectedOriginUuids);
+  });
+
+  it('finds all origin ids for attribution with multiple root resources', () => {
+    const attributionSource = faker.opossum.source();
+    const file1 = faker.opossum.resourceName();
+    const file2 = faker.opossum.resourceName();
+    const pathFile1 = faker.opossum.filePath(file1);
+    const pathFile2 = faker.opossum.filePath(file2);
+    const uuid1 = faker.string.uuid();
+    const uuid2 = faker.string.uuid();
+    const resources = faker.opossum.resources({
+      [file1]: 1,
+      [file2]: 1,
+    });
+    const pathsToRootResources = [pathFile1, pathFile2];
+    const resourcesToExternalAttributions =
+      faker.opossum.resourcesToAttributions({
+        [pathFile1]: [uuid1],
+        [pathFile2]: [uuid2],
+      });
+
+    const originUuid1 = faker.string.uuid();
+    const originUuid2 = faker.string.uuid();
+    const externalAttributions = faker.opossum.externalAttributions({
+      [uuid1]: { originIds: [originUuid1], source: attributionSource },
+      [uuid2]: { originIds: [originUuid2], source: attributionSource },
+    });
+
+    const externalAttributionSource = faker.opossum.externalAttributionSource({
+      isRelevantForPreferred: true,
+    });
+    const externalAttributionsSources =
+      faker.opossum.externalAttributionSources({
+        [attributionSource.name]: externalAttributionSource,
+      });
+
+    const resourcesToManualAttributions = {};
+
+    expect(
+      getOriginIdsToPreferOver(
+        pathsToRootResources,
+        resources,
+        resourcesToExternalAttributions,
+        resourcesToManualAttributions,
+        externalAttributions,
+        externalAttributionsSources,
+      ),
+    ).toEqual([originUuid1, originUuid2]);
   });
 });
--- a/src/Frontend/state/actions/resource-actions/preference-actions.ts
+++ b/src/Frontend/state/actions/resource-actions/preference-actions.ts
@@ -25,6 +25,7 @@ import {
   getResourcesToExternalAttributions,
   getResourcesToManualAttributions,
 } from '../../selectors/all-views-resource-selectors';
+import { getResourceIdsOfSelectedAttribution } from '../../selectors/attribution-view-resource-selectors';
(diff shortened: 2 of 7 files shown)


src/Frontend/state/actions/resource-actions/__tests__/preference-actions.test.ts
```diff
@@ -155,4 +156,53 @@ describe('getOriginIdsToPreferOver', () => {
     const expectedOriginUuids = ['originUuid1'];
     expect(actualOriginUuids).toEqual(expectedOriginUuids);
   });
+
+  it('finds all origin ids for attribution with multiple root resources', () => {
+    const testSource = faker.opossum.source();
+    const file1 = faker.opossum.resourceName();
+    const file2 = faker.opossum.resourceName();
+    const pathFile1 = `/${file1}`;
+    const pathFile2 = `/${file2}`;
```
mstykow, line 165: the e2e tests had the same problem of needing to create file paths from resource names. that's why faker actually provides a method `.filePath`. all you have to do is put the resource names into it where needed.

src/Frontend/state/actions/resource-actions/__tests__/preference-actions.test.ts
```diff
@@ -155,4 +156,53 @@ describe('getOriginIdsToPreferOver', () => {
...
+    const uuid1 = faker.string.uuid();
+    const uuid2 = faker.string.uuid();
+    const resources: Resources = faker.opossum.resources({
```
mstykow, line 168: the function is already fully typed so there is no need for re-declaring the type
```suggestion
    const resources = faker.opossum.resources({
```

src/Frontend/state/actions/resource-actions/__tests__/preference-actions.test.ts
```diff
@@ -155,4 +156,53 @@ describe('getOriginIdsToPreferOver', () => {
     const expectedOriginUuids = ['originUuid1'];
     expect(actualOriginUuids).toEqual(expectedOriginUuids);
   });
+
+  it('finds all origin ids for attribution with multiple root resources', () => {
+    const testSource = faker.opossum.source();
```
mstykow, line 161: we don't gain anything by prefixing variables in test with "test" or "mock".
```suggestion
    const testSource = faker.opossum.source();
```

src/Frontend/state/actions/resource-actions/__tests__/preference-actions.test.ts
```diff
@@ -155,4 +156,53 @@ describe('getOriginIdsToPreferOver', () => {
...
+    const actualOriginUuids = getOriginIdsToPreferOver(
+      pathsToRootResources,
+      resources,
+      resourcesToExternalAttributions,
+      resourcesToManualAttributions,
+      externalAttributions,
+      externalAttributionsSources,
+    );
+
+    const expectedOriginUuids = [originUuid1, originUuid2];
```
mstykow, line 205: why not just inline this in the next line?

src/Frontend/state/actions/resource-actions/__tests__/preference-actions.test.ts
```diff
@@ -155,4 +156,53 @@ describe('getOriginIdsToPreferOver', () => {
...
+      isRelevantForPreferred: true,
+    });
+    const externalAttributionsSources =
+      faker.opossum.externalAttributionSources({
+        [testSource.name]: externalAttributionSource,
+      });
+
+    const resourcesToManualAttributions = {};
+
+    const actualOriginUuids = getOriginIdsToPreferOver(
```
mstykow, line 196: how about just inlining this in the assertion? so
```
expect(getOriginIdsToPreferOver(...)).toEqual([..., ...])
```

src/Frontend/state/actions/resource-actions/preference-actions.ts
```diff
@@ -57,35 +63,64 @@ export function toggleIsSelectedPackagePreferred(
   };
 }
 
+export function setOriginIdsToPreferOverGlobally(
```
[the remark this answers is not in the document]
mstykow: it's fine. you're following the established pattern...

src/Frontend/state/actions/resource-actions/preference-actions.ts
```diff
@@ -57,35 +58,64 @@ export function toggleIsSelectedPackagePreferred(
...
+  packageInfo: DisplayPackageInfo,
+): AppThunkAction {
+  return (dispatch: AppThunkDispatch, getState: () => State): void => {
+    const state = getState();
+
+    const newTemporaryDisplayPackageInfo = cloneDeep(packageInfo);
```
mstykow, line 67: i don't think we need to clone anything deeply here, do we? could this work as well?
```
dispatch(
      setTemporaryDisplayPackageInfo({
        ...packageInfo,
        preferredOverOriginIds: getOriginIdsToPreferOver(
          getResourceIdsOfSelectedAttribution(state) ?? [],
          getResources(state) ?? {},
          getResourcesToExternalAttributions(state),
          getResourcesToManualAttributions(state),
          getExternalAttributions(state),
          getExternalAttributionSources(state),
        ),
      }),
    );
```

mstykow (changes requested):
Thank you for creating this separate PR and associated issue. I understand the problem now. However, could you also explain why the current behavior needs fixing at all? After all, if this bug cannot be seen inside OpossumUI then at what point is there actually any issue at all?

src/Frontend/state/actions/resource-actions/__tests__/preference-actions.test.ts
alexzurbonsen: true, was a left over of the old test naming conventions...

src/Frontend/state/actions/resource-actions/preference-actions.ts
alexzurbonsen: I wasn't sure either. My assumption was that it is not necessary but then I wasn't sure I understood enough of react hooks and if this change could trigger any rerendering etc. I then followed the example from the action `toggleIsSelectedPackagePreferred`.

alexzurbonsen: Though about it again, and I think I am a bit confused about what changing a field in an object can actually trigger... In case there is anything to keep in mind, should we have a short chat? Otherwise I would just remove the deepclone.

PR #2376, opened 2023-12-04
 feat: enable preferring attributions in attribution view

alexzurbonsen:
### Summary of changes

In this PR we enable preferring attributions in the attribution view. If an attribution has multiple resources, a warning is shown and the user is asked for confirmation.

This PR also fixes a bug in the previous implementation of preferring globally. The `originIds`, that an attribution is preferred over, weren't computed for all resources of the attribution. Instead, only the resource that was selected when the attribution was preferred globally, was taken into account.

In #2270 preferring attributions globally was enabled. Consequently, it should be possible to prefer attributions in attribution view as well.

added two e2e tests
and unit test that checks if the correct origin ids are computed if an attribution is attached to multiple resources


src/Frontend/Components/AttributionDetailsViewer/AttributionDetailsViewer.tsx
```diff
@@ -57,26 +58,43 @@ export function AttributionDetailsViewer(): ReactElement | null {
...
+      resourceIdsOfSelectedAttributionId.length > 1
+    ) {
+      dispatch(openPopup(PopupType.ChangePreferredStatusGloballyPopup));
     } else {
       dispatch(
         savePackageInfoIfSavingIsNotDisabled(
           null,
           selectedAttributionId,
-          temporaryDisplayPackageInfo,
+          convertDisplayPackageInfoToPackageInfo(temporaryDisplayPackageInfo),
```
alexzurbonsen, line 78: this fixes the comment saving bug in attribution view

src/Frontend/state/selectors/__tests__/all-views-resource-selectors.test.ts
```diff
@@ -195,3 +199,93 @@ describe('get displayPackageInfo', () => {
     expect(testDisplayPackageInfo).toEqual(expectedDisplayPackageInfo);
   });
 });
+
```
alexzurbonsen, line 202: This test is copied over, since I moved the action `getResourceIdsForSelectedAttributionId` to all views (where I think it belongs to).

src/Frontend/state/selectors/all-views-resource-selectors.ts
```diff
@@ -248,12 +249,6 @@ export function getIsPreferenceFeatureEnabled(state: State): boolean {
   return state.resourceState.allViews.isPreferenceFeatureEnabled;
 }
 
-export function getSelectedAttributionIdInAttributionView(
-  state: State,
-): string {
-  return state.resourceState.attributionView.selectedAttributionId;
-}
-
```
alexzurbonsen, line 256: moving this to attribution view selectors.

src/Frontend/Components/AttributionColumn/AttributionColumn.tsx
```diff
@@ -222,7 +222,7 @@ export function AttributionColumn(props: AttributionColumnProps): ReactElement {
     {
       buttonText: ButtonText.MarkAsPreferred,
       onClick: (): void => {
-        if (selectedPackage) {
+        if (selectedManualAttributionIdInCurrentView) {
```
alexzurbonsen, line 225: selectedPackage is an audit view state, thus I changed it to use the attribution id directly. In audit view this also returns a manual attribution id if the package is from 'All Attributions' or 'Attributions contained in folder content' panel. But since the hamburger menu is only visible if a manual attribution in attribution panel is modified, this should be fine, I think. Or should we be more specific here? i.e. just return an id for attributions from "Attributions" panel in audit view.

mstykow: in my open PRs i deleted the hamburger menu. i think we should discuss what to do here in a call.

mstykow (changes requested):

mstykow (approved):

src/e2e-tests/__tests__/saving-preferred-attributions-globally.test.ts
```diff
@@ -268,3 +268,58 @@ test('popup is only shown if necessary and undo reverts preference status only',
     );
   });
 });
+
+test('popup if attribution is marked as preferred and saved in attribution view', async ({
```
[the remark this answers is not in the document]
alexzurbonsen: good point, missed that.

alexzurbonsen:
fixed test names

PR #2378, opened 2023-12-04
fix: warn user of unsaved changes if file is opened via button in top bar

### LINKED ISSUE #2234: Unsaved popup not shown when opening a new file
**Describe the bug**
When unsave changes are present and the user clicks on open file, the unsaved popup is not shown.

**To Reproduce**

1. Open a file
2. open/create an attribution
3. modify the content, without saving
4. Click on the open file icon (top-left)
5. No unsaved popup is shown

**Expected behavior**
By step 5. the usaved popup is shown, allowing the user not to loose their changes, if so desired.
 
**Desktop (please complete the following information):**
All

alexzurbonsen:
### Summary of changes

The user is warned about unsaved changes if he clicks the "open file" button in the top bar.
Buttons of the warning popup show the following behaviour:
Cancel -> changes are not saved and open file dialog is not displayed
Undo -> changes are discarded and open file dialog is displayed
Save -> popups that apply are shown (was preferred or prefer globally warning), changes are saved eventually and open file dialog is displayed in the end

Closes #2234.

This is a full solution for the bug described in #2234, but only a partial solution to the problem. There is also the possibility to open a file via the menu. Opened a follow up ticket for this: #2379. (It seems to be a bit more effort, because it is handled entirely in the backend. Channels for frontend communication need to be introduced.)

Added an e2e test that tests paths leading up to the system dialog for opening a file, but doesn't include it in the test.

### DIFF
--- a/src/Frontend/state/actions/popup-actions/popup-actions.ts
+++ b/src/Frontend/state/actions/popup-actions/popup-actions.ts
@@ -39,7 +39,11 @@ import {
   getResolvedExternalAttributions,
   getSelectedResourceId,
 } from '../../selectors/audit-view-resource-selectors';
-import { getSelectedView, getTargetView } from '../../selectors/view-selector';
+import {
+  getOpenFileRequest,
+  getSelectedView,
+  getTargetView,
+} from '../../selectors/view-selector';
 import { AppThunkAction, AppThunkDispatch } from '../../types';
 import { setTemporaryDisplayPackageInfo } from '../resource-actions/all-views-simple-actions';
 import {
@@ -73,6 +77,7 @@ import {
   closePopup,
   navigateToView,
   openPopup,
+  setOpenFileRequest,
   setShowNoSignalsLocatedMessage,
   setTargetView,
 } from '../view-actions/view-actions';
@@ -183,7 +188,7 @@ export function unlinkAttributionAndSavePackageInfoAndNavigateToTargetViewIfSavi
         temporaryDisplayPackageInfo,
       ),
     );
-    dispatch(navigateToTargetResourceOrAttribution());
+    dispatch(navigateToTargetResourceOrAttributionOrOpenFileDialog());
   };
 }
 
@@ -228,7 +233,7 @@ export function saveTemporaryDisplayPackageInfoAndNavigateToTargetViewIfSavingIs
         convertDisplayPackageInfoToPackageInfo(temporaryDisplayPackageInfo),
       ),
     );
-    dispatch(navigateToTargetResourceOrAttribution());
+    dispatch(navigateToTargetResourceOrAttributionOrOpenFileDialog());
   };
 }
 
@@ -267,10 +272,19 @@ export function checkIfWasPreferredOrPreferredStatusChangedAndShowWarningOrSave(
   };
 }
 
-export function navigateToTargetResourceOrAttribution(): AppThunkAction {
+export function navigateToTargetResourceOrAttributionOrOpenFileDialog(): AppThunkAction {
   return (dispatch: AppThunkDispatch, getState: () => State): void => {
     const targetView = getTargetView(getState());
     const view = getSelectedView(getState());
+    const openFileRequest = getOpenFileRequest(getState());
+
+    dispatch(closePopup());
+    if (openFileRequest) {
+      void window.electronAPI.openFile();
+      dispatch(setOpenFileRequest(false));
+      return;
+    }
+
     dispatch(setSelectedResourceOrAttributionIdToTargetValue());
     if (targetView) {
       dispatch(navigateToView(targetView));
@@ -282,7 +296,6 @@ export function navigateToTargetResourceOrAttribution(): AppThunkAction {
       ),
     );
 
-    dispatch(closePopup());
     if (view === View.Report) {
       dispatch(closePopup());
     }
@@ -295,6 +308,7 @@ export function closePopupAndUnsetTargets(): AppThunkAction {
     dispatch(setTargetSelectedResourceId(''));
     dispatch(setTargetSelectedAttributionId(''));
     dispatch(closePopup());
+    dispatch(setOpenFileRequest(false));
   };
 }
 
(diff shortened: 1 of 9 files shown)


src/e2e-tests/__tests__/saving-attributions.test.ts
```diff
@@ -313,3 +313,58 @@ test('warns user of unsaved changes if user attempts to navigate away before sav
   await topBar.gotoAuditView();
   await notSavedPopup.assert.isVisible();
 });
+
+const [attributionId3, packageInfo3] = faker.opossum.manualAttribution({
+  packageType: undefined,
+  wasPreferred: true,
+});
+
+test.use({
```
alexzurbonsen, line 322: is it okay to redeclare test data in a file or is that a sign to move the test to a new file?

src/e2e-tests/__tests__/open-file-warnings.test.ts
```diff
@@ -0,0 +1,61 @@
...
+import { test } from '../utils';
+
+const [resourceName1, resourceName2] = faker.opossum.resourceNames({
+  count: 4,
+});
+const license1 = faker.opossum.license();
+const license2 = faker.opossum.license();
+const [attributionId1, packageInfo1] = faker.opossum.manualAttribution({
+  packageType: undefined,
+  wasPreferred: true,
```
mstykow, line 15: why is this test also mixing in the `wasPreferred` attribute? if this is important, i'd say we should mention something about it in the test title. if it's not important, i'd suggest to simplify the test by removing this attribute here.

src/e2e-tests/__tests__/open-file-warnings.test.ts
```diff
@@ -0,0 +1,61 @@
+// SPDX-FileCopyrightText: Meta Platforms, Inc. and its affiliates
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+import { faker } from '../../shared/Faker';
+import { test } from '../utils';
+
+const [resourceName1, resourceName2] = faker.opossum.resourceNames({
+  count: 4,
```
mstykow, line 9:
```suggestion
  count: 2,
```

src/e2e-tests/__tests__/open-file-warnings.test.ts
```diff
@@ -0,0 +1,61 @@
...
+});
+const license1 = faker.opossum.license();
+const license2 = faker.opossum.license();
+const [attributionId1, packageInfo1] = faker.opossum.manualAttribution({
```
mstykow, line 13: nitpick:
```suggestion
const [attributionId, packageInfo] = faker.opossum.manualAttribution({
```

mstykow (approved):

src/Frontend/state/actions/popup-actions/popup-actions.ts
```diff
@@ -270,10 +275,19 @@ export function checkIfWasPreferredAndShowWarningOrSave(): AppThunkAction {
...
+export function navigateToTargetResourceOrAttributionOrOpenFileDialog(): AppThunkAction {
   return (dispatch: AppThunkDispatch, getState: () => State): void => {
     const targetView = getTargetView(getState());
     const view = getSelectedView(getState());
+    const openFileRequest = getOpenFileRequest(getState());
+
+    dispatch(closePopup());
+    if (openFileRequest) {
+      void window.electronAPI.openFile();
+      dispatch(setOpenFileRequest(false));
```
mstykow, line 287: isn't there a problem here? what if the popup opens but i click cancel. then `openFileRequest` in the reducer remains `true` although no open-file request is is progress any longer.

so the next time i have unsaved changes and just want to navigate somewhere without opening a file, the open-file dialog will still appear.

mstykow (changes requested):

src/Frontend/state/actions/popup-actions/popup-actions.ts
alexzurbonsen: Oh yes, you are right...

alexzurbonsen: Added the same logic to the cancel handler that is called in all the popups involved in saving (NotSaved, ModifyWasPreferred, ChangePreferredGlobally).

src/e2e-tests/__tests__/open-file-warnings.test.ts
alexzurbonsen: I thought it might be valueable to test if the process is working if a "follow up" popup is shown after the not saved popup. Setting `wasPreferred` is an easy way to do it.
I will clarify the test title.

src/e2e-tests/__tests__/open-file-warnings.test.ts
```diff
@@ -0,0 +1,56 @@
...
+import { faker } from '../../shared/Faker';
+import { test } from '../utils';
+
+const [resourceName1, resourceName2] = faker.opossum.resourceNames({
+  count: 2,
+});
+const license1 = faker.opossum.license();
+const license2 = faker.opossum.license();
+const [attributionId, packageInfo] = faker.opossum.manualAttribution({
+  packageType: undefined,
```
mstykow, line 14: is it important the package type is undefined here? if not, i'd say we also remove this line.
same with the licenses: it seems that we don't need them. by defining them we're making it seem like the test somehow depends on this setup.

src/Frontend/state/actions/popup-actions/popup-actions.ts
```diff
@@ -282,19 +296,23 @@ export function navigateToTargetResourceOrAttribution(): AppThunkAction {
...
 export function closePopupAndUnsetTargets(): AppThunkAction {
-  return (dispatch: AppThunkDispatch): void => {
+  return (dispatch, getState): void => {
+    const openFileRequest = getOpenFileRequest(getState());
+
     dispatch(setTargetView(null));
     dispatch(setTargetSelectedResourceId(''));
     dispatch(setTargetSelectedAttributionId(''));
     dispatch(closePopup());
+    if (openFileRequest) {
```
mstykow, line 313: is this if-check really necessary? it seems we could always set the value back to false.

mstykow (approved):

src/Frontend/state/actions/popup-actions/popup-actions.ts
alexzurbonsen: True, that may be better.

src/e2e-tests/__tests__/open-file-warnings.test.ts
alexzurbonsen: Copy paste... sorry.

PR #2393, opened 2023-12-06
feat: run tests with babel compiler

mstykow:
### Summary of changes

- faster and simpler: runs unit tests almost twice as fast as `ts-jest`

Locally: `yarn test:unit`
Notice that pipeline unit test step only takes 55 secs to complete (previously 1:35 min ).


meretp (approved):
LGTM

PR #2398, opened 2023-12-08
feat: merge accordion web workers into signals worker

mstykow:
### Summary of changes

- merge accordion workers that calculate signals into signals worker to save on memory (they all need the same input data)
- enable useVariable to provide the previous store value when setting the new value (similar to useState dispatch)

#2384 

As you're switching between resources in audit view, look at the memory tab in the devtools to see that the memory footprint has been reduced compared to the base branch.


davidmundelius (approved):
Looks good to me.

PR #2402, opened 2023-12-08
feat: merge progress data workers into signal worker

mstykow:
### Summary of changes

- merge progress data workers into signals worker to save on memory (they mostly need the same input data)
- delete any remnants of separate workers and their context providers

#2384 

As you're switching between resources in audit view and check that the progress bars are still working as expected, look at the memory tab in the devtools to see that the memory footprint has been reduced compared to the base branch.


src/Frontend/web-workers/signals-worker.ts
```diff
@@ -61,9 +77,11 @@ export class SignalsWorker {
 
   public processInput(input: SignalsWorkerInput) {
     this.setData(input);
-    this.dispatchAutocompleteSignals(input);
+    this.dispatchOverallProgressData(input);
+    this.dispatchFolderProgressData(input);
     this.dispatchAttributionsInFolderContent(input);
     this.dispatchSignalsInFolderContent(input);
+    this.dispatchAutocompleteSignals(input);
```
MarkusObendrauf, line 84: we should re-arrange this to do the most important stuff first. suggestion:
```
    this.dispatchSignalsInFolderContent(input);
    this.dispatchAttributionsInFolderContent(input);
    this.dispatchFolderProgressData(input);
    this.dispatchOverallProgressData(input);
    this.dispatchAutocompleteSignals(input);
```

MarkusObendrauf (changes requested):
Nice change! Performance is good.

src/Frontend/web-workers/signals-worker.ts
mstykow: i already ordered these items the way i thought they are the most important. my reason for putting the progress data first was that i can see the top progress bar first, independent of the selected resource. but thinking about it again now, i think that you're right that computing the signals for the middle column is more important so i'll change it.

src/Frontend/Components/ProgressBar/__tests__/TopProgressBar.test.tsx
```diff
@@ -2,238 +2,84 @@
...
+          filesWithOnlyExternalAttributionCount: 3,
+          filesWithOnlyPreSelectedAttributionCount: 1,
+          resourcesWithMediumCriticalExternalAttributions: [],
+          resourcesWithNonInheritedExternalAttributionOnly: [],
+          resourcesWithHighlyCriticalExternalAttributions: [],
+        }),
+      ],
+    });
+
+    await userEvent.click(screen.getByRole('checkbox'), {
```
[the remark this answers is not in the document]
MarkusObendrauf: I still dont like it but ill live with it

MarkusObendrauf (approved):

Issue #2394, opened 2023-12-07
Add package name suggestions for packages (via deps.dev)


mstykow:
- use this endpoint to load suggestions based on query string: https://deps.dev/_/search/suggest?q=react
- note: this always returns at most 10 results which include package name and package type
- ignore any projects for now (deps.dev distinguishes between packages and projects, where projects are GitHub, GitLab, Bitbucket)
- ignore version and license name for now
- add the suggestions coming from deps.dev to the end of the autocomplete suggestions
- to handle the loading state, use a simple skeleton placeholder for three search results based on the existing suggestion card
- if an error occurs during the network request, just return an empty array of results

davidmundelius:
Will split this in three PRs:

- [ ] Bypass CORS policy in backend
- [ ] Set up react query hook for loading suggestions based on search term
- [ ] Show suggestions or loading state

PR #2415, opened 2023-12-13
feat: offer package name suggestions

davidmundelius:
### Summary of changes

We add the suggestions from deps.dev to the autocomplete for the package name field.

#2394 

Create a new attribution; type "react" for the package name and check that there are nine suggestions from deps.dev.


src/Frontend/Components/AttributionColumn/PackageSubPanel.tsx
```diff
@@ -96,6 +121,7 @@ export function PackageSubPanel(props: PackageSubPanelProps) {
         highlight={'dark'}
         disabled={!props.isEditable}
         showHighlight={props.showHighlight}
+        defaults={packageNames ? packageNames : []}
```
mstykow, line 124:
```suggestion
        defaults={packageNames}
```

src/Frontend/util/package-search-hooks.ts
```diff
@@ -7,11 +7,10 @@ import { useQuery } from '@tanstack/react-query';
 import { PackageInfo } from '../../shared/shared-types';
 import PackageSearchApi from './package-search-api';
 
-function usePackageNames({ packageName }: PackageInfo) {
+function usePackageNames(packageName?: string) {
   const { data, error, isLoading } = useQuery({
     queryKey: ['package-name-suggestions', packageName],
-    queryFn: () =>
-      packageName ? PackageSearchApi.getPackages(packageName) : {},
+    queryFn: () => PackageSearchApi.getPackages(packageName),
```
mstykow, line 13: 👍

src/Frontend/util/package-search-api.ts
```diff
@@ -63,25 +63,30 @@ export class PackageSearchApi {
   }
 
   public async getPackages(
-    packageName: string,
+    packageName?: string,
```
mstykow, line 66: let's change this to have the same signature as other methods in this class:
```suggestion
     props: PackageInfo
```
and then use `this.sanitize` to sanitize your input.

mstykow (changes requested):

src/Frontend/util/package-search-api.ts
davidmundelius: Done.

src/Frontend/Components/AttributionColumn/PackageSubPanel.tsx
davidmundelius: Done.

src/Frontend/Components/AttributionColumn/PackageSubPanel.tsx
```diff
@@ -53,6 +53,22 @@ interface PackageSubPanelProps {
   showHighlight?: boolean;
 }
 
+function useDebounceInput<T>(input: T, delay?: number): T {
```
mstykow, line 56: let's move this to a util file in the `util` directory and then import it here and also in the package search popup. this is quite a common use case and it's likely to come up again elsewhere.

mstykow (changes requested):

src/Frontend/Components/AttributionColumn/PackageSubPanel.tsx
mstykow: i'd also suggest you rename this to `useDebouncedInput`. the idea is this: `const foo = useFoo`, and the returned value from this hook is a "debounced" input and not an action or function that "debounce"s.

src/Frontend/util/__tests__/package-search-api.test.ts
```diff
@@ -111,4 +112,39 @@ describe('PackageSearchApi', () => {
       );
     });
   });
+
+  describe('getPackages', () => {
+    it('provides packages but no projects', async () => {
+      const packageInfo = faker.opossum.externalPackageInfo({
+        packageType: faker.packageSearch.system(),
+      });
+      const testPackage = faker.packageSearch.searchResponse({
```
mstykow, line 121: there is nothing gained here by prefixing variables with "test" or "mock".
```suggestion
      const testPackage = faker.packageSearch.searchResponse({
```

src/Frontend/util/use-debounced-input.ts
```diff
@@ -0,0 +1,21 @@
+// SPDX-FileCopyrightText: Meta Platforms, Inc. and its affiliates
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+import { useEffect, useState } from 'react';
+
+export function useDebouncedInput<T>(input: T, delay?: number): T {
```
mstykow, line 7: without a positive default delay this function really does not make much sense. this will then allow us to avoid defining this constant everywhere where we want to use the hook.
```suggestion
export function useDebouncedInput<T>(input: T, delay = 500): T {
```

mstykow (changes requested):

src/Frontend/util/use-debounced-input.ts
mstykow: also, this might be a better default: https://humanbenchmark.com/tests/reactiontime

davidmundelius: Done. Had do rewrite this a little bit to avoid a no-magic-numbers linter error.

src/Frontend/util/use-debounced-input.ts
```diff
@@ -0,0 +1,23 @@
+// SPDX-FileCopyrightText: Meta Platforms, Inc. and its affiliates
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+import { useEffect, useState } from 'react';
+
+const default_delay = 500;
```
mstykow, line 7: global constants in typescript are usually all uppercase:
```suggestion
const DEFAULT_DELAY = 500;
```

mstykow (approved):

PR #2416, opened 2023-12-13
fix: CORS headers may be duplicated

mstykow:
### Summary of changes

- due to casing, we may set the access-control-allow-origin header to "*" twice
- fix this by overwriting one of the casings

Bug introduced in #2404.

Open the package search popup and notice that no errors are happening compared to the base branch.


davidmundelius (approved):
Looks good to me.

PR #2421, opened 2023-12-14
feat: Create SPDX document and add it to release

antonbauhofer:
Create an SPDX document for the project and add it to the release

### Summary of changes

- Add the Yarn plugin [@yarnpkg/plugin-spdx](https://github.com/spdx/yarn-plugin-spdx)
- Use the plugin in CI/CD workflow `publish-release` to generate an SPDX document for the project and add it to the release
- Use the plugin in CI/CD workflow `build-and-e2e-test`

It is desirable for the OpossumUI releases to contain an SBOM document.

Run the command `yarn spdx` verify that an SPDX document with the name opossum-ui.spdx.json is created in the root of the project and inspect it.
To visualize the generated document, the [SPDX Python tooling](https://github.com/spdx/tools-python) can be used. To do so, use the command pyspdxtools -I <local path to generated spdx document> --graph -o <output path>.
opossum-ui

### Note

The generated SPDX document is currently invalid because the dependency `babel-plugin-transform-import-meta` has the license `BSD` declared, which is not in the [SPDX license list](https://spdx.org/licenses/).
There is a ticket for this in the plugin implementation: https://github.com/spdx/yarn-plugin-spdx/issues/12
We should wait for this to be fixed before landing this PR.


mstykow:
@antonbauhofer how can i tell that the SPDX doc is invalid, as you say? i would have expected an error to occur in the pipeline here telling me that `babel-plugin-transform-import-meta` uses an unexpected license.

antonbauhofer:
> @antonbauhofer how can i tell that the SPDX doc is invalid, as you say? i would have expected an error to occur in the pipeline here telling me that `babel-plugin-transform-import-meta` uses an unexpected license.

The plugin itself does currently not validate the created SPDX documents (at least not completely). So, the pipeline only verifies, that the creation of the SPDX document is successful and not the validity of the document itself.
To check the validity of the SPDX document, one can for example use the [SPDX Python tooling](https://github.com/spdx/tools-python).
For now, I think we can still use it as it is (an invalid SPDX document is still better than none).
This issue will be fixed in the future in the [@spdx/tools](https://github.com/spdx/tools-ts) library that the yarn plugin uses.

antonbauhofer:
Issue for the feature request that would make the SPDX document valid: 
https://github.com/spdx/tools-ts/issues/227

mstykow (approved):

.yarn/plugins/@yarnpkg/plugin-spdx.cjs
```diff
@@ -0,0 +1,17 @@
+// SPDX-FileCopyrightText: Meta Platforms, Inc. and its affiliates
```
maxhbr, line 1: this are the wrong copyrights. This should just be `//  SPDX-FileCopyrightText: 2023 SPDX contributors`, as it just contains IP from SPDX.

.github/workflows/build-and-e2e-test.yml
```diff
@@ -44,6 +44,8 @@ jobs:
 
       - run: yarn ship:auto
 
+      - run: yarn spdx
```
maxhbr, line 47: Why is it ran in the `e2e-test` without any assertions or check?

maxhbr (changes requested):
Please fix the copyright.

.github/workflows/build-and-e2e-test.yml
antonbauhofer: The idea here was to verify that running the command `yarn spdx` does not fail (same as `yarn ship:auto`).
It would be nice to validate the created document here, but at this point I think it would be overkill to use another tool for validation.
Alternatively, we could also remove it from here.
What do you think?

maxhbr: You could validate that the expected file exists or that the file contains the name of the opossum tool. But might not be necessary.

.yarn/plugins/@yarnpkg/plugin-spdx.cjs
antonbauhofer: So, the complete header will then be

// SPDX-FileCopyrightText: 2023 SPDX contributors
//
// SPDX-License-Identifier: MIT

Since the plugin is licensed under MIT.
Consequently, we then need to add MIT to the licenses.

Do you agree?

.github/workflows/build-and-e2e-test.yml
antonbauhofer: Sounds good. I'll do that 👍

maxhbr (approved):
LGTM

Issue #2426, opened 2023-12-15
Offer namespace suggestions


mstykow:
Provide namespace suggestions based on package name and type. Only relevant for maven, github, gitlab, bitbucket.

PR #2427, opened 2023-12-15
feat: offer namespace suggestions

mstykow:
### Summary of changes

- offer namespace suggestions, when relevant, based on package name and type
- do not highlight package version and URL in attribution view since they are not essential information when determining legal information (it's rare albeit not impossible that different versions have different licenses)
- remove advisories from search results

image

#2426 

Namespace suggestions only occur for maven, github, gitlab, bitbucket.


src/Frontend/util/is-important-attribution-information-missing.ts
```diff
@@ -32,8 +32,6 @@ export function isImportantAttributionInformationMissing(
     case 'licenseName':
     case 'packageName':
     case 'packageType':
-    case 'packageVersion':
-    case 'url':
```
mstykow, line 36: discuss with @benedikt-richter

alexzurbonsen (approved):

PR #2432, opened 2023-12-18
chore: simplify unit test setup

mstykow:
### Summary of changes

- remove unnecessary test blocks and setup from setupTests.ts
- add linter rules to ensure that it/test and jest.spyon is correctly used

More test consistency and transparency.

Run unit tests.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -37,6 +37,8 @@ const sharedRules = {
     { allowBoolean: true, allowNumber: true, allowNever: true },
   ],
   '@typescript-eslint/switch-exhaustiveness-check': 'error',
+  'jest/consistent-test-it': 'error',
+  'jest/prefer-spy-on': 'error',
   'filenames/match-regex': [
     'error',
     '^[a-zA-Z0-9\\-]+(.d|.test|.style)?$',
--- a/src/ElectronBackend/main/__tests__/listeners.test.ts
+++ b/src/ElectronBackend/main/__tests__/listeners.test.ts
@@ -566,10 +566,6 @@ describe('getSelectBaseURLListener', () => {
 });
 
 describe('getSaveFileListener', () => {
-  beforeEach(() => {
-    (writeFile as jest.Mock).mockReset();
-  });
-
   it('throws error when projectId is not set', async () => {
     const mockCallback = jest.fn();
     const mainWindow = {
@@ -830,10 +826,6 @@ describe('getExportBomListener', () => {
 });
 
 describe('getExportSpdxDocumentListener', () => {
-  beforeEach(() => {
-    (writeSpdxFile as jest.Mock).mockReset();
-  });
-
   it('throws if path is not set', async () => {
     const mainWindow = await prepareBomSPdxAndFollowUpit();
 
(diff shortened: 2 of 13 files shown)


src/ElectronBackend/main/__tests__/listeners.test.ts
```diff
@@ -566,10 +566,6 @@ describe('getSelectBaseURLListener', () => {
 });
 
 describe('getSaveFileListener', () => {
-  beforeEach(() => {
```
alexzurbonsen, line 569: just checking: from reading the documentation I learned that .mockReset() is doing more than .mockClear(), i.e. also setting the mock definition to undefined. But it seems our test don't rely on this?

src/Frontend/Components/ResourcesTree/__tests__/resources-tree-helpers.test.ts
```diff
@@ -34,63 +34,33 @@ describe('splitResourceItToCurrentAndOtherFolder', () => {
...
-    },
-    'DCO.md': 1,
-  },
-};
-
-it('correctly converts a list of resource ids (paths) into a Resources object', () => {
-  expect(getResourcesFromResourcePaths(testResourcesList)).toEqual(
-    expectedResources,
-  );
-});
```
alexzurbonsen, line 96: Was this a duplicate? I can't spot the difference between the two.

.eslintrc.js
```diff
@@ -37,6 +37,8 @@ const sharedRules = {
     { allowBoolean: true, allowNumber: true, allowNever: true },
   ],
   '@typescript-eslint/switch-exhaustiveness-check': 'error',
+  'jest/consistent-test-it': 'error',
```
alexzurbonsen, line 40: from reading the docs I would expect `['error', {'fn': 'it'}]` here? (https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/consistent-test-it.md)

src/shared/setupTests.ts
```diff
@@ -20,63 +21,45 @@ const SUBSTRINGS_TO_SUPPRESS_IN_CONSOLE_ERROR = [
...
-});
-
-function mockConsoleImplementation(
-  selectors: Array<string>,
-  type: 'info' | 'warn' | 'error',
-): void {
-  jest
-    .spyOn(console, type)
-    .mockImplementation(getMockConsoleImplementation(selectors, type));
-}
```
alexzurbonsen, line 79: I am not entirely sure I understand the changes in this file. Can we quickly chat?

src/ElectronBackend/main/__tests__/listeners.test.ts
mstykow: yes, in fact our tests wouldn't work anymore if we were to use `reset` because then our mock implementation would be thrown away but some of them are set up for the entire test suite or even entire test run.

src/Frontend/Components/ResourcesTree/__tests__/resources-tree-helpers.test.ts
mstykow: yes

.eslintrc.js
mstykow: no, the default behavior is to use `it` inside `describe` blocks but allow `test` outside of them (which also makes sense to me).

src/shared/setupTests.ts
mstykow: yes

alexzurbonsen (approved):

PR #2436, opened 2023-12-19
chore: remove fetch-license button

### LINKED ISSUE #2435: Remove fetch-license button
With the new mechanism in place to fetch license information via the "+" button of autocompletes, the fetch-license button on the repo URL input field is no longer required. In fact, it cannot extract copyright information and it is also somewhat counterproductive since it may convert packages with a valid ecosystem such as NPM to a GitHub-type attribution instead.

mstykow:
### Summary of changes

- no longer needed since the package autocompletes can fetch licenses more effectively (via the "+" button)

Closes #2435.


vasily-pozdnyakov (approved):
Nice!

Issue #2333, opened 2023-11-21
Improve package search


mstykow:
**Is your feature request related to a problem? Please describe.**
The current package search may yield a lot of irrelevant results.

**Describe the solution you'd like**
1. To search for packages by name, replace the current search engine based on [clearlydefined.io](https://clearlydefined.io/) by [deps.dev](https://deps.dev/). Also allow searching for versions given a package name and type.
2. ~~Allow filtering search results by ecosystem (similar to what the deps.dev website does).~~
3. ~~Simplify and beautify the "search for packages" popup.~~
4. ~~Once a search result is selected, use clearlydefined.io to load additional metadata which deps.dev is unaware of to fill the attribution details.~~
5. Handle the resulting loading and possible error state between selecting a package from the deps.dev search and the filling of the input fields through data from clearlydefined.io.

**Additional context**
If you search for "lodash" you'll see that none of the hits are related to the actual NPM `lodash` library's current version, let alone the lodash libraries on NuGet, PyPI, etc. This is because clearlydefined is better at listing rather than prioritizing search results.

As to deps.dev, we can use the same endpoints that the deps.dev website uses to perform searches. This endpoint does not require any API key and yields exactly what we need.

- [x] #2394
- [x] #2396
- [x] #2395
- [x] #2397
- [x] #2426
- [x] #2428
- [x] #2429
- [x] #2433
- [x] #2435

Note: We may add filtering results by ecosystem at a later point.

PR #2438, opened 2023-12-20
feat: fine-tune autocomplete suggestions

mstykow:
### Summary of changes

1. ensure that the autocomplete suggestion "+" button overwrites all previously entered data so that there cannot be confusion about what was and what wasn't replaced by the user's action
2. make autocomplete suggestion "+" button more visually distinct from the primary suggestion action of just adding the single attribute so that it becomes clearer that there really are two separate actions on the suggestion card
image
3. ensure click on entire autocomplete input field opens option popper and not just a click on the underlying HTML input element
4. catch unhandled errors produced by the PackageURL constructor and instead display "INVALID PURL"
5. do not set a default version when user adopts a suggestion via autocomplete "+" button because the default (latest) version will almost always be the wrong one

Finalizing #2333.

See summary of changes.

For 3: click on the margin area of the input element (highlighted in green here): 
image

For 4: chose "pub" as package type and type a package name which includes an extended symbol. See [here](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst) for details.
image


meretp (changes requested):
Changes look good to me. Only one thing I noticed: When using the autocomplete with a manual attribution the URL sometimes gets overwritten by the caclulated URL. I would expect that no data is overwritten here.

meretp (approved):

PR #2443, opened 2023-12-21
Fix integration of @yarnpkg/plugin-spdx

antonbauhofer:
- Add spdx file to .gitignore in order to avoid accidentally checking the file in when creating it locally.
- Remove verification of name in SPDX file from CI workflow. If the document creation does not work correctly, the plugin should throw an error. So, there is no need to test that here. Besides, this test would fail if the app name would be changed in the package.json.

### Summary of changes

- Add spdx file to .gitignore
- Remove verification of name in SPDX file from CI workflow

- We add the spdx file to .gitignore in order to avoid accidentally checking the file in when creating it locally.
- If the document creation does not work correctly, the plugin should throw an error. So, there is no need to test that in the CI workflow. Besides, this test would fail if the app name would be changed in the package.json.

- Create an SPDX file with `yarn spdx` and verify that it's not checked in.
- Run the CI workflow.

### DIFF
--- a/.github/workflows/build-and-e2e-test.yml
+++ b/.github/workflows/build-and-e2e-test.yml
@@ -44,29 +44,7 @@ jobs:
 
       - run: yarn ship:auto
 
-      - name: Create SPDX document
-        run: yarn spdx
-
-      - name: Verify name in SPDX document on MacOS and Ubuntu
-        if: matrix.os == 'macos-latest' || matrix.os == 'ubuntu-latest'
-        run: |
-          set -e
-          name=$(jq .name opossum-ui.spdx.json -r)
-          echo "name=$name"
-          if [[ "$name" != "opossum-ui" ]]; then
-            echo "Name in SPDX document is not opossum-ui"
-            exit 1
-          fi
-
-      - name: Verify name in SPDX document on Windows
-        if: matrix.os == 'windows-latest'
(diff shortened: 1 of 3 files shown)


mstykow (approved):

.github/workflows/build-and-e2e-test.yml
```diff
@@ -44,29 +44,7 @@ jobs:
...
-      - name: Verify name in SPDX document on Windows
-        if: matrix.os == 'windows-latest'
-        run: |
-          $name=jq .name opossum-ui.spdx.json -r
-          echo "name=$name"
-          if ("$name" -ne "opossum-ui") {
-            echo "Name in SPDX document is not opossum-ui"
-            exit 1
-          }
+      - run: yarn spdx
```
mstykow, line 47: i just noticed there is a mistake in the release workflow: it should say "opossum-ui.spdx.json" but currently is "opossum-up.spdx.json" which is why the file is not uploaded as part of the release.

mstykow (changes requested):

.github/workflows/build-and-e2e-test.yml
antonbauhofer: fixed it.

mstykow (approved):

PR #2446, opened 2023-12-21
feat: add button to complete missing URL and legal information

### LINKED ISSUE #2445: Add button to complete missing URL and legal information
After selecting a signal from the autocomplete suggestions, there may be missing information such as repo URL or copyright. Add a magic wand button to try to fill this missing information from the web.

The logic to do this already exists inside `PackageSearchApi.enrichPackageInfo` and just needs to be called imperatively.

mstykow:
### Summary of changes

- add magic wand button in URL input field
- show success or error toasts when user clicks magic wand button
- also show toasts on copy/paste clipboard buttons
- fix tryit function to wait for promise
- throw for error status by default to avoid needing to perform `response.ok` checks

image

image

Closes #2445.

Select a signal with only partial information and try to fill the missing information with the new button. Also note that the button should disappear after the attempt, whether successful or not.


vasily-pozdnyakov:
1. Sometimes the button does not disappear after the first click but after the second.
2. Disappearing button is quite confusing for the user (especially when it comes back after some actions). Instead of hiding the button I would propose to use a toast with the relevant info. Or maybe consider some other alternatives.

mstykow:
> 1. Sometimes the button does not disappear after the first click but after the second.
> 2. Disappearing button is quite confusing for the user (especially when it comes back after some actions). Instead of hiding the button I would propose to use a toast with the relevant info. Or maybe consider some other alternatives.

I'm using a toaster now so that the hiding is no longer necessary.

antonbauhofer:
First of all, this feature is really awesome!
I'm not sure if the usage is quite as expected, though:
- Sometimes I get the toast "Added information where possible" even though seemingly nothing was added to the attribution.
- Maybe we should communicate the requirements for the icon to work in some way. E.g. that it requires a package type (it does, right?). I could imagine pointing that out in the red toast or in the tooltip of the icon.

src/Frontend/Components/AttributionColumn/PackageSubPanel.tsx
```diff
@@ -72,7 +75,12 @@ interface PackageSubPanelProps {
   confirmEditWasPreferred: Confirm;
 }
 
-export function PackageSubPanel(props: PackageSubPanelProps) {
+export function PackageSubPanel({
+  confirmEditWasPreferred,
+  displayPackageInfo,
+  isEditable,
+  showHighlight,
+}: PackageSubPanelProps) {
```
antonbauhofer, line 83: 👍

mstykow:
> First of all, this feature is really awesome! I'm not sure if the usage is quite as expected, though:
> 
> * Sometimes I get the toast "Added information where possible" even though seemingly nothing was added to the attribution.
> * Maybe we should communicate the requirements for the icon to work in some way. E.g. that it requires a package type (it does, right?). I could imagine pointing that out in the red toast or in the tooltip of the icon.

I've improved the error message for when we cannot fetch information because the package type is unknown.
I've also added a check to see if more information was found. If not, there is now a special message for this case:
image

antonbauhofer (approved):
Looks good to me.

Issue #2078, opened 2023-09-29
test temp folders are not deleted automatically in Windows


leslielazzarino:
They just cumulate and need to be deleted manually

PR #2449, opened 2023-12-22
chore: improve unit test cleanup

mstykow:
### Summary of changes

- create test output directory in global setup
- remove test output directory in global teardown
- make sure unit tests write artifacts into output folder by providing output path via new faker method
- convert jest config to TypeScript

fixes issue #2078

Run `yarn test:unit` in terminal and note that the test folder, "test-output" should be created and then deleted again at the end of the run. Should also work on Windows.


package.json
```diff
@@ -103,6 +103,7 @@
     "prettier-plugin-sh": "^0.13.1",
     "run-script-os": "^1.1.6",
     "start-server-and-test": "^2.0.3",
+    "ts-node": "^10.9.2",
```
mstykow, line 106: needed to compile `jest.config.ts`

davidmundelius (approved):
Looks good to me. Also I can confirm that the automatic deletion of the test-output folder works on Windows.

PR #2451, opened 2023-12-22
chore: remove "extracted" directory

mstykow:
### Summary of changes

- move VirtualizedTree back into components folder
- allow components to have a "util" sidecar

It is unclear why we have one component inside an "extracted" directory while all others are in the "components" directory. If extraction is desired in the future, there is no need for any such directory and the PR to extract the component (presumably to a different repository) should simply be opened.


antonbauhofer (approved):
Looks good to me.
I tested it and it works as expected.

PR #2468, opened 2024-01-04
chore: refactor attribution filter dropdown

### LINKED ISSUE #2464: Refactor attribution filter dropdown
There is a UI bug in the filter dropdown that causes the menu items to be shifted to the right. There are also some other issues:
- chip is misshapen
- click on menu item does not close menu

image

Refactor the code to use the Autocomplete component and remove the current FilterMultiSelect component.

mstykow:
### Summary of changes

- handle active filters via simpler redux structure
- replace FilterMultiSelect by Autocomplete
- fix: click on autocomplete clear button or popup indicator was not properly opening/closing the popup

BEFORE:
image

AFTER:
image

closes #2464 

Try the filter dropdown in both attribution and report view.


src/Frontend/Components/AttributionView/AttributionView.tsx
```diff
@@ -42,51 +40,39 @@ const classes = {
   clickableIcon,
 };
 
-export function AttributionView(): ReactElement {
```
antonbauhofer, line 45: Why do you remove these annotations?

antonbauhofer (approved):
Looks good to me.

src/Frontend/Components/AttributionView/AttributionView.tsx
mstykow: a) defining the return type of a react FC is quite unnecessary: almost anything can be a react node
b) the current annotation is incorrect since a react node is more than `ReactElement`, e.g., `null` is acceptable as a react node but excluded by this unnecessarily strict return type.

PR #2471, opened 2024-01-05
feat: add more attribution filters

### LINKED ISSUE #2465: Add more attribution filters
Check which filters are missing that may make sense to add. Suggestions: wasPreferred (grey stars), confidence.

mstykow:
### Summary of changes

- extract common icons to Icons files and use them consistently
- use icon for each filter option
- sort filters alphabetically
- add new filter options: exclude from notice, low confidence, pre-selected, previously preferred

image

closes #2465

Use the new filters in attribution and report view. Also check that icons still look correct on package card, auditing options, filter menu.


src/Frontend/Components/AttributionColumn/AuditingOptions.tsx
```diff
@@ -435,7 +389,7 @@ function useChips({
             highlightSelectedOnly
           />
         ),
-        active: packageInfo.attributionConfidence !== undefined,
+        active: true,
```
mstykow, line 392: allow setting confidence even when it is initially not defined

antonbauhofer (approved):
It looks good to me.
I checked that the filters work in general, but didn't test all filters in the app.
In general, I think it would be easier to review, if we make separate commits or pull requests for refactorings.

Issue #2466, opened 2024-01-04
Move multi-select actions out of context menu


mstykow:
The actions to remove or confirm multiple attributions in attributions view are currently hidden inside context menus. In order to make these features more transparent, add them as icons above the attribution list similar to how an email client works, for example.

Also add an option to select/de-select all attributions.

Finally, convert the filter select to a menu in order to not need more vertical space for the new actions, filters, sortings etc. While doing so, it is important to somehow indicate to the user that filters are active even if the menu is not currently open.

PR #2479, opened 2024-01-09
chore: extract multi-select menu

mstykow:
### Summary of changes

- extract from auditing options the multi-select menu in order to re-use it later in other components

work towards #2466

See that auditing options still work correctly.


antonbauhofer (approved):
Looks good and works as expected.
Thanks for making this a separate pull request!

PR #2480, opened 2024-01-09
feat: move multi-select actions out of context menu

### LINKED ISSUE #2466: Move multi-select actions out of context menu

mstykow:
### Summary of changes

- create row for attribution actions
- convert filter select to filter menu
- indicate active/loading state of filters on filter menu button
- move filtering logic into web worker to unblock the UI when expensive filters are running
- persist search text in attribution view in Redux

SELECTING MULTIPLE FILTERS WITH COUNTS:
image

MULTI-SELECT ACTIONS WHILE SOME FILTERS ARE ACTIVE:
image

FILTERED ATTRIBUTIONS LOADING:
image

SELECTING ALL WHILE SOME FILTERS ACTIVE:
image

closes #2466

Use a large Opossum file to select multiple filter options. Notice that especially the "modified previous preferred" option that previously blocked the UI is now unblocked while a loading spinner is shown.


src/Frontend/Components/ReportView/ReportView.tsx
mstykow, line 1: the changes in this file are here to keep report view unchanged for now.
report view has usability issues at the moment, especially when used with large opossum files.
we'll contact stakeholders to determine if we should delete or fix report view. the essential question would be: what features does report view provide that are not found elsewhere?

antonbauhofer:
The changes look good in the app.
Do we also want to change the sorting dropdown in the same way or how would you deal with that?

src/Frontend/Components/AttributionList/AttributionList.style.tsx
```diff
@@ -0,0 +1,26 @@
...
+export const ActionBar = styled(MuiBox)({
+  border: '1px solid transparent',
+  display: 'flex',
+  justifyContent: 'space-between',
+});
+
+export const ButtonGroup = styled(MuiBox)({
+  display: 'flex',
+  gap: '4px',
+});
```
antonbauhofer, line 26: This is a new pattern to have the styles in a separate file.
I personally like the idea, but maybe we should get some feedback from the team on that.
If we decide to go that way, we should aim at being consistent and adjust it in the other components as well, in the future.

src/Frontend/Components/AttributionList/AttributionList.tsx
```diff
@@ -2,58 +2,139 @@
...
 
 export function AttributionList() {
   const dispatch = useAppDispatch();
   const selectedPackageCardIdInAttributionView = useAppSelector(
     getSelectedAttributionIdInAttributionView,
   );
+  const multiSelectSelectedAttributionIds = useAppSelector(
+    getMultiSelectSelectedAttributionIds,
+  );
+  const selectedAttributionIdAttributionView = useAppSelector(
```
antonbauhofer, line 48: These variable names tend to get a bit long / hard to read.
I suggest to replace `multiSelectSelectedAttributionIds` with `multiSelectAttributionIds `
`selectedAttributionIdAttributionView` with `selectedAttributionId`

src/Frontend/Components/AttributionList/AttributionList.tsx
```diff
@@ -2,58 +2,139 @@
...
+    getSelectedAttributionIdInAttributionView,
+  );
+  const manualAttributions = useAppSelector(getManualAttributions);
+
+  const [anchorEl, setAnchorEl] = useState<HTMLElement>();
```
antonbauhofer, line 53: In general, we try to avoid abbreviations.
```suggestion
  const [anchorElement, setAnchorElement] = useState<HTMLElement>();
```

src/Frontend/Components/AttributionList/AttributionList.tsx
```diff
@@ -2,58 +2,139 @@
...
+    useFilterMenuOptions();
+  const someSelectedAttributionsArePreSelected = useMemo(
+    () =>
+      multiSelectSelectedAttributionIds.some(
+        (id) => filteredAndSortedAttributions[id]?.preSelected,
+      ),
+    [filteredAndSortedAttributions, multiSelectSelectedAttributionIds],
+  );
+
+  const areAllAttributionsSelected = useMemo(
```
antonbauhofer, line 91: Variable names for booleans should not be phrased as questions.
```suggestion
  const allAttributionsAreSelected = useMemo(
```

src/Frontend/Components/AttributionList/AttributionList.tsx
```diff
@@ -74,57 +155,114 @@ export function AttributionList() {
...
-function getFilteredAndSortedPackageCardIdsAndDisplayPackageInfos(
-  attributions: Attributions,
-): {
-  filteredAndSortedPackageCardIds: Array<string>;
-  filteredDisplayPackageInfos: DisplayPackageInfos;
-} {
-  const sortedAttributionIds = Object.keys(attributions).sort(
-    getAlphabeticalComparerForAttributions(attributions),
-  );
+  function renderSelectAll() {
```
antonbauhofer, line 171: Why do we call these `render...`? Aren't they just normal react components? As far as I know, we don't use the `render` preamble in other places.

mstykow: a function name describes what the function does. this function renders the "select-all" checkbox. this is also the naming react uses still in class components: the main function there is called "render". hence the name here.

src/Frontend/Components/AttributionList/AttributionList.tsx
mstykow: it's common practice in many programming languages to start booleans with "is". this PR is not the right place to start such a fundamental discussion.

src/Frontend/Components/AttributionList/AttributionList.tsx
mstykow: "anchorEl" is an exception to the rule because it's a common variable name in React for the anchor element. you can also search the code base and you'll find many other hits for this usage.

src/Frontend/Components/AttributionList/AttributionList.tsx
mstykow: i noticed the same issue but if i were to make your change i'd also have to change the selector names for consistency. so for the purpose of this PR i'm not willing to make such a large refactoring.

src/Frontend/Components/AttributionList/AttributionList.style.tsx
mstykow: this has already been discussed and decided upon while you were gone. so it's not new from this PR. you can search the repo and you'll find many places that use this pattern now.

antonbauhofer:
My observations from testing it on a large file:
It's a bit funny that the spinner icon only shows up next to the filter icon when switching to the attribution view. In this case, I would expect a large spinner at the location of the cards. Other than that, it's snappy and I could barely ever see the spinner.

antonbauhofer (approved):

PR #2478, opened 2024-01-09
feat: offer multiple sorting options for attributions

antonbauhofer:
### Summary of changes

- add toggle state for critical signals progress bar to store
- extend sorting algorithm to consider criticality
- extend test of sorting

### Context and reason for change

There is highlighting for critical signals but before this change it was still hard to find critical signals in long lists of signals coming from e.g. ScanCode.

### How can the changes be tested

Open an input file that contains critical signals and look at the sorting.

PR #2481, opened 2024-01-09
feat: offer multiple sorting options for attributions

antonbauhofer:
Summary of changes

add toggle state for critical signals progress bar to store
extend sorting algorithm to consider criticality
extend test of sorting
Context and reason for change

There is highlighting for critical signals but before this change it was still hard to find critical signals in long lists of signals coming from e.g. ScanCode.

How can the changes be tested

Open an input file that contains critical signals and look at the sorting.

### DIFF
--- a/src/Frontend/Components/AllAttributionsPanel/AllAttributionsPanel.tsx
+++ b/src/Frontend/Components/AllAttributionsPanel/AllAttributionsPanel.tsx
@@ -5,6 +5,7 @@
 import MuiPaper from '@mui/material/Paper';
 import { ReactElement } from 'react';
 
+import { text } from '../../../shared/text';
 import { PackagePanelTitle } from '../../enums/enums';
 import { OpossumColors } from '../../shared-styles';
 import { selectPackageCardInAuditViewOrOpenUnsavedPopup } from '../../state/actions/popup-actions/popup-actions';
@@ -12,6 +13,8 @@ import { addToSelectedResource } from '../../state/actions/resource-actions/save
 import { useAppDispatch } from '../../state/hooks';
 import { DisplayPackageInfos, PackageCardConfig } from '../../types/types';
 import { convertDisplayPackageInfoToPackageInfo } from '../../util/convert-package-info';
+import { getAlphabeticalComparerForAttributions } from '../../util/get-alphabetical-comparer';
+import { useActiveSortingInAuditView } from '../../util/use-active-sorting';
 import { PackageCard } from '../PackageCard/PackageCard';
 import { PackageList } from '../PackageList/PackageList';
 
@@ -35,6 +38,7 @@ export function AllAttributionsPanel(
   props: AllAttributionsPanelProps,
 ): ReactElement {
   const dispatch = useAppDispatch();
+  const [activeSorting] = useActiveSortingInAuditView();
 
   function getPackageCard(packageCardId: string): ReactElement | null {
     const displayPackageInfo = props.displayPackageInfos[packageCardId];
@@ -74,11 +78,18 @@ export function AllAttributionsPanel(
     );
   }
 
+  const sortedPackageCardIds = Object.keys(props.displayPackageInfos).sort(
+    getAlphabeticalComparerForAttributions(
+      props.displayPackageInfos,
+      activeSorting === text.auditViewSorting.byCriticality,
+    ),
+  );
+
   return (
     <MuiPaper sx={classes.root} elevation={0} square={true}>
       <PackageList
         displayPackageInfos={props.displayPackageInfos}
-        sortedPackageCardIds={Object.keys(props.displayPackageInfos)}
+        sortedPackageCardIds={sortedPackageCardIds}
         getAttributionCard={getPackageCard}
         listTitle={PackagePanelTitle.AllAttributions}
         fullHeight
--- a/src/Frontend/Components/AttributionList/AttributionList.tsx
+++ b/src/Frontend/Components/AttributionList/AttributionList.tsx
@@ -3,10 +3,12 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 import FilterAltIcon from '@mui/icons-material/FilterAlt';
+import SortIcon from '@mui/icons-material/Sort';
 import MuiBox from '@mui/material/Box';
-import { useState } from 'react';
+import { ChangeEvent, useState } from 'react';
 
 import { Attributions } from '../../../shared/shared-types';
+import { text } from '../../../shared/text';
 import { clickableIcon, disabledIcon } from '../../shared-styles';
 import { changeSelectedAttributionIdOrOpenUnsavedPopup } from '../../state/actions/popup-actions/popup-actions';
 import { useAppDispatch, useAppSelector } from '../../state/hooks';
@@ -14,15 +16,23 @@ import { getSelectedAttributionIdInAttributionView } from '../../state/selectors
 import { DisplayPackageInfos } from '../../types/types';
 import { convertPackageInfoToDisplayPackageInfo } from '../../util/convert-package-info';
 import { getAlphabeticalComparerForAttributions } from '../../util/get-alphabetical-comparer';
+import {
+  AttributionViewSorting,
+  useActiveSortingInAttributionView,
+} from '../../util/use-active-sorting';
 import { AttributionCountsPanel } from '../AttributionCountsPanel/AttributionCountsPanel';
 import { FilterMultiSelect } from '../Filter/FilterMultiSelect';
 import { useFilteredAttributions } from '../Filter/FilterMultiSelect.util';
 import { IconButton } from '../IconButton/IconButton';
+import { Dropdown, MenuItem } from '../InputElements/Dropdown';
 import { PackageCard } from '../PackageCard/PackageCard';
 import { AttributionsViewPackageList } from '../PackageList/AttributionsViewPackageList';
 import { ResizableBox } from '../ResizableBox/ResizableBox';
 
 const classes = {
+  dropdown: {
+    marginTop: '8px',
+  },
   topElements: {
     display: 'flex',
     alignItems: 'center',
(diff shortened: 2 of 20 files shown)


src/Frontend/Components/AllAttributionsPanel/AllAttributionsPanel.tsx
```diff
@@ -74,11 +78,18 @@ export function AllAttributionsPanel(
     );
   }
 
+  const sortedPackageCardIds = Object.keys(props.displayPackageInfos).sort(
+    getAlphabeticalComparerForAttributions(
+      props.displayPackageInfos,
+      activeSorting === text.auditViewSorting.byCriticality,
+    ),
+  );
```
antonbauhofer, line 86: [Original comment](https://github.com/opossum-tool/OpossumUI/pull/2478/files#r1445875580)
this doesn't seem right: in the case that we are not sorting by criticality we are now sorting alphabetically here although before your changes no sorting was necessary at all.

antonbauhofer: Correct, before this pull request we didn't sort the attributions at this point. I think, however, that this was a bug or at least not the desired behavior, so I decided to introduce sorting in any case.

src/Frontend/Components/AttributionList/AttributionList.tsx
```diff
@@ -46,8 +69,23 @@ export function AttributionList() {
   const hasActiveFilters = !!activeFilters.length;
   const [showMultiSelect, setShowMultiselect] = useState(hasActiveFilters);
 
+  const [activeSorting, setActiveSorting] = useActiveSortingInAttributionView();
+  const [showSortingSelect, setShowSortingSelect] = useState<boolean>(false);
+  const defaultSortingIsActive = activeSorting === defaultSorting;
+
+  if (!defaultSortingIsActive && !showSortingSelect) {
+    setShowSortingSelect(!showSortingSelect);
+  }
```
antonbauhofer, line 78: [original comment](https://github.com/opossum-tool/OpossumUI/pull/2478/files#r1446222486)
i think we talked about how we wanted to move this logic into the initial value of useState to avoid unnecessary setting of state on each render.

antonbauhofer: Correct. I fixed this in the ResourceDetailsTab, but overlooked it here. Will fix it.

src/Frontend/Components/AttributionList/AttributionList.tsx
```diff
@@ -82,18 +120,42 @@ export function AttributionList() {
...
+            tooltipPlacement={'right'}
+            onClick={() => setShowMultiselect((prev) => !prev)}
+            disabled={hasActiveFilters}
+            icon={
+              <FilterAltIcon
+                sx={hasActiveFilters ? disabledIcon : clickableIcon}
+              />
+            }
+          />
+        </MuiBox>
```
antonbauhofer, line 147: [original comment](https://github.com/opossum-tool/OpossumUI/pull/2478/files/9d4b63197be8331fba5026be9f794c6ce9f550f4#r1446225129)
we're taking up a lot of vertical space now and also the padding between the elements is not equal. i think we should fix at least the basic layout/padding. the fact that we're taking up so much space i'll fix later in #2480.

i also suggest we just write "sorting" without "algorithm"

antonbauhofer: I fixed the padding and removed "algorithm".

src/Frontend/util/use-active-sorting.ts
```diff
@@ -0,0 +1,43 @@
...
+import { useVariable } from './use-variable';
+
+export const auditViewSorting = Object.values(text.auditViewSorting);
+export const attributionViewSorting = Object.values(
+  text.attributionViewSorting,
+);
+export type AuditViewSorting = (typeof auditViewSorting)[number];
+export type AttributionViewSorting = (typeof attributionViewSorting)[number];
+
+export function useActiveSorting(): [
```
antonbauhofer, line 15: [original comment](https://github.com/opossum-tool/OpossumUI/pull/2478#discussion_r1445862711)
i think we said you'd change the naming here so that it reflects the fact that this sorting is only used in audit view.
please remove the return type annotation on both of these hooks. they are just repeating the annotation of useVariable which means that if useVariable is refactored, we'd break the typing here for no good reason because this function is nothing but a thin wrapper around useVariable.
Member
Author
@[antonbauhofer](https://github.com/antonbauhofer) antonbauhofer [43 minutes ago](https://github.com/opossum-tool/OpossumUI/pull/2481/files#discussion_r1446234493)
Without the type annotation the type checker was not happy.
Member
@[mstykow](https://github.com/mstykow) mstykow [41 minutes ago](https://github.com/opossum-tool/OpossumUI/pull/2481/files#discussion_r1446237584)
this is again just a local problem on your end, i believe
Member
Author
@[antonbauhofer](https://github.com/antonbauhofer) antonbauhofer [10 minutes ago](https://github.com/opossum-tool/OpossumUI/pull/2481/files#discussion_r1446280319)
Fixed the naming, but it the typing problem remains, if I remove the annotations, even though I fixed the local setup.

mstykow: ah, you need to write `as const` behind the returned array. then you can delete the annotation. by doing so, you're telling typescript not to widen the type which was causing the error.

src/Frontend/Components/AttributionList/AttributionList.tsx
```diff
@@ -82,18 +117,42 @@ export function AttributionList() {
...
-          disabled={hasActiveFilters}
-          icon={
-            <FilterAltIcon
-              sx={hasActiveFilters ? disabledIcon : clickableIcon}
-            />
-          }
-        />
+        <MuiBox>
+          <IconButton
+            tooltipTitle="Sort"
```
mstykow, line 122: please extract user-facing text

src/Frontend/Components/AttributionList/AttributionList.tsx
```diff
@@ -82,18 +117,42 @@ export function AttributionList() {
...
+            disabled={!defaultSortingIsActive}
+            icon={
+              <SortIcon
+                aria-label={'Sort icon'}
+                sx={!defaultSortingIsActive ? disabledIcon : clickableIcon}
+              />
+            }
+          />
+          <IconButton
+            tooltipTitle={'Filters'}
```
mstykow, line 134: extract please since you're touching it

src/Frontend/Components/ResourceDetailsTabs/ResourceDetailsTabs.tsx
```diff
@@ -165,13 +207,39 @@ export function ResourceDetailsTabs(
...
-      <IconButton
-        tooltipTitle="Search signals by name, license name, copyright text and version"
-        tooltipPlacement="right"
-        onClick={onSearchToggleClick}
-        icon={<SearchPackagesIcon sx={classes.largeClickableIcon} />}
-        iconSx={classes.searchToggle}
-      />
+      <MuiBox sx={classes.icons}>
+        <IconButton
+          tooltipTitle="Choose sorting algorithm for the displayed signals"
```
mstykow, line 212: extract...

src/Frontend/Components/ResourceDetailsTabs/ResourceDetailsTabs.tsx
```diff
@@ -165,13 +207,39 @@ export function ResourceDetailsTabs(
...
+          tooltipPlacement="right"
+          onClick={onSortToggleClick}
+          disabled={!defaultSortingIsActive}
+          icon={
+            <SortAttributionsIcon
+              sx={
+                !defaultSortingIsActive
+                  ? classes.disabledLargeClickableIcon
+                  : classes.largeClickableIcon
```
mstykow, line 221: it would be easier to understand code if you simply drop the `!` and turn the ternary around

src/shared/text.ts
```diff
@@ -96,4 +96,17 @@ export const text = {
...
+    alphabetical: 'Alphabetical',
+    byCriticality: 'By Criticality',
+  },
+  auditViewSorting: {
+    sorting: 'Sorting',
+    byOccurrence: 'By Occurrence',
+    byCriticality: 'By Criticality',
+  },
+  tooltips: {
+    search: 'Search signals by name, license name, copyright text and version',
```
mstykow, line 110: better to group semantically than by technical function. so this would be part of "attributionColumn".

mstykow (changes requested):

src/Frontend/util/use-active-sorting.ts
antonbauhofer: Implemented it as suggested.

src/shared/text.ts
antonbauhofer: Changed it as suggested. However, I don't quite agree with you here. The name "attributionColumn" indicates grouping the texts by components. Hence, we will either eventually end up defining the same text multiple times or using texts from other components. The name "tooltips", on the other hand, can be understood as the user-facing behavior in the frontend. So these texts could, for example be used for the tooltips in the auditView and in the attributionView.

mstykow: yes, good point. one needs to see if the text is reused or not. it seems to me this one is not. what do you think?

src/shared/text.ts
```diff
@@ -44,6 +44,8 @@ export const text = {
     source: 'Source',
     useAutocompleteSuggestion:
       'Adopt all coordinates and legal information from suggestion',
+    sortTooltip: 'Sort',
```
mstykow, line 47: I think it would be better to say "Sorting" here because then we match the noun "Filter" instead of using the verb "sort".

mstykow (approved):

src/shared/text.ts
antonbauhofer: In this case it probably doesn't matter, since we are using different texts anyway. But in general I don't think that it is a good idea to start grouping the texts by components.

src/shared/text.ts
mstykow: ok, but it currently says "filters", not "filter"

antonbauhofer: I removed the "s"

Issue #2412, opened 2023-12-13
State is mutated inside Redux reducers


mstykow:
**Describe the bug**
There are several reducers which are mutating the Redux state which is a gross error in using Redux and leads to hard-to-debug issues. To fix them, we need to turn on the `immutableCheck` in `configure-store.ts` and run the unit tests. The ones that are failing will give hints where to look for the problem.

Also see #2381.

**To Reproduce**

1. Turn on `immutableCheck` in `configure-store.ts`
2. Run unit tests
3. See errors

**Expected behavior**
No mutation should take place inside reducers.
https://redux.js.org/style-guide/#priority-a-rules-essential

PR #2487, opened 2024-01-11
fix: remove mutations of the redux state in reducers

antonbauhofer:
### Summary of changes

This fixes some of our reducers, so that they don't mutate the state.

Before this change, there were a few places in our code, where reducers were mutating the state. This is strongly discouraged according to [the Redux guide](https://redux.js.org/style-guide/#priority-a-rules-essential).

- `yarn test:unit`
- check that the app performance is still acceptable, by opening a large file in OpossumUI

Fixes https://github.com/opossum-tool/OpossumUI/issues/2412

### DIFF
--- a/src/Frontend/state/configure-store.ts
+++ b/src/Frontend/state/configure-store.ts
@@ -19,7 +19,7 @@ export function createAppStore() {
         // TECH DEBT: we should not be putting sets into the store
         // https://redux.js.org/style-guide/#do-not-put-non-serializable-values-in-state-or-actions
         serializableCheck: false,
-        // TECH DEBT: we are mutating the redux state inside reducers
+        // We set this to false because it significantly reduces performance in development mode.
         immutableCheck: false,
       }),
   });
--- a/src/Frontend/state/helpers/save-action-helpers.ts
+++ b/src/Frontend/state/helpers/save-action-helpers.ts
@@ -56,7 +56,13 @@ export function createManualAttribution(
       [newAttributionId]: [selectedResourceId],
     },
     resourcesWithAttributedChildren: {
-      ...manualData.resourcesWithAttributedChildren,
+      paths: [...manualData.resourcesWithAttributedChildren.paths],
+      pathsToIndices: {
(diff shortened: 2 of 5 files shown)


src/Frontend/state/helpers/save-action-helpers.ts
```diff
@@ -595,7 +602,13 @@ function getAttributionDataShallowCopy(
...
     attributionsToResources: { ...attributionData.attributionsToResources },
     resourcesWithAttributedChildren: {
-      ...attributionData.resourcesWithAttributedChildren,
+      paths: [...attributionData.resourcesWithAttributedChildren.paths],
+      pathsToIndices: {
+        ...attributionData.resourcesWithAttributedChildren.pathsToIndices,
+      },
+      attributedChildren: {
+        ...attributionData.resourcesWithAttributedChildren.attributedChildren,
+      },
```
antonbauhofer, line 611: This is called in several places and not always is every one of these parts mutated. To improve performance, we could just make a shallow copy here and then make copies of the respective parts, where they are mutated.
If we decide to do that, I think we should generally establish that pattern and also fix it at other places, though.

src/Frontend/state/configure-store.ts
```diff
@@ -19,7 +19,7 @@ export function createAppStore() {
         // TECH DEBT: we should not be putting sets into the store
         // https://redux.js.org/style-guide/#do-not-put-non-serializable-values-in-state-or-actions
         serializableCheck: false,
-        // TECH DEBT: we are mutating the redux state inside reducers
+        // We set this to false because it significantly reduces performance in development mode.
```
antonbauhofer, line 22: https://github.com/reduxjs/redux-toolkit/issues/415

mstykow (approved):
Very nice. Thank you @antonbauhofer! 👍

antonbauhofer:
I did a performance test with a large OpossumUI file and the performance does not seem to be significantly impacted by these changes. Also, I loosely checked that no other state mutations, not covered by tests, are present.

PR #2491, opened 2024-01-11
fix: correctly set selected ID on package checkbox click

mstykow:
### Summary of changes

- use the correct thunk action to set selected ID to ensure that temporary package info is also correctly shown

Previously the displayed package info was not shown correctly.

To reproduce the bug on main:
1. Open a file.
2. Go to attribution view.
3. Click on one of the package card checkboxes without first clicking on the package card.
4. Notice that the displayed package info in the right column is completely empty although the package card is selected.


mstykow:
fixed here together with a blinker that caused tests to fail: https://github.com/opossum-tool/OpossumUI/commit/30324045993922f2b0af8e54cc6fc6a3f1516da2

mstykow:
fixed here together with a blinker that caused tests to fail: https://github.com/opossum-tool/OpossumUI/commit/30324045993922f2b0af8e54cc6fc6a3f1516da2

PR #2497, opened 2024-01-12
chore: enable redux immutability check in ci

### LINKED ISSUE #2493: Activate immutability redux check in the CI
The serializable and immutable checks are important, but if they need to be performed manually they might be forgotten.

We could setthe parameters *serializableCheck* and *immutableCheck* in *src/Frontend/state/configure-store.ts* through environment variables. That would allow to use different values locally or in the CI.

src/Frontend/state/configure-store.ts could have a different version, that replaces the regular one in the CI.

mstykow:
### Summary of changes

- conditionally enable check in CI but keep it off in development mode because of performance issues

closes #2493

Run `yarn test:unit` locally and notice that the immutability check is off, so no error occurs when you mutate state inside one of the reducers. Now look at this pipeline run to see that the CI detects mutations: https://github.com/opossum-tool/OpossumUI/actions/runs/7506410149/job/20437774813?pr=2497

### DIFF
--- a/.github/workflows/check-code-quality.yml
+++ b/.github/workflows/check-code-quality.yml
@@ -20,6 +20,7 @@ jobs:
     if: |
       !(github.event_name == 'pull_request' && startsWith(github.head_ref, 'renovate/'))
     env:
+      CI: true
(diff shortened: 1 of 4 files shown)


benedikt-richter (approved):
👍

PR #2501, opened 2024-01-15
fix: disable suggestions loading when viewing a non-editable signal/attribution

mstykow:
### Summary of changes

- when viewing a non-editable signal or attribution it is not necessary to load any suggestions
- the input field was already correctly disabled before

Avoid unnecessary background requests.

While in audit view, open the dev tools network tab. Then select once an attribution that is editable and then one that is not (for example a signal). Notice that only the editable attribution should trigger network requests.


MarkusObendrauf (approved):
Nice

PR #2503, opened 2024-01-16
fix: avoid unnecessary initial filter loading state

### LINKED ISSUE #2502: Loading attribution list in attribution view for the first time is very slow
**Describe the bug**
Loading attribution list in attribution view for the first time is very slow in the case that there are several thousand attributions of which are many pre-selected.

**To Reproduce**

1. Open a opossum input file that contains thousands of attributions
2. Navigate to the attribution view
3. Observe that the filter's loading spinner spins for a long time, while most attributions are only loaded after approx 25 seconds 

**Expected behavior**
Loading the list of attributions should be reasonably fast.

mstykow:
### Summary of changes

- use all manual attributions as filtered attributions when no filter is set
- separate concerns of signals worker hook from those of the hooks accessing variables from the Redux store

closes #2502

Open a large Opossum file and notice that the attribution list immediately displays attributions without any filter loading spinner spinning.

### DIFF
--- a/src/Frontend/state/variables/__tests__/use-filtered-attributions.test.ts
+++ b/src/Frontend/state/variables/__tests__/use-filtered-attributions.test.ts
@@ -1,0 +1,48 @@
+// SPDX-FileCopyrightText: Meta Platforms, Inc. and its affiliates
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+import { faker } from '../../../../testing/Faker';
+import { renderHook } from '../../../test-helpers/render';
+import { setManualData } from '../../actions/resource-actions/all-views-simple-actions';
+import { setVariable } from '../../actions/variables-actions/variables-actions';
+import {
+  FILTERED_ATTRIBUTIONS,
+  FilteredAttributions,
+  initialFilteredAttributions,
+  useFilteredAttributions,
+} from '../use-filtered-attributions';
+
+describe('useFilteredAttributions', () => {
+  it('returns all manual attributions when no filters set', () => {
+    const manualAttributions = faker.opossum.manualAttributions();
+    const { result } = renderHook(() => useFilteredAttributions(), {
+      actions: [
+        setManualData(
+          manualAttributions,
+          faker.opossum.resourcesToAttributions(),
+        ),
+      ],
+    });
+
+    expect(result.current[0].attributions).toEqual(manualAttributions);
+  });
+
+  it('returns filtered manual attributions when filters set', () => {
+    const manualAttributions = faker.opossum.manualAttributions();
+    const { result } = renderHook(() => useFilteredAttributions(), {
+      actions: [
+        setManualData(
+          manualAttributions,
+          faker.opossum.resourcesToAttributions(),
+        ),
+        setVariable<FilteredAttributions>(FILTERED_ATTRIBUTIONS, {
+          ...initialFilteredAttributions,
+          selectedFilters: ['First Party'],
+        }),
+      ],
+    });
+
+    expect(result.current[0].attributions).toEqual({});
+  });
+});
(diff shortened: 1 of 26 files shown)


src/Frontend/state/variables/use-folder-progress-data.ts
```diff
@@ -0,0 +1,12 @@
+// SPDX-FileCopyrightText: Meta Platforms, Inc. and its affiliates
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+import { ProgressBarData } from '../../types/types';
+import { useVariable } from './use-variable';
+
+export const FOLDER_PROGRESS_DATA = 'folder-progress-data';
+
+export function useFolderProgressData() {
```
antonbauhofer, line 10: We could also have the simple hooks accumulated in one file, but I think it is nice to have them separated like you did it.

src/Frontend/state/variables/use-filtered-attributions.ts
```diff
@@ -0,0 +1,56 @@
...
+import { Attributions } from '../../../shared/shared-types';
+import {
+  Filter,
+  FilterCounts,
+} from '../../web-workers/scripts/get-filtered-attributions';
+import { useAppSelector } from '../hooks';
+import { getManualAttributions } from '../selectors/all-views-resource-selectors';
+import { useVariable } from './use-variable';
+
+export const FILTERED_ATTRIBUTIONS = 'filtered-attributions';
```
antonbauhofer, line 16: We should always do that 👍

src/Frontend/state/variables/use-filtered-attributions.ts
```diff
@@ -0,0 +1,56 @@
...
+      FILTERED_ATTRIBUTIONS,
+      initialFilteredAttributions,
+    );
+
+  useEffect(() => {
+    if (!filteredAttributions.selectedFilters.length) {
+      setFilteredAttributions((prev) => ({
+        ...prev,
+        attributions,
+      }));
```
antonbauhofer, line 47: Why do we pass an arrow function here instead of just passing `attributions`?
```suggestion
      setFilteredAttributions(
        attributions,
      );
```

src/Frontend/web-workers/__tests__/use-signals-worker.test.ts
```diff
@@ -104,4 +110,35 @@ describe('useSignalsWorker', () => {
...
+    const metadata = faker.opossum.metadata();
+    renderHook(useSignalsWorker, { actions: [setProjectMetadata(metadata)] });
+
+    expect(mockPostMessage).not.toHaveBeenCalledWith<[SignalsWorkerInput]>({
+      name: 'selectedFilters',
+      data: expect.anything(),
+    });
+  });
+
+  it('sends selected filters when some are set', () => {
```
antonbauhofer, line 124: 👍

src/Frontend/state/variables/__tests__/use-filtered-attributions.test.ts
```diff
@@ -0,0 +1,48 @@
...
+import { setManualData } from '../../actions/resource-actions/all-views-simple-actions';
+import { setVariable } from '../../actions/variables-actions/variables-actions';
+import {
+  FILTERED_ATTRIBUTIONS,
+  FilteredAttributions,
+  initialFilteredAttributions,
+  useFilteredAttributions,
+} from '../use-filtered-attributions';
+
+describe('useFilteredAttributions', () => {
```
antonbauhofer, line 16: 👍

antonbauhofer (approved):
Looks good to me and works as expected.
For a more pleasant review, I would have preferred to have the refactorings in a separate pull request, though.

src/Frontend/state/variables/use-folder-progress-data.ts
mstykow: yes, i was considering along the same lines

src/Frontend/state/variables/use-filtered-attributions.ts
mstykow: because i need to preserve the other attributes that are inside `prev` but not part of `attributions`

Issue #2508, opened 2024-01-18
Add option to filter for modified previously preferred attributions


mstykow:
**Is your feature request related to a problem? Please describe.**
As a user, I'd like to filter for attributions that were previously preferred but have been modified.

**Ideas**
Calculate the origin of each attribution in the main thread when the file is initially opened and save this information in a new data structure in the Redux store. Then, keep this structure up-to-date, whenever attributions are modified (part of the ACTION_UPDATE_ATTRIBUTION action in the resource reducer).

**Additional context**
The feature had already been implemented but was removed in #2507 because of performance issues.  We need to find a more performant solution.

PR #2509, opened 2024-01-18
feat: enable the same sortings for attributions and signals

### LINKED ISSUE #2507: Harmonize sorting options
As a user, I'd like to sort by alphabet, frequency, criticality regardless of whether I'm looking at a list of attributions or signals.

**Additional context**
Currently there are two different kinds of sortings defined. Harmonize them.

mstykow:
### Summary of changes

- move all attribution sorting and filtering logic into the web worker (single place of responsibility)
- remove "modified preferred" filter because it performs too slowly (will be added again in #2508)
- consolidate different functions responsible for sorting packages into a single function
- use the same filtering options for attributions and signals

closes #2507

Check the sorting in audit and attribution view.


src/shared/text.ts
```diff
@@ -111,26 +111,22 @@ export const text = {
...
   },
-  attributionViewSorting: {
-    alphabetical: 'Alphabetically',
-    byCriticality: 'By Criticality',
-  },
-  auditViewSorting: {
-    byOccurrence: 'By Occurrence',
-    byCriticality: 'By Criticality',
+  sortings: {
+    alphabet: 'By Alphabet',
```
antonbauhofer, line 127:
```suggestion
    alphabet: 'Alphabetically',
```

src/shared/text.ts
```diff
@@ -111,26 +111,22 @@ export const text = {
...
+    criticality: 'By Criticality',
+    frequency: 'By Frequency',
```
antonbauhofer, line 129: I disagree with the name. I don't think frequency captures it correctly, so I originally named it "by occurrence". It doesn't need to be "occurrence", but let's try and find something different than "frequency".

src/Frontend/util/get-alphabetical-comparer.ts
```diff
@@ -1,88 +0,0 @@
-// SPDX-FileCopyrightText: Meta Platforms, Inc. and its affiliates
```
antonbauhofer, line 1: Awesome 👍

src/Frontend/util/__tests__/get-package-sorter.test.tsx
antonbauhofer, line 1: We should also add a test for sorting by "frequency"

src/Frontend/Components/ReportView/ReportView.tsx
antonbauhofer, line 1: After looking at the report view for a real example, I think it would be nice to have the sorting options here as well. In the example I looked at, there were lots of attributions with only copyright or comments and thus, the table looked very empty. This could be done independently of this diff, though. I have not opened a ticket yet, what do you think?

antonbauhofer (approved):
Looks good to me and performance is much better after this change.

src/shared/text.ts
mstykow: Thinking more about this, I think "alphabet" is not accurate anyways because it's not clear which attribute we're sorting alphabetically by.
Now I'm starting to think we should be more explicit here and say "By package name". What do you think?

src/shared/text.ts
mstykow: OK, I'll change it back to occurrence.

src/Frontend/Components/ReportView/ReportView.tsx
mstykow: The report view will be redesigned at a later point. At the moment it's barely functional when you open a big file (performance issues), so that would be the first problem to solve.

PR #2518, opened 2024-01-22
chore: extract components that display attribution info

alexzurbonsen:
### Summary of changes

This diff extracts components from the `AttributionColumn` that display package details into a reusable component.

We want to create a popup to show a diff between and attribution/package info and its origin. This popup should reuse the parts from `AttributionColumn` that display package info.

Run unit and e2e tests (CI).

Open Opossum on this commit and check that nothing changed in the looks and behaviour of audit and attribution view.


src/Frontend/Components/AttributionColumn/AttributionColumn.tsx
```diff
@@ -147,50 +126,12 @@ export function AttributionColumn(props: AttributionColumnProps): ReactElement {
...
-            displayPackageInfo={temporaryDisplayPackageInfo}
-            confirmEditWasPreferred={confirmEditWasPreferred}
-          />
-        </MuiBox>
+        <AttributionForm
+          packageInfo={temporaryDisplayPackageInfo}
+          isEditable={props.isEditable}
+          showHighlight={showHighlight}
+          confirmEditWasPreferred={confirmEditWasPreferred}
+        ></AttributionForm>
```
mstykow, line 134:
```suggestion
        />
```

src/Frontend/Components/AttributionColumn/AttributionForm.tsx
```diff
@@ -0,0 +1,118 @@
...
+};
+
+interface AttributionFormProps {
+  packageInfo: DisplayPackageInfo;
+  isEditable: boolean;
+  showHighlight?: boolean;
+  confirmEditWasPreferred?: Confirm;
+}
+
+export function AttributionForm(props: AttributionFormProps): ReactElement {
```
mstykow, line 42: no need for a return type of react components (`ReactElement` is also not correct since a React FC can also return null, undefined, etc.)
```suggestion
export function AttributionForm(props: AttributionFormProps) {
```

mstykow (approved):

Issue #2514, opened 2024-01-22
Create diff popup to display original and current attribution state


mstykow:
- add "COMPARE TO ORIGIN" button to the button row
- conditions for showing the button:
  - attribution has an attribution origin (`AuditingOptions.util` has logic for that; look up `originalPreferred`)

- render popup next to the "COMPARE TO ORIGIN" button

Popup
- Title: "Compare to original signal"
- Background light blue as the input fields are white
- Two columns 
  - left: original signal
  - right: latest confirmed attribution
- auditing option on top
- the fields for each attribution / signal are listed below the auditing options
- Button row with three buttons in the bottom right (from left to right)
  - "Apply changes" button 
    - Text: `APPLY CHANGES`
    - disabled
  - "Revert to original" button
    - Text: `REVERT TO ORIGINAL`
    - disbaled   
  - "Cancel" button
    - Text: `CANCEL`
    - onclick:
      - closes the popup

PR #2521, opened 2024-01-23
feat: add static diff popup

alexzurbonsen:
### Summary of changes

Add popup that compares the displayed package info to the package info of its original signal if it has one. Since comments of otherwise equal package infos are merged in the UI, the comments are not diffed when comparing to the original package info.

At this stage the popup is static and can only be opened and closed. All displayed info is not editable. Colors to highlight diffs and revert buttons will be added later.  

Closes [#2514](https://github.com/opossum-tool/OpossumUI/issues/2514). 

Added an e2e test.

Test manually. New "Compare to origin" is visible in the bottom right in Audit View.

### DIFF
--- a/src/Frontend/Components/AttributionColumn/AttributionForm.tsx
+++ b/src/Frontend/Components/AttributionColumn/AttributionForm.tsx
@@ -35,16 +35,26 @@ interface AttributionFormProps {
   packageInfo: DisplayPackageInfo;
   showHighlight?: boolean;
   onEdit?: Confirm;
+  variant?: 'default' | 'diff';
+  label?: string;
 }
 
-export function AttributionForm(props: AttributionFormProps) {
+export function AttributionForm({
+  packageInfo,
+  label,
+  onEdit,
+  showHighlight,
+  variant = 'default',
+}: AttributionFormProps) {
   const dispatch = useAppDispatch();
+  const isDiff = variant === 'diff';
 
   return (
-    <MuiBox sx={classes.formContainer}>
+    <MuiBox sx={classes.formContainer} aria-label={label}>
       <AuditingOptions
-        packageInfo={props.packageInfo}
-        isEditable={!!props.onEdit}
+        packageInfo={packageInfo}
+        isEditable={!!onEdit}
+        sx={{ flex: isDiff ? 1 : undefined }}
       />
       <MuiDivider variant={'middle'}>
         <MuiTypography>
@@ -52,45 +62,44 @@ export function AttributionForm(props: AttributionFormProps) {
         </MuiTypography>
       </MuiDivider>
       <PackageSubPanel
-        displayPackageInfo={props.packageInfo}
-        showHighlight={props.showHighlight}
-        onEdit={props.onEdit}
+        displayPackageInfo={packageInfo}
+        showHighlight={showHighlight}
+        onEdit={onEdit}
       />
       <MuiDivider variant={'middle'}>
         <MuiTypography>{text.attributionColumn.legalInformation}</MuiTypography>
       </MuiDivider>
       {renderAttributionType()}
-      {props.packageInfo.firstParty ? null : (
-        <>
-          <CopyrightSubPanel
-            displayPackageInfo={props.packageInfo}
-            showHighlight={props.showHighlight}
-            onEdit={props.onEdit}
-          />
-          <LicenseSubPanel
-            displayPackageInfo={props.packageInfo}
-            showHighlight={props.showHighlight}
-            onEdit={props.onEdit}
-          />
-        </>
-      )}
-      <CommentStack
-        displayPackageInfo={props.packageInfo}
-        onEdit={props.onEdit}
+      <CopyrightSubPanel
+        displayPackageInfo={packageInfo}
+        showHighlight={showHighlight}
+        onEdit={onEdit}
+        expanded={isDiff}
+        hidden={packageInfo.firstParty}
       />
+      <LicenseSubPanel
+        displayPackageInfo={packageInfo}
+        showHighlight={showHighlight}
+        onEdit={onEdit}
+        expanded={isDiff}
+        hidden={packageInfo.firstParty}
+      />
+      {isDiff ? null : (
+        <CommentStack displayPackageInfo={packageInfo} onEdit={onEdit} />
+      )}
     </MuiBox>
   );
 
   function renderAttributionType() {
     return (
       <MuiToggleButtonGroup
-        value={props.packageInfo.firstParty || false}
+        value={packageInfo.firstParty || false}
         exclusive
         onChange={(_, newValue) =>
-          props.onEdit?.(() =>
+          onEdit?.(() =>
             dispatch(
               setTemporaryDisplayPackageInfo({
-                ...props.packageInfo,
+                ...packageInfo,
                 firstParty: newValue,
                 wasPreferred: undefined,
               }),
@@ -99,6 +108,7 @@ export function AttributionForm(props: AttributionFormProps) {
         }
         size={'small'}
         fullWidth
+        disabled={!onEdit}
       >
         <MuiToggleButton value={false} disableRipple>
           {AttributionType.ThirdParty}
--- a/src/Frontend/Components/AttributionColumn/AuditingOptions.tsx
+++ b/src/Frontend/Components/AttributionColumn/AuditingOptions.tsx
@@ -5,7 +5,7 @@
 import AddIcon from '@mui/icons-material/Add';
 import MuiBox from '@mui/material/Box';
 import MuiChip from '@mui/material/Chip';
-import { SxProps } from '@mui/system';
+import { SxProps, Theme } from '@mui/system';
 import { useState } from 'react';
 
 import { DisplayPackageInfo } from '../../../shared/shared-types';
@@ -21,9 +21,10 @@ const classes = {
 interface Props {
   packageInfo: DisplayPackageInfo;
   isEditable: boolean;
+  sx?: SxProps<Theme>;
 }
 
-export function AuditingOptions({ packageInfo, isEditable }: Props) {
+export function AuditingOptions({ packageInfo, isEditable, sx }: Props) {
   const options = useAuditingOptions({ packageInfo, isEditable });
   const [anchorEl, setAnchorEl] = useState<HTMLElement>();
   const unselectedOptions = options.filter(
@@ -32,7 +33,7 @@ export function AuditingOptions({ packageInfo, isEditable }: Props) {
 
   return options.length ? (
     <>
-      <MuiBox sx={classes.container}>
+      <MuiBox sx={{ ...classes.container, ...sx }}>
         {renderTriggerButton()}
         {renderSelectedOptions()}
       </MuiBox>
--- a/src/Frontend/Components/AttributionColumn/ButtonRow.tsx
+++ b/src/Frontend/Components/AttributionColumn/ButtonRow.tsx
@@ -3,19 +3,24 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 import MuiBox from '@mui/material/Box';
+import { useMemo, useState } from 'react';
 
 import { DisplayPackageInfo } from '../../../shared/shared-types';
+import { text } from '../../../shared/text';
 import { ButtonText } from '../../enums/enums';
 import { EMPTY_DISPLAY_PACKAGE_INFO } from '../../shared-constants';
 import { setTemporaryDisplayPackageInfo } from '../../state/actions/resource-actions/all-views-simple-actions';
 import { useAppDispatch, useAppSelector } from '../../state/hooks';
 import {
+  getExternalAttributions,
   getIsGlobalSavingDisabled,
   getIsSavingDisabled,
   getManualDisplayPackageInfoOfSelected,
   wereTemporaryDisplayPackageInfoModified,
 } from '../../state/selectors/all-views-resource-selectors';
+import { convertPackageInfoToDisplayPackageInfo } from '../../util/convert-package-info';
 import { Button, ButtonProps } from '../Button/Button';
+import { DiffPopup } from '../DiffPopup/DiffPopup';
 import { SplitButton } from '../SplitButton/SplitButton';
 
 interface ButtonRowProps {
@@ -50,6 +55,19 @@ export function ButtonRow({
     EMPTY_DISPLAY_PACKAGE_INFO;
   const isSavingDisabled = useAppSelector(getIsSavingDisabled);
   const isGlobalSavingDisabled = useAppSelector(getIsGlobalSavingDisabled);
+  const externalAttributions = useAppSelector(getExternalAttributions);
+  const [isDiffPopupOpen, setIsDiffPopupOpen] = useState(false);
+
+  const originalDisplayPackageInfo = useMemo(() => {
+    const originalPackageInfo = !!displayPackageInfo.originIds?.length
+      ? Object.values(externalAttributions).find(({ originIds }) =>
+          originIds?.some((id) => displayPackageInfo.originIds?.includes(id)),
+        )
+      : undefined;
+    return originalPackageInfo
+      ? convertPackageInfoToDisplayPackageInfo(originalPackageInfo, [])
+      : undefined;
+  }, [externalAttributions, displayPackageInfo.originIds]);
 
   return (
     !areButtonsHidden && (
@@ -64,7 +82,9 @@ export function ButtonRow({
         {renderSaveButton()}
         {renderDeleteButton()}
         {renderRevertButton()}
+        {renderCompareButton()}
         {renderAdditionalActions()}
+        {renderDiffPopup()}
       </MuiBox>
     )
   );
@@ -140,6 +160,16 @@ export function ButtonRow({
     );
   }
 
+  function renderCompareButton() {
+    return originalDisplayPackageInfo ? (
+      <Button
+        onClick={() => setIsDiffPopupOpen(true)}
+        color={'secondary'}
+        buttonText={text.buttons.compareToOrigin}
+      />
+    ) : null;
+  }
+
   function renderAdditionalActions() {
     return additionalActions.map((action) => (
       <Button
@@ -150,5 +180,16 @@ export function ButtonRow({
         disabled={action.disabled}
       />
     ));
+  }
+
+  function renderDiffPopup() {
+    return originalDisplayPackageInfo ? (
+      <DiffPopup
+        original={originalDisplayPackageInfo}
+        current={displayPackageInfo}
+        isOpen={isDiffPopupOpen}
+        setOpen={setIsDiffPopupOpen}
+      />
+    ) : null;
   }
 }
--- a/src/Frontend/Components/AttributionColumn/CopyrightSubPanel.tsx
+++ b/src/Frontend/Components/AttributionColumn/CopyrightSubPanel.tsx
@@ -3,7 +3,6 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 import MuiBox from '@mui/material/Box';
-import { ReactElement } from 'react';
 
 import { DisplayPackageInfo } from '../../../shared/shared-types';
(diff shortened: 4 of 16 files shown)


src/Frontend/Components/AttributionColumn/ButtonRow.tsx
```diff
@@ -151,4 +180,23 @@ export function ButtonRow({
       />
     ));
   }
+
+  function renderDiffPopup() {
+    // TODO unpacking just tricks the compiler? asserting non-null is not allowed by linter rule
```
alexzurbonsen, line 185: still to be solved...

src/Frontend/Components/DiffPopup/DiffPopup.tsx
```diff
@@ -0,0 +1,93 @@
...
+  originalDisplayPackageInfo: DisplayPackageInfo;
+}
+
+interface DiffPopupProps {
+  packageInfosForDiff: DiffViewProps;
+  isOpen: boolean;
+  setOpen: Dispatch<SetStateAction<boolean>>;
+}
+
+function DiffView(props: DiffViewProps): ReactElement {
```
mstykow, line 37: it's not great having multiple react comps inside the same file. if you really think it's worth extracting this, then let's do it properly and create a sub-folder of `DiffPopup` for a component called `DiffView`.

alternatively, you could inline this component at the end of `DiffPopup` as a function that starts with the word "render", as I've done for example in `AttributionList.tsx`. this has the advantage, that you can access all the parent's variables directly without needing to define yet another interface like the one in line 31.

when the component is small-ish, as it is in your case, i personally prefer this latter option.

src/Frontend/enums/enums.ts
```diff
@@ -100,6 +103,7 @@ export enum PieChartCriticalityNames {
 export enum HighlightingColor {
   LightOrange = 'Light Orange',
   DarkOrange = 'Dark Orange',
+  Grey = 'Grey',
```
mstykow, line 106: do we still need this? i thought you are using `visiblity: none` now.

src/Frontend/enums/enums.ts
```diff
@@ -69,6 +71,7 @@ export enum ButtonText {
   Ok = 'Ok',
   OpenDotOpossumFile = 'Open ".opossum" file',
   Revert = 'Revert',
+  RevertAll = 'Revert all',
```
mstykow, line 74: using this enum is deprecated in favor of the `text.ts` file. i should probably open a PR to mark it as such but in the meanwhile, please move your new user-facing texts to that file instead.

src/Frontend/Components/AttributionColumn/AuditingOptions.tsx
```diff
@@ -16,14 +16,22 @@ import { useAuditingOptions } from './AuditingOptions.util';
...
   container: { display: 'flex', gap: '8px', flexWrap: 'wrap' },
+  containerDiffView: {
+    flexGrow: 1,
+  },
 } satisfies SxProps;
 
 interface Props {
   packageInfo: DisplayPackageInfo;
   isEditable: boolean;
+  isDiffView?: boolean;
```
mstykow, line 27: this prop is too specific for a component that knows nothing about a "diff view". instead, please introduce an `sx` prop that you can use as the parent to set `flex: 1`.

this comment applies also to the other sub-components where you introduced this prop.

src/Frontend/Components/AttributionColumn/ButtonRow.tsx
mstykow: there is no need for a trick here. both the compare button and the diff popup should not be rendered when the original package info does not exist. so you're missing if statements that return null when there is no such package info.

src/Frontend/Components/AttributionColumn/CopyrightSubPanel.tsx
```diff
@@ -17,24 +17,32 @@ interface CopyrightSubPanelProps {
...
   onEdit?: Confirm;
+  isDiffView?: boolean;
+  isHidden?: boolean;
 }
 
 export function CopyrightSubPanel({
   displayPackageInfo,
   onEdit,
   showHighlight,
+  isDiffView,
```
mstykow, line 28: here, a diff-view-agnostic name for this prop could be "expanded".

maybe even better: create a prop called "variant" which has the three options `'default' | 'expanded' | 'hidden'` then you don't need to introduce two separate, but related, props.

src/Frontend/Components/AttributionColumn/AttributionForm.tsx
```diff
@@ -66,18 +69,36 @@ export function AttributionForm(props: AttributionFormProps) {
...
-      <CommentStack
-        displayPackageInfo={props.packageInfo}
-        onEdit={props.onEdit}
-      />
+      {props.packageInfo.firstParty && props.isDiffView ? (
+        <>
+          <CopyrightSubPanel
+            displayPackageInfo={props.packageInfo}
+            isDiffView={props.isDiffView}
+            isHidden={true}
```
mstykow, line 87: you can see here that something is quite confusing about the new props: if the component is always hidden, why would i need to render it at all?

src/Frontend/Components/AttributionColumn/AttributionForm.tsx
```diff
@@ -66,18 +69,36 @@ export function AttributionForm(props: AttributionFormProps) {
...
             onEdit={props.onEdit}
+            isDiffView={props.isDiffView}
           />
         </>
       )}
-      <CommentStack
-        displayPackageInfo={props.packageInfo}
-        onEdit={props.onEdit}
-      />
+      {props.packageInfo.firstParty && props.isDiffView ? (
```
mstykow, line 82: introducing this second but related ternary to the one in line 66 is making the code hard to understand. i think once you change your props to use the `variant` i suggested in another comment, you won't need this entire block anymore.

src/Frontend/Components/DiffPopup/DiffPopup.tsx
```diff
@@ -0,0 +1,93 @@
...
+import MuiDivider from '@mui/material/Divider';
+import { Dispatch, ReactElement, SetStateAction } from 'react';
+
+import { DisplayPackageInfo } from '../../../shared/shared-types';
+import { text } from '../../../shared/text';
+import { ButtonText } from '../../enums/enums';
+import { AttributionForm } from '../AttributionColumn/AttributionForm';
+import { NotificationPopup } from '../NotificationPopup/NotificationPopup';
+
+const classes = {
```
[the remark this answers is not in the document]
alexzurbonsen: Sure, I was wondering about that at some point but must have forgotten.

src/Frontend/enums/enums.ts
alexzurbonsen: Right. Oversight on my side.

src/Frontend/Components/AttributionColumn/AttributionForm.tsx
alexzurbonsen: True, that is hidden in the if statement around it.

src/Frontend/Components/AttributionColumn/AuditingOptions.tsx
alexzurbonsen: Good idea.

src/Frontend/Components/AttributionColumn/CopyrightSubPanel.tsx
alexzurbonsen: good point.

src/shared/text.ts
```diff
@@ -53,9 +53,12 @@ export const text = {
     replaceSelected: 'Replace Selected',
   },
   buttons: {
+    applyChanges: 'Apply changes',
     cancel: 'Cancel',
+    compareToOrigin: 'Compare to origin',
     confirm: 'Confirm',
     filter: 'Filter',
+    revertAll: 'Revert all',
```
mstykow, line 61: i would prefer if we create a new object "diffPopup" here instead of merging these diff-popup-specific texts with the generic "buttons" object.

src/Frontend/Components/DiffPopup/DiffPopup.tsx
```diff
@@ -0,0 +1,77 @@
...
+          flexItem={true}
+          orientation={'vertical'}
+        />
+        <AttributionForm
+          packageInfo={{
+            ...props.packageInfosForDiff.displayedDisplayPackageInfo,
+            comments: undefined,
+          }}
+          isDiffView={true}
+          label={'diff displayed'}
```
mstykow, line 72: "displayed"? i don't quite understand that term. do you mean "current"?

src/Frontend/Components/DiffPopup/DiffPopup.tsx
```diff
@@ -0,0 +1,77 @@
...
+        />
+        <MuiDivider
+          variant={'middle'}
+          flexItem={true}
+          orientation={'vertical'}
+        />
+        <AttributionForm
+          packageInfo={{
+            ...props.packageInfosForDiff.displayedDisplayPackageInfo,
+            comments: undefined,
```
mstykow, line 69: this seems hacky and unnecessary: in diff view, the attribution form already does not render any comments. so why do i need to set them to undefined here?

src/Frontend/Components/AttributionColumn/AttributionForm.tsx
```diff
@@ -35,16 +35,38 @@ interface AttributionFormProps {
   packageInfo: DisplayPackageInfo;
   showHighlight?: boolean;
   onEdit?: Confirm;
+  isDiffView?: boolean;
+  label?: string;
+}
+
+function getVariant({
```
mstykow, line 42: instead of extracting this function outside the FC, defining (repeating) variables names and types, and calling it multiple times, i'd put it inside the component and wrap it in a `useMemo`. in fact, you then don't need a function at all since it'll be a memoized constant.

src/Frontend/Components/AttributionColumn/AttributionForm.tsx
```diff
@@ -35,16 +35,38 @@ interface AttributionFormProps {
...
 export function AttributionForm(props: AttributionFormProps) {
   const dispatch = useAppDispatch();
 
   return (
-    <MuiBox sx={classes.formContainer}>
+    <MuiBox sx={classes.formContainer} aria-label={props.label}>
       <AuditingOptions
         packageInfo={props.packageInfo}
         isEditable={!!props.onEdit}
+        {...(props.isDiffView ? { sx: { flexGrow: 1 } } : {})}
```
mstykow, line 69:
```suggestion
        sx={{ flex: props.isDiffView ? 1 : undefined }}
```

src/Frontend/Components/DiffPopup/DiffPopup.tsx
```diff
@@ -0,0 +1,77 @@
...
+  originalDisplayPackageInfo: DisplayPackageInfo;
+}
+
+interface DiffPopupProps {
+  packageInfosForDiff: DiffViewProps;
+  isOpen: boolean;
+  setOpen: Dispatch<SetStateAction<boolean>>;
+}
+
+export function DiffPopup(props: DiffPopupProps): ReactElement {
```
mstykow, line 25:
```suggestion
export function DiffPopup(props: DiffPopupProps) {
```

src/Frontend/Components/DiffPopup/DiffPopup.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { NotificationPopup } from '../NotificationPopup/NotificationPopup';
+import { DiffPopupContainer } from './DiffPopup.style';
+
+interface DiffViewProps {
+  displayedDisplayPackageInfo: DisplayPackageInfo;
+  originalDisplayPackageInfo: DisplayPackageInfo;
+}
+
+interface DiffPopupProps {
+  packageInfosForDiff: DiffViewProps;
```
mstykow, line 20:
```suggestion
  original: DisplayPackageInfo;
  current: DisplayPackageInfo;
  ...
```

src/Frontend/Components/DiffPopup/DiffPopup.tsx
```diff
@@ -0,0 +1,77 @@
+// SPDX-FileCopyrightText: Meta Platforms, Inc. and its affiliates
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+import MuiDivider from '@mui/material/Divider';
+import { Dispatch, ReactElement, SetStateAction } from 'react';
```
mstykow, line 6: since these are all types, the import is unnecessary if you write `React.X` everywhere instead (global namespace).

src/Frontend/Components/AttributionColumn/PackageAutocomplete.tsx
```diff
@@ -51,6 +48,7 @@ type AutocompleteAttribute = Extract<
 interface Props {
   title: string;
   attribute: AutocompleteAttribute;
+  temporaryPackageInfo: DisplayPackageInfo;
```
mstykow, line 51: i don't think we gain much by calling this "temporary".
```suggestion
  packageInfo: DisplayPackageInfo;
```

src/Frontend/Components/AttributionColumn/ButtonRow.tsx
```diff
@@ -50,6 +55,18 @@ export function ButtonRow({
     EMPTY_DISPLAY_PACKAGE_INFO;
   const isSavingDisabled = useAppSelector(getIsSavingDisabled);
   const isGlobalSavingDisabled = useAppSelector(getIsGlobalSavingDisabled);
+  const externalAttributions = useAppSelector(getExternalAttributions);
+  const [openDiffPopup, setOpenDiffPopup] = useState(false);
```
mstykow, line 59: "openDiffPopup" sounds like an action but it's actually a boolean. i could see either "isDiffPopupOpen" or "diffPopupOpen".

src/Frontend/Components/AttributionColumn/ButtonRow.tsx
```diff
@@ -151,4 +181,24 @@ export function ButtonRow({
       />
     ));
   }
+
+  function renderDiffPopup() {
+    if (originalPackageInfo) {
+      const originDisplayedPackageInfo = convertPackageInfoToDisplayPackageInfo(
```
mstykow, line 187: could you do this conversion as part of the `useMemo` above?

src/Frontend/Components/AttributionColumn/AuditingOptions.tsx
```diff
@@ -32,7 +33,14 @@ export function AuditingOptions({ packageInfo, isEditable }: Props) {
 
   return options.length ? (
     <>
-      <MuiBox sx={classes.container}>
+      <MuiBox
+        sx={[
```
mstykow, line 37:
```suggestion
        sx={{...classes.container, ...sx}}
```

src/shared/shared-types.ts
```diff
@@ -283,3 +283,9 @@ export type AutocompleteSignal = DisplayPackageInfo & {
   suffix?: string;
   default?: boolean;
 };
+
+export type VariantProp =
```
mstykow, line 287: this variant is specific to the attribution form and thus deserves a specific name such as "AttributionFormVariant". also, this folder is not the right one for it since the `src/shared` folder is for shared stuff across all parts of the repo but you only need this type as part of the frontend.
since this type belongs to the attribution form, why not just keep it there and import from there instead of putting it into such a high-level shared place?

mstykow (changes requested):

src/shared/text.ts
alexzurbonsen: Okay :)

src/Frontend/Components/DiffPopup/DiffPopup.tsx
alexzurbonsen: Yes. I am distinguishing `original` and `displayed` (which is the currently displayed package info). At some point I thought that was more clear, but I don't mind to change this.

src/Frontend/Components/DiffPopup/DiffPopup.tsx
alexzurbonsen: Right. Good catch.

src/Frontend/Components/DiffPopup/DiffPopup.tsx
alexzurbonsen: What is the advantage of doing that by the way? As I mentioned in another PR, my IDE always displays a warning about usage of global variables.

src/Frontend/Components/AttributionColumn/PackageAutocomplete.tsx
alexzurbonsen: True.

src/Frontend/Components/AttributionColumn/AuditingOptions.tsx
alexzurbonsen: I looked this up and they said it is the recommended way to do it because `sx` can be an array. But at least the compiler does not complain, seems to work.

src/shared/shared-types.ts
alexzurbonsen: Sounds reasonable. For some reason I thought that was a pattern that we used. But now it doesn't seem so reasonable to me either. Moved it.

alexzurbonsen: About the naming: the two components that use it are Copyright- and LicenseSubPanel. Are you okay with `PanelVariantProp`? I named it such for now.

src/shared/text.ts
mstykow: some of my comments are opinionated. feel free to disagree!

src/Frontend/Components/DiffPopup/DiffPopup.tsx
mstykow: btw: what you did is of course not incorrect. it's just my opinionated suggestion to make my life easier in future refactorings, for example if suddenly the component returns null in a certain situation i'll have to touch this line for no good reason since the actual return value of an FC is `React.ReactNode` which is wider than `React.ReactElement`. but instead of arguing over the correct return type, why not ignore it completely since it doesn't add much value. it's almost like saying that the function can return anything...

src/Frontend/Components/DiffPopup/DiffPopup.tsx
mstykow: again, my opinionated suggestion. i personally prefer to write fewer lines of code and not deal with adding/removing or re-ordering imports. so if i don't _need_ to import something, why would i? it would be good if you could find out why your IDE is complaining. is it not correctly configured or does it actually think it's an antipattern to use the global namespace?

src/Frontend/Components/AttributionColumn/AuditingOptions.tsx
mstykow: yes, i saw the MUI docs too but it's just so much simpler doing it this way...

mstykow (changes requested):

src/shared/text.ts
alexzurbonsen: I think it is good idea to start this way. As long as the buttons are all next to each other it should be fine.

src/Frontend/Components/AttributionColumn/ButtonRow.tsx
```diff
@@ -50,6 +55,19 @@ export function ButtonRow({
...
+
+  const originalDisplayPackageInfo = useMemo(() => {
+    const originalPackageInfo = !!displayPackageInfo.originIds?.length
+      ? Object.values(externalAttributions).find(({ originIds }) =>
+          originIds?.some((id) => displayPackageInfo.originIds?.includes(id)),
+        )
+      : undefined;
+    return originalPackageInfo
+      ? convertPackageInfoToDisplayPackageInfo(originalPackageInfo, [])
+      : undefined;
```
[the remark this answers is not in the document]
alexzurbonsen: We decided to always show the button if there is a diff for two reasons:
- stability of the UI, no buttons jump around
- hidding it when there is no diff, means that the button can be hidden for two reasons. Either there is no origin or no changes were made. This may be confusing to the user who is not sure whether the attribution was edited already.

PR #2531, opened 2024-01-25
fix: cannot save comments by hotkey

### LINKED ISSUE #2525: Saving comment by keyboard shortcut is not working in Audit View
**Describe the bug**
Saving by keyboard shortcut is not working in Audit View.

**To Reproduce**

1. Go to Audit View.
2. Modify the comment on an attribution.
3. Save using "Ctrl + S"
4. Notice that the comment field is emptied and saved.

**Expected behavior**
The changes to the comment are saved.

mstykow:
### Summary of changes

- code forgot to convert display package info to package info which caused an error during serialization

closes #2525

Adjust a comment in audit view and use the keyboard shortcut to save.


src/Frontend/Components/ResourceDetailsAttributionColumn/ResourceDetailsAttributionColumn.tsx
```diff
@@ -141,15 +141,15 @@ export function ResourceDetailsAttributionColumn(
         unlinkAttributionAndSavePackageInfoIfSavingIsNotDisabled(
           selectedResourceId,
           attributionIdOfSelectedPackageInManualPanel,
-          temporaryDisplayPackageInfo,
+          convertDisplayPackageInfoToPackageInfo(temporaryDisplayPackageInfo),
```
alexzurbonsen, line 144: Since we had this problem a couple of times: is there a way to fix this with types?

Could we for example type the functions involved in saving differently? Say,
```
export function unlinkAttributionAndSavePackageInfoIfSavingIsNotDisabled<T extends PackageInfo>(
  resourceId: string,
  attributionId: string,
  packageInfo: T & Record<Exclude<keyof T, keyof PackageInfo>, never>,
)
...
```
or
```
export function unlinkAttributionAndSavePackageInfoIfSavingIsNotDisabled<T extends PackageInfo>(
  resourceId: string,
  attributionId: string,
  packageInfo: T & Record<'comments', never>,
)
...
```

alexzurbonsen (approved):

src/Frontend/Components/ResourceDetailsAttributionColumn/ResourceDetailsAttributionColumn.tsx
mstykow: good point. i've made a separate commit to address this issue which indeed caught more problematic places.

src/Frontend/Components/ResourceDetailsAttributionColumn/__tests__/ResourceDetailsAttributionColumn.test.tsx
```diff
@@ -19,26 +20,25 @@ import { setSelectedResourceId } from '../../../state/actions/resource-actions/a
 import { loadFromFile } from '../../../state/actions/resource-actions/load-actions';
 import { getParsedInputFileEnrichedWithTestData } from '../../../test-helpers/general-test-helpers';
 import { renderComponent } from '../../../test-helpers/render';
+import { convertPackageInfoToDisplayPackageInfo } from '../../../util/convert-package-info';
 import { ResourceDetailsAttributionColumn } from '../ResourceDetailsAttributionColumn';
 
 const testManualLicense = 'Manual attribution license.';
 const testManualLicense2 = 'Another manual attribution license.';
-const testTemporaryDisplayPackageInfo: DisplayPackageInfo = {
+const testTemporaryDisplayPackageInfo: PackageInfo = {
```
alexzurbonsen, line 28: nit: we should probably delete `Display` from the variable names where necessary.

alexzurbonsen:
Really like the typing approach!

PR #2533, opened 2024-01-25
fix: correct multiline textfield padding

### LINKED ISSUE #2530: Scroll bar in multi-line textfields is misaligned
**Describe the bug**
The scroll bar in multi-line textfields should be at the edge of the input field but is currently separated from this edge by some padding.

**To Reproduce**

Image

mstykow:
### Summary of changes

- move scroll bar to edge of input field and apply padding around input element instead

BEFORE:
image

AFTER:
image

closes #2530

Check that both single and multiline textfields have the correct padding.

### DIFF
--- a/src/Frontend/Components/InputElements/TextBox.tsx
+++ b/src/Frontend/Components/InputElements/TextBox.tsx
@@ -2,27 +2,23 @@
 // SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
 //
 // SPDX-License-Identifier: Apache-2.0
-import { SxProps } from '@mui/material';
 import MuiBox from '@mui/material/Box';
 import MuiInputAdornment from '@mui/material/InputAdornment';
 import MuiTextField from '@mui/material/TextField';
-import { ReactElement } from 'react';
 
 import { HighlightingColor } from '../../enums/enums';
-import { getSxFromPropsAndClasses } from '../../util/get-sx-from-props-and-classes';
 import { inputElementClasses, InputElementProps } from './shared';
 
 interface TextProps extends InputElementProps {
-  textFieldInputSx?: SxProps;
   minRows?: number;
   maxRows?: number;
-  endIcon?: ReactElement;
+  endIcon?: React.ReactElement;
   multiline?: boolean;
   highlightingColor?: HighlightingColor;
   error?: boolean;
 }
 
-export function TextBox(props: TextProps): ReactElement {
+export function TextBox(props: TextProps) {
   const isDefaultHighlighting =
     props.highlightingColor === HighlightingColor.LightOrange ||
     props.highlightingColor === undefined;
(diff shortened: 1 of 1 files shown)


src/Frontend/Components/InputElements/TextBox.tsx
```diff
@@ -33,24 +29,30 @@ export function TextBox(props: TextProps): ReactElement {
...
         disabled={!props.isEditable}
         error={props.error}
-        sx={textBoxSx}
+        sx={{
+          ...(props.isHighlighted ? highlightedStyling : {}),
+          ...inputElementClasses.textField,
+        }}
         label={props.title}
         InputProps={{
+          slotProps: {
```
alexzurbonsen, line 43: Seems logical and all works nicely. But I couldn't find this in the documentation (https://mui.com/material-ui/api/outlined-input/) Do you have a link for reference maybe?

src/Frontend/Components/InputElements/TextBox.tsx
```diff
@@ -2,27 +2,23 @@
...
 import MuiInputAdornment from '@mui/material/InputAdornment';
 import MuiTextField from '@mui/material/TextField';
-import { ReactElement } from 'react';
 
 import { HighlightingColor } from '../../enums/enums';
-import { getSxFromPropsAndClasses } from '../../util/get-sx-from-props-and-classes';
 import { inputElementClasses, InputElementProps } from './shared';
 
 interface TextProps extends InputElementProps {
-  textFieldInputSx?: SxProps;
```
alexzurbonsen, line 16: This prop was unused?

src/Frontend/Components/InputElements/TextBox.tsx
```diff
@@ -2,27 +2,23 @@
...
   minRows?: number;
   maxRows?: number;
-  endIcon?: ReactElement;
+  endIcon?: React.ReactElement;
```
alexzurbonsen, line 15: Why is it preferred to take these from a global namespace? My IDE complains mildly if I do this.

alexzurbonsen (approved):

src/Frontend/Components/InputElements/TextBox.tsx
mstykow: yes

src/Frontend/Components/InputElements/TextBox.tsx
mstykow: because it saves an import and thus also javascript lines of code. if the IDE is complaining, then it's not correctly configured to honor the global react namespace.

src/Frontend/Components/InputElements/TextBox.tsx
mstykow: https://mui.com/material-ui/api/input/#input-prop-slotProps

src/Frontend/Components/InputElements/TextBox.tsx
alexzurbonsen: With IDE is complaining I mean, that it is complaining about the usage of a global namespace. Apparently, it is configured to not appreciate that. It does recognize the variable though.

src/Frontend/Components/InputElements/TextBox.tsx
alexzurbonsen: Thanks!

PR #2534, opened 2024-01-25
fix: "license text modified" overlaps license text

### LINKED ISSUE #2529: "License text modified" overlaps with license name on smaller screens
**Describe the bug**
On smaller screens, the "license text modified" label overlaps with the actual license name.

**To Reproduce**

1. Go to Audit View
2. Modify the license text
3. Reduce screen size until overlap occurs

**Expected behavior**
Either move the label out of the input field in general, or find a way to start hiding it as the screen shrinks.

mstykow:
### Summary of changes

- replace end adornment label by end adornment icon in order to save space

BEFORE:
image

AFTER:
image

closes #2529

Inspect an attribution with a modified (non-empty-string) license text.

### DIFF
--- a/src/Frontend/Components/AttributionColumn/LicenseSubPanel.tsx
+++ b/src/Frontend/Components/AttributionColumn/LicenseSubPanel.tsx
@@ -2,21 +2,22 @@
 // SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
 //
 // SPDX-License-Identifier: Apache-2.0
+import EditNoteIcon from '@mui/icons-material/EditNote';
 import ExpandMoreIcon from '@mui/icons-material/ExpandMore';
 import MuiAccordion from '@mui/material/Accordion';
 import MuiAccordionDetails from '@mui/material/AccordionDetails';
 import MuiAccordionSummary from '@mui/material/AccordionSummary';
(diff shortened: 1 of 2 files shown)


src/Frontend/Components/AttributionColumn/LicenseSubPanel.tsx
```diff
@@ -124,9 +121,12 @@ export function LicenseSubPanel({
             onEdit={onEdit}
             endAdornment={
               displayPackageInfo.licenseText ? (
-                <MuiInputAdornment position="end" sx={classes.endAdornment}>
-                  {text.attributionColumn.licenseTextModified}
-                </MuiInputAdornment>
+                <MuiTooltip title={text.attributionColumn.licenseTextModified}>
+                  <EditNoteIcon
```
alexzurbonsen, line 125: subjective remark: my first impression in the UI was that I have to press the icon to edit the license. But I couldn't find a more suitable icon quickly. Apparently people also uses asterisks to mark change. But MUI doesn't have one (?).

Do you think that is worth a second thought?

mstykow: Good point. I think we should take a fundamentally different approach then.

mstykow:
Will take different approach.

PR #2542, opened 2024-01-26
feat: add colours to the static diff popup

alexzurbonsen:
### Summary of changes

Add colours to the package info fields that indicate change.

If there is a diff between two relevant fields the original value is displayed in red and the currently displayed value is green.
Two fields differ if they are not equal. More sophisticated diffing is kept for later efforts.

Only if original and displayed package are of type third party, we also diff the copyright and license fields.

Closes [#2517](https://github.com/opossum-tool/OpossumUI/issues/2517)

Added e2e test.

Manually. Open the Diff popup and play around with changing the package info.

### DIFF
--- a/src/Frontend/Components/DiffPopup/DiffPopup.tsx
+++ b/src/Frontend/Components/DiffPopup/DiffPopup.tsx
@@ -1,0 +1,66 @@
+// SPDX-FileCopyrightText: Meta Platforms, Inc. and its affiliates
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+import MuiDivider from '@mui/material/Divider';
+
+import { DisplayPackageInfo } from '../../../shared/shared-types';
+import { text } from '../../../shared/text';
+import { AttributionForm } from '../AttributionColumn/AttributionForm';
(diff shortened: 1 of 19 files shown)


src/Frontend/Components/DiffPopup/DiffPopup.tsx
```diff
@@ -17,7 +22,96 @@ interface DiffPopupProps {
...
+>;
+type AttributionFormConfigThirdParty = Pick<
+  PackageInfoCore,
+  'copyright' | 'licenseName' | 'licenseText'
+>;
+export type AttributionFormConfig = AttributionFormConfigBase &
+  AttributionFormConfigThirdParty & { isDiffView?: boolean };
+export type AttributionFormConfigAttribute = {
+  colorValue?: string;
+};
```
alexzurbonsen, line 37: Two questions:
- I put the types here since that is where we use them. But since they are a property of the attribution form they should may be live in `AttributionForm`?
- I could add a layer in `AttributionFromConfig`, that provides information about what is configured here. Like so:
```
export type AttributionFormConfig = {color: AttributionFormConfigBase &
  AttributionFormConfigThirdParty} & { isDiffView?: boolean };
```
It might be overkill since we are only using the config for this currently, but it might provide more clarity.
Or we call the whole thing `AttributionFormColorConfig` instead. What do you think?

alexzurbonsen:
Closing in favour of v2 #2549

PR #2549, opened 2024-01-30
feat: add colours to the static diff popup

alexzurbonsen:
### Summary of changes

Add colours to the package info fields that indicate change.

If there is a diff between two relevant fields the original value is displayed in red and the currently displayed value is green.
Two fields differ if they are not equal. More sophisticated diffing is kept for later efforts.

Only if original and displayed package are of type third party, we also diff the copyright and license fields.

Closes [#2517](https://github.com/opossum-tool/OpossumUI/issues/2517)

Added e2e test.

Manually. Open the Diff popup and play around with changing the package info.

### DIFF
--- a/src/Frontend/Components/DiffPopup/__test__/DiffPopup.util.test.tsx
+++ b/src/Frontend/Components/DiffPopup/__test__/DiffPopup.util.test.tsx
@@ -1,0 +1,117 @@
+// SPDX-FileCopyrightText: Meta Platforms, Inc. and its affiliates
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+import { waitFor } from '@testing-library/react';
+import { without } from 'lodash';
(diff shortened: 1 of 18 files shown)


src/Frontend/Components/DiffPopup/__test__/DiffPopup.util.test.tsx
```diff
@@ -0,0 +1,128 @@
...
+        { original: DisplayPackageInfo; current: DisplayPackageInfo },
+        [AttributionFormConfig, AttributionFormConfig]
+      >(() =>
+        useAttributionFormConfigs({
+          original: originalPackageInfo,
+          current: currentPackageInfo,
+        }),
+      );
+      await waitFor(() =>
+        expect(result.current[0]).toStrictEqual(expectedOriginalConfig),
```
alexzurbonsen, line 121: With out naming it is a bit unfortunate that we have to use `.current` here. But I didn't change them. You think we should?

mstykow (approved):

PR #2551, opened 2024-01-30
chore: simplify follow-up attribute type

mstykow:
### Summary of changes

- convert follow-up from string enum to boolean since it only contains one value
- remove unused criticality boolean calculated during deserialization

Reduce code complexity: "followUp" behaves like a boolean and is similar to other attributes like "wasPreferred".

Check that the "needs follow-up" as part of the auditing options still works as expected.

### DIFF
--- a/src/ElectronBackend/input/__tests__/importFromFile.test.ts
+++ b/src/ElectronBackend/input/__tests__/importFromFile.test.ts
@@ -11,7 +11,6 @@ import { AllowedFrontendChannels } from '../../../shared/ipc-channels';
 import {
   Criticality,
   DiscreteConfidence,
-  FollowUp,
   ParsedFileContent,
 } from '../../../shared/shared-types';
 import { writeFile, writeOpossumFile } from '../../../shared/write-file';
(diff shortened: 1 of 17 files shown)


meretp (approved):
Looks good to me. It probably makes sense to also update the input file schema to use a boolean for followUp, do you know if we already have a ticket to do so?

mstykow:
> Looks good to me. It probably makes sense to also update the input file schema to use a boolean for followUp, do you know if we already have a ticket to do so?

Since this attribute is used by the pipeline generating the Opossum file, I cannot change the schema here without introducing a breaking change. Thus, I would prefer to keep this change in OpossumUI only.

meretp:
> Since this attribute is used by the pipeline generating the Opossum file, I cannot change the schema here without introducing a breaking change. Thus, I would prefer to keep this change in OpossumUI only.
Yeah, sure, I actually had the pipeline in mind, we could update it everywhere. But as we would probably need to keep some of the code to be backwards compatible, this might not be worth it.

PR #2552, opened 2024-01-31
fix: updating attribution with original signal or suggestion

mstykow:
### Summary of changes

- attributes "id", "preSelected" should not be used when updating an attribution via an original signal or suggestion
- also ensure multiple comments are dropped

1. Modify a previously preferred and pre-selected attribution, save the changes, and then click on the chip to revert to the original state.
2. Pick a suggestion from an autocomplete via the "+" button and notice that id, pre-selected do not change while an suggested signal that had multiple comments drops these comments.


alexzurbonsen (approved):
Nice improvement!

Issue #2515, opened 2024-01-22
Expand diff popup functionality to allow resetting individual attributes


mstykow:
Global: 
- Revert All Button (if diff exists)
- Apply changes (if changes exist)
Local:
- revert buttons for fields and auditing options (if diff exists)

State changes:
- store locally in `useState` hook

Details: #2488

Issue #2488, opened 2024-01-11
Create diff popup for comparing an attribution with its original signal


FlorianSchepers:
**Is your feature request related to a problem? Please describe.**
As a user I want to be able to see the changes made to an attribution after it was created from a signal. This is currently not possible in a lightweight and convinient mannor. 

**Describe the solution you'd like**
A popup is introduced that brings the content of the current attribution face to face with the content of the original signal. 

Open diff view popup button
-  "COMPARE TO ORIGIN" button
- name: `COMPARE TO ORIGIN` 
- location: in buttom right corner below the attribution details in AUDIT and ATTRIBUTION view
- show: 
  - only if there are no unconfirmed changes
  - attribution has an `attribution origin`

Popup
- Title: "Compare to original signal"
- Background light blue as the input fields are white
- Two columns 
  - left: original signal
  - right: latest confirmed attribution
- auditing option on top
  - revert button on the right hand side reverts all user controlled chips to the state of the original signal
- the fields for each attribution / signal are listed below the auditing options
  - default: show the values of the fields as text in a non-editable text field 
  - exception: 
    - Third / First Party field (see below)
      - use same visualization as currently used in audit / attribution view
- If the value of two matching fields are identical
  - text color: black
  - do not show a revert button on the right side of the current attribution field (see below)
- If the value of two corresponding fields deviates
  - text color: 
    - of the original signal field is red (overwritten value)
    - of the current attribution field is green (new value)
  - show a `revert` button on the right side of the field of the current attribution
    - icon: some "revert" icon, e.g., a circulating arrow (see Mock below)
    - on-click: 
      - resets the value of the current attribution to the value of the original signal 
      - the button vanishes
      - both fields are shown as described above for identical fields (e.g. text color black)      
- seperator lines "Package Coordinates" and "Legal Information" as headlines for the corresponding sections
  - we use them here as the user is used to them from audit and attribution view
-Button row with three buttons in the bottom right (from left to right)
  - "Apply changes" button 
    - Text: `APPLY CHANGES`
    - disabled if no changes have been made
    - Onclick: 
      - closes popup
      - stores changes made to the temporary state
      - the user will need to actively save the changes in audit / attribution view
  - "Revert to original" button
    - Text: `REVERT TO ORIGINAL`
    - onclick:
      - only enabled if there are differences between the original signal and current attribution 
      - reverts all fields of the current attribution to the values of the original signal
[...]

PR #2557, opened 2024-02-06
feat: add revert functionality to diff popup

alexzurbonsen:
### Summary of changes

This diff adds revert functionality to the diff popup. Every field that is changed compared to the original value, can be reverted individually with a button. You can also restore the modified value again with a redo button, that replaces the revert/undo button once the change was reverted.

There is also a button to revert all changes. Finally, changes can be written into the temporary package info state by using the apply button.

We decided to not show the auditing options in the Diff popup since they are related to a manual attribution and not to a signal.

We also decided to always show license and copyright fields, also in the first party case. 

Story: https://github.com/opossum-tool/OpossumUI/issues/2488

Closes https://github.com/opossum-tool/OpossumUI/issues/2515

Adapted unit tests. 
Added e2e tests.

Manually play around with the new feature and test if it behaves as expected.


src/Frontend/Components/AttributionColumn/AttributionForm.util.tsx
```diff
@@ -0,0 +1,43 @@
+// SPDX-FileCopyrightText: Meta Platforms, Inc. and its affiliates
```
alexzurbonsen, line 1: Introduced this file to resolve the circular dependency chain I had.

Previously the types were in `AttritbutionForm.tsx` and `getComparableAttributes` was in `DiffPopup.util.tsx`.
That resulted in a circular dependency because AttributionForm was importing from DiffPopup.util via AuditingOptions.util and DiffPopup.util was importing from AttributionForm.

src/Frontend/Components/DiffPopup/DiffPopup.util.tsx
```diff
@@ -0,0 +1,247 @@
...
+          ? {
+              ...current,
+              copyright: undefined,
+              licenseName: undefined,
+              licenseText: undefined,
+            }
+          : current,
+      );
+    }
+  }, [isOpen, current, setBufferPackageInfo]);
```
alexzurbonsen, line 42: I kept the `useEffect` for now. The only other way I found to update the state when the popup is opened was forcing a re-mount via the key prop. If we want to do that we would probably want to do something like `key={someHashFunction(getComparableAttributes(packageInfo))}`.

src/Frontend/Components/AttributionColumn/LicenseSubPanel.tsx
```diff
@@ -154,7 +149,9 @@ export function LicenseSubPanel({
             showHighlight={showHighlight}
             onEdit={onEdit}
             endAdornment={
-              packageInfo.licenseText ? (
+              !!config?.licenseName?.endIcon ? (
+                config.licenseName.endIcon
+              ) : packageInfo.licenseText ? (
```
mstykow, line 154:
```suggestion
              config?.licenseName?.endIcon || packageInfo.licenseText ? (
```

src/Frontend/Components/AttributionColumn/LicenseSubPanel.tsx
```diff
@@ -194,6 +191,11 @@ export function LicenseSubPanel({
                 ),
               )
             }
+            endIcon={
+              !!config?.licenseText?.endIcon
+                ? config.licenseText.endIcon
+                : undefined
+            }
```
mstykow, line 198:
```suggestion
            endIcon={config?.licenseText?.endIcon}
```

src/Frontend/Components/AttributionColumn/PackageSubPanel.tsx
```diff
@@ -139,6 +141,11 @@ export function PackageSubPanel({
         onEdit={onEdit}
         color={config?.packageName?.color}
         focused={config?.packageName?.focused}
+        endAdornment={
+          !!config?.packageName?.endIcon
+            ? config.packageName.endIcon
+            : undefined
+        }
```
mstykow, line 148:
```suggestion
        endAdornment={config?.packageName?.endIcon}
```

src/Frontend/Components/AttributionColumn/PackageSubPanel.tsx
```diff
@@ -156,6 +163,11 @@ export function PackageSubPanel({
         onEdit={onEdit}
         color={config?.packageNamespace?.color}
         focused={config?.packageNamespace?.focused}
+        endAdornment={
+          !!config?.packageNamespace?.endIcon
+            ? config.packageNamespace.endIcon
+            : undefined
+        }
```
mstykow, line 170:
```suggestion
        endAdornment={config?.packageNamespace?.endIcon}
```

src/Frontend/Components/AttributionColumn/PackageSubPanel.tsx
```diff
@@ -172,6 +184,11 @@ export function PackageSubPanel({
         onEdit={onEdit}
         color={config?.packageVersion?.color}
         focused={config?.packageVersion?.focused}
+        endAdornment={
+          !!config?.packageVersion?.endIcon
+            ? config.packageVersion.endIcon
+            : undefined
+        }
```
mstykow, line 191:
```suggestion
        endAdornment={config?.packageVersion?.endIcon}
```

src/Frontend/Components/AttributionColumn/PackageSubPanel.tsx
```diff
@@ -189,6 +206,11 @@ export function PackageSubPanel({
         onEdit={onEdit}
         color={config?.packageType?.color}
         focused={config?.packageType?.focused}
+        endAdornment={
+          !!config?.packageType?.endIcon
+            ? config.packageType.endIcon
+            : undefined
+        }
```
mstykow, line 213:
```suggestion
        endAdornment={config?.packageType?.endIcon}
```

src/Frontend/Components/AttributionColumn/PackageSubPanel.tsx
```diff
@@ -267,45 +295,49 @@ export function PackageSubPanel({
...
-              onClick={() => openUrl(packageInfo.url)}
-              hidden={!packageInfo.url}
-              icon={
-                <OpenInNewIcon aria-label={'Url icon'} sx={clickableIcon} />
-              }
-            />
-          </>
+          !!config?.url?.endIcon ? (
+            config.url.endIcon
+          ) : (
```
mstykow, line 300:
```suggestion
          config?.url?.endIcon || (
```

src/Frontend/Components/DiffPopup/DiffPopup.tsx
```diff
@@ -20,31 +27,86 @@ interface DiffPopupProps {
...
+    buffer: PackageInfo,
+  ) {
+    if (
+      isEqual(
+        getComparableAttributes(buffer),
+        getComparableAttributes(original),
+      ) &&
+      original.wasPreferred
+    ) {
+      buffer.wasPreferred = true;
```
mstykow, line 58: this is a cardinal error in react: you are mutating state that's managed by `useState`. this means react has no idea a change occurred and thus cannot display the most recent state. the only reason we don't see errors is because immediately after this mutation you're unmounting the component that manages the state, but any refactoring to the code will introduce bugs and spooky-action-at-a-distance outside the react framework.

mutation of reactive state must be avoided at all cost. in this situation, you'd use the spread operator inside `setTemporaryDisplayPackageInfo`.

src/Frontend/Components/DiffPopup/DiffPopup.tsx
```diff
@@ -20,31 +27,86 @@ interface DiffPopupProps {
...
+          );
+          setOpen(false);
+        },
       }}
       centerRightButtonConfig={{
-        disabled: true,
+        disabled: isEqual(
+          getComparableAttributes(bufferPackageInfo),
+          getComparableAttributes(original),
+        ),
```
mstykow, line 91: this is the same check as in line 73-76. i suggest we extract into a named const.

src/Frontend/Components/DiffPopup/DiffPopup.tsx
```diff
@@ -20,31 +27,86 @@ interface DiffPopupProps {
...
+          getComparableAttributes(current),
+        ),
         buttonText: text.buttons.diffPopup.applyChanges,
+        onClick: () => {
+          checkIfWasPreferredAndLicenseInfoResetNecessaryAndApply(
+            original,
+            current,
+            bufferPackageInfo,
+          );
+          setOpen(false);
```
mstykow, line 84: not a fan of these long names that already don't make sense anymore as soon as i refactor the code even just slightly.
let's rename `checkIfWasPreferredAndLicenseInfoResetNecessaryAndApply` to `handleApplyChanges` and include `setOpen(false)` in it.

src/Frontend/Components/DiffPopup/DiffPopup.tsx
```diff
@@ -20,31 +27,86 @@ interface DiffPopupProps {
...
       centerRightButtonConfig={{
-        disabled: true,
+        disabled: isEqual(
+          getComparableAttributes(bufferPackageInfo),
+          getComparableAttributes(original),
+        ),
         buttonText: text.buttons.diffPopup.revertAll,
+        onClick: () => {
+          setBufferPackageInfo({
+            ...FORM_ATTRIBUTES.reduce(
```
mstykow, line 95: there is no need for the spread operator here, i.e., we don't need to make a copy of the return value of `reduce`.

src/Frontend/Components/DiffPopup/DiffPopup.tsx
```diff
@@ -20,31 +27,86 @@ interface DiffPopupProps {
...
+              (buffer, attribute) => {
+                // @ts-expect-error original[attribute] will always have a value that is allowed for buffer[attribute] but typescript does not recognize that
+                buffer[attribute] = original[attribute];
+                return buffer;
+              },
+              { ...bufferPackageInfo },
+            ),
+          });
+        },
```
mstykow, line 104: this will do the job without type errors or accidentally mutating reactive state:
```suggestion
        onClick: () => {
          setBufferPackageInfo({
            ...bufferPackageInfo,
            ...FORM_ATTRIBUTES.reduce(
              (acc, attribute) => ({
                ...acc,
                [attribute]: original[attribute],
              }),
              {},
            ),
          });
        },
```

src/Frontend/Components/DiffPopup/DiffPopup.util.tsx
```diff
@@ -0,0 +1,246 @@
...
+    margin: 0,
+  },
+};
+
+function getEndIcon(
+  variant: 'undo' | 'redo',
+  tooltip: string,
+  onClick: () => void,
+  size: 'small' | 'big' = 'small',
+) {
```
mstykow, line 227: please use object notation for the arguments. reading `getEndIcon('undo', 'foo', () => {}, 'small')` is not very helpful when it comes to function signatures and forces me to look up the order of the arguments, or worse, make a mistake confusing them. same as kwargs in python...

src/Frontend/Components/DiffPopup/DiffPopup.util.tsx
```diff
@@ -0,0 +1,246 @@
...
+      icon={
+        variant === 'undo' ? (
+          <UndoIcon
+            sx={size === 'small' ? clickableIcon : classes.bigRevertIcon}
+          />
+        ) : (
+          <RedoIcon
+            sx={size === 'small' ? clickableIcon : classes.bigRevertIcon}
+          />
+        )
```
mstykow, line 242: To reduce the number of lines of code unnecessarily repeated, I suggest initially defining:
```
const Icon = variant === 'undo' ? UndoIcon : RedoIcon
```

src/Frontend/Components/DiffPopup/DiffPopup.util.tsx
```diff
@@ -0,0 +1,246 @@
...
+const classes = {
+  bigRevertIcon: {
+    ...clickableIcon,
+    width: 24,
+    height: 24,
+    margin: 0,
+  },
+};
+
+function getEndIcon(
```
mstykow, line 222: this is actually a full-blooded react component. so why not extract it into a separate sub-component (file) and just call it `DiffEndIcon` or something like that.

src/Frontend/Components/DiffPopup/DiffPopup.util.tsx
mstykow: if the trick with `key` works, it would be preferable i'd say. can you give it a try? maybe using the `packageInfo.id` could work as key?

src/Frontend/Components/DiffPopup/DiffPopup.util.tsx
```diff
@@ -0,0 +1,246 @@
...
+
+  const [originalFormConfig, bufferFormConfig] = useMemo(
+    () =>
+      FORM_ATTRIBUTES.reduce<[AttributionFormConfig, AttributionFormConfig]>(
+        ([originalFormConfig, bufferFormConfig], attribute) => {
+          switch (attribute) {
+            case 'copyright':
+            case 'licenseName':
+            case 'licenseText': {
+              const isThirdPartyUnchanged =
```
mstykow, line 52: always try to use as few NOTs as possible to reduce complexity. so can we turn the logic around and call this "isThirdPartyChanged"?

src/Frontend/Components/DiffPopup/DiffPopup.util.tsx
mstykow: EDIT: once you convert this to a real FC, you'll be forced to give names to your args anyways.

src/Frontend/Components/DiffPopup/DiffPopup.util.tsx
```diff
@@ -0,0 +1,246 @@
...
+                          'redo',
+                          text.diffPopup.undoRevert[attribute],
+                          () => {
+                            setBufferPackageInfo({
+                              ...bufferPackageInfo,
+                              [attribute]: current[attribute],
+                            });
+                          },
+                        )
+                      : undefined,
```
mstykow, line 98: nested ternaries are best avoided as they make the code more difficult to understand. i think a chain of if-statements would be easier here. alternatively, you can probably move your ternaries inside the first two props, as the third prop is the same, and handle the `undefined` case upfront.

mstykow, line 98: nested ternaries are best avoided as they make the code more difficult to understand. i think a chain of if-statements would be easier here. alternatively, you can probably move your ternaries inside the first two props, as the third prop is the same, and handle the `undefined` case upfront.

src/Frontend/Components/DiffPopup/__test__/DiffPopup.util.test.tsx
```diff
@@ -25,6 +27,18 @@ describe('useAttributionFormConfigs', () => {
...
+    copyright: { color: undefined, focused: false, endIcon: undefined },
+    firstParty: { color: undefined, endIcon: undefined },
+    licenseName: { color: undefined, focused: false, endIcon: undefined },
+    licenseText: { color: undefined, focused: false, endIcon: undefined },
+    packageName: { color: undefined, focused: false, endIcon: undefined },
+    packageNamespace: { color: undefined, focused: false, endIcon: undefined },
+    packageVersion: { color: undefined, focused: false, endIcon: undefined },
+    url: { color: undefined, focused: false, endIcon: undefined },
+    packageType: { color: undefined, focused: false, endIcon: undefined },
+  };
```
mstykow, line 41: please generate these using `reduce`. same for the object in line 18.

src/shared/text.ts
```diff
@@ -140,5 +140,31 @@ export const text = {
...
+      firstParty:
+        'Recover modified attribution type, license and copyright information',
+      licenseName: 'Recover modified license name',
+      licenseText: 'Recover modified license text',
+      packageName: 'Recover modified package name',
+      packageNamespace: 'Recover modified package namespace',
+      packageType: 'Recover modified package type',
+      packageVersion: 'Recover modified package version',
+      url: 'Recover modified URL',
+    },
```
mstykow, line 168: i'm not convinced we need any of these. isn't it clear without a tooltip what the revert/undo buttons do?

mstykow (changes requested):

src/Frontend/Components/AttributionColumn/AttributionForm.util.tsx
alexzurbonsen: The problem here is `FORM_ATTRIBUTES` I think. It is a `const`. To resolve the loop I put `getComparableAttributes` and `FORM_ATTRIBUTES` next to each other. And since the types are related, I put them all in the same place.
Just moving `getComparableAttributes` doesn't help unfortunately.

src/Frontend/Components/AttributionColumn/ButtonRow.tsx
```diff
@@ -21,14 +22,20 @@ import {
...
   packageInfo: PackageInfo;
+
   onSaveButtonClick?(): void;
+
   onSaveGloballyButtonClick?(): void;
+
   onDeleteButtonClick?(): void;
+
   onDeleteGloballyButtonClick?(): void;
+
```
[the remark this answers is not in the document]
alexzurbonsen: Missed that, not sure how they got here...

src/Frontend/Components/DiffPopup/DiffPopup.tsx
alexzurbonsen: Okay, sure. I just thought that it might be okay since this is the last thing this state does for us. But of course that might change...

src/Frontend/Components/DiffPopup/DiffPopup.tsx
alexzurbonsen: Just realized: didn't do this because one is comparing with `current`, the other one with `original`. Could introduce a function for that, but I don't think that is worth it?

src/Frontend/Components/DiffPopup/DiffPopup.tsx
alexzurbonsen: Nice.

src/Frontend/Components/DiffPopup/DiffPopup.util.tsx
alexzurbonsen: What did you have in mind here? I introduced the `Unchanged` since it removed some NOTs and it seemed to me that made it easier to understand.
We have an expression like `!isThirdPartyChanged && isAttributeValueChanged` were we want to check if one side changed and the other one didn't. I also considered something like `bothAreThirdParty`? But that breaks the "something changed" naming schema.

src/Frontend/Components/DiffPopup/DiffPopup.util.tsx
alexzurbonsen: After pulling creating the `DiffEndIcon` component I created a new `getEndIcon` function which does the choosing of the correct icon. What do you think about that?

src/shared/text.ts
alexzurbonsen: Fair point. I am not sure either. I guess it might be reassuring to some users, but I don't think it is terribly necessary.

src/Frontend/Components/DiffPopup/DiffPopup.util.tsx
alexzurbonsen: I implemented a variant of this. `packageInfo.id` is not enough since it only changes when we change the attribution that is shown, but not if we change something in the attribution form.

src/Frontend/Components/DiffPopup/DiffPopup.tsx
mstykow: yes, nevermind then

src/Frontend/Components/DiffPopup/DiffPopup.util.tsx
mstykow: ok, nevermind then

src/shared/text.ts
mstykow: i feel like excessive use of tooltips causes noise and distraction

src/Frontend/Components/DiffPopup/DiffPopup.util.tsx
alexzurbonsen: It indeed works with `key={isDiffPopupOpen.toString()}`.

mstykow (approved):

Issue #2558, opened 2024-02-08
fix: alignment of scroll bar and end adornment TextField


alexzurbonsen:
**Describe the bug**
Currently, if there a scroll bar and an icon as end adornment are present in a text field at the same time, the scroll bar shifts to the left, into the text field, making room for the end adornment. This is not what we want. The scroll bar should be aligned with the right edge of the text field and the icon should be shifted inward.

This can be solved for example with styles like this
```
endAdornment: {
    position: 'absolute',
    right: 0,
    paddingRight: '14px',
  }, 
```
but then you need to handle the case where the icons start overlapping with the text cleanly.

This problem occurred in the context of the Diff Popup #2488.

PR #2559, opened 2024-02-09
fix: scroll bar alignment in text fields

alexzurbonsen:
### Summary of changes

Render end icon left of the scroll bar and keep scroll bar aligned with right edge in multiline text fields with a scroll bar.

Closes [#2558](https://github.com/opossum-tool/OpossumUI/issues/2558)

Open OpossumUI, put longer strings into the URL and fill the relevant fields to get a PURL. Shrink the window size such that end icons start overlapping with the text. The text overflowing text should be replaced by ellipsis and there should be no overlap between text and icons.
Do the same in the Diff Popup.


alexzurbonsen:
Closing in favour of #2560

PR #2560, opened 2024-02-09
fix: shift end icons in text fiels inwards instead of scroll bar

alexzurbonsen:
### Summary of changes

Render end icon left of the scroll bar and keep scroll bar aligned with right edge in multiline text fields with a scroll bar.

Closes [#2558](https://github.com/opossum-tool/OpossumUI/issues/2558)

Open OpossumUI, put longer strings into the URL and fill the relevant fields to get a PURL. Shrink the window size such that end icons start overlapping with the text. The text overflowing text should be replaced by ellipsis and there should be no overlap between text and icons.
Do the same in the Diff Popup.

### DIFF
--- a/src/Frontend/Components/InputElements/shared.ts
+++ b/src/Frontend/Components/InputElements/shared.ts
@@ -49,6 +49,12 @@ export const inputElementClasses = {
       padding: '1px 3px',
     },
   },
+  endAdornmentRoot: {
+    position: 'absolute',
+    right: 0,
+    marginRight: '14px',
+    height: 0,
+  },
 } satisfies SxProps;
 
 export interface InputElementProps {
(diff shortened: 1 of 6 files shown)


src/Frontend/Components/InputElements/shared.ts
```diff
@@ -48,6 +48,12 @@ export const inputElementClasses = {
       padding: '1px 3px',
     },
   },
+  endAdornmentRoot: {
+    position: 'absolute',
+    right: 0,
+    marginRight: '14px',
+    height: 0,
```
alexzurbonsen, line 55: This is the fix for the weird white horizontal line, that was crossing the scrollbar. It disappears when setting the height of the root to 0. Before it was at 0.01em or something like that.

mstykow (approved):

PR #2571, opened 2024-02-14
feat: show labels for original and current attribution in diff popup

mstykow:
image


alexzurbonsen (approved):
Thanks for taking care of this. Nice change.

PR #2582, opened 2024-02-23
fix: introduce proper error boundary

### LINKED ISSUE #2564: Create a proper error boundary
**Describe the bug**
When the app crashes unexpectedly, I as a user want to see an in-app error message instead of having to deal with system dialogs that permanently close the app. In addition, after the error was caught, I want to be able to re-open the file and try again.

**To Reproduce**

1. Trigger an unexpected error
2. Notice that there are system dialogs which need to be confirmed until the app terminates

**Additional context**
There is already an error boundary component in the react app but it doesn't seem to be doing anything. The component name is `ErrorBoundary`.

Use this library instead of a custom component: https://github.com/bvaughn/react-error-boundary
Come up with an appropriate generic user-facing error message to display when an unexpected error occurs.

The error boundary should be able to catch both frontend and backend errors.

mstykow:
### Summary of changes

- add methods to quit or relaunch the app
- create error boundary that displays error and offers options to relaunch or quit

image

closes #2564

You must produce an unexpected error. For example, you could start the app locally and change the code somewhere to break it, e.g., put this into `ReportView` to trigger an unexpected error when the user switches to report view: `console.log(null.map())`


alexzurbonsen:
Nice. Lgtm (Before the most recent rebase at least...)

alexzurbonsen (approved):

PR #2583, opened 2024-02-23
feat: add keyboard navigation to attributions and signals lists

### LINKED ISSUE #971: Enable keyboard navigation in package lists

mstykow:
### Summary of changes

- merge PackageCard and ListCard components
- track keyboard selected item in list and grouped-list
- add focus styling to package card

closes #971

Check keyboard navigation after clicking on a package card in the attributions or signals lists. Pay special attention also to switching tabs: the previously focused package card should not carry over to another tab.


src/Frontend/util/use-virtuoso-refs.ts
```diff
@@ -0,0 +1,67 @@
...
+export function useVirtuosoRefs<T extends VirtuosoHandle, D = unknown>({
+  data,
+  selectedIndex,
+}: {
+  data: ReadonlyArray<D> | null | undefined;
+  selectedIndex: number | undefined;
+}) {
+  const ref = useRef<T>(null);
+  const listRef = useRef<Window | HTMLElement>();
+  const [focusedIndex, setFocusedIndex] = useState(selectedIndex ?? -1);
```
MarkusObendrauf, line 17: I would prefer not to use -1 here

src/e2e-tests/__tests__/selecting-attributions.test.ts
```diff
@@ -0,0 +1,102 @@
...
+  await window.keyboard.press('Enter');
+  await attributionDetails.attributionForm.assert.matchesPackageInfo(
+    packageInfo3,
+  );
+
+  await window.keyboard.press('ArrowUp');
+  await window.keyboard.press('Space');
+  await attributionDetails.attributionForm.assert.matchesPackageInfo(
+    packageInfo2,
+  );
```
MarkusObendrauf, line 101: this test is really really satisfying

MarkusObendrauf (approved):
Cool. Sometimes using arrow keys is a bit slow (still less that 0.5s but its noticeable). For example, if you're on the second-last item and press down, it takes a while for the highlighted item to move.

mstykow:
> Hey, finally figured out that a bug I'm encountering in the follow-up already exists in this version.
> 
> To reproduce:
> 
> 1. Select a resource without attributions
> 2. Focus on the attributions panel, eg. by clicking on the scrollbar
> 3. Scroll down using the down arrow
> 4. Select a different resource without attributions
> 5. Focus on the attributions panel, eg. by clicking on the scrollbar
> 6. Scroll down using the down arrow
> 
> Notice that the focused index was preserved from the previous list.

I think I found a pretty simple solution to this issue that should also help you with the focus-stealing problems you faced in your follow-up task.

src/Frontend/util/use-virtuoso-refs.ts
```diff
@@ -0,0 +1,73 @@
...
+          ref.current?.scrollIntoView({
+            index,
+            behavior: 'auto',
+            done: () => setFocusedIndex(index),
+          });
+          event.preventDefault();
+        }
+      }
+    },
+    [data?.length, focusedIndex],
```
MarkusObendrauf, line 52: what's the `data?.length` for? is it for the file tree?

MarkusObendrauf (changes requested):
It's still a little buggy:
1. Click a resource with no attached attributions
2. Click on an attribution
3. Click on a resource with an attached attribution
4. Click on a resource with no attached attributions
Notice that the focus index from step 2 was preserved.

mstykow:
> It's still a little buggy:
> 
> 1. Click on a resource with no attached attributions
> 2. Click on an attribution
> 3. Click on a resource with an attached attribution
> 4. Click on a resource with no attached attributions
>    Notice that the focus index from step 2 was preserved.

Good catch. Please try again now.

src/Frontend/util/use-virtuoso-refs.ts
mstykow: since this is a generic hook, it has nothing to do with a specific component such as a file tree.

in this implementation of `keyDownCallback` you can see that we need to know the number of items the list is dealing with because we don't want to set the focused index to something lower than the first or higher than the last possible index.

mstykow:
> Member

The slowness resulted from only changing the focused index once the scrolling was complete. But there is actually no need to wait. So I changed the behavior to immediately set the new index regardless of the scroll state.

mstykow:
> Cool. Sometimes using arrow keys is a bit slow (still less that 0.5s but its noticeable). For example, if you're on the second-last item and press down, it takes a while for the highlighted item to move.

The slowness resulted from only changing the focused index once the scrolling was complete. But there is actually no need to wait. So I changed the behavior to immediately set the new index regardless of the scroll state.

MarkusObendrauf (approved):
Seems to work, I'm just praying that I never have to open up `use-virtuoso-refs.ts` again.

Issue #2588, opened 2024-02-29
Automatically map license names to SPDX license identifiers


FlorianSchepers:
**Is your feature request related to a problem? Please describe.**
When selecting a common license for an attribution from the dropdown menue the full name will be shown in the "license name" field. However, we want that the SPDX license identifier is shown instead.

**Describe the solution you'd like**
Automatically map the license names of the common licenses to the corresponding SPDX license identifiers when selecting them from the dropdown menue.

**Additional context**
This is a feature request for OpossumUI 2.0

mstykow:
Won't do for now.

PR #2589, opened 2024-03-04
feat: use short name of common licenses as license name if selected

meretp:
### Summary of changes

If the user selects a common license as license of the attribution, the short name (SPDX identifier) is stored as license name of the attribution.

fixes issue #2588

See the new e2e test: Update an attribution by selecting one of the common licenses and note that the short name (the name in brackets in the list of common licenses) is set as license name for the attribution.


mstykow:
After discussing with stakeholder and PO, we'll keep things as they are for now and instead communicate to our users that it's not necessary to prefer short names over long.

PR #2590, opened 2024-03-04
feat: add filter for attributions that were preferred but have been modified

alexzurbonsen:


src/Frontend/state/helpers/save-action-helpers.ts
```diff
@@ -86,13 +87,23 @@ export function updateManualAttribution(
...
+    packageInfo.wasPreferred = isModifiedPreferred ? undefined : true;
+  }
+
   return {
     ...manualData,
     attributions: {
       ...manualData.attributions,
       [attributionIdToUpdate]: {
         ...getStrippedPackageInfo(packageInfo),
+        modifiedPreferred: isModifiedPreferred ? true : undefined,
```
alexzurbonsen, line 106: This is a bit of a tricky choice: 
I decided to remove `modifiedPreferred` when stripping `packageInfo` because of the replace and link case in `save-package-info`. My reasoning: we would not recognize a `matchingAttribution` if it contains the attribute `modifiedPreferred` already but the attribution we are saving will only get it during the update. So we would never replace or link such attributions otherwise.

The problem with this choice: it breaks the intuitive connection that ephemeral package props are not saved.

The alternative would be to change the matching logic.

src/shared/__tests__/get-is-modified-preferred.test.ts
```diff
@@ -0,0 +1,69 @@
+// SPDX-FileCopyrightText: Meta Platforms, Inc. and its affiliates
```
alexzurbonsen, line 1: This is not executed currently. Probably change jest config (?)

PR #2591, opened 2024-03-05
feat: add keyboard navigation for resource tree

### LINKED ISSUE #2572: Enable keyboard navigation in resource browser
Currently, navigating the resource tree requires the users to click on each resource to expand/collapse.
For the user that implies the need to click a lot.

Navigation:
- up-down navigation with arrows
- use horizontal navigation to expand/collapse leaves
- experiment if it's better to select the item with space or if an item should be immediately selected after the navigation event

As a user navigating in the file tree in resource view, I want to be able to use keyboard navigation.

is blocked by #971

mstykow:
### Summary of changes

- keep track of focused ID instead of focused index in Virtuoso
- add focus state for tree nodes and define keyboard actions

closes #2572

Navigate the resource browser up and down and left and right and check that focus behaves as expected.


mstykow (commented):
Co-authored this PR with @MarkusObendrauf and I approve.

PR #2592, opened 2024-03-06
feat: update readme

mstykow:
### Summary of changes

Update readme.


MarkusObendrauf (approved):
Cool. Why did we remove the data? Do we have these test files somewhere else? It looks useful to keep it around for when we need to update our screenshots.

mstykow:
> Cool. Why did we remove the data? Do we have these test files somewhere else? It looks useful to keep it around for when we need to update our screenshots.

I would expect that the data in `example-files` in enough for that purpose and don't think we need to store sample data in multiple locations. What do you think?

PR #2593, opened 2024-03-06
feat: add filter for modified previously pereferred attributions

alexzurbonsen:


mstykow (changes requested):
The approach you're taking here seems quite complicated to me. I've tried an alternative in #2596 which entirely avoids the need to touch any Redux related logic. Please let me know what you think. If you like it, we're just missing some (E2E) tests there.

alexzurbonsen:
Closing in favor of https://github.com/opossum-tool/OpossumUI/pull/2596

PR #2601, opened 2024-03-11
feat: filter for modified previously preferred attributions V2

### LINKED ISSUE #2508: Add option to filter for modified previously preferred attributions

mstykow:
### Summary of changes

- introduce new ephemeral attributes to keep track of ID, source, and previously-preferred of original attribution
- introduce filter and package card icon for modified-previously-preferred attributions

closes #2508

Check that filtering for modified previously preferred attributions works as expected as you add/remove/edit previously-preferred attributions.
Also check that original sources are still correctly displayed.


src/Frontend/state/helpers/save-action-helpers.ts
```diff
@@ -641,3 +643,23 @@ function _addPathAndParentsToResourcesWithAttributedChildren(
     );
   });
 }
+
+export function getWasPreferred({
```
mstykow, line 647: 1. test `getWasPreferred`
2. add test to `get-filtered-attributions`
3. extend e2e-test which removes previously preferred label by asserting that modified-previously-preferred label shows up instead
4. add tests for `parseInputData`

alexzurbonsen (approved):

PR #2603, opened 2024-03-13
fix: license text modified overlaps license name

### LINKED ISSUE #2529: "License text modified" overlaps with license name on smaller screens

mstykow:
### Summary of changes

- always show license text input
- display default license text as placeholder when a default license name is selected

image

closes #2529

Check out the license text field when
- creating a new attribution
- modifying an attribution with/without license text
- looking at a signal with/without license


alexzurbonsen:
I like the approach.

I am just wondering if it might be more convenient to who an option to expand the license text instead of showing it whenever it is present? I have the feeling that the small text field might be too small for longer texts.

mstykow:
> I like the approach.
> 
> I am just wondering if it might be more convenient to who an option to expand the license text instead of showing it whenever it is present? I have the feeling that the small text field might be too small for longer texts.

I guess the same argument could be made for copyright and comment?

src/shared/text.ts
```diff
@@ -25,7 +25,10 @@ export const text = {
     invalidPurl: 'INVALID PURL',
     legalInformation: 'Legal Information',
     licenseName: 'License Name',
-    licenseTextModified: '(License text modified)',
+    licenseText: 'License Text',
+    useCustomLicenseText: 'Use custom license text',
+    useCustomLicenseTextHint:
+      'Specify a custom license text if your license does not match one of the common licenses from the list above.',
```
alexzurbonsen, line 31: I am not sure I entirely understand the tooltip. Is the license text input filled automatically if I select a common license? I thought not?

alexzurbonsen: And I agree, as we discussed before, I might make sense to drop the expandable approach entirely and just always show the field.

alexzurbonsen (approved):

src/shared/text.ts
mstykow: tried an approach where we always show the license text input now.

alexzurbonsen (approved):
I think that is a good solution. We might need to communicate that the palceholder is something that shows the real license text, that is associated with the attribution, instead of some "real" placeholder that won't affect the output file. But I think that is okay.

Issue #2099, opened 2023-10-05
Improve error message in case of "outdated" OpossumUI versions


benedikt-richter:
The schema of Opossum files changes over time. While we guarantee that old Opossum files can be opened with all future versions of the app, it is not true that old versions of the app can open opossum files that follow the most recent schema.
E.g. there can be the case that a required (by the schema) field is missing such that the schema validation fails on opening this file.

Currently, we expose the technical error to the user. Instead we should add the minimun required version of the app to each opossum file and check that the app that is used for opening satisfies the requirement. If not, an easy to understand  error should be shown to the user.

PR #2615, opened 2024-04-18
feat: improve error message in case of parsing issues with the input file

meretp:
### Summary of changes

If the app is not able to parse the opossum file it might be that the version of the app used is outdated. This PR adds a hint to the error message to use the newest version of the app if there is a problem parsing the input file. 

fixes issue #2099

See CI, I updated the correpsonding unit tests.


alexzurbonsen:
I am wondering: the ticket proposes a more involved solution than the one contained in the diff. Did we decide to go the alternative route of this PR?

meretp:
I talked with Maxim about this and we agreed that the proposed solution is too involved and that we can keep it simple by just adding this note.

PR #2617, opened 2024-04-22
feat: ignore hidden signals in project statistics

### LINKED ISSUE #2227: Project statistics popup should ignore hidden signals
User can hide signals by right-click -> hide. This should be reflected at the Project statistics popup.

Do not take into account hidden signals at all the data at Project statistics popup -> in this case numbers should decrease.

Maybe introduce one more column "hidden signals" or something like that.

**Additional context**
image

meretp:
### Summary of changes

If a user hides a signal this signal will no longer be taken into account for the project statistics popup. I also changed the displayed icon in resource tree.  If all critical signals linked to a resource are hidden the resource no longer has the criticality icon next to it. If in general all signals linked to the resource are hidden, the flag indicating  that the resource contains signals is no longer shown.

fixes #2227

Play around with hiding critical signals and look at the project statistics and the respective icons in the resource tree.

### DIFF
--- a/src/e2e-tests/page-objects/ProjectStatisticsPopup.ts
+++ b/src/e2e-tests/page-objects/ProjectStatisticsPopup.ts
@@ -4,15 +4,28 @@
 // SPDX-License-Identifier: Apache-2.0
 import { expect, type Locator, type Page } from '@playwright/test';
 
+import { text } from '../../shared/text';
+
 export class ProjectStatisticsPopup {
   private readonly node: Locator;
   readonly title: Locator;
   readonly closeButton: Locator;
+  readonly totalCriticalLicensesCount: Locator;
 
   constructor(window: Page) {
     this.node = window.getByLabel('project statistics');
-    this.title = this.node.getByText('Project Statistics');
+    this.title = this.node.getByRole('heading').getByText('Project Statistics');
     this.closeButton = this.node.getByRole('button', { name: 'Close' });
+    const signalsCount = window.getByText(
+      text.projectStatisticsPopup.criticalLicensesSignalCountColumnName,
+    );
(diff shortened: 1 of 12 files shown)


meretp:
As I had to fix two tests that expected the criticality icon in the resource tree for the resolved signals I am no longer sure about that change. But for me it feels inconsistent to show this icon if there actually is no critical signal because the signal was resolved / hidden before. What do you think?

src/e2e-tests/page-objects/ProjectStatisticsPopup.ts
```diff
@@ -8,11 +8,19 @@ export class ProjectStatisticsPopup {
...
-    this.title = this.node.getByText('Project Statistics');
+    this.title = this.node.getByRole('heading').getByText('Project Statistics');
     this.closeButton = this.node.getByRole('button', { name: 'Close' });
+    this.totalCriticalLicensesCount = this.node
+      .getByRole('table')
+      .nth(1)
+      .getByRole('row')
+      .last()
+      .getByRole('cell')
+      .last();
```
meretp, line 23: This locator feels a bit weird and not really reliable, but it's the best I could come up with. Any suggestions on how to improve here are very welcome.

meretp:
The test succeeded now and I've also run it once more locally where it also passed, I assume this was just some random failure, nothing to worry about.

meretp:
I had another look at the locator for the total signals count on the project statistics popup and decided to rewrite it a bit to make the location more stable.

davidmundelius:
> I had another look at the locator for the total signals count on the project statistics popup and decided to rewrite it a bit to make the location more stable.

Nice.

davidmundelius (approved):
Looks good to me now.

Issue #2069, opened 2023-09-28
Make click action on progress bar more visible


benedikt-richter:
Currently, there is the feature that clicking the progress bar navigates through resources in the resource tree. This feature is hard to discover for users.
Therefore, we want to introduce an icon that on click triggers the navigation. This makes this feature more explicit for the user.

mstykow:
Closed in favor of #2622.

PR #2618, opened 2024-04-22
feat: add button to jump through unresolved files

alexzurbonsen:
### Summary of changes

This diff adds a button next to the progress bar, that performs the same action as the progress bar if it is clicked: jumping to the next resource with unresolved signals or critical signals, depending on the setting of the progress bar. 

Closes [#2069](https://github.com/opossum-tool/OpossumUI/issues/2069).

Unit tests
e2e test
Manually


src/Frontend/Components/ProgressBarWithButtons/ProgressBarWithButtons.tsx
```diff
@@ -0,0 +1,108 @@
...
+    progressBarWithButtonsData.resources.withHighlyCriticalExternalAttributions.concat(
+      progressBarWithButtonsData.resources
+        .withMediumCriticalExternalAttributions,
+    );
+  const onCriticalityBarClick = useOnProgressBarClick(
+    resourcesWithCriticalExternalAttributions.length > 0
+      ? resourcesWithCriticalExternalAttributions
+      : progressBarWithButtonsData.resources
+          .withNonInheritedExternalAttributionOnly,
+  );
```
alexzurbonsen, line 68: Initially, I wanted to move these hooks into `TopBar` instead of creating a new component. But they only work if `progressBarWithButtonsData` is non-null. So it seemed more sensible to me to put all this into its own component and only render it if `progressBarWithButtonsData` is non-null.

alexzurbonsen:
Closing in favour of stack #2620

PR #2619, opened 2024-04-22
chore: refactor progress bar component and data to prepare additional jump button

alexzurbonsen:
### Summary of changes

Refactoring ProgressBar to prepare adding a button next to it with the same functionality as the progress bar. 

#2069

Unit tests.
Manually, everything should be the same.

### DIFF
--- a/src/Frontend/Components/ProgressBarWithButtons/ProgressBarWithButtons.tsx
+++ b/src/Frontend/Components/ProgressBarWithButtons/ProgressBarWithButtons.tsx
@@ -1,0 +1,81 @@
+// SPDX-FileCopyrightText: Meta Platforms, Inc. and its affiliates
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+import { SxProps } from '@mui/material';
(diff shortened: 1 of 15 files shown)


src/Frontend/Components/ProgressBarWithButtons/ProgressBarWithButtons.tsx
```diff
@@ -0,0 +1,81 @@
...
+    progressBarWithButtonsData.resources.withHighlyCriticalExternalAttributions.concat(
+      progressBarWithButtonsData.resources
+        .withMediumCriticalExternalAttributions,
+    );
+  const onCriticalityBarClick = useOnProgressBarClick(
+    resourcesWithCriticalExternalAttributions.length > 0
+      ? resourcesWithCriticalExternalAttributions
+      : progressBarWithButtonsData.resources
+          .withNonInheritedExternalAttributionOnly,
+  );
```
alexzurbonsen, line 53: Initially, I wanted to move these hooks into `TopBar` instead of creating a new component. But they only work if `progressBarWithButtonsData` is non-null. So it seemed more sensible to me to put all this into its own component and only render it if `progressBarWithButtonsData` is non-null.

alexzurbonsen:
Closing in favour of a new approach.

PR #2620, opened 2024-04-22
feat: add button to jump through unresolved resources

### LINKED ISSUE #2069: Make click action on progress bar more visible

alexzurbonsen:
### Summary of changes

This diff adds a button next to the progress bar, that performs the same action as the progress bar if it is clicked: jumping to the next resource with unresolved signals or critical signals, depending on the setting of the progress bar.

Closes #2069.

Unit tests.
Added e2e test.
Manually.


alexzurbonsen:
Closing in favour of a new approach.

PR #2621, opened 2024-04-23
chore: dynamically enable menu options

### LINKED ISSUE #2411: Top bar menu enabled even when no project is opened
**Describe the bug**
Right after opening of the app (before opening any project) the top bar menu is enabled and user can choose any option. Some of them are fine (qa mode), some of them do nothing (Project metadata or project statistics) and some of them result in error (file->export)

**To Reproduce**
Start opossumUI and try the menu items.

**Expected behavior**
Some options should be disabled.

**Screenshots**
image

image

alexzurbonsen:
### Summary of changes

This PR deactivates some menu options initially and activates them once a file is opened.

Ideally, I wanted to disable the entire `Edit` menu, but that does not come with nice UX. Instead I disabled individual options. The ones that come with a role, like `cut` and `copy` cannot be disabled in the same way. I left them as is.

I would have also liked to disable the `Export` submenu in the `File` menu. But if I do that I can't enable it in the same way as individual items. As an easy solution I just disabled the submenu items instead.

grafik

grafik

Closes #2411 

Manually check that options are disabled and will be enabled once a file is opened.

### DIFF
--- a/src/ElectronBackend/main/menu.ts
+++ b/src/ElectronBackend/main/menu.ts
@@ -25,6 +25,61 @@ import {
 } from './notice-document-helpers';
 import { UserSettings } from './user-settings';
 
+const INITIALLY_DISABLED_MENU_ITEMS = [
+  'save',
+  'projectMetadata',
+  'projectStatistics',
+  'followUp',
+  'compactComponentList',
+  'detailedComponentList',
+  'spdxYAML',
+  'spdxJSON',
+  'selectAll',
+  'searchAttributions',
+  'searchSignals',
+  'searchResourcesAll',
+  'searchResourceLinked',
+] as const;
+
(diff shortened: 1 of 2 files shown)


src/ElectronBackend/main/menu.ts
```diff
@@ -176,6 +243,8 @@ export async function createMenu(mainWindow: BrowserWindow): Promise<Menu> {
               );
             }
           },
+          id: MENU_ITEMS.projectStatistics.id,
+          enabled: false,
         },
         {
           icon: getIconBasedOnTheme(
```
alexzurbonsen, line 250: I was not sure about this option: 'Set Path to Sources'. Disabled it as well?

leslielazzarino (approved):

PR #2636, opened 2024-05-17
feat: collapse license text by default

mstykow:
### Summary of changes

- prevent license text from taking up vertical space by introducing toggle button to show/hide it
- rename "package" to "component" for OSS naming consistency
- make sure icon button keyboard-focus style is separate from click-focus style

image

User feedback on smaller screens revealed that we are unnecessarily taking up vertical space which forces frequent scrolling to see comments.

Try the new toggle button with and without changed license texts.


src/Frontend/Components/AttributionForm/LicenseSubPanel/LicenseSubPanel.tsx
```diff
@@ -71,47 +74,69 @@ export function LicenseSubPanel({
...
+            size={'small'}
+            aria-label="license-text-toggle-button"
+          >
+            <Badge
+              anchorOrigin={{ vertical: 'bottom', horizontal: 'right' }}
+              color={'info'}
+              variant={'dot'}
+              invisible={!packageInfo.licenseText}
+            >
+              <NotesIcon />
```
alexzurbonsen, line 104: Subjective: I think it could also be nice to just give the color to the Icon, instead of adding a badge. But go with what you think is best.

alexzurbonsen (approved):

src/Frontend/Components/AttributionForm/LicenseSubPanel/LicenseSubPanel.tsx
mstykow: i tried color first but that's not how the MUI toggle button is designed: it will only display the color once it's active

Issue #2622, opened 2024-04-23
Improve progress bar


mstykow:
**Is your feature request related to a problem? Please describe.**
The progress bar has a few areas that could be improved:
- the rightmost color in the stacked bar looks like empty space rather than a color
- hovering over the progress bar, there is no message what will happen when I click on the progress bar

**Describe the solution you'd like**
I see two options:
1. Replace the entire bar by a proper library to render a stacked bar chart. The one I like most is this one: https://nivo.rocks/bar/ Notice how this bar chart has nice (customizable) tooltips when you hover over a part of the stacked bar chart. This means we could offer different kinds of click actions depending on which part of the bar is clicked, in addition to having just a better display.
2. Sticking with the current progress bar design, we could exchange just the current tooltip for a [MUI Popover](https://mui.com/material-ui/react-popover/) which contains information similar to how the tooltip of option 1. looks like, i.e., a colored box followed by a title and a count. The popover would allow us to render custom content and not just text, unlike the current tooltip.

LMayTNG:
I wouldn't add a new library for this, this should be possible with the current dependencies.

LMayTNG:
I actually think we already have a pretty good user experience here. We already show the counts of all files shown in the progress bar, you don't need to hover over the individual sections.

We could also show "Click to show next relevant file" in the progress bar, or we could change the click behavior depending on which part of the progress bar you click, but I don't know if that's worth the effort.

PR #2639, opened 2024-05-29
feat: use nivo responsive bar chart for progress and criticality bar

meretp:
### Summary of changes

To improve the progress bar this PR adds a new library nivo and introduces a ResponsiveBar. 

fixes part of #2622 

Open OpossumUI and play around with the progress/criticality bar.


src/Frontend/Components/ProgressBar/__tests__/ProgressBar.test.tsx
meretp, line 1: I decided to replace this unit test with an e2e test.

src/Frontend/Components/ProgressBar/ProgressBar.util.tsx
```diff
@@ -37,27 +37,95 @@ export function useOnProgressBarClick(resourceIds: Array<string>) {
...
+    borderRadius: '4px',
+    color: 'white',
+    fontFamily: 'Karla Variable, sans-serif',
+    padding: 4,
+    fontSize: 10,
+    margin: 2,
+    fontWeight: 500,
+    position: 'relative',
+    top: 100,
+    'z-index': 100,
```
mstykow, line 53: this is not a valid css prop as is and should be `zIndex`. but given that it doesn't do anything right now we should probably just delete it then.

src/Frontend/Components/ProgressBar/ProgressBar.util.tsx
```diff
@@ -37,27 +37,95 @@ export function useOnProgressBarClick(resourceIds: Array<string>) {
...
-      )}
+    <MuiBox aria-label={'tooltip'} style={classes.tooltip}>
+      Number of resources...
+      <div style={classes.div}>
+        <span
+          style={{
+            ...classes.colorBlock,
+            background: OpossumColors.pastelDarkGreen,
+          }}
+        ></span>
```
mstykow, line 79: here and everywhere else:

```suggestion
        <span
          style={{
            ...classes.colorBlock,
            background: OpossumColors.pastelDarkGreen,
          }}
        />
```

src/Frontend/Components/ProgressBar/ProgressBar.util.tsx
```diff
@@ -70,129 +138,47 @@ export function getCriticalityBarTooltipText(
...
-      {new Intl.NumberFormat().format(filesWithNonCriticalExternalAttributions)}
+      <div style={classes.div} role="tooltip" aria-label="tooltip">
+        <span
+          style={{ ...classes.colorBlock, background: criticalityColor.high }}
+        ></span>
+        <span>
+          …containing highly critical signals:{' '}
+          {new Intl.NumberFormat().format(
+            progressBarData.filesWithHighlyCriticalExternalAttributionsCount,
+          )}{' '}
```
mstykow, line 151: here and below: i don't see why we need a trailing space after the item

src/Frontend/Components/ProgressBar/ProgressBar.util.tsx
```diff
@@ -37,27 +37,95 @@ export function useOnProgressBarClick(resourceIds: Array<string>) {
...
+    background: 'rgba(97, 97, 97, 0.92)',
+    borderRadius: '4px',
+    color: 'white',
+    fontFamily: 'Karla Variable, sans-serif',
+    padding: 4,
+    fontSize: 10,
+    margin: 2,
+    fontWeight: 500,
+    position: 'relative',
+    top: 100,
```
mstykow, line 52: the positioning is not working consistently: depending on the bar's mode, the tooltip appears higher or lower below the cursor. could you find a way to make the positioning consistent?

src/Frontend/Components/ProgressBar/ProgressBar.util.tsx
```diff
@@ -37,27 +37,95 @@ export function useOnProgressBarClick(resourceIds: Array<string>) {
...
 
+const classes = {
+  tooltip: {
+    flex: 1,
+    background: 'rgba(97, 97, 97, 0.92)',
+    borderRadius: '4px',
+    color: 'white',
+    fontFamily: 'Karla Variable, sans-serif',
+    padding: 4,
+    fontSize: 10,
```
mstykow, line 48: to me this seems quite small. i'm not aware of us using 10px anywhere else. i've seen 12px for captions and i think that's the smallest size we should use.

src/Frontend/Components/ProgressBar/ProgressBar.util.tsx
```diff
@@ -37,27 +37,95 @@ export function useOnProgressBarClick(resourceIds: Array<string>) {
...
+    margin: 2,
+    fontWeight: 500,
+    position: 'relative',
+    top: 100,
+    'z-index': 100,
+  } as CSSProperties,
+  div: { display: 'flex', 'align-items': 'center' },
```
mstykow, line 55: the name "div" for a css class that's applied to any of the many divs an app may have is really not the best. it's better to choose class names by the function they fulfill.

src/Frontend/Components/ProgressBar/ProgressBar.tsx
```diff
@@ -49,31 +43,78 @@ export function ProgressBar(props: ProgressBarProps): ReactElement {
...
+
   return (
-    <MuiBox sx={props.sx}>
-      <MuiTooltip
-        title={
+    <MuiBox sx={{ ...props.sx, ...classes.bar }} aria-label={'ProgressBar'}>
+      <ResponsiveBar
+        data={data}
+        keys={Object.keys(data[0])}
+        margin={{ top: 5, right: 1, bottom: 5, left: 6 }}
```
mstykow, line 82: these margins are asymmetrical and deviate from the multiples-of-4 rule. i think there is no reason to break the rules and introduce asymmetry here:
```suggestion
        margin={{ top: 4, bottom: 4 }}
```

src/Frontend/Components/ProgressBar/ProgressBar.tsx
```diff
@@ -49,31 +43,78 @@ export function ProgressBar(props: ProgressBarProps): ReactElement {
       ? resourcesWithCriticalExternalAttributions
       : props.progressBarData.resourcesWithNonInheritedExternalAttributionOnly,
   );
+  const data: Array<{ [key: string]: number }> = props.showCriticalSignals
+    ? [
+        {
+          high_critical:
```
mstykow, line 49: here and below: JavaScript uses camel case

src/Frontend/Components/ProgressBar/ProgressBar.tsx
```diff
@@ -4,28 +4,22 @@
...
 const classes = {
   bar: {
-    flex: 1,
-    border: `2px solid ${OpossumColors.white}`,
-    marginTop: '6px',
-    height: '20px',
     '&:hover': {
       cursor: 'pointer',
-      opacity: 0.75,
+      opacity: 1,
```
mstykow, line 22: do we still need this opacity declaration?

src/Frontend/Components/ProgressBar/ProgressBar.tsx
```diff
@@ -49,31 +43,78 @@ export function ProgressBar(props: ProgressBarProps): ReactElement {
       ? resourcesWithCriticalExternalAttributions
       : props.progressBarData.resourcesWithNonInheritedExternalAttributionOnly,
   );
+  const data: Array<{ [key: string]: number }> = props.showCriticalSignals
```
mstykow, line 46: could you use nivo's built-in type (`BarDatum`) to ensure that the object we're defining here always matches what nivo expects, even if nivo gets refactored?

src/Frontend/Components/ProgressBar/ProgressBar.util.tsx
```diff
@@ -37,27 +37,95 @@ export function useOnProgressBarClick(resourceIds: Array<string>) {
   };
 }
 
+const classes = {
+  tooltip: {
+    flex: 1,
```
mstykow, line 42: this prop doesn't seem to have any effect

src/Frontend/Components/ProgressBar/ProgressBar.util.tsx
```diff
@@ -37,27 +37,95 @@ export function useOnProgressBarClick(resourceIds: Array<string>) {
   };
 }
 
+const classes = {
+  tooltip: {
```
mstykow, line 41: could you give the interior of the tooltip a little more room to breathe? right now the lines of text have hardly any space between them and are also really close to the edge of the container.

mstykow (changes requested):

src/Frontend/Components/ProgressBar/ProgressBar.util.tsx
meretp: This is still to do, but so far I couldn't find a way.

src/Frontend/Components/ProgressBar/ProgressBar.util.tsx
```diff
@@ -37,27 +37,93 @@ export function useOnProgressBarClick(resourceIds: Array<string>) {
...
+    background: 'rgba(97, 97, 97, 0.92)',
+    borderRadius: '4px',
+    color: 'white',
+    fontFamily: 'Karla Variable, sans-serif',
+    padding: 8,
+    fontSize: 12,
+    margin: 2,
+    fontWeight: 500,
+    // position: 'relative',
+    // top: 'auto',
```
mstykow, line 51: can we delete these commented props?

src/Frontend/Components/ProgressBar/ProgressBar.util.tsx
```diff
@@ -37,27 +37,93 @@ export function useOnProgressBarClick(resourceIds: Array<string>) {
...
+      Number of resources...
+      <div style={classes.tooltipLine}>
+        <span
+          style={{
+            ...classes.colorBlock,
+            background: OpossumColors.pastelDarkGreen,
+          }}
+        />
+        <span>
+          ...with attributions:{' '}
```
mstykow, line 79: here and above: let's use either three separate dots "..." or the special three dot symbol "…" consistently.

src/Frontend/Components/ProgressBar/ProgressBar.util.tsx
mstykow: ok, i'll have a look

src/Frontend/Components/ProgressBar/ProgressBar.util.tsx
meretp: sure, I was trying to debug and find a solution for the positioning, I reverted it to the previous state.

src/Frontend/Components/ProgressBar/ProgressBar.util.tsx
meretp: done

mstykow:
The following issues need to be solved before this PR can merge:
1. the position differs between tooltip modes
2. when resizing the window down, the new bar chart does not resize itself thus causing x-overflow

meretp:
Added some commits that fixes the second issue and updates nivo bar - there has been a new release in the meantime. Also I updated the diff to use the ariaLabel from ResponsiveBar.

meretp:
To fix the first issue the only idea I could come up with is to use sepearate CSS styling like
 ```
transformProgress: { transform: 'translate(0px, 168px)' } as CSSProperties,
 transformCritical: { transform: 'translate(0px, 144px)' } as CSSProperties,
```
to transform the tooltips separately. This does not feel like a stable solution so I only add this as a comment not commit yet, what do you think?

meretp:
Have to take a look at why the unit tests are failing now

mstykow:
Closing for now since no work done in a while.

Issue #2689, opened 2024-12-15
Change field "license name" to "license expression"


mstykow:
**Is your feature request related to a problem? Please describe.**
The field "license name" can be used to enter license expressions. Since not every license expression is a license name but the converse is true, it would be more correct to name the field "license expression".

Don't change the underlying data model but just the display name.

PR #2701, opened 2025-01-03
fix: change field "license name" to "license expression"

Hellgartner:
### Summary of changes

Change the label for the license name field to license expression as it better describes the actual possibilities

The field "license name" can be used to enter license expressions. Since not every license expression is a license name but the converse is true, it would be more correct to name the field "license expression".

Fixes https://github.com/opossum-tool/OpossumUI/issues/2689

* Open the new version of Opossum UI
* Load any file
* Check any attribution
image

See also 

### DIFF
--- a/src/shared/text.ts
+++ b/src/shared/text.ts
@@ -25,6 +25,7 @@ export const text = {
     invalidPurl: 'INVALID PURL',
     legalInformation: 'Legal Information',
     licenseName: 'License Name',
+    licenseExpression: 'License Expression',
     licenseText: 'License Text',
     licenseTextDefault: 'License Text (inferred from license name)',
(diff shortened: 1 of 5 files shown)


src/shared/text.ts
```diff
@@ -25,6 +25,7 @@ export const text = {
     invalidPurl: 'INVALID PURL',
     legalInformation: 'Legal Information',
     licenseName: 'License Name',
```
mstykow, line 27: this now seems unused and can be deleted.

mstykow (approved):

src/shared/text.ts
Hellgartner: fixed

PR #2728, opened 2025-01-29
Feat: Enhance import process

### LINKED ISSUE #2714: Enhance import process for non-.opossum files
Importing non-.opossum files should no longer be possible through "Open File", but should rather be handled by a new import action that improves on the workflow which is currently triggered when opening non-.opossum files through "Open File".

- [x] Restrict "Open File" to only allow .opossum files
- [x] Add new "Import" menu entry to the "File" menu, with a submenu entry for the outdated opossum file format
- [x] Create a dialog that is opened when the submenu entry is clicked. The dialog should allow to select an input file and specify where the derived .opossum file should be saved. It should also provide a brief explanation about the import process.
- [x] When the user clicks to proceed with the import, the creation and loading of the .opossum file should proceed as in the current version.

PhilippMa:
### Summary of changes

Added an import dialog for importing non-.opossum files into OpossumUI. The import dialog is accessible through a new "Import file" menu item with submenus for all supported file types. In the dialog, users select the file to import and the location where the newly created .opossum file should be saved, either through native file select dialogs or through text field input. The file paths are validated, invalid file paths cause error messages to show and the import button only works when the file paths are valid. Clicking the import button uses the existing functions for loading files into OpossumUI, but the loading log information is displayed directly in the import dialog instead of the ProcessPopup used when opening files. As non-.opossum files are now handled through this import workflow, the open file action now only allows selection of .opossum files.

We want OpossumUI to be able to import files with multiple different formats. For this purpose, separating importing from just plainly opening files creates more clarity for the user. Also, it was deemed confusing that the existing workflow for opening legacy opossum files displayed a warning over having to create a new .opossum file, and that it would be better to allow the user to select the location of this new .opossum file instead.

In future issues, the import dialog will be extended with supporting more file types and handling errors during file conversion.

For viewing that the loading logs are displayed correctly, either a large file should be imported or some manual delays put into the loadInputAndOutputFromFilePath function in importFromFile.ts, as small files are loaded so quickly that the loading logs are only visible for a very short time before the dialog closes.

closes #2714


src/ElectronBackend/main/__tests__/listeners.test.ts
```diff
@@ -99,201 +104,13 @@ jest.mock('../../input/importFromFile', () => ({
...
-
-    await getOpenDotOpossumFileInsteadListener(mainWindow)();
-
-    expect(loadInputAndOutputFromFilePath).toHaveBeenCalledWith(
-      expect.anything(),
-      opossumFilePath,
-    );
-  });
-});
+// TODO: add tests for import listeners
```
mstykow, line 113: there is still a TODO here

src/ElectronBackend/main/dialogs.ts
```diff
@@ -4,21 +4,49 @@
...
     : undefined;
 }
 
+export function openOpossumFileDialog(): Array<string> | undefined {
+  const filters = [
+    {
+      name: 'Opossum File',
+      extensions: ['opossum'],
+    },
+  ];
```
mstykow, line 27: this is an untyped object. easiest way to fix this issue is by inlining this in line 28.

src/ElectronBackend/main/__tests__/listeners.test.ts
```diff
@@ -638,6 +433,228 @@ describe('_exportFileAndOpenFolder', () => {
...
+    );
+
+    expect(result).toStrictEqual([
+      FilePathValidity.VALID,
+      FilePathValidity.OVERWRITE_WARNING,
+    ]);
+  });
+});
+
 function prepareBomSPdxAndFollowUpit(): Promise<BrowserWindow> {
```
mstykow, line 658: not new, but what's going on with the name of this function? what does "FollowUpit" mean? also, "SPdx" should probably be "SPDX"?

src/ElectronBackend/main/__tests__/listeners.test.ts
```diff
@@ -638,6 +433,228 @@ describe('_exportFileAndOpenFolder', () => {
...
+
+    expect(mainWindow.webContents.send).toHaveBeenCalledWith(
+      AllowedFrontendChannels.ImportFileShowDialog,
+      fileFormat,
+    );
+  });
+});
+
+describe('getImportFileSelectInputListener', () => {
+  it('calls openNonOpossumFileDialog and returns received file path', async () => {
```
mstykow, line 454: this test is testing an implementation detail. also, as soon as i rename the function `openNonOpossumFileDialog` the title of the test is no longer valid. i would question the value of this test as long as it has this title. perhaps you mean to test that the listener returns the expected file path? if so, i may be convinced that this test is useful.

src/shared/shared-types.ts
```diff
@@ -229,9 +229,15 @@ export interface ExternalAttributionSources {
   [source: string]: ExternalAttributionSource;
 }
 
-export interface FileSupportPopupArgs {
-  showFileSupportPopup: boolean;
-  dotOpossumFileAlreadyExists: boolean;
+export type FileFormatInfo = [string, Array<string>];
```
mstykow, line 232: this type is somewhat obscure. i can see that it's a tuple but i have no idea what the first and second value are supposed to represent by looking at it. if i had to guess, perhaps the first value is the file ending and the second value is a list of paths? better to use an object type with clear key names.

src/Frontend/Components/ImportDialog/ImportDialog.tsx
```diff
@@ -0,0 +1,258 @@
...
+                overflow: 'hidden',
+                textOverflow: 'ellipsis',
+                whiteSpace: 'nowrap',
+              }}
+            >
+              {processInfo}
+            </MuiTypography>
+          </>
+        ) : (
+          <></>
```
mstykow, line 241: unnecessary empty fragment. just return undefined

src/Frontend/Components/ImportDialog/ImportDialog.tsx
```diff
@@ -0,0 +1,258 @@
...
+import { LoggingListener, useIpcRenderer } from '../../util/use-ipc-renderer';
+import { FilePathInput } from '../FilePathInput/FilePathInput';
+import { NotificationPopup } from '../NotificationPopup/NotificationPopup';
+import { Spinner } from '../Spinner/Spinner';
+
+const explanationTextLine1 =
+  'OpossumUI will convert the selected file into a new opossum file.';
+
+const explanationTextLine2 =
+  'All changes made to the project in OpossumUI will be saved in this opossum file.';
```
mstykow, line 22: please move all user-facing texts to `text.ts` and import from there.

src/Frontend/Components/ImportDialog/ImportDialog.tsx
```diff
@@ -0,0 +1,258 @@
...
+
+interface ImportDialogProps {
+  fileFormat: FileFormatInfo;
+}
+
+export const ImportDialog: React.FC<ImportDialogProps> = ({ fileFormat }) => {
```
mstykow, line 28: is there only one import dialog? i think in the future we'll have all sorts of imports so should this one be named more specifically to be related to the outdated opossum file format? or do you want to create a generic component that works the same for all imports and just has some props to exchange the various helper texts?

src/Frontend/Components/ImportDialog/ImportDialog.tsx
```diff
@@ -0,0 +1,258 @@
...
+    switch (opossumFilePathValidity) {
+      case FilePathValidity.EMPTY_STRING:
+        return 'No save location selected';
+      case FilePathValidity.WRONG_EXTENSION:
+        return 'File extension has to be .opossum';
+      case FilePathValidity.PATH_DOESNT_EXIST:
+        return 'The path contains a non-existent directory';
+      case FilePathValidity.VALIDATION_FAILED:
+        return 'Invalid file path';
+      default:
```
mstykow, line 105: instead of introducing a `FilePathValidity` enum, why not directly send a text error message from the backend to the frontend?

src/Frontend/shared-constants.ts
```diff
@@ -73,3 +74,7 @@ export const thirdPartyKeys: Array<keyof PackageInfo> = [
   'licenseName',
   'licenseText',
 ];
+
+export const importFileFormats: Array<FileFormatInfo> = [
+  ['Legacy Opossum File', ['json', 'json.gz']],
+];
```
mstykow, line 80: the only reason you have to put this const into the frontend shared constants is because you're using it in the global popup tests. i don't think that's enough justification to put it here.

src/e2e-tests/__tests__/import-dialog.test.ts
```diff
@@ -0,0 +1,64 @@
+// SPDX-FileCopyrightText: Meta Platforms, Inc. and its affiliates
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+import { expect } from '@playwright/test';
+
+import { faker, test } from '../utils';
+
+const [resourceName1] = faker.opossum.resourceNames({ count: 1 });
```
mstykow, line 9: you have no other resource so there is no need for "1"

src/shared/ipc-channels.ts
```diff
@@ -7,14 +7,13 @@
 export enum IpcChannel {
   DeleteFile = 'delete-file',
   ExportFile = 'export-file',
-  KeepFile = 'keep-file',
   OpenFile = 'open-file',
+  ImportFileSelectInput = 'import-file-select-input',
+  ImportFileSelectSaveLocation = 'import-file-select-save-location',
+  ImportFileConvertAndLoad = 'import-file-convert-and-load',
+  ImportFileValidatePaths = 'import-file-validate-paths',
```
mstykow, line 14: did you make your backend logic generic enough so that it'll be able to use the same channels for other import dialogs in the coming stories? i don't want to see 5 new channels for each import dialog we create.

src/shared/shared-types.ts
```diff
@@ -229,9 +229,15 @@ export interface ExternalAttributionSources {
...
+
+export enum FilePathValidity {
```
mstykow, line 234: as mentioned elsewhere: i think we should delete this enum but we can discuss.

src/shared/write-file.ts
```diff
@@ -9,6 +9,27 @@ export const OPOSSUM_FILE_EXTENSION = '.opossum';
...
+  possibleExtensions: Array<string>,
+): string {
+  const fileExtensionLength = possibleExtensions
+    .map((extension) => {
+      if (resourceFilePath.endsWith(`.${extension}`)) {
+        return extension.length;
+      }
+      return 0;
+    })
+    .reduce((a, b) => Math.max(a, b), 0);
```
mstykow, line 23: i'm not sure why you're using `reduce` here: `Math.max` can be applied directly to arrays.

src/shared/write-file.ts
```diff
@@ -9,6 +9,27 @@ export const OPOSSUM_FILE_EXTENSION = '.opossum';
 export const INPUT_FILE_NAME = 'input.json';
 export const OUTPUT_FILE_NAME = 'output.json';
 
+export function getDotOpossumFilePath(
```
mstykow, line 12: this whole function seems overly complicated to me. let's use a library to simplify this process. we are already using `upath` to manipulate paths in the frontend. it would go something like this:
```
upath.changeExt(originalFilePath, 'opossum')
```

src/ElectronBackend/errorHandling/errorHandling.ts
```diff
@@ -26,23 +49,22 @@ export function createListenerCallbackWithErrorHandling(
...
-      }
+      await reportListenerError(mainWindow, error);
+    }
+  };
+}
+
+export function createListenerCallbackWithErrorHandling<T>(
+  mainWindow: BrowserWindow,
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-function-type
+  func: Function,
```
mstykow, line 60: i wonder if we can improve the type here: the first arg of this function is always the same just the following args are variable. there should be a way to type that correctly without disabling the linter.

src/Frontend/Components/FilePathInput/FilePathInput.tsx
```diff
@@ -0,0 +1,46 @@
...
+  return (
+    <FormControl sx={{ display: 'flex', flexDirection: 'column' }}>
+      <MuiBox sx={{ display: 'flex', alignItems: 'center', pt: '10px' }}>
+        <TextBox
+          title={props.label}
+          text={props.text}
+          error={props.errorMessage !== null}
+          handleChange={(event) => props.onEdit(event.target.value)}
+          onBlur={props.onBlur}
+          sx={{ width: 600, mr: '10px' }}
```
mstykow, line 33: i don't think any other place in this app uses the MUI specific shorthands ("pt", "mr", ...) for styling.

src/Frontend/Components/FilePathInput/FilePathInput.tsx
```diff
@@ -0,0 +1,46 @@
...
+        />
+        <IconButton
+          icon={<Folder fontSize="medium" />}
+          onClick={props.onButtonClick}
+          tooltipTitle={props.buttonToolTip}
+        />
+      </MuiBox>
+      <FormHelperText aria-label={'file path helper text'} error={true}>
+        {props.errorMessage ?? props.warnMessage ?? ' '}
```
mstykow, line 42: do you really want to render a whitespace when there is neither error nor warn message? that seems rather unusual. expected would be to not render the entire `FormHelperText`.

src/Frontend/Components/GlobalPopup/GlobalPopup.tsx
```diff
@@ -24,18 +24,19 @@ function getPopupComponent(popupType: PopupType | null) {
...
-    case PopupType.FileSupportDotOpossumAlreadyExistsPopup:
-      return <FileSupportDotOpossumAlreadyExistsPopup />;
     case PopupType.UpdateAppPopup:
       return <UpdateAppPopup />;
-    case null:
+    case PopupType.ImportDialog:
+      if (popupInfo?.fileFormat) {
+        return <ImportDialog fileFormat={popupInfo.fileFormat} />;
+      }
+      return null;
```
mstykow, line 33: i don't think we should integrate your new import dialog with `GlobalPopup`. you couldn't have know this but `GlobalPopup` is a deprecated mechanism that we're trying to replace bit by bit. that's also why the open-file dialog is not here.

src/Frontend/Components/ImportDialog/ImportDialog.tsx
```diff
@@ -0,0 +1,258 @@
...
+    opossumFilePathValidity === FilePathValidity.OVERWRITE_WARNING;
+
+  const [showInputFilePathErrors, setShowInputFilePathErrors] =
+    useState<boolean>(false);
+  const [showOpossumFilePathErrors, setShowOpossumFilePathErrors] =
+    useState<boolean>(false);
+
+  const [processInfo, setProcessInfo] = useState<string>('');
+
+  const [isLoading, setIsLoading] = useState<boolean>(false);
```
mstykow, line 53: i think there is too many states here and we should think about how to get rid of some of them. removing FilePathValidity will already help a little bit.

src/Frontend/Components/ImportDialog/ImportDialog.tsx
```diff
@@ -0,0 +1,258 @@
...
+  }
+
+  async function onConfirm(): Promise<void> {
+    setIsLoading(true);
+
+    setShowInputFilePathErrors(true);
+    setShowOpossumFilePathErrors(true);
+
+    if (inputFilePathIsValid && opossumFilePathIsValid) {
+      const success = await window.electronAPI.importFileConvertAndLoad(
```
mstykow, line 173: can you return a possible error message here? then we don't need to make another BE call to validate the inputs. after all, validating the inputs could be unrelated to why the confirmation step was unsuccessful.

src/Frontend/Components/FilePathInput/FilePathInput.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { FormControl, FormHelperText } from '@mui/material';
+import MuiBox from '@mui/system/Box';
+
+import { IconButton } from '../IconButton/IconButton';
+import { TextBox } from '../TextBox/TextBox';
+
+interface FilePathInputProps {
+  label: string;
+  text: string;
+  buttonToolTip: string;
```
mstykow, line 15: tooltip is one word
```suggestion
  buttonTooltip: string;
```

mstykow (changes requested):
general comments:
1. great first PR! it's clear you've thought deeply about the problem and addressed many edge cases.
2. this PR is too large. it's intimidating to review it, you are getting feedback only after days of work, and it will take a long time to get it to a mergable state because it changes so many things. next time, please think about how to chop your changes into smaller chunks.

src/ElectronBackend/main/__tests__/listeners.test.ts
PhilippMa: resolved

src/ElectronBackend/main/dialogs.ts
PhilippMa: resolved

src/ElectronBackend/main/dialogs.ts
```diff
@@ -4,21 +4,49 @@
...
+  return openFileDialog(filters);
+}
+
+export function openNonOpossumFileDialog(
+  fileFormat: FileFormatInfo,
+): Array<string> | undefined {
+  const filters = [
```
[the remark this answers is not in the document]
PhilippMa: resolved

src/Frontend/Components/ImportDialog/ImportDialog.tsx
PhilippMa: resolved

src/Frontend/Components/ImportDialog/ImportDialog.tsx
PhilippMa: My understanding is that the UI workflow for importing different formats will be the same. My idea was to just create a single dialog component that is parameterized over the file format that will be imported, and use this file format info to adjust helper text, configure the file filter of the native filesystem dialog and communicate to the backend which converter has to be used.

src/Frontend/shared-constants.ts
PhilippMa: moved to menu.ts

src/e2e-tests/__tests__/import-dialog.test.ts
PhilippMa: resolved

src/shared/ipc-channels.ts
PhilippMa: The plan is to use only one import dialog and only one set of IPC channels for all supported file types. The coming stories might potentially have to adjust the payload of these channels but there should be no need to add any further channels.

src/shared/write-file.ts
PhilippMa: Just a typescript noob looking for an Array.max function, not finding one and deciding to implement it myself instead. I wasn't aware Math.max can be applied to arrays and does what I was looking for, thanks for the tip!

src/Frontend/Components/FilePathInput/FilePathInput.tsx
PhilippMa: changed to full attribute names

src/Frontend/Components/FilePathInput/FilePathInput.tsx
PhilippMa: resolved

src/ElectronBackend/main/__tests__/listeners.test.ts
PhilippMa: The function name doesn't make much sense to me even in the commit where its was first introduced. It appears to do just some general setup before the tests are started, maybe something like initWindowAndBackendState would be a more appropriate name?

src/ElectronBackend/main/__tests__/listeners.test.ts
PhilippMa: changed test to focus on the listener returning the file path selected by the user

src/shared/shared-types.ts
PhilippMa: resolved

src/shared/write-file.ts
PhilippMa: Weird error:
`Uncaught TypeError: y.extname is not a function`

PhilippMa: Unfortunately not, I got the error when using upath.changeExt directly from my ImportDialog component

src/shared/shared-types.ts
PhilippMa: resolved

src/Frontend/Components/FilePathInput/FilePathInput.tsx
PhilippMa: resolved

src/Frontend/Components/GlobalPopup/GlobalPopup.tsx
PhilippMa: resolved

src/Frontend/Components/ImportDialog/ImportDialog.tsx
PhilippMa: resolved

src/Frontend/Components/ImportDialog/ImportDialog.tsx
PhilippMa: error messages now processed through error logs sent from backend

src/Frontend/Components/ImportDialog/ImportDialog.tsx
PhilippMa: resolved

src/Frontend/Components/ImportDialog/ImportDialog.tsx
PhilippMa: resolved

src/Frontend/Components/FilePathInput/FilePathInput.tsx
mstykow, line 1: as discussed, will replace the folder button by a click action on the input field.

mstykow (changes requested):

src/Frontend/Components/LogDisplay/LogDisplay.tsx
```diff
@@ -0,0 +1,61 @@
...
+  { Component: React.FC<SvgIconProps>; color: string }
+> = {
+  error: { Component: ErrorIcon, color: 'red' },
+  warn: { Component: WarningIcon, color: 'orange' },
+  info: { Component: CheckIcon, color: 'green' },
+};
+
+interface LogDisplayProps {
+  log: Log;
+  isActive: boolean;
```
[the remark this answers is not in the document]
PhilippMa: After looking at the code again, this prop is used to decide whether to display a spinner **instead** of the usual info/warn/error icon, so I would prefer something that highlights that the current step that was logged is still running, e.g. `isRunning` or with reversed semantics `hasFinished`

src/Frontend/Components/TextBox/TextBox.tsx
```diff
@@ -72,7 +79,9 @@ export interface TextBoxProps {
   sx?: SxProps;
   text?: string;
   title: string;
-  endIcon?: React.ReactElement<unknown> | Array<React.ReactElement<unknown>>;
+  startIcon?: React.ReactElement | Array<React.ReactElement>;
+  endIcon?: React.ReactElement | Array<React.ReactElement>;
+  cursor?: string;
```
mstykow, line 84: the type here should be more restrictive: the CSS cursor prop isn't just any string. you should be able to import the correct CSS type here.

src/Frontend/Components/FilePathInput/FilePathInput.tsx
```diff
@@ -0,0 +1,26 @@
...
+  onClick: () => void;
+}
+
+export const FilePathInput: React.FC<FilePathInputProps> = (props) => {
+  return (
+    <TextBox
+      title={props.label}
+      text={props.text}
+      onClick={props.onClick}
+      startIcon={<AttachFile />}
```
mstykow, line 21: the icon here is too big. if you check how we've used icons in the rest of OpossumUI, we always make them smaller. for example, places which use `endIcon` use the shared CSS class "clickableIcon".

src/Frontend/Components/FilePathInput/FilePathInput.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+import { TextBox } from '../TextBox/TextBox';
+
+interface FilePathInputProps {
+  label: string;
+  text: string;
+  onClick: () => void;
+}
+
+export const FilePathInput: React.FC<FilePathInputProps> = (props) => {
```
mstykow, line 15: The input is cut off here. I think this will fix itself when you use `readOnly={true}`. See my other comment.

image

src/Frontend/Components/FilePathInput/FilePathInput.tsx
```diff
@@ -0,0 +1,26 @@
...
+  return (
+    <TextBox
+      title={props.label}
+      text={props.text}
+      onClick={props.onClick}
+      startIcon={<AttachFile />}
+      cursor={'pointer'}
+      sx={{ width: 600, marginTop: '20px' }}
```
mstykow, line 23: using the dialog, i noticed that it's width jumps (gets larger) when you click "import". not sure why but that should be fixed.

src/Frontend/Components/FilePathInput/FilePathInput.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+interface FilePathInputProps {
+  label: string;
+  text: string;
+  onClick: () => void;
+}
+
+export const FilePathInput: React.FC<FilePathInputProps> = (props) => {
+  return (
+    <TextBox
```
mstykow, line 17: in an earlier version, you used `readOnly={true}` but now you removed this prop again. the result of this is that when a file path has been entered, there are two actions available: the primary action which opens the dialog to select a file, and the secondary action which activates the cursor in the input field. this is confusing UX. an input field either has a click action, or a cursor action, not both. i don't really see a reason to still allow cursor interactions so i would use `readOnly` again.

did you change this behavior because you weren't sure how to handle overflowing input lengths to allow the user to "move to the right"? if you're concerned about the length of the input, perhaps it's enough to simply show the file name instead of the full path?

mstykow (changes requested):

src/Frontend/Components/FilePathInput/FilePathInput.tsx
PhilippMa: Good catch, that wasn't on purpose. Seems like I accidentally deleted it while cleaning up the unused properties of FilePathInput

src/Frontend/Components/TextBox/TextBox.tsx
PhilippMa: resolved

src/Frontend/Components/FilePathInput/FilePathInput.tsx
PhilippMa: resolved

src/Frontend/Components/FilePathInput/FilePathInput.tsx
PhilippMa: resolved

src/Frontend/Components/TextBox/TextBox.tsx
PhilippMa: `yarn unused-check` complains that `import { Property } from 'csstype'` is an unresolved import, what is that about? The project builds without any errors

mstykow: maybe you need a type-only import?

src/Frontend/Components/TextBox/TextBox.tsx
```diff
@@ -72,7 +80,9 @@ export interface TextBoxProps {
   sx?: SxProps;
   text?: string;
   title: string;
-  endIcon?: React.ReactElement<unknown> | Array<React.ReactElement<unknown>>;
+  startIcon?: React.ReactElement | Array<React.ReactElement>;
+  endIcon?: React.ReactElement | Array<React.ReactElement>;
+  cursor?: Property.Cursor;
```
mstykow, line 85: you want this:
```suggestion
  cursor?: React.CSSProperties['cursor'];
```

mstykow (changes requested):

src/Frontend/Components/TextBox/TextBox.tsx
PhilippMa: That works, thanks!

src/Frontend/Components/ImportDialog/ImportDialog.tsx
```diff
@@ -0,0 +1,181 @@
...
+
+  function updateInputFilePath(filePath: string) {
+    setInputFilePath(filePath);
+    setCurrentLog(null);
+  }
+
+  function updateOpossumFilePath(filePath: string) {
+    setOpossumFilePath(filePath);
+    setCurrentLog(null);
+  }
```
mstykow, line 66: #nit these two functions are not adding any value. i would just inline their contents where they are called.

src/ElectronBackend/main/listeners.ts
```diff
@@ -103,43 +103,130 @@ async function writeOutputJsonToFile(
...
+  mainWindow: BrowserWindow,
+): (
+  _: Electron.IpcMainInvokeEvent,
+  fileFormat: FileFormatInfo,
+) => Promise<string | null> {
+  return createListenerCallbackWithErrorHandling(
+    mainWindow,
+    (_: Electron.IpcMainInvokeEvent, fileFormat: FileFormatInfo) => {
+      const filePaths = openNonOpossumFileDialog(fileFormat);
+      if (!filePaths || filePaths.length < 1) {
```
mstykow, line 152: #nit an array won't have negative length
```suggestion
      if (!filePaths?.length) {
```

src/Frontend/Components/ImportDialog/ImportDialog.tsx
```diff
@@ -0,0 +1,181 @@
...
+import {
+  LoggingListener,
+  ShowImportDialogListener,
+  useIpcRenderer,
+} from '../../util/use-ipc-renderer';
+import { FilePathInput } from '../FilePathInput/FilePathInput';
+import { LogDisplay } from '../LogDisplay/LogDisplay';
+import { NotificationPopup } from '../NotificationPopup/NotificationPopup';
+
+export const ImportDialog: React.FC = () => {
```
mstykow, line 22: the only somewhat major thing that still bothers me is that the input is still cut off when the input field is focussed:
image

mstykow (changes requested):

src/Frontend/Components/FilePathInput/FilePathInput.tsx
```diff
@@ -0,0 +1,28 @@
...
+      title={props.label}
+      text={props.text}
+      onClick={props.onClick}
+      startIcon={<AttachFile sx={clickableIcon} />}
```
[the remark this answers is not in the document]
PhilippMa: resolved

src/Frontend/Components/TextBox/TextBox.tsx
```diff
@@ -101,10 +113,17 @@ export function TextBox(props: TextBoxProps) {
               textOverflow: 'ellipsis',
               paddingTop: '8.5px',
               paddingBottom: '8.5px',
-              paddingLeft: '14px',
+              paddingLeft: `calc(${ensureArray(props.startIcon).length} * 28px)`,
               paddingRight: `calc(14px + ${ensureArray(props.endIcon).length} * 28px)`,
```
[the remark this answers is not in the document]
PhilippMa: resolved

src/Frontend/Components/TextBox/TextBox.tsx
```diff
@@ -47,6 +47,12 @@ export const classes = {
       padding: '1px 3px',
     },
   },
+  startAdornmentRoot: {
+    position: 'absolute',
+    left: 0,
+    marginLeft: '14px',
```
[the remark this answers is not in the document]
PhilippMa: resolved

src/Frontend/Components/TextBox/TextBox.tsx
```diff
@@ -90,6 +99,9 @@ export function TextBox(props: TextBoxProps) {
         color={props.color}
         InputLabelProps={{
           shrink: !!props.placeholder || !!props.text,
+          sx: {
+            marginLeft: `calc(${ensureArray(props.startIcon).length} * 28px)`,
```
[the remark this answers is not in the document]
PhilippMa: Shold I then also adjust the padding to multiply by 20px instead of 28px? Otherwise the label of an empty textbox is misaligned with the text of a non-empty textbox
image

src/Frontend/Components/TextBox/TextBox.tsx
```diff
@@ -101,10 +113,17 @@ export function TextBox(props: TextBoxProps) {
               textOverflow: 'ellipsis',
               paddingTop: '8.5px',
               paddingBottom: '8.5px',
-              paddingLeft: '14px',
+              paddingLeft: `calc(${ensureArray(props.startIcon).length} * 28px)`,
               paddingRight: `calc(14px + ${ensureArray(props.endIcon).length} * 28px)`,
+              cursor: props.cursor,
             },
           },
+          sx: { cursor: props.cursor },
```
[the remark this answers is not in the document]
PhilippMa: From my testing, the cursor prop in the InputProps.inputProps.sx changes the cursor on everything but the start icon, while setting the cursor prop in InputProps.sx seems to only affect the start icon. I also tried putting the cursor prop directly into the sx prop of the MuiTextField but that seems to have no effect at all

src/Frontend/Components/ImportDialog/ImportDialog.tsx
PhilippMa: resolved

src/ElectronBackend/main/listeners.ts
PhilippMa: resolved

src/Frontend/Components/ImportDialog/ImportDialog.tsx
PhilippMa: Passing a custom input component to MuiTextField seems like a good solution here, that also disables horizontal scrolling in the textboxes and makes it easy to disable text selection as well.

src/Frontend/Components/TextBox/TextBox.tsx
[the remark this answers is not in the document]
PhilippMa: Turns out this was the correct place to put the cursor prop, it rather doesn't need to be applied in the other places I previously also had it. The issue with the padding shifting when adding the cursor prop here was because it was overwriting the sx that was indirectly given through slotProps.root, which could be fixed by combining these into one sx object

src/Frontend/Components/ImportDialog/ImportDialog.tsx
PhilippMa: fixed

mstykow (approved):

PR #2740, opened 2025-01-31
Feat: Import ScanCode Json Files

### LINKED ISSUE #2718: Good case flow for scancode json
* New sumbmenu entry in import menu
* Open inport dialog (adapt with correct options/namings)
* call interface from last story + spinner
* Load generated opossum file

PhilippMa:
### Summary of changes

By connecting the previously implemented import dialog with the existing backend interface for opossum-file, it is now possible to import ScanCode Json files into OpossumUI.

We want OpossumUI to be usable with different foreign file formats.

closes #2718


src/ElectronBackend/main/listeners.ts
```diff
@@ -213,11 +217,17 @@ export function getImportFileConvertAndLoadListener(
...
+        case FileType.LEGACY_OPOSSUM:
+          await writeOpossumFile({
+            path: opossumFilePath,
+            input: getInputJson(resourceFilePath),
+            output: getOutputJson(resourceFilePath),
+          });
+          break;
+        case FileType.SCANCODE_JSON:
+          await convertScancodeToOpossum(resourceFilePath, opossumFilePath);
+      }
```
Hellgartner, line 230: For safety for extending this code, I would recommend to add another break at the end of the second case statement

src/ElectronBackend/main/openFileFromCliOrEnvVariableIfProvided.ts
```diff
@@ -45,9 +45,13 @@ export async function openFileFromCliOrEnvVariableIfProvided(
   const inputScanCodeFileFromEnvVariable: string | undefined =
     process.env.SCANCODE_JSON;
   if (!inputFileName && inputScanCodeFileFromEnvVariable) {
-    inputFileName = await convertScancodeToOpossum(
+    inputFileName = getFilePathWithAppendix(
```
Hellgartner, line 48: Is this really the input filename or more like the output filename?
That is the file from the resulting opossum file which is later the input file again, correct?
So better be named something along the lines of `createdOpossumFile`?

src/ElectronBackend/main/openFileFromCliOrEnvVariableIfProvided.ts
Hellgartner, line 1: do we actually need the possibility to trigger the conversion via command line now that we can do it via frontend?

src/ElectronBackend/opossum-file/__tests__/convertToOpossum.test.ts
```diff
@@ -12,21 +12,19 @@ import { isOpossumFileFormat } from '../../utils/isOpossumFileFormat';
 import { convertScancodeToOpossum } from '../convertScancodeToOpossum';
 
 function getTempPath(): string {
-  return join(tmpdir(), uniqueId('opossum_'));
+  return join(tmpdir(), `${uniqueId('opossum_')}.opossum`);
```
Hellgartner, line 15: The function actually no longer returns a tem path but a filename

src/Frontend/Components/ImportDialog/ImportDialog.tsx
```diff
@@ -22,6 +22,7 @@ import { NotificationPopup } from '../NotificationPopup/NotificationPopup';
 export const ImportDialog: React.FC = () => {
   const [isOpen, setIsOpen] = useState<boolean>(false);
   const [fileFormat, setFileFormat] = useState<FileFormatInfo>({
+    fileType: FileType.LEGACY_OPOSSUM,
```
Hellgartner, line 25: Why is legacy the default value here, what makes that special?

src/e2e-tests/page-objects/MenuBar.ts
```diff
@@ -26,12 +26,8 @@ export class MenuBar {
     await clickMenuItem(this.window.app, 'label', 'Project Statistics');
   }
 
-  async openImportDialog(): Promise<void> {
-    await clickMenuItem(
-      this.window.app,
-      'label',
-      'Legacy Opossum File (.json/.json.gz)',
-    );
+  async openByLabel(label: string): Promise<void> {
```
Hellgartner, line 29: instead of having that arbitrary string, maybe provide just two mehtods: One for importing the old file format and one for scancode

src/e2e-tests/utils/fixtures.ts
```diff
@@ -67,7 +67,7 @@ export const test = base.extend<{
   linkedResourcesTree: LinkedResourcesTree;
   menuBar: MenuBar;
   notSavedPopup: NotSavedPopup;
-  openFileManually: boolean;
+  testImport: boolean;
```
Hellgartner, line 70: I cannot follow this name change tbh.

src/shared/shared-types.ts
Hellgartner, line 1: Why is this a `.ts` file and not a `.d.ts` file?

Hellgartner (changes requested):
Impressive!
That needed way less code changes that I had feared for --> good work from the previous tickets!

src/ElectronBackend/opossum-file/__tests__/convertToOpossum.test.ts
PhilippMa: Good catch, that refactoring was kind of incomplete. I just inlined the function body since we now only use it at one point anyway

src/Frontend/Components/ImportDialog/ImportDialog.tsx
PhilippMa: This is just a dummy value, as the assumption is that this fileFormat state will always be set to a proper value when the dialog is opened. I mainly wanted to avoid doing explicit null value handling when that should never be important, but if you have any ideas for how to handle this in a cleaner way then I'm open for suggestions!

src/e2e-tests/utils/fixtures.ts
PhilippMa: I introduced that field in my previous PR, while working on this issue I just felt that the name was a bit misleading because the field is only used in the import dialog tests and has nothing to do with plain opening of files. I'm also open for other naming suggestions though

src/shared/shared-types.ts
PhilippMa: Prehaps investigate in future issue

src/ElectronBackend/main/openFileFromCliOrEnvVariableIfProvided.ts
PhilippMa: currently there is no active use case for that, it might be interesting to implement it as part of a future issue

Hellgartner (approved):

PR #2745, opened 2025-02-03
test: fix resolution to reduce flakiness of e2e tests

Hellgartner:
### Summary of changes

Fixed resolution for e2e tests

Some tests where found to be flaky and one difference of local test (passing) and remote tests (failing) was the difference in resolution.

Testing should be covered by ci


abraemer (approved):
I guess time will tell whether this truly fixed the issue :)

PR #2749, opened 2025-02-03
Feat improve install-deps script

### LINKED ISSUE #2746: CI fails due to Github rate limit
Since #2725 our CI pipelines sometimes fails due to "exceeding the Github rate limit". This PR added 2 GitHub-API calls to each pipeline run. Apparently this causes rate limiting if the jobs start all at once.

abraemer:
### Summary of changes

* remove dependency on `gh-release-fetch`
* use `GITHUB_TOKEN` for authentification during CI pipelines

Using authentificated request to GitHub API allows significantly more requests (at [least 5000/h](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28) instead of 60/h for unauthenticated requests)

Closes #2746

CI pipelines shouldn't fail anymore.

### DIFF
--- a/tools/downloadOpossumFile.mjs
+++ b/tools/downloadOpossumFile.mjs
@@ -8,116 +8,66 @@ import { env } from 'process';
 import { pipeline } from 'stream';
 
 const EXECUTE_PERMISSIONS = 0o755;
-const HTTP_FORBIDDEN = 403;
 
-async function getLatestRelease(request_params = undefined) {
-  try {
-    const res = await fetch(
-      'https://api.github.com/repos/opossum-tool/opossum-file/releases/latest',
-      request_params,
+async function downloadOpossumFile() {
+  const osSuffix = process.argv[2];
+
+  if (!osSuffix) {
+    console.error(
+      'Please specify one of the following options: mac, ubuntu, windows.exe',
     );
-    const data = await res.json();
-    if (
-      res.status === HTTP_FORBIDDEN &&
-      typeof data.message === 'string' &&
-      data.message.includes('API rate limit exceeded')
-    ) {
-      throw new Error('API rate limit exceeded, please try again later');
-    }
-    return data.tag_name;
-  } catch (err) {
-    throw new Error(`Fetching the latest release failed: ${err.message}`);
+    process.exit(1);
   }
-}
 
-async function downloadBinary(
-  version,
-  filename,
-  savepath,
-  request_params = undefined,
-) {
-  const url = `https://github.com/opossum-tool/opossum-file/releases/download/${version}/${filename}`;
-  try {
-    const res = await fetch(url, request_params);
-    if (!res.ok) {
-      throw new Error(`HTTP error! status: ${res.status}`);
-    }
-    await new Promise((resolve, reject) => {
-      pipeline(res.body, fs.createWriteStream(savepath), (err) => {
-        if (err) {
-          reject(err);
-        } else {
-          resolve();
-        }
-      });
-    });
-    console.log(`File downloaded and saved to ${savepath}`);
-  } catch (err) {
-    console.error(`Error downloading file: ${err.message}`);
-  }
-}
-
-function prepareDownloadLocation(path) {
-  const folder = dirname(path);
-  if (!fs.existsSync(folder)) {
-    console.info(`Creating folder ${folder}`);
-    fs.mkdirSync(folder);
-  }
-  if (fs.existsSync(path)) {
-    console.info("Deleting existing 'opossum-file'.");
-    fs.rmSync(path);
-  }
-}
-
-async function downloadOpossumFile(osSuffix, headers = undefined) {
-  const TARGET_NAME = 'bin/opossum-file';
+  const destinationPath = 'bin/opossum-file';
   const opossumFileBinaryName = `opossum-file-for-${osSuffix}`;
-  prepareDownloadLocation(TARGET_NAME);
 
-  try {
-    const currentVersion = await getLatestRelease(headers);
-    await downloadBinary(
-      currentVersion,
-      opossumFileBinaryName,
-      TARGET_NAME,
-      headers,
-    );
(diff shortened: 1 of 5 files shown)


tools/downloadOpossumFile.mjs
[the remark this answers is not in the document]
abraemer: mostly done. There is an issue though: The linter reports two errors because `'fetch' is not defined  no-undef` which I believe are just wrong because fetch is global since NodeJS 18 and the script works.
Maybe the linter is somehow configured wrong? If not please advice how to fix.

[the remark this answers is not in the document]
mstykow: you need `fetch: true` under `globals` in the eslint config.

mstykow (changes requested):

tools/downloadOpossumFile.mjs
```diff
@@ -3,48 +3,90 @@
...
 
 const EXECUTE_PERMISSIONS = 0o755;
+const HTTP_FORBIDDEN = 403;
 
-async function downloadOpossumFile(osSuffix, downloadDestination = 'bin') {
-  const TARGET_NAME = 'bin/opossum-file';
-  const opossumFileBinaryName = `opossum-file-for-${osSuffix}`;
-  const downloadedFileName = join(downloadDestination, opossumFileBinaryName);
-
+async function getLatestRelease(request_params = undefined) {
```
mstykow, line 13: 1. javascript uses camel-case
2. why are you setting a default value if you never call this function without explicitly passing a parameter?

tools/downloadOpossumFile.mjs
```diff
@@ -3,48 +3,90 @@
...
+    const data = await res.json();
+    if (
+      res.status === HTTP_FORBIDDEN &&
+      typeof data.message === 'string' &&
+      data.message.includes('API rate limit exceeded')
+    ) {
+      throw new Error('API rate limit exceeded, please try again later');
+    }
+    return data.tag_name;
+  } catch (err) {
```
mstykow, line 28: general comment about your use of try/catch: it seems that in most cases your try catch does nothing more than wrap the error message in a few additional words. is this really adding any value? i personally would be just as happy with the raw error and stack trace without you telling me in addition that "Fetching the latest release failed".

tools/downloadOpossumFile.mjs
```diff
@@ -3,48 +3,90 @@
...
 
-  if (fs.existsSync(TARGET_NAME)) {
-    console.info('Found opossum-file binary. Overwriting.');
+async function downloadBinary(
+  version,
+  filename,
+  savepath,
+  request_params = undefined,
+) {
+  const url = `https://github.com/opossum-tool/opossum-file/releases/download/${version}/${filename}`;
```
mstykow, line 39: #nit here the url is extracted but in line 16 it was not. i would just always inline it.

tools/downloadOpossumFile.mjs
```diff
@@ -3,48 +3,90 @@
...
+        if (err) {
+          reject(err);
+        } else {
+          resolve();
+        }
+      });
+    });
+    console.log(`File downloaded and saved to ${savepath}`);
+  } catch (err) {
+    console.error(`Error downloading file: ${err.message}`);
```
mstykow, line 56: 1. not entirely true: the file may have downloaded but the write stream may have failed. another example of why i would drop the try/catch.
2. you're not re-raising the error. so if `downloadBinary` fails, the program just keeps running.

tools/downloadOpossumFile.mjs
```diff
@@ -3,48 +3,90 @@
...
+    await new Promise((resolve, reject) => {
+      pipeline(res.body, fs.createWriteStream(savepath), (err) => {
+        if (err) {
+          reject(err);
+        } else {
+          resolve();
+        }
+      });
+    });
+    console.log(`File downloaded and saved to ${savepath}`);
```
mstykow, line 54: this log offers no information that the log in line 86 doesn't already provide.
a similar thing is happening in the error case where you log twice that downloading failed.
in general, i would use fewer logs.

tools/downloadOpossumFile.mjs
```diff
@@ -3,48 +3,90 @@
...
+    console.info(`Creating folder ${folder}`);
+    fs.mkdirSync(folder);
+  }
+  if (fs.existsSync(path)) {
+    console.info("Deleting existing 'opossum-file'.");
+    fs.rmSync(path);
+  }
+}
+
+async function downloadOpossumFile(osSuffix, headers = undefined) {
```
mstykow, line 72: this function is never called without headers and yet you provide a default arg. also: why not construct the headers as part of the function instead of passing them from outside. it just seems arbitrary to me right now why some things are passed from the outside while other things such as the target are constructed inside.
i also think there are too many small functions calling each other in this script in general. this causes a lot of jumping up and down when i'd rather just like to read your (already quite small) script like a page in a book. but that's probably just my personal preference.

mstykow (changes requested):

abraemer:
I restructured the file putting everything into a single function and removed most of the error handling/logging. 
The linter is satisfied now after adding `fetch: true` to the config as suggested.

tools/downloadOpossumFile.mjs
```diff
@@ -3,66 +3,72 @@
...
 // SPDX-License-Identifier: Apache-2.0
 import fs from 'fs';
-import { fetchLatest } from 'gh-release-fetch';
-import { join } from 'path';
+import { dirname } from 'path';
+import { env } from 'process';
+import { pipeline } from 'stream';
 
 const EXECUTE_PERMISSIONS = 0o755;
+const HTTP_OK = 200;
```
mstykow, line 11: unused now

abraemer: removed

mstykow (approved):

Issue #2768, opened 2025-02-10
e2e test pipeline doesn't work properly on macos-latest runners


PhilippMa:
**Describe the bug**
When running the "Build and E2E Test" action on macos-latest, the run seems to not terminate (manually cancelled after running for more than 20 minutes).

**To Reproduce**

1. On a new branch: change the macos version in build-and-e2e-test.yml from macos-13 to macos-latest
2. Create a pull request for that branch to trigger the Github action to run.

**Expected behavior**
The "Build and E2E Test" action should terminate after ~5-10 minutes, since that is the amount of time it takes on windows-latest and ubuntu-latest runners.

**Additional context**
Could possibly be related to the fact that macos-latest runners have lower RAM than macos-13 runners, see https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories.

Issue #2767, opened 2025-02-10
Publish OpossumUI releases for both MacOS architectures


PhilippMa:
**Is your feature request related to a problem? Please describe.**
The OpossumUI release for MacOS currently only supports the ARM64 architecture, not the older Intel architecture.

**Describe the solution you'd like**
The Github action for releases should also build and publish an OpossumUI binary for MacOS with Intel architecture.

**Describe alternatives you've considered**
-

**Additional context**
-

PR #2770, opened 2025-02-10
Architecture-aware MacOS build and test

### LINKED ISSUE #2766: Choose correct MacOS binary during bundling
opossum-file will soon publish two different MacOS executables for Intel and ARM64 architectures, but the OpossumUI bundler can currently handle only one executable per OS.

On MacOS, the OpossumUI bundler needs to download the correct executable based on the local architecture.

\-

**Additional context**
\-

PhilippMa:
### Summary of changes

Adds proper support for both Intel and ARM64 based MacOS systems.

* bundle correct opossum-file binary based on architecture
* invoke electron-builder with the correct architecture cli switch
* adjust Github actions for e2e tests and release to run on both macos-13 and macos-latest

For MacOS systems, we need to test and build OpossumUI targeting the correct architecture to ensure that it works correctly.

* An opossum-file binary built for the wrong architecture cannot be properly executed
* Giving the wrong architecture cli switch to electron-builder was what caused the issue described in #2768 

* The adjusted Github actions will already be running on this PR
* The publish-release action could be tested by creating a temporary pre-release with this branch as its target

closes #2766, #2767, #2768


package.json
```diff
@@ -131,19 +131,22 @@
     "update-commit-hash:darwin:linux": "COMMIT_INFO=$(git describe --exact-match --tags 2> /dev/null || git rev-parse --short HEAD); echo \"{\\\"commitInfo\\\" : \\\"$COMMIT_INFO\\\" }\" > \"src/commitInfo.json\"",
     "update-commit-hash:win32": "tools/get_app_version_for_windows.bat",
     "generate-notice": "mkdirp notices && yarn licenses generate-disclaimer --production > notices/notices.txt && yarn node tools/generateNotices.mjs && ./tools/checkNotices.sh",
+    "run-script-arch": "yarn node tools/runScriptArch.mjs",
```
mstykow, line 134: what's the purpose of this new package.json-script? we could just inline this in line 141, i.e.,
```
"ship:auto:darwin": "yarn node tools/runScriptArch.mjs ship-mac",
```
this means one indirection less.

mstykow (changes requested):
As discussed, there is an issue with code signing.

package.json
PhilippMa: Good point, this was meant to make the command definitions more readable when I use runScriptArch in multiple commands, but that never happened.

PR #2771, opened 2025-02-11
Improve error reporting for import with opossum-file

### LINKED ISSUE #2719: Error handling of including SPDX files
* on exit of python script show error
* Investigate what can/should be shown
* Might need a lib.py story for better exit codes/error reporting
* delete all temp files -- does that exist?
* User back to main screen

PhilippMa:
### Summary of changes

Clean up error reporting in import dialog and report errors occurring during file conversion with opossum-file.

Users should be properly informed about potential errors that occurr during the import process.

Try to import an invalid file through the import ScanCode Json functionality.

closes #2719

### DIFF
--- a/src/ElectronBackend/main/listeners.ts
+++ b/src/ElectronBackend/main/listeners.ts
@@ -29,6 +29,7 @@ import { LoadedFileFormat } from '../enums/enums';
 import {
   createListenerCallbackWithErrorHandling,
   createVoidListenerCallbackWithErrorHandling,
+  ListenerErrorReporter,
 } from '../errorHandling/errorHandling';
 import { loadInputAndOutputFromFilePath } from '../input/importFromFile';
 import { serializeAttributions } from '../input/parseInputData';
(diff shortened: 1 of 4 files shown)


src/ElectronBackend/main/listeners.ts
```diff
@@ -239,6 +236,7 @@ export function getImportFileConvertAndLoadListener(
 
       return true;
     },
+    ListenerErrorReporter.Frontend,
```
abraemer, line 239: This is the only place where you use this new error handler right now. Do you expect to have more usage for it in future commits? If not, then I'd say, you could just do the conversion error handling right here in this function and only have minimal changes in the rest of the project (i.e. no new error handlers, convertScancodeToOpossum doesn't need to repackage the error)

abraemer (changes requested):

src/ElectronBackend/opossum-file/convertScancodeToOpossum.ts
```diff
@@ -19,11 +19,15 @@ export async function convertScancodeToOpossum(
...
+  try {
+    await execFile(OPOSSUM_FILE_EXECUTABLE, [
+      'generate',
+      '-o',
+      pathToOpossum,
+      '--scan-code-json',
+      pathToScanCode,
+    ]);
+  } catch (error) {
+    throw new Error('Conversion of ScanCode file to .opossum file failed');
```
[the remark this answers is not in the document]
PhilippMa: Yup, just giving a nicer error message than the exit code

src/ElectronBackend/main/listeners.ts
PhilippMa: As discussed, I expect error messages to be shown directly in the frontend more often in the future. The introduction of these error reporters allows separation of raising errors and the methods of how they are displayed to the users.

abraemer (approved):

PR #2780, opened 2025-02-12
chore: enable dotenv github token

mstykow:
### Summary of changes

This PR enables setting the GitHub token via a .env file. This allows setting a permanent GitHub token for downloading releases without rate limit.

Get a GH token by running `gh auth token` and set it in the .env file. Then run `yarn install-deps` and notice that there is no rate-limit warning.


tools/generateNotices.mjs
abraemer, line 1: What's up with these 4 files in `tools/`? Why do they show up empty?

abraemer: Ah I think you changed the permissions of these. Was that intended?

abraemer (approved):
LGTM
Just 1 minor question.

PR #2782, opened 2025-02-13
Check for unsaved changes before import/open/export

### LINKED ISSUE #2775: Show save dialog before open/import/merge/export
Open/import/merge/export actions can lead to changes being lost or otherwise incorrect results when they are executed while the currently open file has unsaved changes.

The NotSavedPopup should be opened when starting any of the above workflows. If the user clicks on "Discard and Proceed" then the workflow continues as before, if the user clicks cancel then the workflow is cancelled.

Another option would be to directly give the user the option to save their changes from this dialog. However, the currently existing Popups in OpossumUI do not yet support this feature.

PhilippMa:
### Summary of changes

If unsaved changes are present when the user clicks on import/open/export, the NotSavedPopup is shown and they can only proceed if they explicitly discard their current changes.

Importing or opening a file while there are unsaved changes implicitly discards these changes. Exported documents do not reflect unsaved changes.

Change any field of an attribution, then click on import/open/export.

closes #2775


src/Frontend/util/attribution-utils.ts
```diff
@@ -16,7 +18,7 @@ import {
   isIdOfResourceWithChildren,
 } from './can-resource-have-children';
 
-export function getAttributionsWithResources(
+export function attributionUtils(
```
mstykow, line 21: this does not seem like a good name for a function that returns attributions

PhilippMa: good catch, I just wanted to rename the file and apparently didn't unselect that webstorm would also rename the function

src/Frontend/state/actions/popup-actions/popup-actions.ts
```diff
@@ -119,9 +144,49 @@ export function navigateToTargetResourceOrAttributionOrOpenFileDialog(): AppThun
...
     dispatch(setOpenFileRequest(false));
+    dispatch(setImportFileRequest(null));
+    dispatch(setExportFileRequest(null));
+  };
+}
+
+export function showImportDialogWithUnsavedCheck(
+  fileFormat: FileFormatInfo,
+): AppThunkAction {
+  return (dispatch, _) => {
```
mstykow, line 159: #nit
```suggestion
  return (dispatch) => {
```

src/Frontend/state/actions/popup-actions/popup-actions.ts
```diff
@@ -119,9 +144,49 @@ export function navigateToTargetResourceOrAttributionOrOpenFileDialog(): AppThun
...
+    dispatch(
+      withUnsavedCheck(
+        () =>
+          dispatch(openPopup(PopupType.ImportDialog, undefined, fileFormat)),
+        () => dispatch(setImportFileRequest(fileFormat)),
+      ),
+    );
+  };
+}
```
mstykow, line 168: something weird is going on here: you're wrapping `withUnsavedCheck` in a dispatch but then again calling `dispatch` on `showImportDialogWithUnsavedCheck` where you're using it. this is not how you use, e.g., `navigateToSelectedPathOrOpenUnsavedPopup` which is not wrapped in a dispatch.
so shouldn't this just be:
```suggestion
  return withUnsavedCheck(
        (dispatch) =>
          dispatch(openPopup(PopupType.ImportDialog, undefined, fileFormat)),
        (dispatch) => dispatch(setImportFileRequest(fileFormat)),
      );
}
```
same questions about the actions below.

src/Frontend/util/__tests__/attribution-utils.test.ts
```diff
@@ -438,3 +440,68 @@ describe('removeSlashesFromFilesWithChildren', () => {
     ).toEqual(expectedAttributionsWithResources);
   });
 });
+
+describe('getBomAttributions', () => {
+  it('filters the correct BOM attributions', () => {
```
mstykow, line 445: could you please split this rather long test up into one test per part. having multiple setups and asserts as part of a single test makes it harder to find out where things actually go wrong when the test fails.

src/e2e-tests/__tests__/import-dialog.test.ts
```diff
@@ -37,11 +37,13 @@ test('opens, displays and closes import dialog', async ({
   await importDialog.assert.titleIsHidden();
 });
 
-test('imports legacy opossum file', async ({
+test('imports legacy opossum file and works with NotSavedPopup', async ({
```
mstykow, line 40: please keep test titles free from implementation details such as component names. focus on describing behavior rather than referencing names. names change quickly and then it's hard to remember to update the test title.

mstykow (changes requested):

src/Frontend/state/actions/popup-actions/popup-actions.ts
Hellgartner, line 1: TBH for me the whole pattern seems to be flawed. Why do we need a dedicated state flag for each and every method which at the end also needs to be reset in the close method?

src/Frontend/state/actions/resource-actions/export-actions.ts
```diff
@@ -0,0 +1,140 @@
...
+  getResources,
+} from '../../selectors/resource-selectors';
+import { AppThunkAction } from '../../types';
+
+export function exportFile(exportType: ExportType): AppThunkAction {
+  return (_, getState) => {
+    switch (exportType) {
+      case ExportType.SpdxDocumentJson:
+      case ExportType.SpdxDocumentYaml:
+        return getSpdxDocumentExportListener(getState(), exportType);
```
Hellgartner, line 34: Compared to where we came from, the state handling changed. 
* Before: useSelector was used in a reactive context
* Now: we are explicitly getting the state at the time the function is called

Unsure but might this be a cause for subtle bugs, especially re-renders which no longer happen?

src/Frontend/state/actions/resource-actions/export-actions.ts
Hellgartner, line 1: As this has quite some logic I would prefer to have a test covering this

src/e2e-tests/__tests__/import-dialog.test.ts
```diff
@@ -61,6 +63,15 @@ test('imports legacy opossum file', async ({
...
   await resourcesTree.assert.resourceIsVisible(resourceName);
+
+  const comment = faker.lorem.sentences();
+  await resourcesTree.goto(resourceName);
+  await attributionDetails.attributionForm.comment.fill(comment);
+
+  await menuBar.openImportLegacyOpossumFile();
+  await notSavedPopup.assert.isVisible();
+  await notSavedPopup.discardButton.click();
+  await importDialog.assert.titleIsVisible();
```
Hellgartner, line 74: Do we need to check that this really passes through, i.e. really imports at the end?

src/e2e-tests/__tests__/import-dialog.test.ts
Hellgartner, line 1: Should we also cover the new flows for export/open?

Hellgartner (changes requested):

src/Frontend/Components/GlobalPopup/GlobalPopup.tsx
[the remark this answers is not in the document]
PhilippMa: As discussed, there doesn't really exist a good replacement for this yet. Getting rid of the GlobalPopup mechanism would require a larger refactoring that would have to be done in a separate issue.

src/Frontend/Components/GlobalPopup/GlobalPopup.tsx
```diff
@@ -25,6 +26,10 @@ function getPopupComponent(popupInfo: PopupInfo | null) {
       return <ProjectStatisticsPopup />;
     case PopupType.UpdateAppPopup:
       return <UpdateAppPopup />;
+    case PopupType.ImportDialog:
```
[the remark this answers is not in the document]
PhilippMa: As above, I would consider this out of scope for this PR

src/Frontend/Components/ImportDialog/ImportDialog.tsx
```diff
@@ -74,7 +74,7 @@ export const ImportDialog: React.FC<ImportDialogProps> = ({
   }
 
   function onCancel(): void {
-    closeDialog();
+    dispatch(closePopup());
```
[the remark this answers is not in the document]
PhilippMa: Before, the ImportDialog only needed to be opened when receiving an IPC message from the backend, which could be handled "locally" by the ImportDialogProvider. However, as there are now situations where the ImportDialog has to be opened as a result of clicking "proceed" in the NotSavedPopup, we need a more global way of managing the open state of the ImportDialog, i.e. GlobalPopup (unfortunately)

src/Frontend/state/actions/popup-actions/__tests__/popup-actions.test.ts
```diff
@@ -303,66 +313,120 @@ describe('The actions checking for unsaved changes', () => {
...
+        name: '',
+      }),
+    );
+    testStore.dispatch(openPopup(PopupType.NotSavedPopup));
+    testStore.dispatch(proceedFromUnsavedPopup());
+
+    expect(getOpenPopup(testStore.getState())?.popup).toBe(
+      PopupType.ImportDialog,
+    );
+    expect(getImportFileRequest(testStore.getState())).toBeNull();
```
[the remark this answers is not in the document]
PhilippMa: good point, I updated the test

src/Frontend/state/actions/popup-actions/popup-actions.ts
PhilippMa: I totally agree (and maxim also sees it this way I think), I would also prefer just passing a continuation callback to the NotSavedPopup which it could then execute when the proceed button is clicked, but unfortunately this would likely again require quite a large refactoring and would potentially also require getting rid of GlobalPopup, so for now it seems that this pattern will have to stay until we can afford to do a larger refactoring in the future.

mstykow (approved):

src/Frontend/state/actions/resource-actions/export-actions.ts
PhilippMa: This is how I understand it:

Before, state was read in the BackendCommunication component using useAppSelector and stored in local variables. When state changed, the component was "rerendered" (or reevaluated, since it doesn't actually return any JSX), so the variables would be kept up to date. Then, these local variables were used in the export functions and the main export function was registered as an IPC listener, with the local variables listed as dependencies so that it would get re-registered every time one of the local variables changed in value.
Now, the main export function is registered only once, and it is implemented as an AppThunkAction which always has access to the current state through its getState parameter.

I think these two approaches should be equivalent, or at least I can't think of any situation where they would lead to different behavior.

src/e2e-tests/__tests__/import-dialog.test.ts
PhilippMa: My goal was to make sure that the workflow actually continues with opening the import dialog when clicking "Discard & Proceed" in the NotSavedPopup here, especially because the logic for continuing execution after the NotSavedPopup is closed is quite easy to break currently, because (as discussed above) the patterns employed here aren't great.

src/Frontend/state/actions/resource-actions/export-actions.ts
Hellgartner: Agreed so far. So we do have to pray that nothing else implicitly relied on the re-rendering. But probably not much we can do here

src/ElectronBackend/main/menu.ts
```diff
@@ -155,8 +152,6 @@ export async function createMenu(mainWindow: BrowserWindow): Promise<Menu> {
                 'icons/follow-up-black.png',
               ),
               click: () => {
-                setLoadingState(mainWindow.webContents, true);
-                logger.info('Preparing data for follow-up export');
                 webContents.send(
```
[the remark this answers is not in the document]
PhilippMa: I was willing to 'sacrifice' the log messages in order to avoid the (actually not so small) work it would require to send them after the NotSavedPopup, but I definitely missed that it's important to set the loading state properly. So I ended up putting both back in, now triggered from export-actions.ts

src/Frontend/state/actions/resource-actions/export-actions.ts
PhilippMa: This logic is not new (I just moved it), so I would say that writing tests for it is out of scope for this PR (especially considering that I have no idea what these functions do in detail). I opened a new issue for this instead at #2795

src/e2e-tests/__tests__/import-dialog.test.ts
PhilippMa: I discovered that there was already a some tests for warning about unsaved changes in updating-attributions.test.ts, so I moved the import case here and added/updated test cases for export and open there.

src/Frontend/Components/ProcessPopup/__tests__/ProcessPopup.test.tsx
```diff
@@ -2,84 +2,51 @@
...
-      () =>
-        void electronAPI.send(AllowedFrontendChannels.FileLoading, {
-          isLoading: true,
         }),
-    );
+      ],
+    });
+
+    store.dispatch(setLoading(false));
+    rerender(popup);
```
Hellgartner, line 47: What worries me a bit here is that the re-render is triggered manually while it should be triggered by the state change. This might hide a bug.
Does that work without the rerender?

src/Frontend/state/actions/resource-actions/export-actions.ts
```diff
@@ -0,0 +1,166 @@
...
+import { setLoading, writeInfoLogMessage } from '../view-actions/view-actions';
+
+export function exportFile(exportType: ExportType): AppThunkAction {
+  return (dispatch, getState) => {
+    dispatch(setLoading(true));
+
+    switch (exportType) {
+      case ExportType.SpdxDocumentJson:
+        dispatch(writeInfoLogMessage('Preparing data for SPDX (json) export'));
+        exportSpdxDocument(getState(), ExportType.SpdxDocumentJson);
```
Hellgartner, line 37: my gut feeling tells me to move the logging into the e.g. `exportSpdxDocument` functions but no strong feelings here

src/Frontend/Components/ProcessPopup/__tests__/ProcessPopup.test.tsx
PhilippMa: I do not know if this is an issue with the way we set up our tests or if this is a general problem, but changing the state (i.e. dispatching an action on the store) after the initial render does not automatically trigger a rerender in the tests for some reason, so the test does not work without manual rerendering. When actually running the app, changes to the loading state do properly trigger the component to rerender.

Hellgartner (approved):

src/Frontend/Components/ProcessPopup/__tests__/ProcessPopup.test.tsx
PhilippMa: Yes that works, thanks!

PR #2784, opened 2025-02-13
Build all MacOS releases on macos-latest runners

PhilippMa:
### Summary of changes

On release, the build for macos with intel architecture is now also executed on a macos-latest runner.

Electron-builder is capable of cross-compilation and building on maco-latest runners is faster on potentially more future proof than macos-13 runners.

Create a temporary release targeting this branch.

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -51,6 +51,7 @@ jobs:
           files: ${{ matrix.FILES }}
 
       - name: Upload user guide
+        continue-on-error: true
         uses: softprops/action-gh-release@v2
         with:
           files: USER_GUIDE.md
@@ -59,6 +60,7 @@ jobs:
         run: yarn spdx
 
       - name: Upload SPDX document
+        continue-on-error: true
         uses: softprops/action-gh-release@v2
         with:
           files: opossum-ui.spdx.json
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -17,7 +17,6 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest, macos-latest, windows-latest]
         include:
           - os: ubuntu-latest
             build-command: yarn ship:auto
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -17,17 +17,21 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-13, macos-latest, windows-latest]
+        os: [ubuntu-latest, macos-latest, macos-latest, windows-latest]
         include:
           - os: ubuntu-latest
+            build-command: yarn ship:auto
             FILES: |
               release/linux/OpossumUI-for-linux.AppImage
               release/linux/OpossumUI-for-linux.snap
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-release.yml
```diff
@@ -40,14 +43,15 @@ jobs:
...
-      - run: yarn ship:auto
+      - run: ${{ matrix.build-command }}
 
       - name: Upload release asset
         uses: softprops/action-gh-release@v2
         with:
           files: ${{ matrix.FILES }}
 
       - name: Upload user guide
+        continue-on-error: true
```
mstykow, line 54: i think we can do better: let's extract the upload of user guide and SBOM to a separate job in the same workflow. that is, this job will only be responsible for uploading those two files, run only on ubuntu-latest, and there will never be a case where two runners upload the same file multiple times, regardless of which type of runner they are.

mstykow (changes requested):

.github/workflows/publish-release.yml
```diff
@@ -17,17 +17,20 @@ jobs:
...
+            build-command: yarn ship:auto
             FILES: |
               release/linux/OpossumUI-for-linux.AppImage
               release/linux/OpossumUI-for-linux.snap
-          - os: macos-13
+          - os: macos-latest
+            build-command: yarn ship-mac:x64
             FILES: release/mac/OpossumUI-for-mac-intel.zip
           - os: macos-latest
+            build-command: yarn ship:auto
```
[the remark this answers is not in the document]
PhilippMa: agreed

.github/workflows/publish-release.yml
PhilippMa: good idea, it was weird anyways that these documents were uploaded multiple times by different runners

.github/workflows/publish-release.yml
```diff
@@ -40,13 +43,25 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-node@v4
+        with:
+          cache: 'yarn'
+          node-version: 22
+
+      - run: yarn install --immutable
```
mstykow, line 63: do you even need to install the dependencies? it seems like for the upload you don't need anything.

.github/workflows/publish-release.yml
```diff
@@ -40,13 +43,25 @@ jobs:
...
+
       - name: Upload user guide
         uses: softprops/action-gh-release@v2
```
mstykow, line 66: instead of calling the upload action twice, let's combine this into a single command (see the documentation for `action-gh-release`. in that case, you just need to make sure you run `yarn spdx` before running the upload.

mstykow (changes requested):

.github/workflows/publish-release.yml
PhilippMa: I just tried it, but without yarn install the yarn spdx commands produces an error:
`Internal Error: ENOENT: no such file or directory, open '/home/runner/work/OpossumUI/OpossumUI/node_modules/.yarn-state.yml'`

.github/workflows/publish-release.yml
PhilippMa: done

mstykow (approved):

PR #2794, opened 2025-02-14
Refactor/menu

Hellgartner:
## Summary of changes

Splitting up the menu.ts file

## Context and reason for change

Reason: It got too large over time

## How can the changes be tested

Double check the menu
This PR should not NOT introduce any functional changes


src/ElectronBackend/main/menu/editMenu.ts
```diff
@@ -0,0 +1,153 @@
...
+import { MenuItemConstructorOptions } from 'electron';
+
+import { AllowedFrontendChannels } from '../../../shared/ipc-channels';
+import { isFileLoaded } from '../../utils/getLoadedFile';
+import { getGlobalBackendState } from '../globalBackendState';
+import { getIconBasedOnTheme } from '../iconHelpers';
+import { INITIALLY_DISABLED_ITEMS_INFO } from './initiallyDisabledMenuItems';
+
+const UNDO: MenuItemConstructorOptions = {
+  icon: getIconBasedOnTheme('icons/undo-white.png', 'icons/undo-black.png'),
```
PhilippMa, line 15: getIconBasedOnTheme uses the electron.nativeTheme property which is only available once electron has finished initialization, however these const definitions are evaluated before that, so the icons currently always use the black icons regardless of the theme. I would suggest fixing this by changing these constants into functions that get called from getEditMenu. Same goes for the menu entry constants in the other menu files

PhilippMa (changes requested):

src/ElectronBackend/main/menu/editMenu.ts
Hellgartner: Nice catch !
Could you retest?
I somehow cannot do so on my machine

PhilippMa (approved):

src/ElectronBackend/main/__tests__/menu.test.ts
```diff
@@ -0,0 +1,58 @@
...
+  },
+  Menu: {
+    buildFromTemplate: jest.fn(),
+  },
+}));
+
+jest.mock('electron-settings');
+
+const getUsedIcons = (
+  menuImput: Array<MenuItemConstructorOptions>,
```
PhilippMa, line 23: typo: menuInput

src/ElectronBackend/main/__tests__/menu.test.ts
```diff
@@ -0,0 +1,58 @@
...
+      );
+    } else if (menuItem.icon) {
+      result.push(menuItem.icon as string);
+    }
+  });
+  return result;
+};
+
+describe('create menu', () => {
+  it('needs to evaluate dark and light mode properly', async () => {
```
PhilippMa, line 39: This test currently only tests icon selection in dark mode. I think it would make sense to test light mode as well

PhilippMa (approved):

Issue #2730, opened 2025-01-30
Merging of input files


abraemer:
Expose merge functionality in OpossumUI.
Here are some ideas how this could work:
* Support opening an "additional" file the contents of which are then simply merged into the currently open document
* Extend the import dialog with the possibility of selecting multiple files (either in one dialog but then they'd need to be of the same kind) or by letting the user add additional fields for more input files.

PR #2796, opened 2025-02-17
Merging of input files (Backend logic)

### LINKED ISSUE #2797: Merging of input files (Backend logic)
Next to the function for converting files with opossum-file, we need a new function for merging two files with opossum-file.

PhilippMa:
### Summary of changes

Added a new function for merging a second file into an existing .opossum file using opossum-file. Organized the opossum-file related code into a class hierarchy to better deal with the growing "dimensions" (convert/merge, legacy/scancode/owasp) of provided functionality.

Preparation for full integration of merging input files in OpossumUI (#2730)

Unit tests for the new functionality are in opossum-file.test.ts

closes #2797


src/Frontend/shared-constants.ts
```diff
@@ -73,3 +73,7 @@ export const thirdPartyKeys: Array<keyof PackageInfo> = [
   'licenseName',
   'licenseText',
 ];
+
+export const outputFileEnding = '_attributions.json';
```
abraemer, line 77: maybe `legacyOutputFileEnding`? I was confused which output file is meant when I first found it in the code

src/Frontend/shared-constants.ts
```diff
@@ -73,3 +73,7 @@ export const thirdPartyKeys: Array<keyof PackageInfo> = [
...
+export const jsonGzipFileExtension = '.json.gz';
+export const jsonFileExtension = '.json';
```
abraemer, line 79: I don't think these add much value and would prefer deleting them.
Note that both strings also appear in another location where this constant is not used.

src/ElectronBackend/opossum-file/LegacyFileConverter.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  private tryToGetInputFileFromOutputFile(filePath: string): string {
+    const outputFilePattern = `(${outputFileEnding})$`;
+    const outputFileRegex = new RegExp(outputFilePattern);
+
+    return fs.existsSync(filePath.replace(outputFileRegex, jsonFileExtension))
+      ? filePath.replace(outputFileRegex, jsonFileExtension)
+      : fs.existsSync(filePath.replace(outputFileRegex, jsonGzipFileExtension))
+        ? filePath.replace(outputFileRegex, jsonGzipFileExtension)
+        : filePath;
```
abraemer, line 31: I think this would be easier to read/understand if written a bit more explicitly:
```ts
const outputAsJson = filePath.replace(outputFileRegex, jsonFileExtension);
if (fs.existsSync(outputAsJson)) return outputAsJson;
const outputAsJsonGzip = filePath.replace(outputFileRegex, jsonGzipFileExtension);
if (fs.existsSync(outputAsJsonGzip)) return outputAsJsonGzip;
return filePath;
```

src/ElectronBackend/opossum-file/LegacyFileConverter.ts
```diff
@@ -0,0 +1,96 @@
...
+} from '../../Frontend/shared-constants';
+import { writeOpossumFile } from '../../shared/write-file';
+import { getFilePathWithAppendix } from '../utils/getFilePathWithAppendix';
+import { FileConverter } from './FileConverter';
+
+export class LegacyFileConverter extends FileConverter {
+  readonly fileTypeName: string = 'Legacy Opossum';
+  readonly fileTypeSwitch: string = '--opossum';
+
+  private tryToGetInputFileFromOutputFile(filePath: string): string {
```
abraemer, line 23: This name does not make a lot of sense to me. Especially when it is called (below) with `pathToInputFile`

src/ElectronBackend/opossum-file/LegacyFileConverter.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  override async convertFile(
+    pathToInputFile: string,
+    pathToOpossumFile: string,
+  ): Promise<void> {
+    let pathToInputJson = pathToInputFile;
+
+    if (pathToInputFile.endsWith(outputFileEnding)) {
+      pathToInputJson = this.tryToGetInputFileFromOutputFile(pathToInputFile);
+    }
```
abraemer, line 70: Is the logic of this function correct? It seems that it accepts either a "legacy input" file (just a json possibly gzipped) or a "legacy output" file (which seems to be `*_attributions.json` file). If it is the latter, then it tries to find the accompanied "legacy input" file. However if it doesn't find that, then it will use the "legacy output" file both as input and output file for the new opossum? Does that make sense? Would an error be more reasonable if we can't find the "legacy input" file?

src/ElectronBackend/opossum-file/FileConverter.ts
```diff
@@ -0,0 +1,59 @@
...
+    pathToOpossumFile: string,
+  ): Promise<void>;
+
+  async mergeFiles(
+    pathToInputFile: string,
+    pathToOpossumFile: string,
+  ): Promise<void> {
+    try {
+      const pathToPreConvertedInputFile =
+        await this.preConvertInputFile(pathToInputFile);
```
abraemer, line 38: I think it might make sense to move this out of the try catch. Because then errors inside preConvert can have separate messages (I am thinking of something like "legacy input file could not be found")

src/ElectronBackend/opossum-file/FileConverter.ts
```diff
@@ -0,0 +1,59 @@
...
+  ): Promise<string | null>;
+
+  abstract convertFile(
+    pathToInputFile: string,
+    pathToOpossumFile: string,
+  ): Promise<void>;
+
+  async mergeFiles(
+    pathToInputFile: string,
+    pathToOpossumFile: string,
```
abraemer, line 34: Something that confused me with these function signatures is the slightly different meaning of `pathToOpossumFile`. In `convertFile` it means the save location.
In `mergeFiles` it is both the second input and the save location.
I think it might be clearer to make the naming a bit longer:
`convertFile(pathToInputFile, pathToOpossumFile)` -> `convertFileToOpossum(pathToInputFile, pathToSaveLocation)`
`mergeFiles(pathToInputFile, pathToOpossumFile)` -> 
```js
async mergeFileWithOpossum(pathToInputFile, pathToOpossumFile, pathToSaveLocation=undefined) {
pathToSaveLocation = pathToSaveLocation ? pathToSaveLocation : pathToOpossumFile
...
```

abraemer (changes requested):

src/ElectronBackend/opossum-file/LegacyFileConverter.ts
```diff
@@ -0,0 +1,96 @@
...
+  ): Promise<string | null> {
+    let tempFilePath;
+    try {
+      tempFilePath = path.join(app.getPath('temp'), 'temp.opossum');
+    } catch (error) {
+      // When executing as part of unit tests, app.getPath('temp') throws an error
+      tempFilePath = path.join(__dirname, 'temp.opossum');
+    }
+
+    console.log(`Temp file path: ${tempFilePath}`);
```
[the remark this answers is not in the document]
PhilippMa: Good catch, that was a leftover debug statement

src/Frontend/shared-constants.ts
PhilippMa: :+1:

src/ElectronBackend/opossum-file/LegacyFileConverter.ts
PhilippMa: Good point

src/ElectronBackend/opossum-file/LegacyFileConverter.ts
PhilippMa: There were two different notions of 'input file' here, I renamed one of them to make this a bit more clear

src/ElectronBackend/opossum-file/LegacyFileConverter.ts
PhilippMa: Good question, this is preexisting logic and I honestly don't know if it makes any sense to continue with only the legacy output here

src/ElectronBackend/opossum-file/FileConverter.ts
PhilippMa: For `convertFile` I agree. for `mergeFiles`, how about renaming it to `mergeFileIntoOpossum` and keeping it with only two parameters? Otherwise we end up adding a third parameter that is currently never used

abraemer (approved):

PR #2800, opened 2025-02-18
feat: update input file schema to include configurable classifications for signals

abraemer:
### Summary of changes

Introduces a new field on `externalAttributions` called `classification`. It is a positive integer and its meaning (i.e. text label) can be configured via a new mapping `classifications`. In anticipation of further configurable items in the future, a new top-level key `config` is introduced with `classification` being its only subkey currently.

We want to replace criticality by something more universally usable. Thus we introduce its successor in this PR and will build on this is in following work.

No further changes are required at this point. Opening files with OpossumUI and saving them preserves the `classification` information out-of-the-box since the input file is never changed.

#### Compatibility
This change is backwards compatible: new versions of OpossumUI can open old opossum files. However, new opossum files (with config and/or classification) cannot be opened by old OpossumUI releases.

There should be no functional changes because we just added optional fields. Everything that worked before still works.

### DIFF
--- a/src/ElectronBackend/input/OpossumInputFileSchema.json
+++ b/src/ElectronBackend/input/OpossumInputFileSchema.json
@@ -11,6 +11,18 @@
   ],
   "additionalProperties": false,
   "properties": {
+    "config": {
+      "type": "object",
+      "description": "Additional configuration for OpossumUI.",
+      "additionalProperties": true,
+      "properties": {
+        "classifications": {
+          "type": "object",
+          "description": "Mapping of classification values to human-readable labels. See \"classification\" field of \"externalAttribution\".",
+          "additionalProperties": true
+        }
+      }
+    },
     "resources": {
       "id": "#ResourcesSchema",
       "type": "object",
@@ -120,6 +132,11 @@
           "criticality": {
             "type": "string",
             "description": "Indication on how critical a signal is. Possible values are \"high\" and \"medium\"."
+          },
+          "classification": {
+            "type": "number",
+            "description": "Signal classification by usage, usually based on the license. Zero corresponds to unrestricted usage ehile higher levels mean more and more restrictions. Exact values and labels need to be configured in the config section.",
+            "minimum": 0
           },
           "copyright": {
             "type": "string",
--- a/src/ElectronBackend/input/OpossumInputFileSchema.json
+++ b/src/ElectronBackend/input/OpossumInputFileSchema.json
@@ -13,13 +13,17 @@
   "properties": {
     "config": {
       "type": "object",
-      "description": "Additional configuration for OpossumUI.",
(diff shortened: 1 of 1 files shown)


src/ElectronBackend/input/OpossumInputFileSchema.json
```diff
@@ -121,6 +133,11 @@
             "type": "string",
             "description": "Indication on how critical a signal is. Possible values are \"high\" and \"medium\"."
           },
+          "classification": {
+            "type": "number",
+            "description": "Signal classification by usage, usually based on the license. Zero corresponds to unrestricted usage ehile higher levels mean more and more restrictions. Exact values and labels need to be configured in the config section.",
```
Hellgartner, line 138: typo: ehile --> while

src/ElectronBackend/input/OpossumInputFileSchema.json
```diff
@@ -11,6 +11,18 @@
   ],
   "additionalProperties": false,
   "properties": {
+    "config": {
+      "type": "object",
+      "description": "Additional configuration for OpossumUI.",
```
vasily-pozdnyakov, line 16: let's keep it simple "Configuration for OpossumUI"

src/ElectronBackend/input/OpossumInputFileSchema.json
```diff
@@ -11,6 +11,18 @@
...
+      "additionalProperties": true,
+      "properties": {
+        "classifications": {
+          "type": "object",
+          "description": "Mapping of classification values to human-readable labels. See \"classification\" field of \"externalAttribution\".",
+          "additionalProperties": true
```
vasily-pozdnyakov, line 22: this is not necessary

vasily-pozdnyakov (changes requested):

vasily-pozdnyakov:
I also generated a new .opossum file with the fields and it works fine with your changes.

src/ElectronBackend/input/OpossumInputFileSchema.json
abraemer: I read a bit more about JSON Schema: We could make this more restrictive by using `patternProperties`:
```json
    "classifications": {
          ....
          "patternProperties": {
            "^[0-9]+$": {
              "type":"string"
            }
    }
```
What do you think about that?

vasily-pozdnyakov: I would propose to keep it simple for now. I also tried to test it with the pattern, and it does not work as expected.

vasily-pozdnyakov: As discussed, let's proceed this way.

vasily-pozdnyakov (approved):

PR #2810, opened 2025-02-24
Add classification to internal data model

PhilippMa:
### Summary of changes

Added classification to internal data model:
* Add field for classification config in resource state
* Add classification field on attributions

Upcoming tickets will need access to classification data included in .opossum files.

* Check that files can be parsed and opened without error
* (Optional) log parts of internal state to console after load has finished, check that classification data in internal state is correct


abraemer (approved):
LGTM

abraemer (approved):
LGTM

PR #2813, opened 2025-02-24
fix: config is optional

abraemer:
### Summary of changes

When loading an input file without `config` field, we replace the resulting `undefined` with an object that represents an empty config properly.

Since config is optional, we need to handle the case when it is not present in the opossum file

Not directly since nothing uses `getClassifications` right now. However my next PR cause errors on opossum files without `config` section.

### DIFF
--- a/src/ElectronBackend/types/types.ts
+++ b/src/ElectronBackend/types/types.ts
@@ -46,7 +46,7 @@ export interface RawFrequentLicense {
 export interface ParsedOpossumInputFile {
   metadata: ProjectMetadata;
   resources: Resources;
-  config: ProjectConfig;
+  config?: ProjectConfig;
   externalAttributions: RawAttributions;
   resourcesToAttributions: ResourcesToAttributions;
   frequentLicenses?: Array<RawFrequentLicense>;
--- a/src/ElectronBackend/input/importFromFile.ts
+++ b/src/ElectronBackend/input/importFromFile.ts
@@ -8,6 +8,7 @@ import fs from 'fs';
 import { cloneDeep } from 'lodash';
 import { v4 as uuid4 } from 'uuid';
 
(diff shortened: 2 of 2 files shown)


src/ElectronBackend/types/types.ts
```diff
@@ -46,7 +46,7 @@ export interface RawFrequentLicense {
 export interface ParsedOpossumInputFile {
   metadata: ProjectMetadata;
   resources: Resources;
-  config: ProjectConfig;
+  config?: ProjectConfig;
```
mstykow, line 49: this doesn't seem right: once you've parsed the opossum file, you're falling back to the empty project config so that the config here can no longer be nullable.

mstykow (changes requested):

src/ElectronBackend/types/types.ts
abraemer: Does seem correct to me but the situation is a bit confusing as we have very similar types:
- Once in `ElectronBackend/types/types.ts`: Here is `ParsedOpossumInputFile` with nullable `config` used as type directly after reading and parsing the file
- Additionally there is `shared/shared-types.ts` which has `ParsedFileContent` which is similar (but also contains information from the output file), has non-nullable `config` and is used as type to communicate the information from backend to frontend.

So I see 2 possible points where I could normalize a missing `ProjectConfig` into an empty one. Either:
- When we combine `ParsedOpossumInputFile` with `ParsedOpossumOutputFile` to `ParsedFileContent` and send that to the frontend (this is what I chose)
- Or, we during the `setConfig`-ResourceAction-pathway in the frontend, that it is normalized when the global state is updated
I think normalization is best done as early as possible. That's why I chose the former.

mstykow (approved):

Issue #2812, opened 2025-02-24
Move export preprocessing to backend


PhilippMa:
**Is your feature request related to a problem? Please describe.**
The data preprocessing currently happening in export-actions.ts is so CPU intensive that the UI gets unresponsive while it is running.

**Describe the solution you'd like**
The preprocessing step should be moved to the backend and executed there as part of the exportFileListener. To this end, either the raw redux state data required has to be sent to the backend via IPC payload, or the backend has to read the relevant data again from the output file. It should be evaluated which of these approaches is faster, and whether this ends up being significantly slower than the current implementation.
After the preprocessing has been move to the backend, the code for setting the loading state to true and writing the first progress log message should be moved from the menu entries' onClick to the exportFileListener. Similarly, the loading state should then be set to false at the end of the exportFileListener, making the StopLoading IPC channel (which was only introduced as a temporary fix) and all its uses obsolete.

**Describe alternatives you've considered**
Offloading the preprocessing to the SignalsWorker would also be possible, but using the backend makes more sense here.

PR #2815, opened 2025-02-24
Unify handling of loading state

PhilippMa:
### Summary of changes

* the loading state is set/unset from the backend only
* frontend components directly listen on the FileLoading ipc channel again, no more loading state in redux
* add StopLoading ipc channel as a temporary solution for handling loading state on export (see #2812)

* Heavy operations requiring the use of loading state should all run in the backend to guarantee that the UI stays responsive, thus the loading state should also be controlled from the backend
* Inconsistent handling of loading state in different components is confusing and can lead to bugs

Check that the ProcessPopup is shown for open/export and that the import dialog properly displays logs during loading.

### DIFF
--- a/src/shared/ipc-channels.ts
+++ b/src/shared/ipc-channels.ts
@@ -5,7 +5,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
 export enum IpcChannel {
-  DeleteFile = 'delete-file',
   ExportFile = 'export-file',
   OpenFile = 'open-file',
   ImportFileSelectInput = 'import-file-select-input',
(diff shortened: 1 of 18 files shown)


src/shared/ipc-channels.ts
```diff
@@ -5,14 +5,14 @@
...
 export enum IpcChannel {
-  DeleteFile = 'delete-file',
   ExportFile = 'export-file',
   OpenFile = 'open-file',
   ImportFileSelectInput = 'import-file-select-input',
   ImportFileSelectSaveLocation = 'import-file-select-save-location',
   ImportFileConvertAndLoad = 'import-file-convert-and-load',
   OpenLink = 'open-link',
   SaveFile = 'save-file',
+  StopLoading = 'stop-loading',
```
mstykow, line 15: code comment that this shouldn't be used anywhere else

mstykow (changes requested):

mstykow (approved):

PR #2819, opened 2025-02-26
Add classification column to "Signals per Sources" table in project statistics popup

### LINKED ISSUE #2804: Add classification column to signals per sources table
Users do not have an overview of what signals have which classifications.

Add a column to the signals per sources table for displaying the classification of each signal.

\-

**Additional context**
\-

PhilippMa:
### Summary of changes

Added a new column to the "Signals per Sources" table in the project statistics popup for displaying the classification of each license.

see #2804 

Check that classification information is shown correctly in the project statistics popup.

closes #2804

### DIFF
--- a/src/Frontend/Components/AttributionCountPerSourcePerLicenseTable/AttributionCountPerSourcePerLicenseTable.tsx
+++ b/src/Frontend/Components/AttributionCountPerSourcePerLicenseTable/AttributionCountPerSourcePerLicenseTable.tsx
@@ -8,7 +8,11 @@ import MuiTableBody from '@mui/material/TableBody';
 import MuiTableContainer from '@mui/material/TableContainer';
 import MuiTypography from '@mui/material/Typography';
 
-import { LicenseCounts, LicenseNamesWithCriticality } from '../../types/types';
+import {
+  LicenseCounts,
+  LicenseNamesWithClassification,
+  LicenseNamesWithCriticality,
+} from '../../types/types';
 import { AttributionCountPerSourcePerLicenseTableFooter } from './AttributionCountPerSourcePerLicenseTableFooter/AttributionCountPerSourcePerLicenseTableFooter';
 import { AttributionCountPerSourcePerLicenseTableHead } from './AttributionCountPerSourcePerLicenseTableHead/AttributionCountPerSourcePerLicenseTableHead';
 import { AttributionCountPerSourcePerLicenseTableRow } from './AttributionCountPerSourcePerLicenseTableRow/AttributionCountPerSourcePerLicenseTableRow';
@@ -23,6 +27,7 @@ const classes = {
 interface AttributionCountPerSourcePerLicenseTableProps {
   licenseCounts: LicenseCounts;
   licenseNamesWithCriticality: LicenseNamesWithCriticality;
+  licenseNamesWithClassification: LicenseNamesWithClassification;
   title: string;
 }
 
--- a/src/Frontend/Components/AttributionCountPerSourcePerLicenseTable/AttributionCountPerSourcePerLicenseTable.tsx
+++ b/src/Frontend/Components/AttributionCountPerSourcePerLicenseTable/AttributionCountPerSourcePerLicenseTable.tsx
@@ -61,10 +61,14 @@ export const AttributionCountPerSourcePerLicenseTable: React.FC<
                   licenseCriticality={
                     props.licenseNamesWithCriticality[licenseName]
                   }
+                  licenseClassification={
+                    props.licenseNamesWithClassification[licenseName]
+                  }
                   totalSignalCount={
                     props.licenseCounts.totalAttributionsPerLicense[licenseName]
                   }
                   key={rowIndex}
+                  rowIndex={rowIndex}
                 />
               ))}
           </MuiTableBody>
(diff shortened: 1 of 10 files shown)


abraemer (approved):
very nice and minimal changes

src/Frontend/Components/AttributionCountPerSourcePerLicenseTable/AttributionCountPerSourcePerLicenseTable.tsx
```diff
@@ -122,6 +133,25 @@ export const AttributionCountPerSourcePerLicenseTable: React.FC<
...
+      <MuiTableCell
+        sx={{
+          ...tableClasses.body,
+        }}
+        key={1}
+        align={'center'}
+      >
+        <span>
+          {licenseClassification
+            ? (classifications[licenseClassification] ?? '-')
```
vasily-pozdnyakov, line 149: please use UNKNOWN_SOURCE_PLACEHOLDER for this.

vasily-pozdnyakov (changes requested):
I also think that the titles of the extra columns (criticality and classification) have very similar (the same) style as sources columns, which is a bit misleading. Can we think about a way how to better separate them, e.g. by vertical line between license info columns and sources?

src/Frontend/Components/AttributionCountPerSourcePerLicenseTable/AttributionCountPerSourcePerLicenseTable.tsx
PhilippMa: I understand why you would want to use a text constant here, but since UNKNOWN_SOURCE_PLACEHOLDER has quite a specific name that doesn't really apply here, how about introducing a new text constant in text.ts and using it here instead?

src/Frontend/Components/AttributionCountPerSourcePerLicenseTable/AttributionCountPerSourcePerLicenseTable.tsx
```diff
@@ -73,12 +86,46 @@ export const AttributionCountPerSourcePerLicenseTable: React.FC<
...
   );
 
+  const headerVerticalSeparator = {
+    borderRight: '2px solid lightgray',
+  };
+
+  const headerHorizontalSeparator = {
+    borderBottom: '1.5px solid lightgray',
+  };
+
```
mstykow, line 96: these are untyped styles, not components (their name "Separator" suggests they would be components). probably best to make styled components out of these or use `sx`.

src/Frontend/Components/AttributionCountPerSourcePerLicenseTable/AttributionCountPerSourcePerLicenseTable.tsx
vasily-pozdnyakov: could be an option

src/Frontend/Components/AttributionCountPerSourcePerLicenseTable/AttributionCountPerSourcePerLicenseTableHead/AttributionCountPerSourcePerLicenseTableHead.tsx
```diff
@@ -19,20 +28,49 @@ export const AttributionCountPerSourcePerLicenseTableHead: React.FC<
...
-    componentText.columnNames.totalSources,
-  ].map(
-    (sourceName) => sourceName.charAt(0).toUpperCase() + sourceName.slice(1),
-  );
+    componentText.columns.licenseName,
+    componentText.columns.criticality.title,
+    componentText.columns.classification,
+    ...props.sourceNames.map(
+      (sourceName) => sourceName.charAt(0).toUpperCase() + sourceName.slice(1),
+    ),
```
mstykow, line 36:
```suggestion
    ...props.sourceNames.map(upperFirst),
```

src/shared/text.ts
```diff
@@ -187,15 +187,19 @@ export const text = {
...
+      licenseName: 'Name',
       criticality: {
         title: 'Criticality',
         medium: 'Medium Criticality',
         high: 'High Criticality',
       },
+      classification: 'Classification',
       totalSources: 'Total',
     },
+    absent: '-',
```
mstykow, line 202: do you mean "none"?

src/Frontend/Components/AttributionCountPerSourcePerLicenseTable/AttributionCountPerSourcePerLicenseTableHead/AttributionCountPerSourcePerLicenseTableHead.tsx
```diff
@@ -9,6 +9,15 @@ import MuiTableRow from '@mui/material/TableRow';
...
 
+const classes = {
+  headerCellWithVerticalSeparator: {
+    borderRight: '2px solid lightgray',
+  },
+  headerCellWithHorizontalSeparator: {
+    borderBottom: '1.5px solid lightgray',
+  },
+};
+
```
mstykow, line 20: this is untyped. i know various other places in opossumui are following this pattern which isn't the best. i would either use a styled component or `sx` instead.

mstykow (approved):

PR #2822, opened 2025-02-27
Change menu structure on MacOS

PhilippMa:
### Summary of changes

Add standard app menu on MacOS. This also has the effect that the "File" menu is actually seen with the name "File" and is no longer sort of hidden behind the app name.

This kind of menu structure follows the standard for MacOS apps.

Open OpossumUI on a MacOS device and check the menu structure. Also check that the menu structure is not affected when opening OpossumUI on non-MacOS devices.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,6 @@
 {
   "name": "opossum-ui",
+  "productName": "OpossumUI",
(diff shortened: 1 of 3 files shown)


mstykow (approved):
Top 👍

PR #2823, opened 2025-02-27
Refactor ProjectStatisticsPopup.util.ts

abraemer:
### Summary of changes

Small cleanups around PrjectStatisticsPopup.util.ts
* delete unused function `getLicenseNameVariants`
* move getUniqueLicenseNameToAttribution inside of PrjectStatisticsPopup.util.ts to make using the methods a bit easier

Boy Scout rule

### DIFF
--- a/src/Frontend/Components/ProjectStatisticsPopup/ProjectStatisticsPopup.tsx
+++ b/src/Frontend/Components/ProjectStatisticsPopup/ProjectStatisticsPopup.tsx
@@ -30,7 +30,6 @@ import {
   getIncompleteAttributionsCount,
   getMostFrequentLicenses,
   getSignalCountByClassification,
-  getUniqueLicenseNameToAttribution,
 } from './ProjectStatisticsPopup.util';
 
 const classes = {
@@ -50,17 +49,12 @@ export const ProjectStatisticsPopup: React.FC = () => {
     getUnresolvedExternalAttributions,
   );
 
-  const strippedLicenseNameToAttribution = getUniqueLicenseNameToAttribution(
-    unresolvedExternalAttribution,
-  );
-
   const {
     licenseCounts,
     licenseNamesWithCriticality,
     licenseNamesWithClassification,
   } = aggregateLicensesAndSourcesFromAttributions(
     unresolvedExternalAttribution,
-    strippedLicenseNameToAttribution,
     attributionSources,
   );
 
--- a/src/Frontend/Components/ProjectStatisticsPopup/ProjectStatisticsPopup.util.ts
+++ b/src/Frontend/Components/ProjectStatisticsPopup/ProjectStatisticsPopup.util.ts
@@ -35,13 +35,14 @@ export const ATTRIBUTION_TOTAL = 'Total Attributions';
 
 export function aggregateLicensesAndSourcesFromAttributions(
   attributions: Attributions,
-  strippedLicenseNameToAttribution: UniqueLicenseNameToAttributions,
   attributionSources: ExternalAttributionSources,
 ): {
   licenseCounts: LicenseCounts;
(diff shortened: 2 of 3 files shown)


src/Frontend/Components/AttributionCountPerSourcePerLicenseTable/AttributionCountPerSourcePerLicenseTable.tsx
```diff
@@ -40,34 +57,163 @@ export const AttributionCountPerSourcePerLicenseTable: React.FC<
...
-      title={props.title}
-      containerStyle={classes.container}
-      columnHeaders={headerRow}
-      columnNames={headerRow}
-      rowNames={sortBy(Object.keys(props.licenseNamesWithCriticality))}
-      tableContent={props.licenseCounts.attributionCountPerSourcePerLicense}
-      tableFooter={footerRow}
-      licenseNamesWithCriticality={props.licenseNamesWithCriticality}
-    />
+    <MuiBox>
```
mstykow, line 204: why a box that does nothing? can you use a fragment instead?

src/Frontend/Components/AttributionCountPerSourcePerLicenseTable/AttributionCountPerSourcePerLicenseTable.tsx
```diff
@@ -40,34 +57,163 @@ export const AttributionCountPerSourcePerLicenseTable: React.FC<
...
   Object.entries(
     props.licenseCounts.attributionCountPerSourcePerLicense,
   ).forEach(
     ([licenseName, value]) =>
-      (value[TOTAL_SOURCES_TITLE] =
+      (value[TOTAL_SOURCE_NAME] =
         props.licenseCounts.totalAttributionsPerLicense[licenseName]),
   );
 
+  const tableHead = (
```
mstykow, line 86: ok, i could write a long essay here but probably better if talk about some react fundamentals. :)
it would be a good start, if you look how other components in OpossumUI are built. you will then notice some things:
1. component functions are never just consts but real functions
2. component functions either use the react component syntax `<TableHead />` or they are inside some other component function where their name starts with "render" (not with "build").
3. react components read like a book from top to bottom. so rather than trying to find the main `return` statement at the bottom of the function and then scrolling back up to see which other functions are part of the main `return`, the sub-functions follow below the main `return`.

src/Frontend/Components/AttributionCountPerSourcePerLicenseTable/AttributionCountPerSourcePerLicenseTable.tsx
```diff
@@ -40,34 +57,163 @@ export const AttributionCountPerSourcePerLicenseTable: React.FC<
...
+    </MuiTableHead>
+  );
+
+  const buildTableRow = (licenseName: string, rowIndex: number) => {
+    const licenseNameCell = (
+      <MuiTableCell
+        sx={{
+          ...tableClasses.body,
+        }}
+        key={0}
```
mstykow, line 108: a constant key is unnecessary
```suggestion
        sx={tableClasses.body}
```

mstykow (changes requested):
i started reviewing this diff but will stop now because what it does has nothing to do with the title or the description of the diff. how are all these changes, including major refactorings of existing components part of a "small util refactoring"?

abraemer:
Something went wrong with the base branch. Maybe philipp force pushed it? As you can see in the commits included in this PR only the last 2 are from me and theses indeed are only a minor refactor :)
Will rebase in the morning.

abraemer:
I think what happened is that Philipp addressed the comments in the first PR of this stack and then rebased all of his PRs but not mine of course.

PR #2829, opened 2025-02-28
Make Criticality enum complete

PhilippMa:
### Summary of changes

Add non-critical criticality as explicit enum case and use criticality enum without `| undefined` throughout the application.

It's better to have an explicit representation of non-critical criticality.

Check that usage of criticality throughout is semantically the same.


src/shared/shared-types.ts
```diff
@@ -13,10 +13,17 @@ export interface Resources {
 }
 
 export enum Criticality {
-  High = 'high',
-  Medium = 'medium',
+  NonCritical,
```
mstykow, line 16: let's make this "None"

mstykow (approved):
Looks fine except that one renaming.

PR #2830, opened 2025-02-28
Introduce sorting for signals per source table

PhilippMa:
### Summary of changes

Make the signals per source table sortable by each column.

For larger .opossum files, being able to sort the table by e.g. criticality can help reviewers get a better overview of the signals need to look at.

Check that sorting works correctly for the signals per source table.

### DIFF
--- a/src/Frontend/Components/AttributionCountPerSourcePerLicenseTable/AttributionCountPerSourcePerLicenseTable.tsx
+++ b/src/Frontend/Components/AttributionCountPerSourcePerLicenseTable/AttributionCountPerSourcePerLicenseTable.tsx
@@ -5,38 +5,32 @@
 import MuiBox from '@mui/material/Box';
 import MuiTable from '@mui/material/Table';
 import MuiTableBody from '@mui/material/TableBody';
-import MuiTableCell from '@mui/material/TableCell';
 import MuiTableContainer from '@mui/material/TableContainer';
-import MuiTableFooter from '@mui/material/TableFooter';
-import MuiTableHead from '@mui/material/TableHead';
-import MuiTableRow from '@mui/material/TableRow';
 import MuiTypography from '@mui/material/Typography';
+import { upperFirst } from 'lodash';
 import { useMemo, useState } from 'react';
 
+import { text } from '../../../shared/text';
 import {
   LicenseCounts,
   LicenseNamesWithClassification,
   LicenseNamesWithCriticality,
 } from '../../types/types';
-import { orderLicenseNames } from './AttributionCountPerSourcePerLicenseTable.util';
-import { AttributionCountPerSourcePerLicenseTableFooter } from './AttributionCountPerSourcePerLicenseTableFooter/AttributionCountPerSourcePerLicenseTableFooter';
 import {
-  AttributionCountPerSourcePerLicenseTableHead,
+  ColumnConfig,
+  orderLicenseNames,
+  SingleColumn,
   TableOrdering,
-} from './AttributionCountPerSourcePerLicenseTableHead/AttributionCountPerSourcePerLicenseTableHead';
+} from './AttributionCountPerSourcePerLicenseTable.util';
+import { AttributionCountPerSourcePerLicenseTableFooter } from './AttributionCountPerSourcePerLicenseTableFooter/AttributionCountPerSourcePerLicenseTableFooter';
+import { AttributionCountPerSourcePerLicenseTableHead } from './AttributionCountPerSourcePerLicenseTableHead/AttributionCountPerSourcePerLicenseTableHead';
 import { AttributionCountPerSourcePerLicenseTableRow } from './AttributionCountPerSourcePerLicenseTableRow/AttributionCountPerSourcePerLicenseTableRow';
-import { tableClasses } from '../../shared-styles';
 
 const classes = {
   container: {
     maxHeight: '400px',
     marginBottom: '3px',
   },
-};
-
-const DEFAULT_ORDERING: TableOrdering = {
-  orderDirection: 'asc',
-  orderedColumn: 0,
 };
 
 interface AttributionCountPerSourcePerLicenseTableProps {
@@ -49,10 +43,66 @@ interface AttributionCountPerSourcePerLicenseTableProps {
 export const AttributionCountPerSourcePerLicenseTable: React.FC<
   AttributionCountPerSourcePerLicenseTableProps
 > = (props) => {
-  const [ordering, setOrdering] = useState(DEFAULT_ORDERING);
+  const componentText = text.attributionCountPerSourcePerLicenseTable;
 
-  const handleRequestSort = (columnIndex: number) => {
-    if (ordering.orderedColumn === columnIndex) {
+  const sourceNames = Object.keys(
+    props.licenseCounts.totalAttributionsPerSource,
+  );
+
+  const columnConfig: ColumnConfig = useMemo(
+    () =>
+      new ColumnConfig([
+        {
+          groupName: componentText.columns.licenseInfo,
+          columns: [
+            {
+              columnName: componentText.columns.licenseName,
+              columnType: SingleColumn.NAME,
+              columnId: SingleColumn.NAME,
+              align: 'left',
+            },
+            {
+              columnName: componentText.columns.criticality.title,
+              columnType: SingleColumn.CRITICALITY,
+              columnId: SingleColumn.CRITICALITY,
+              align: 'center',
+            },
+            {
+              columnName: componentText.columns.classification,
+              columnType: SingleColumn.CLASSIFICATION,
+              columnId: SingleColumn.CLASSIFICATION,
+              align: 'center',
+            },
+          ],
+        },
+        {
+          groupName: componentText.columns.signalCountPerSource,
+          columns: [
+            ...sourceNames.map((sourceName) => ({
+              columnName: upperFirst(sourceName),
+              columnType: { sourceName },
+              columnId: sourceName,
+              align: 'center' as const,
+            })),
+            {
+              columnName: componentText.columns.totalSources,
+              columnType: SingleColumn.TOTAL,
+              columnId: SingleColumn.TOTAL,
+              align: 'center',
+            },
+          ],
+        },
+      ]),
+    [sourceNames, componentText],
+  );
+
+  const [ordering, setOrdering] = useState<TableOrdering>({
+    orderDirection: 'asc',
+    orderedColumn: SingleColumn.NAME,
+  });
+
+  const handleRequestSort = (columnId: string) => {
+    if (ordering.orderedColumn === columnId) {
       setOrdering({
         ...ordering,
         orderDirection: ordering.orderDirection === 'asc' ? 'desc' : 'asc',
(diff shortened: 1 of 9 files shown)


Hellgartner:
The select behavior for the Name column seems to differ from the other columns visually. Is that desired?
image
image
(compare the text color and maybe also font properties)

src/Frontend/Components/AttributionCountPerSourcePerLicenseTable/AttributionCountPerSourcePerLicenseTable.tsx
```diff
@@ -34,43 +38,94 @@ interface AttributionCountPerSourcePerLicenseTableProps {
 export const AttributionCountPerSourcePerLicenseTable: React.FC<
   AttributionCountPerSourcePerLicenseTableProps
 > = (props) => {
+  const [order, setOrder] = useState<Order>('asc');
+  const [orderedColumn, setOrderedColumn] = useState(0);
+
+  const handleRequestSort = (columnIndex: number) => {
+    setOrder(orderedColumn === columnIndex && order === 'asc' ? 'desc' : 'asc');
```
Hellgartner, line 45: nice one liner
Question is whether this is understandable in the long run.
If I get this correctly this means
```
if (clickedColumn === sortedColumn)
    toggleSortOrder
else
   setAscendingOrder
```

PhilippMa:
> The select behavior for the Name column seems to differ from the other columns visually. Is that desired? image image (compare the text color and maybe also font properties)

That was not intended and I have no idea what caused this, as I am using the same component for all table headers with sorting functionality... Adding `!important` to the relevant color property seems to have fixed it though

src/Frontend/Components/AttributionCountPerSourcePerLicenseTable/AttributionCountPerSourcePerLicenseTable.tsx
```diff
@@ -34,43 +38,94 @@ interface AttributionCountPerSourcePerLicenseTableProps {
 export const AttributionCountPerSourcePerLicenseTable: React.FC<
   AttributionCountPerSourcePerLicenseTableProps
 > = (props) => {
+  const [order, setOrder] = useState<Order>('asc');
+  const [orderedColumn, setOrderedColumn] = useState(0);
```
Hellgartner, line 42: Idea: Create SearchState object and do a state on that? I guess those properties will always be used in together.

src/Frontend/Components/AttributionCountPerSourcePerLicenseTable/AttributionCountPerSourcePerLicenseTable.tsx
```diff
@@ -34,43 +38,94 @@ interface AttributionCountPerSourcePerLicenseTableProps {
...
+  };
+
   const sourceNames = Object.keys(
     props.licenseCounts.totalAttributionsPerSource,
   );
 
+  const sortedLicenseNames = useMemo(
+    () =>
+      orderBy(
+        Object.keys(props.licenseNamesWithCriticality).toSorted(),
```
Hellgartner, line 56: is this toSorted() required here?

src/Frontend/Components/AttributionCountPerSourcePerLicenseTable/AttributionCountPerSourcePerLicenseTable.tsx
```diff
@@ -34,43 +38,94 @@ interface AttributionCountPerSourcePerLicenseTableProps {
...
+  const handleRequestSort = (columnIndex: number) => {
+    setOrder(orderedColumn === columnIndex && order === 'asc' ? 'desc' : 'asc');
+    setOrderedColumn(columnIndex);
+  };
+
   const sourceNames = Object.keys(
     props.licenseCounts.totalAttributionsPerSource,
   );
 
+  const sortedLicenseNames = useMemo(
```
Hellgartner, line 53: IMHO this function is complex enough to warrant its own file and consequently its own tests

src/Frontend/Components/AttributionCountPerSourcePerLicenseTable/AttributionCountPerSourcePerLicenseTable.tsx
```diff
@@ -34,43 +38,94 @@ interface AttributionCountPerSourcePerLicenseTableProps {
...
+    () =>
+      orderBy(
+        Object.keys(props.licenseNamesWithCriticality).toSorted(),
+        (licenseName) => {
+          const numStartingColumns = 3;
+
+          if (orderedColumn === 0) {
```
Hellgartner, line 60: there are quite some "magic numbers" in here which create a very implicit dependency on `headerRow` in `AttributionCountPerSourcePerLicenseTableHead` as well as to the inner workings of `AttributionCountPerSourcePerLicenseTableRow`. 
I wonder if there is any possibility to collect that logic in one place

src/Frontend/Components/AttributionCountPerSourcePerLicenseTable/AttributionCountPerSourcePerLicenseTable.tsx
```diff
@@ -34,43 +38,94 @@ interface AttributionCountPerSourcePerLicenseTableProps {
...
+    setOrderedColumn(columnIndex);
+  };
+
   const sourceNames = Object.keys(
     props.licenseCounts.totalAttributionsPerSource,
   );
 
+  const sortedLicenseNames = useMemo(
+    () =>
+      orderBy(
```
Hellgartner, line 55: nit: consider using `Array.prototype.toSorted`

src/Frontend/Components/TableCellWithSorting/TableCellWithSorting.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+export const TableCellWithSorting: React.FC<TableCellWithSortingProps> = (
+  props,
+) => {
+  return (
+    <MuiTableCell
+      {...props}
+      sx={{
+        '.Mui-active': { color: 'white' },
+        '& :hover': { color: 'white !important' },
```
Hellgartner, line 26: Is there any possibility to avoid styling with `!important` ? (see also below). It breaks the cascade.

src/Frontend/Components/TableCellWithSorting/TableCellWithSorting.tsx
```diff
@@ -0,0 +1,48 @@
...
+      <TableSortLabel
+        sx={{
+          '& .MuiTableSortLabel-icon': { color: 'white !important' },
+        }}
+        active={props.isSortedColumn}
+        direction={props.isSortedColumn ? props.order : 'asc'}
+        onClick={props.onRequestSort}
+      >
+        {props.children}
+        {props.isSortedColumn ? (
```
Hellgartner, line 40: Nice :100: Thinking of a11y

Hellgartner (changes requested):
On the whole, does this needs some automated testing.

src/Frontend/Components/AttributionCountPerSourcePerLicenseTable/AttributionCountPerSourcePerLicenseTable.tsx
PhilippMa: Good point, I basically just copied this line from the table sorting example in the mui docs, but looking at it again I would agree that splitting the conditions would make this a lot easier to understand

src/Frontend/Components/AttributionCountPerSourcePerLicenseTable/AttributionCountPerSourcePerLicenseTable.tsx
PhilippMa: Elements that are equal in respect to the ordering are kept in their original order by orderBy, so I used this to have secondary odering by the license name when another column was selected. , it's probably better to pass a explicitly secondary ordering to orderBy instead

src/Frontend/Components/AttributionCountPerSourcePerLicenseTable/AttributionCountPerSourcePerLicenseTable.tsx
PhilippMa: Makes sense, that way I also need one less property on the table head for passing down the ordering

src/Frontend/Components/AttributionCountPerSourcePerLicenseTable/AttributionCountPerSourcePerLicenseTable.tsx
PhilippMa: Isn't orderBy a good fit here because it already nicely separates the ordered property from the order direction, and has support for secondary orderings?

src/Frontend/Components/TableCellWithSorting/TableCellWithSorting.tsx
PhilippMa: (Actually just copy paste from the sorting example on mui docs :sweat_smile:)

src/Frontend/Components/TableCellWithSorting/TableCellWithSorting.tsx
PhilippMa: I would also love to know if there is any way to avoid `!important` here, but from my own ideas and the AI suggestions I tried this was the only thing that actually worked...

src/Frontend/Components/AttributionCountPerSourcePerLicenseTable/AttributionCountPerSourcePerLicenseTable.tsx
PhilippMa: done (hopefully not overengineered)

src/Frontend/Components/AttributionCountPerSourcePerLicenseTable/AttributionCountPerSourcePerLicenseTable.tsx
Hellgartner: it at least works

src/Frontend/Components/AttributionCountPerSourcePerLicenseTable/AttributionCountPerSourcePerLicenseTable.tsx
```diff
@@ -34,46 +43,142 @@ interface AttributionCountPerSourcePerLicenseTableProps {
...
+  const [ordering, setOrdering] = useState<TableOrdering>({
+    orderDirection: 'asc',
+    orderedColumn: SingleColumn.NAME,
+  });
+
+  const handleRequestSort = (columnId: string) => {
+    if (ordering.orderedColumn === columnId) {
+      setOrdering({
+        ...ordering,
+        orderDirection: ordering.orderDirection === 'asc' ? 'desc' : 'asc',
```
Hellgartner, line 108: to toggle a value, you should use the other way of using the state setter in react:
`setOrdering( (currentOrdering) => ...)`

Hellgartner (changes requested):
A small nitpick concerning the tests:
At many (but AFAIK not everywhere) place we do consider using faker for testing

Hellgartner (approved):

PR #2834, opened 2025-03-05
Tabs in project statistics

PhilippMa:
### Summary of changes

Divide project statistics popup into two tabs, one for a general overview with charts and one for the signals per sources table with more detailed information.

The project statistics popup was quite cluttered with a lot of elements being displayed at once. Additionally, moving the table to a separate tab allows for the pie charts to be always displayed, without the need for accordions.

Look at the new tabs in the project statistics popup. Note: Layout of the first tab in the project statistics popup will be fixed in a future stacked PR.

### DIFF
--- a/src/Frontend/Components/ProjectStatisticsPopup/ProjectStatisticsPopup.tsx
+++ b/src/Frontend/Components/ProjectStatisticsPopup/ProjectStatisticsPopup.tsx
@@ -3,7 +3,10 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 import MuiBox from '@mui/material/Box';
+import MuiTab from '@mui/material/Tab';
+import MuiTabs from '@mui/material/Tabs';
 import MuiTypography from '@mui/material/Typography';
+import { useState } from 'react';
 
 import { Criticality } from '../../../shared/shared-types';
 import { text } from '../../../shared/text';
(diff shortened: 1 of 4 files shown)


src/Frontend/Components/ProjectStatisticsPopup/ProjectStatisticsPopup.tsx
```diff
@@ -96,66 +99,80 @@ export const ProjectStatisticsPopup: React.FC = () => {
...
-                data={incompleteAttributionsData}
-                title={
-                  text.projectStatisticsPopup.charts
-                    .incompleteAttributionsPieChart
-                }
-              />
+          <MuiTabs
+            value={selectedTab}
+            onChange={(_, tab) => setSelectedTab(tab)}
+            sx={{ marginBottom: '10px' }}
```
Hellgartner, line 111: A possible alternative here would be to use a box with a divider line, similar to the react [mui example](https://mui.com/material-ui/react-tabs/?srsltid=AfmBOorcu-DzyVjFaVYNFCzEEmE0BEQjsbM1G4a2jWpUqWHHkjvhp2Yp#basics). Or does this result in a very long line?

We are using a lot of "sleight of hand" pixel values for styling. Just safety, we do not have a size/distance ladder defined anywhere. Looking at the rest of the project, multiples of 6 seem to be more common.

Hellgartner (changes requested):
While I do like the code I do have some problems when doing a live test:

**Warnings in the javascript log**

On project statistics popup open, I do get 
```
recharts.js?v=51e374da:3418 The width(0) and height(0) of chart should be greater than 0,
       please check the style of container, or the props width(100%) and height(100%),
       or add a minWidth(0) or minHeight(undefined) or use aspect(2) to control the
       height and width.
```
When navigating between tabs, the number of these error messages rises

**Jumps in the UI flow**
When switching between tabs, the size of the popup jumps and thus the tab buttons move away from under my mouse

Compare
image
with
image

As the table on the second tab additionally has scrollbars, I would tend to fix the size of the popup relative to the viewport to avoid this

src/Frontend/Components/ProjectStatisticsPopup/ProjectStatisticsPopup.tsx
PhilippMa: I like the divider line here!

PhilippMa:
> While I do like the code I do have some problems when doing a live test:
> 
> **Warnings in the javascript log**
> 
> On project statistics popup open, I do get
> 
> ```
> recharts.js?v=51e374da:3418 The width(0) and height(0) of chart should be greater than 0,
>        please check the style of container, or the props width(100%) and height(100%),
>        or add a minWidth(0) or minHeight(undefined) or use aspect(2) to control the
>        height and width.
> ```
> 
> When navigating between tabs, the number of these error messages rises
> 
> **Jumps in the UI flow** When switching between tabs, the size of the popup jumps and thus the tab buttons move away from under my mouse
> 
> Compare image with image
> 
> As the table on the second tab additionally has scrollbars, I would tend to fix the size of the popup relative to the viewport to avoid this

These are both issues that I am aware of, I just felt like it would make more sense to make the larger layout changes after I also change the attributions overview table into a bar chart, extract the pie charts from the accordion components and display everything in a grid layout in the first tab.

Hellgartner:
> > While I do like the code I do have some problems when doing a live test:
> > **Warnings in the javascript log**
> > On project statistics popup open, I do get
> > ```
> > recharts.js?v=51e374da:3418 The width(0) and height(0) of chart should be greater than 0,
> >        please check the style of container, or the props width(100%) and height(100%),
> >        or add a minWidth(0) or minHeight(undefined) or use aspect(2) to control the
> >        height and width.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > When navigating between tabs, the number of these error messages rises
> > **Jumps in the UI flow** When switching between tabs, the size of the popup jumps and thus the tab buttons move away from under my mouse
> > Compare image with image
> > As the table on the second tab additionally has scrollbars, I would tend to fix the size of the popup relative to the viewport to avoid this
> 
> These are both issues that I am aware of, I just felt like it would make more sense to make the larger layout changes after I also change the attributions overview table into a bar chart, extract the pie charts from the accordion components and display everything in a grid layout in the first tab.

Fine for me

Hellgartner (approved):

PR #2835, opened 2025-03-05
Convert attribution properties display

PhilippMa:
### Summary of changes

Convert attribution properties display from table to bar chart.

A chart is easier to read than a table here.

Look at the project statistics popup. Note: Layout of the first tab in the project statistics popup will be fixed in a future stacked PR.


src/Frontend/Components/PieChart/PieChart.tsx
```diff
@@ -24,47 +28,32 @@ const defaultPieChartColors = [
...
+  };
+}
+
 interface PieChartProps {
-  segments: Array<PieChartData>;
+  segments: Array<ChartDataItem>;
   colors?: Array<string>;
 }
 
-const classes = {
```
Hellgartner, line 32: You probably want to revamp the Classes object (even if you load the contends from other files) as this is a common pattern in OpossumUI (I recently had a discussion with @mstykow  on that)
And you probably want to add a `satisfies SxProps;` at the end.

src/Frontend/Components/ProjectStatisticsPopup/ProjectStatisticsPopup.tsx
```diff
@@ -120,15 +121,24 @@ export const ProjectStatisticsPopup: React.FC = () => {
...
-                  attributionPropertyCountsEntries={Object.entries(
-                    manualAttributionPropertyCounts,
-                  )}
-                  title={
-                    text.projectStatisticsPopup.charts
-                      .attributionPropertyCountTable
-                  }
-                />
+                <MuiPaper
+                  sx={{
```
Hellgartner, line 125: As this component already has a classes object for styling, you might want to move these styles there.

src/Frontend/Components/BarChart/BarChart.tsx
Hellgartner, line 1: This component does not have a test. As it mainly  wraps the barchart that might be OK but I would prefer to have a "renders" tests to catch any breaking API changes.

Hellgartner (changes requested):
Findings from testing:
* Do we need a label on the x-Axis of the plot? I would love to have one.
* The y axis tick labels are hard to read as they are very close to each other
   image
* Also on tab switch, the bars are always "animated in" which draws attention of the user. Is that desired?

src/Frontend/Components/ProjectStatisticsPopup/ProjectStatisticsPopup.tsx
PhilippMa: I'm actually extracting this into a styled component as part of the next PR

src/Frontend/Components/PieChart/PieChart.tsx
PhilippMa: It seems like git somehow showed you an outdated version of this file :thinking:
I did actually break up the classes object and made sure to add type annotations where necessary

PhilippMa:
> Findings from testing:
> 
> * Do we need a label on the x-Axis of the plot? I would love to have one.
> * The y axis tick labels are hard to read as they are very close to each other
>   image
> * Also on tab switch, the bars are always "animated in" which draws attention of the user. Is that desired?

* Good point, I added a label there
* This will improve with better overall layout in the next PR
* True, that animation is kind of unnecessary, I disabled it

Hellgartner (approved):

PR #2837, opened 2025-03-06
Add attribution form chip for classification 

Hellgartner:
### Summary of changes

Add a chip in the attribution form showing the classification of the signal

We are currently introducing classifications into opossum-ui

* ci,
* Load a file with classifications and select a few of the signals (should show the new chip) 
* Promote a signal to an attribution and observe the chip vanishing

### DIFF
--- a/src/Frontend/Components/Icons/Icons.tsx
+++ b/src/Frontend/Components/Icons/Icons.tsx
@@ -19,7 +19,6 @@ import WhatshotIcon from '@mui/icons-material/Whatshot';
 import WidgetsIcon from '@mui/icons-material/Widgets';
 import { Icon, SxProps } from '@mui/material';
 import MuiTooltip from '@mui/material/Tooltip';
-import { ReactNode } from 'react';
 
 import { Classifications, Criticality } from '../../../shared/shared-types';
 import { text } from '../../../shared/text';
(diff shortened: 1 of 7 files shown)


src/Frontend/Components/Icons/Icons.tsx
```diff
@@ -249,6 +251,18 @@ export function ClassificationIcon({
...
+  props: IconProps & {
+    classification?: number;
+    classification_mapping?: Classifications;
+  },
+) {
+  if (!props.classification || props.classification === 0) {
+    return null;
+  }
+  return <ClassificationIcon {...props} />;
+}
```
abraemer, line 264: Why this split? `ClassificationIcon` now is used only by the tests.
I dislike this for 2 reasons:
 - Testing something that's not used is testing an implementation detail which usually is seen as bad style because it is usually unnecessary.
 - This file now exports 2 almost identical things and I cannot really tell the difference in behavior from their names `CriticalClassificationIcon` vs. `ClassificationIcon`
 
 When I first read this I thought, that you want to use `ClassificationIcon` for the `AuditingOptions` but that is not the case.

abraemer (changes requested):

src/Frontend/Components/Icons/Icons.tsx
Hellgartner: It has been the case before the latest change in business logic. That is why these two were introduced to begin with.
Thanks for the reminder. Forgot to clean that up.

abraemer (approved):

PR #2838, opened 2025-03-07
Improve layout in project statistics popup

PhilippMa:
### Summary of changes

* extract pie charts from their surrounding accordion components
* retain light blue background for charts by adding card backgrounds
* arrange all charts in a 2x3 grid layout
* make popup size consistent between both tabs and make sure that the content grows to fill the available space in the popup

The layout of the popup wasn't looking great after the recent smaller changes, and due to moving the signals per sources table to a separate tab there is now enough space to display all charts at the same time

Look at the project statistics popup.


src/e2e-tests/page-objects/ProjectStatisticsPopup.ts
```diff
@@ -28,6 +33,36 @@ export class ProjectStatisticsPopup {
       .last()
       .getByRole('cell')
       .last();
+    this.attributionsOverviewDiv = this.node.locator('div').filter({
+      hasText: text.projectStatisticsPopup.charts.attributionProperties.title,
+      hasNotText:
+        text.projectStatisticsPopup.charts.mostFrequentLicenseCountPieChart,
+    });
```
Hellgartner, line 40: This is a very strange pattern. Why not using any of the provided [locators](https://playwright.dev/docs/locators) instead of doing this strange filtering. 
This might (unsure) also not play well with some of the better properties of playwright locators (auto retry etc)

src/e2e-tests/page-objects/ProjectStatisticsPopup.ts
```diff
@@ -40,5 +75,44 @@ export class ProjectStatisticsPopup {
     totalSignalCount: async (count: number): Promise<void> => {
       await expect(this.totalSignalCount).toContainText(count.toString());
     },
+    attributionPropertiesIsVisible: async (): Promise<void> => {
```
Hellgartner, line 78: Normally, I would not expect further getBy or similar in these expect statements ...

Hellgartner (commented):
The new popup looks great, testing it I found the following issues:

* On smaller sizes the graphs totally break
   image
   Suggestion: Make the layout responsive and reduce the number of columns for smaller screen sizes, introduce scrolling if needed (probably) (I can help with that if required)

* The headlines are for my feeling a tard hard to read and a bit small (personal opinion)

* The different graphs might get problems with really long license/config names, e.g. 
image

PhilippMa:
> The new popup looks great, testing it I found the following issues:
> 
> * On smaller sizes the graphs totally break
>   image
>   Suggestion: Make the layout responsive and reduce the number of columns for smaller screen sizes, introduce scrolling if needed (probably) (I can help with that if required)
> * The headlines are for my feeling a tard hard to read and a bit small (personal opinion)
> * The different graphs might get problems with really long license/config names, e.g.
>   image

* I made the grid responsive now, so hopefully the graphs should not break anymore
* Which headlines do you mean? The titles of the charts (e.g. "Most Frequent Licenses")?
* I made sure that the color icons don't get squished when names get longer in the legend, is it better to keep it that long names are displayed over multiple lines or should we use text with ellipsis here?

src/Frontend/Components/NotificationPopup/NotificationPopup.tsx
```diff
@@ -30,6 +30,7 @@ interface NotificationPopupProps {
   customAction?: React.ReactNode;
   background?: keyof typeof OpossumColors;
   width?: React.CSSProperties['width'];
+  height?: React.CSSProperties['height'];
```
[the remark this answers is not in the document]
PhilippMa: Possible, I am just not quite sure how what would be good naming for the props then, as we already have an sx prop that gets passed to the MuiDialogContent and the width/height styles are passed through the slotProps of the MuiDialog, so we would have two sx props that are applied to two different subcomponents

src/Frontend/Components/ProjectStatisticsPopup/ProjectStatisticsPopup.tsx
```diff
@@ -118,80 +107,86 @@ export const ProjectStatisticsPopup: React.FC = () => {
...
+            selectedTab={selectedTab}
+          >
+            <ChartCard>
+              <MuiTypography variant="subtitle1">
+                {text.projectStatisticsPopup.charts.attributionProperties.title}
+              </MuiTypography>
+              <BarChart data={attributionBarChartData} />
+            </ChartCard>
+            <ChartCard>
+              {mostFrequentLicenseCountData.length > 0 ? (
```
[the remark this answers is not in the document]
PhilippMa: True, I'm not sure why I was only hiding the title :thinking:

src/Frontend/Components/ProjectStatisticsPopup/ProjectStatisticsPopup.tsx
```diff
@@ -214,12 +209,17 @@ export const ProjectStatisticsPopup: React.FC = () => {
...
 }
 
 const TabPanel: React.FC<TabPanelProps> = (props) => {
   return (
     <div
-      style={props.selectedTab !== props.tabIndex ? { display: 'none' } : {}}
+      style={{
+        ...props.style,
+        ...(props.selectedTab !== props.tabIndex ? { display: 'none' } : {}),
+        flexGrow: 1,
```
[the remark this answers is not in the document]
PhilippMa: After a few recent changes there is no need anymore to receive style through props for the TabPanel, so all styles are directly collected here now

src/Frontend/Components/ProjectStatisticsPopup/ProjectStatisticsPopup.tsx
```diff
@@ -105,7 +92,9 @@ export const ProjectStatisticsPopup: React.FC = () => {
   return (
     <NotificationPopup
       content={
-        <>
+        <div
```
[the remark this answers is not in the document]
PhilippMa: Good question, I don't see a specific reason either. Should I change it?

src/e2e-tests/__tests__/project-statistics.test.ts
```diff
@@ -44,6 +48,24 @@ test('opens, displays, and closes project statistics', async ({
   await projectStatisticsPopup.assert.titleIsHidden();
 });
 
+test('displays bar and pie charts in the overview tab', async ({
+  menuBar,
+  projectStatisticsPopup,
+}) => {
+  await menuBar.openProjectStatistics();
+  await projectStatisticsPopup.assert.titleIsVisible();
+  // hover on title to avoid getting tooltips that mess up locators
```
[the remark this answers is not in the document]
PhilippMa: Is that better or just an equal alternative?

src/e2e-tests/page-objects/ProjectStatisticsPopup.ts
PhilippMa: This was the best way I found for selecting the "chart cards" first and then searching for chart text inside of them, when searching on the full popup I was getting some weird multiple findings with some locators

src/e2e-tests/page-objects/ProjectStatisticsPopup.ts
PhilippMa: Should I then make the locators for the class fields more specific instead?

src/Frontend/Components/ProjectStatisticsPopup/ProjectStatisticsPopup.tsx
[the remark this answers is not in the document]
PhilippMa: I forgot about that, switched to Mui boxes now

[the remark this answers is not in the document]
PhilippMa: Let's do this in a separate task

src/Frontend/Components/ProjectStatisticsPopup/ProjectStatisticsPopup.tsx
```diff
@@ -218,10 +213,65 @@ interface TabPanelProps extends React.PropsWithChildren {
...
+      theme={createTheme({
+        ...useTheme(),
+        breakpoints: {
+          values: {
+            xs: 0,
+            sm: 0,
+            md: 1300,
+            lg: 1850,
+            xl: 2000,
+          },
```
Hellgartner, line 240: These breakpoints are somewhat magical. Probably through experimentation?
Looking at the mui it looks like that this is the recommended way to do it. Still it feels quite old fashioned to me.

src/Frontend/Components/ProjectStatisticsPopup/ProjectStatisticsPopup.tsx
```diff
@@ -218,10 +213,65 @@ interface TabPanelProps extends React.PropsWithChildren {
...
+        alignContent={'flex-start'}
+      >
+        {props.children}
+      </MuiGrid>
+    </ThemeProvider>
+  );
+};
+
+interface ChartGridItemProps extends PropsWithChildren {
+  isHidden?: boolean;
```
Hellgartner, line 261: nit: It actually is not hidden, it is not rendered at all.

src/Frontend/Components/ProjectStatisticsPopup/ProjectStatisticsPopup.tsx
```diff
@@ -218,10 +213,65 @@ interface TabPanelProps extends React.PropsWithChildren {
...
+          },
+        },
+      })}
+    >
+      <MuiGrid
+        height={'100%'}
+        minHeight={'fit-content'}
+        container
+        columns={{ sm: 1, md: 2, lg: 3 }}
+        spacing={{ sm: 3, lg: 3 }}
```
Hellgartner, line 249: why is the spacing missing the `md` breakpoint?

src/Frontend/Components/ProjectStatisticsPopup/ProjectStatisticsPopup.tsx
PhilippMa: Yes, I manually tried to find the smallest screen widths where the charts wouldn't break for 3 and 2 columns. If there is a better way to do this I would be interested

Hellgartner: I do not think so with MUI

Hellgartner (approved):

PR #2839, opened 2025-03-07
fix: adjust casing for Progress Bar Select

Hellgartner:
### Summary of changes

Adjust casing of Classification Bar in Dropdown

To make sure it is consistent

Open the select next to the toolbar

### DIFF
--- a/src/shared/text.ts
+++ b/src/shared/text.ts
@@ -367,8 +367,8 @@ export const text = {
         ariaLabel: 'Progress bar attribution progress',
       },
       classificationProgressBar: {
-        selectLabel: 'Signals by classification',
-        ariaLabel: 'Progress bar classification',
+        selectLabel: 'Signals by Classification',
+        ariaLabel: 'Progress bar Classification',
       },
     },
   },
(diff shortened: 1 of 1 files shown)


src/shared/text.ts
```diff
@@ -367,8 +367,8 @@ export const text = {
         ariaLabel: 'Progress bar attribution progress',
       },
       classificationProgressBar: {
-        selectLabel: 'Signals by classification',
-        ariaLabel: 'Progress bar classification',
```
abraemer, line 371: I think we could take things a bit further and improve the existing name as well. Also I think that Sentence Case would be preferred generally but I checked and other menus use Title Case as well.
Suggestion:
```
selectAriaLabel: 'ProgressBar Selector',
      criticalSignalsBar: {
        selectLabel: 'Signals by Criticality',
        ariaLabel: 'Progress bar showing signals by criticality',
      },
      attributionProgressBar: {
        selectLabel: 'Resources by Attribution Status',
        ariaLabel: 'Progress bar showing attribution progress',
      },
      classificationProgressBar: {
        selectLabel: 'Signals by Classification',
        ariaLabel: 'Progress bar showing signals by classification',
```
I noticed that it is unclear to me whether `attributionProgressBar` and the "signal-based` progress bars perform fundamentally different accounting. Is it true that the total of the `attributionProgressBar` is the number of resources but the total of the other 2 bar is the number of signals (on resources without attribution)?

abraemer (changes requested):

abraemer (approved):

PR #2840, opened 2025-03-07
Better color scale for classifications

Hellgartner:
Goal: Being able to react on the classification

### Summary of changes

A new color scale for showing classifications in 
* the progress bar
* the icons ("C")
* the statistics bar
Note: Only the first item is already implemented in this PR

Current color scale does not match the design

Look at the classifications progress bar


src/Frontend/Components/ProgressBar/ProgressBar.util.tsx
```diff
@@ -232,33 +232,46 @@ function interpolateFromRedToGreen(
   return `hsl(${hue.toFixed(0)} 100 45)`;
 }
 
+function roundPercentagesToAtLeastOnePercentAndNormalize(
+  progressBarSteps: Array<ProgressBarStep>,
+): Array<ProgressBarStep> {
+  //percentages and steps need to be in parallel here
```
mstykow, line 238: not understanding this comment. what does it mean "to be in parallel". do we need a comment at all here or is it clear from the control flow what's going on?

src/Frontend/Components/ProgressBar/ProgressBar.util.tsx
```diff
@@ -232,33 +232,46 @@ function interpolateFromRedToGreen(
...
+  let percentages = progressBarSteps.map((step) => step.widthInPercent);
+  percentages = roundToAtLeastOnePercentAndNormalize(percentages);
```
mstykow, line 240: i don't think we need a mutable variable here:
```suggestion
  const percentages = roundToAtLeastOnePercentAndNormalize(progressBarSteps.map((step) => step.widthInPercent));
```

src/Frontend/Components/ProgressBar/ProgressBar.util.tsx
```diff
@@ -232,33 +232,46 @@ function interpolateFromRedToGreen(
...
+  for (let i = 0; i < percentages.length; i++) {
+    progressBarSteps[i].widthInPercent = percentages[i];
+  }
+  return progressBarSteps;
```
mstykow, line 244: here too, i think we can avoid the mutable variable by using `progressBarSteps.map`.

src/Frontend/Components/ProgressBar/ProgressBar.util.tsx
```diff
@@ -232,33 +232,46 @@ function interpolateFromRedToGreen(
...
-
   const numberOfClassifications = Object.keys(classificationStatistics).length;
-  const progressBarSteps = percentages.map((percentage, index) => {
-    return {
-      widthInPercent: percentage,
-      color: interpolateFromRedToGreen(numberOfClassifications, index),
-    };
+  const progressBarSteps = Object.entries(classificationStatistics)
+    .reverse()
+    .map(([classifictionNumericValue, statisticsEntry], index) => {
```
mstykow, line 254: untyped object, please use a generic: `.map<GENERIC>(...`

src/Frontend/Components/ProgressBar/ProgressBar.util.tsx
```diff
@@ -232,33 +232,46 @@ function interpolateFromRedToGreen(
...
+      return {
+        widthInPercent:
+          (statisticsEntry.correspondingFiles.length * 100) /
+          progressBarData.filesWithOnlyExternalAttributionCount,
+        color:
+          (classifictionNumericValue as unknown as number) === 0
```
mstykow, line 260:
```suggestion
          !Number(classifictionNumericValue)
```

src/Frontend/Components/ProgressBar/ProgressBar.util.tsx
mstykow, line 254:
```suggestion
    .map(([classificationNumericValue, statisticsEntry], index) => {
```

src/Frontend/Components/ProgressBar/__tests__/ProgressBar.util.test.ts
```diff
@@ -99,7 +99,7 @@ describe('ProgressBar helpers', () => {
       const background = getClassificationBarBackground(testProgressBarData);
 
       const expectedBackground =
-        'linear-gradient(to right, hsl(0 100 45) 0% 5% , hsl(49 100 45) 5% 25% , hsl(97 100 45) 25% 40% , hsl(146 100 45) 40% 65% , hsl(220, 41%, 92%) 65% 100% )';
+        'linear-gradient(to right, #ff0000 0% 5% , #ff8565 5% 25% , #ffc6b3 25% 40% , hsl(146, 50%, 80%) 40% 65% , hsl(220, 41%, 92%) 65% 100% )';
```
mstykow, line 102: do these tests still make sense? since these values are calculated by a library this test assertion now smells like a snapshot test.

mstykow (approved):

src/Frontend/Components/ProgressBar/__tests__/ProgressBar.util.test.ts
Hellgartner: I also thought about this ... 
The problem is that at the end we do get a string representation here.
So for me this tests two things:
* the correct percentages
* That we do have color overrides for the last two items (the green and the blue)
* Most importantly: That the linear gradient string is assembled correctly 

What alternative did I consider:
* Testing the function `createBackgroundFromProgressBarSteps` and `createBackgroundFromProgressBarSteps` separately. I did not do so as this would have been an implementation detail at this point
* Mocking the call to the external lib, but then you do have more mocks and at the end would still need to do string comparison

PR #2851, opened 2025-03-12
Fix criticality icons

PhilippMa:
### Summary of changes

Fix bug where criticality icons where not showing in the attribution and signal lists.

Bug fix

Check that the criticality icons are again displayed correctly in the attribution and signal lists.

### DIFF
--- a/src/Frontend/Components/PackageCard/PackageCard.tsx
+++ b/src/Frontend/Components/PackageCard/PackageCard.tsx
@@ -107,7 +107,7 @@ export interface PackageCardConfig {
 
 export interface PackageCardProps {
   packageInfo: PackageInfo;
-  cardConfig?: PackageCardConfig;
+  cardConfig?: Partial<PackageCardConfig>;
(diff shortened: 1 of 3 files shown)


src/Frontend/Components/PackageCard/PackageCard.tsx
```diff
@@ -107,7 +107,7 @@ export interface PackageCardConfig {
 
 export interface PackageCardProps {
   packageInfo: PackageInfo;
-  cardConfig?: PackageCardConfig;
+  cardConfig?: Partial<PackageCardConfig>;
```
Hellgartner, line 110: This makes me nervous TBH, 
IMHO that means we should build two types:

What about 
* PackageCardConfig without criticality
* EffectiveCardConfig: PackageCardConfig & {criticality: Criticality}

Hellgartner (changes requested):

Hellgartner (approved):

Issue #2852, opened 2025-03-12
Make statistics popup use the same color as the progress bar


Hellgartner:
For consistency, use the same color palette everywhere

PR #2855, opened 2025-03-14
feat: statistics popup and progress bar share same colors

Hellgartner:
### Summary of changes

Make the progress bar colors and the statistics popup colors match

closes https://github.com/opossum-tool/OpossumUI/issues/2852

Have a look at the statistics popup

### DIFF
--- a/src/Frontend/Components/ProjectStatisticsPopup/ProjectStatisticsPopup.util.ts
+++ b/src/Frontend/Components/ProjectStatisticsPopup/ProjectStatisticsPopup.util.ts
@@ -13,7 +13,7 @@ import {
   PackageInfo,
 } from '../../../shared/shared-types';
 import { text } from '../../../shared/text';
-import { OpossumColors } from '../../shared-styles';
+import { classificationUnknownColor } from '../../shared-styles';
 import {
   AttributionCountPerSourcePerLicense,
   ChartDataItem,
@@ -378,7 +378,7 @@ export function getSignalCountByClassification(
       name: text.projectStatisticsPopup.charts
         .signalCountByClassificationPieChart.noClassification,
       count: classificationCounts[NO_CLASSIFICATION],
-      color: OpossumColors.lightestBlue,
+      color: classificationUnknownColor,
     });
   }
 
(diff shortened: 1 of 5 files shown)


src/Frontend/Components/ProjectStatisticsPopup/ProjectStatisticsPopup.util.ts
```diff
@@ -356,27 +359,37 @@ export function getSignalCountByClassification(
...
     })
     .filter(({ count }) => count > 0);
 
   if (classificationCounts[NO_CLASSIFICATION]) {
-    return pieChartData.concat({
+    pieChartData.push({
       name: text.projectStatisticsPopup.charts
         .signalCountByClassificationPieChart.noClassification,
       count: classificationCounts[NO_CLASSIFICATION],
+      color: OpossumColors.lightestBlue,
```
PhilippMa, line 381: We need to either change this color or the background color behind the charts, as it is currently the same color so this pie chart segment is not visible
image

src/Frontend/Components/ProjectStatisticsPopup/ProjectStatisticsPopup.util.ts
```diff
@@ -356,27 +359,37 @@ export function getSignalCountByClassification(
...
     });
   }
 
-  return pieChartData;
+  const colorMap = Object.fromEntries(
+    pieChartData.map((colorChartEntry) => [
+      colorChartEntry.name,
+      colorChartEntry.color,
+    ]),
+  );
```
PhilippMa, line 390: I don't really see a reason to include color information in pieChartData and then compute the colorMap in a second step here, couldn't you just compute the colorMap separately from pieChartData through a separate map over Object.keys(classifications)?

PhilippMa (changes requested):

src/Frontend/Components/ProjectStatisticsPopup/ProjectStatisticsPopup.util.ts
Hellgartner: TBH my preferred option would be to just use the pieChartData with color but that is not supported by the interface of the PieChart. 

I just do not like implicitly relying on "things just having the same index"

PhilippMa (approved):

PR #2856, opened 2025-03-14
Chore/user config handling and performance fixes

Hellgartner:
### Summary of changes

Exchange the current setup for using electron settings in the frontend with
* Make sure all of the configuration are always set, this reduces the need for defaults
* Do not use the variables section but create a dedicated redux store for user config
* Sync this store directly at frontend startup -- before any rendering happens
* Register _one_ central handler which forwards any user settings changes to the store
* Using user settings is then only a normal use on the store
* Note that writing user settings from the frontend still requires and IPC call to the backend.

The current setup for the user configuration does not scale to the higher number of uses.
Using `useUserSetting` leads to a lot of 
``` typescript
useIpcRenderer(
    AllowedFrontendChannels.UserSettingsChanged, ...
```
handlers being registered. This has a few unpleasant consequences:
* On _any_ change of _any_ user setting all these handlers fire.
* All these handlers cause a write to the redux store and thus re-renders.
* Hooks for these variables are used in the `App.tsx` causing re-renders of the entire app on change

This makes even handling small files hard to work with and makes the app crash on larger, more realistic, files.

Full regression test


src/shared/ipc-channels.ts
```diff
@@ -18,6 +18,7 @@ export enum IpcChannel {
    */
   StopLoading = 'stop-loading',
   GetUserSettings = 'get-user-settings',
```
mstykow, line 20: the IPC channel for getting a single user setting seems unused now. i would say we actually keep this one and instead make it have the functionality of `GetFullUserSettings`.

src/Frontend/state/selectors/user-settings-selector.ts
mstykow, line 1: does it really make things easier to have one selector per setting? this way, i have to know exactly the name of what i'm looking for. it seems to be it would be easier to have a single selector for the user settings and then just do `settings.MY_SETTING` which comes with autocomplete.

src/Frontend/state/variables/use-user-setting.ts
```diff
@@ -2,92 +2,31 @@
...
+  UserSettingsChangedListener,
+} from '../../util/use-ipc-renderer';
+import {
+  fetchUserSettings,
+  setUserSetting,
+} from '../actions/user-settings-actions/user-settings-actions';
+import { useAppDispatch } from '../hooks';
+
+// should only be called once
+export function useInitialSyncUserSettings() {
```
mstykow, line 20: hook name no longer matches file name. but as mentioned in my other comment: perhaps we still want a single `useUserSettings` hook.

src/Frontend/state/actions/user-settings-actions/user-settings-actions.ts
```diff
@@ -0,0 +1,63 @@
...
+    const currentPanelSizes = getPanelSizes(currentState);
+    dispatch(
+      updateUserSettings({
+        panelSizes: { ...currentPanelSizes, ...panelSizes },
+      }),
+    );
+  };
+}
+
+export function toggleAreHiddenSignalsVisible(
```
mstykow, line 51: you are breaking naming convention. i would have expected "updateAreHiddenSignalsVisible". also, why do only some user settings have a dedicated action while others do not?
suggestion: create only a single action that can be used to update any user setting. then use this action in the single hook for updating any user setting.

src/Frontend/Components/AttributionForm/AuditingOptions/AuditingOptions.util.tsx
```diff
@@ -58,14 +58,14 @@ export function useAuditingOptions({
 }) {
   const dispatch = useAppDispatch();
   const store = useAppStore();
-  const [qaMode] = useUserSetting({ key: 'qaMode' });
+  const qaMode = useAppSelector(getQaMode);
```
mstykow, line 61: it's strange to me that for some user settings we have dedicated hooks for getting and setting but other user settings we call `useAppSelector`. suggestion: have a single hook for getting and setting all user settings such as `useUserSettings`.

src/Frontend/state/actions/user-settings-actions/__tests__/user-settings-actions.test.ts
mstykow, line 1: not a great fan of redux action tests. better to write integration or e2e tests.

src/shared/shared-constants.ts
mstykow, line 1: not sure about the introduction of this file. isn't there some other place to share constants between frontend and backend? if not, why is your feature so special that it's the first one that needs this?

mstykow (changes requested):

src/ElectronBackend/main/user-settings-provider.ts
```diff
@@ -0,0 +1,74 @@
...
+import settings from 'electron-settings';
+
+import { AllowedFrontendChannels } from '../../shared/ipc-channels';
+import { DEFAULT_USER_SETTINGS } from '../../shared/shared-constants';
+import {
+  UserSettings as IUserSettings,
+  UserSettings,
+} from '../../shared/shared-types';
+
+export class UserSettingsProvider {
```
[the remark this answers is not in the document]
Hellgartner: Concerning 1)
It does for me, both in IntelliJ as well as in GitHub
image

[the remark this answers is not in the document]
Hellgartner: 2. Take UserSettingsService

src/ElectronBackend/main/__tests__/user-settings-provider.test.ts
mstykow, line 1: file is still called "user-settings-provider"

src/ElectronBackend/main/__tests__/user-settings-provider.test.ts
```diff
@@ -0,0 +1,151 @@
...
+            },
+          },
+        ];
+      },
+      sendFunction,
+    },
+  };
+});
+
+describe('UserSettings', () => {
```
mstykow, line 41:
```suggestion
describe('UserSettingsService', () => {
```

src/ElectronBackend/main/user-settings-service.ts
```diff
@@ -0,0 +1,74 @@
...
+  public static get<T extends keyof IUserSettings>(
+    path?: T,
+  ): Promise<IUserSettings[T]> | Promise<IUserSettings> {
+    if (path) {
+      return settings.get(path) as Promise<IUserSettings[T]>;
+    }
+    return settings.get() as unknown as Promise<IUserSettings>;
+  }
+
+  public static async update(
```
mstykow, line 43: i'm confused why we need both the .update and the .set methods. can we eliminate one of the two and do all setting/update operations through a single path?

src/Frontend/Components/PackageCard/PackageCard.tsx
```diff
@@ -143,8 +142,9 @@ export const PackageCard = memo(
       }),
       [cardConfig, packageInfo, classification_mapping],
     );
-    const [showClassifications] = useShowClassifications();
-    const [showCriticality] = useShowCriticality();
+    const [userSettings, _] = useUserSettings();
```
mstykow, line 145:
```suggestion
    const [userSettings] = useUserSettings();
```

src/Frontend/state/variables/use-panel-sizes.ts
```diff
@@ -2,19 +2,18 @@
...
-    key: 'panelSizes',
-  });
+export const usePanelSizes = (): {
+  panelSizes: PanelSizes;
+  setPanelSizes: (panelsSizes: Partial<PanelSizes>) => void;
+} => {
+  const [userSettings, updateUserSettings] = useUserSettings();
+  const panelSizes = userSettings.panelSizes;
+  const setPanelSizes = (panelSizes: Partial<PanelSizes>) =>
+    updateUserSettings((currentSettings: UserSettings) => ({
```
mstykow, line 15: please don't repeat inferred types as this makes refactorings harder. comment applies in several other places as well.
```suggestion
    updateUserSettings((currentSettings) => ({
```

src/Frontend/state/actions/user-settings-actions/user-settings-actions.ts
```diff
@@ -0,0 +1,48 @@
...
+  getState: () => State,
+): Partial<UserSettings> {
+  if (typeof userSettings === 'function') {
+    const currentUserSettings: UserSettings = getUserSettings(getState());
+    return userSettings(currentUserSettings);
+  }
+  return userSettings;
+}
+
+export function updateUserSettings(
```
mstykow, line 38: i find it a bit misleading that the action is called "updateUserSettings" when it calls the electron API for "setUserSettings". this adds to the confusion that there is both the "set" and the "update" function. hopefully we can eliminate one of the two and then consistently use either "set" or "update" everywhere.

src/Frontend/state/actions/user-settings-actions/user-settings-actions.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+function getUserSettingsToSet(
+  userSettings:
+    | Partial<UserSettings>
+    | ((currentSettings: UserSettings) => Partial<UserSettings>),
+  getState: () => State,
+): Partial<UserSettings> {
+  if (typeof userSettings === 'function') {
+    const currentUserSettings: UserSettings = getUserSettings(getState());
```
mstykow, line 32: here too: please don't repeat types that are already specified and inferred
```suggestion
    const currentUserSettings = getUserSettings(getState());
```

src/Frontend/state/actions/user-settings-actions/user-settings-actions.ts
```diff
@@ -0,0 +1,48 @@
...
+  }
+  return userSettings;
+}
+
+export function updateUserSettings(
+  userSettings:
+    | Partial<UserSettings>
+    | ((currentSettings: UserSettings) => Partial<UserSettings>),
+): AppThunkAction {
+  return async (dispatch: AppThunkDispatch, getState: () => State) => {
```
mstykow, line 43: here too: types which are already inferred by the return type of the function
```suggestion
  return async (dispatch, getState) => {
```

src/Frontend/state/variables/use-user-setting.ts
```diff
@@ -2,92 +2,50 @@
...
+import {
+  fetchUserSettings,
+  setUserSetting,
+  updateUserSettings as updateUserSettingsThunkAction,
+} from '../actions/user-settings-actions/user-settings-actions';
+import { useAppDispatch, useAppSelector } from '../hooks';
+import { getUserSettings } from '../selectors/user-settings-selector';
+
+// should only be called once
+export function useInitialSyncUserSettings() {
```
mstykow, line 22: how about "useInitUserSettings" and delete the comment?

src/Frontend/state/variables/use-user-setting.ts
```diff
@@ -2,92 +2,50 @@
...
-      }),
-    [readStoredValue],
+    (_, updatedSettings: Partial<UserSettings>) =>
+      dispatch(setUserSetting(updatedSettings)),
+    [dispatch],
   );
+}
 
-  return [storedValue, setStoredValue, hydrated];
+type PartialUserSettings =
```
mstykow, line 36: to me, this type name is not accurate since i would not expect a function here. you can save yourself the naming and creation of a type by inlining `updateUserSettings`. this then also avoids the need to rename the import to `updateUserSettingsThunkAction`. so something like this:
```
export function useUserSettings(): [
  UserSettings,
  (userSettings: Partial<UserSettings> | ((settings: UserSettings) => Partial<UserSettings>)) => void,
] {
  const userSettings = useAppSelector(getUserSettings);

  return [userSettings, (userSettings) => dispatch(updateUserSettings(userSettings))];
}
```

mstykow (changes requested):

src/ElectronBackend/main/main.ts
```diff
@@ -92,11 +93,11 @@ export async function main(): Promise<void> {
...
-    ipcMain.handle(IpcChannel.GetUserSettings, (_, key) =>
-      UserSettings.get(key),
-    );
-    ipcMain.handle(IpcChannel.SetUserSettings, (_, { key, value }) =>
-      UserSettings.set(key, value, { skipNotification: true }),
+    ipcMain.handle(IpcChannel.GetUserSettings, () => UserSettingsService.get());
+    ipcMain.handle(
+      IpcChannel.SetUserSettings,
+      (_, userSettings: Partial<UserSettings>) =>
+        UserSettingsService.update(userSettings, true),
```
mstykow, line 100: more confusion: channel "SetUserSettings" calls "update" instead of the expected "set" on the service.

mstykow (changes requested):
This diff now mixes in changes for classification and criticality that in previous versions were not present. This makes it harder to do a review and I don't think any of this is intended.

src/Frontend/Components/AttributionPanels/SignalsPanel/IncludeExcludeButton/IncludeExcludeButton.tsx
```diff
@@ -22,7 +23,11 @@ export const IncludeExcludeButton: React.FC = () => {
     <MuiIconButton
       aria-label={label}
       size={'small'}
-      onClick={() => setAreHiddenSignalsVisible((prev) => !prev)}
+      onClick={() => {
+        updateUserSettings((currentSettings: UserSettings) => ({
```
mstykow, line 27:
```suggestion
        updateUserSettings((currentSettings) => ({
```

src/Frontend/state/actions/user-settings-actions/__tests__/user-settings-actions.test.ts
mstykow, line 1: aren't we testing this - and better - via E2E?

src/Frontend/state/variables/use-user-setting.ts
mstykow: you resolved this comment but neither rejected it with a reason nor adopted it.

src/Frontend/state/variables/use-user-setting.ts
```diff
@@ -2,92 +2,54 @@
...
+      | ((settings: UserSettings) => Partial<UserSettings>),
+  ) => void,
+] {
+  const userSettings = useAppSelector(getUserSettings);
+  const dispatch = useAppDispatch();
+
+  const updateUserSettings = (
+    userSettings:
+      | Partial<UserSettings>
+      | ((settings: UserSettings) => Partial<UserSettings>),
```
mstykow, line 50: this duplicates the type definition from lines 39ff. if you inline your `updateUserSettings`, it won't be necessary to specify types a second time.

src/shared/shared-types.ts
```diff
@@ -318,16 +313,18 @@ export interface Log {
...
   panelSizes: {
     resourceBrowserWidth: number;
     packageListsWidth: number;
     linkedResourcesPanelHeight: number | null;
     signalsPanelHeight: number | null;
-  } | null;
-  recentlyOpenedPaths: Array<string> | null;
+  };
+  recentlyOpenedPaths: Array<string>;
 }
```
mstykow, line 328: nice improvement 👍

mstykow (changes requested):
I think we're almost good to go now.

mstykow:
Not sure onto which PR this comment belongs but the table in the statistics popup doesn't obey the settings yet:
image

src/Frontend/state/actions/user-settings-actions/__tests__/user-settings-actions.test.ts
Hellgartner: I thought about this.
Actually these are very hard to test via E2E: There are two tests there:
* The initial load from the backend would be flaky to test from e2e as we do not have a good possibility to test the what is arriving there and what the current default values are
* Updating the backend settings from the frontend is also difficult as we do not have (at least not to my knowledge) any setting which could be set from frontend and then easily tested by an e2e test.

mstykow (approved):

PR #2857, opened 2025-03-17
feat: introduce open recent

### LINKED ISSUE #2785: Open recent projects

As a OpossumUI user who works on the same file more than once
I want to be able to open my most recent files directly from the menu
In order to not to waste time selecting it from the file system.

Take care of edge cases. For example, you may want to limit the number of recent projects to 10 so that the list does not eventually become huge.

There is already a library which handles interaction with the Electron app settings, e.g., when setting the checkmark which determines if you will see the project statistics popup on launch.

None at that point

**Additional context**
None

mstykow:
### Summary of changes

This PR allows users to chose from up to 10 recently opened files. It also creates a central method for creating/rebuilding the electron menu and uses it to simplify some code around QA mode.

Closes #2785 

Check out the new option in the file menu.


src/ElectronBackend/main/iconHelpers.ts
```diff
@@ -23,18 +23,3 @@ export function getIconBasedOnTheme(
     ? path.join(getBasePathOfAssets(), white_icon)
     : path.join(getBasePathOfAssets(), black_icon);
 }
-
-export function makeFirstIconVisibleAndSecondHidden(
```
Hellgartner, line 27: That change will cause me some merge conflicts but nothing you need to take care of

src/ElectronBackend/main/listeners.ts
```diff
@@ -130,6 +135,10 @@ export async function handleOpeningFile(
 
   await openFile(mainWindow, filePath, onOpen);
 
+  await updateRecentlyOpenedPaths(filePath);
+
+  await createMenu(mainWindow);
```
Hellgartner, line 140: would it be better to 
* make `updateRecentlyOpenedPaths` to call createMenu internally
* Probably this would mean that `updateRecentlyOpenedPaths` needs to be renamed to `updateRecentlyOpenedPathsMenu` or similar

This would ensure the createMenu call cannot be forgotten

src/ElectronBackend/main/listeners.ts
```diff
@@ -351,6 +360,18 @@ export async function openFile(
   onOpen();
 }
 
+async function updateRecentlyOpenedPaths(filePath: string): Promise<void> {
```
Hellgartner, line 363: This function probably warrants a test

src/ElectronBackend/main/menu/fileMenu.ts
```diff
@@ -43,7 +47,9 @@ export const importFileFormats: Array<FileFormatInfo> = [
   },
 ];
 
-function getOpenFile(mainWindow: Electron.CrossProcessExports.BrowserWindow) {
+function getOpenFile(
+  mainWindow: Electron.CrossProcessExports.BrowserWindow,
```
Hellgartner, line 51: Why not just use the imported BrowserWindow from `electron`. Same below?

src/ElectronBackend/main/menu/fileMenu.ts
```diff
@@ -55,7 +61,45 @@ function getOpenFile(mainWindow: Electron.CrossProcessExports.BrowserWindow) {
...
-function getImportFile(mainWindow: Electron.CrossProcessExports.BrowserWindow) {
+async function getOpenRecent(
+  mainWindow: Electron.CrossProcessExports.BrowserWindow,
+): Promise<MenuItemConstructorOptions> {
+  const recentlyOpenedPaths = await UserSettings.get('recentlyOpenedPaths');
+
+  return {
+    icon: getIconBasedOnTheme('icons/open-white.png', 'icons/open-black.png'),
+    label: text.menu.fileSubmenu.openRecent,
+    submenu: recentlyOpenedPaths
```
Hellgartner, line 72: Maybe extract the sub-menu into its own function. That would also allow to avoid the use of the ternary operator which (IMHO) is a bit hard to read here

Hellgartner (changes requested):

src/ElectronBackend/main/listeners.ts
mstykow: i created an e2e test now.

src/ElectronBackend/main/menu/fileMenu.ts
```diff
@@ -55,7 +64,51 @@ function getOpenFile(mainWindow: Electron.CrossProcessExports.BrowserWindow) {
...
+    submenu: getOpenRecentSubmenu(mainWindow, recentlyOpenedPaths),
+    enabled: !!recentlyOpenedPaths?.length,
+  };
+}
+
+function getOpenRecentSubmenu(
+  mainWindow: BrowserWindow,
+  recentlyOpenedPaths: Array<string> | null,
+): MenuItemConstructorOptions['submenu'] {
+  if (!recentlyOpenedPaths) {
```
Hellgartner, line 84: You probably want `!!recentlyOpenedPaths?.length` as an if condition here as `[]` is truthy in ts/js.
Note that this is currently no visible bug as it is hidden by disabling the menu.

Hellgartner (approved):

Issue #2849, opened 2025-03-12
Make display of classification and criticality optional


Hellgartner:
**Is your feature request related to a problem? Please describe.**
To reduce clutter, make the display of criticality and classification optional

**Describe the solution you'd like**
* Add a two menu items in the view menu which allow toggling display of
  * Criticality (default true)
  * Classification (default true)
* This should survive a restart of OpossumUI --> use a UserSetting here
* Edge case: What if criticality progressBar/sorting is selected and we disable the display

PR #2861, opened 2025-03-19
feat: respect settings for showing also in overview table

Hellgartner:
### Summary of changes

Also adapt the statistics table based on the configuration for showing classifications and criticality

See https://github.com/opossum-tool/OpossumUI/issues/2849

Open the statistics popup and toggle the classifications/criticality on and off

### DIFF
--- a/src/ElectronBackend/main/createWindow.ts
+++ b/src/ElectronBackend/main/createWindow.ts
@@ -9,13 +9,19 @@ import upath from 'upath';
 
 import { getIconPath } from './iconHelpers';
 
+const openDevTools = (mainWindow: BrowserWindow) => {
+  const devtools = new BrowserWindow();
+  mainWindow.webContents.setDevToolsWebContents(devtools.webContents);
+  mainWindow.webContents.openDevTools({ mode: 'detach' });
+};
+
 export async function loadWebApp(
   mainWindow: Electron.CrossProcessExports.BrowserWindow,
 ) {
   if (!app.isPackaged) {
     await mainWindow.loadURL('http://localhost:5173/');
 
-    mainWindow.webContents.openDevTools();
+    openDevTools(mainWindow);
   } else {
     await mainWindow.loadURL(
       `file://${path.join(upath.toUnix(__dirname), '../../index.html')}`,
--- a/src/ElectronBackend/main/createWindow.ts
+++ b/src/ElectronBackend/main/createWindow.ts
@@ -9,19 +9,13 @@ import upath from 'upath';
 
 import { getIconPath } from './iconHelpers';
 
-const openDevTools = (mainWindow: BrowserWindow) => {
-  const devtools = new BrowserWindow();
-  mainWindow.webContents.setDevToolsWebContents(devtools.webContents);
-  mainWindow.webContents.openDevTools({ mode: 'detach' });
-};
-
 export async function loadWebApp(
   mainWindow: Electron.CrossProcessExports.BrowserWindow,
 ) {
   if (!app.isPackaged) {
     await mainWindow.loadURL('http://localhost:5173/');
 
-    openDevTools(mainWindow);
+    mainWindow.webContents.openDevTools();
   } else {
     await mainWindow.loadURL(
       `file://${path.join(upath.toUnix(__dirname), '../../index.html')}`,
--- a/src/Frontend/Components/AttributionCountPerSourcePerLicenseTable/AttributionCountPerSourcePerLicenseTable.tsx
+++ b/src/Frontend/Components/AttributionCountPerSourcePerLicenseTable/AttributionCountPerSourcePerLicenseTable.tsx
@@ -136,9 +136,21 @@ export const AttributionCountPerSourcePerLicenseTable: React.FC<
   );
 
   const [ordering, setOrdering] = useState<TableOrdering>(defaultOrdering);
+  const effectiveOrdering = columnConfig.getColumnById(ordering.orderedColumn)
+    ? ordering
+    : defaultOrdering;
 
   const handleRequestSort = (columnId: string, defaultOrder: Order) => {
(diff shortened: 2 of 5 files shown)


src/Frontend/Components/AttributionCountPerSourcePerLicenseTable/AttributionCountPerSourcePerLicenseTable.tsx
```diff
@@ -119,32 +152,37 @@ export const AttributionCountPerSourcePerLicenseTable: React.FC<
     }
   };
 
+  const effectiveOrdering = columnConfig.getColumnById(ordering.orderedColumn)
+    ? ordering
+    : defaultOrdering;
```
PhilippMa, line 157: We either have to do a hard reset here (actually setting the `ordering` state) when a column gets disabled or we need to also make the `handleRequestSort` function use `effectiveOrdering` instead of `ordering`

src/Frontend/Components/AttributionCountPerSourcePerLicenseTable/__tests__/AttributionCountPerSourcePerLicenseTable.test.tsx
```diff
@@ -0,0 +1,149 @@
...
+function expectHeaderTextsToEqual(
+  screen: Screen,
+  expectedHeaderTexts: Array<string>,
+) {
+  const headerTexts = screen
+    .getAllByTestId('classification-table-row-header')
+    .map((node) => node.textContent);
+
+  expect(headerTexts).toEqual(expectedHeaderTexts);
+}
```
PhilippMa, line 57: There are a few test cases where you still do this manually, I assume you just forgot to also switch to this function there

src/Frontend/Components/AttributionCountPerSourcePerLicenseTable/__tests__/AttributionCountPerSourcePerLicenseTable.test.tsx
```diff
@@ -0,0 +1,149 @@
...
+      'SourceA',
+      'SourceB',
+      'Total',
+    ]);
+
+    fireEvent.click(screen.getByText('Criticality'));
+
+    expectHeaderTextsToEqual(screen, [
+      'Name', //correct, the sorted, ascending is for a11y
+      'Criticalitysorted descending',
```
PhilippMa, line 130: Looks like you forgot to adjust the comment here

src/Frontend/Components/TableCellWithSorting/TableCellWithSorting.tsx
```diff
@@ -32,6 +32,7 @@ export const TableCellWithSorting: React.FC<TableCellWithSortingProps> = (
         ...props.sx,
       }}
       sortDirection={props.isSortedColumn ? props.order : false}
+      data-testid="classification-table-row-header"
```
PhilippMa, line 35: I know this is just a test id, but I would prefer a name that doesn't explicitly refer to the license table here, as the component is defined independently of that table and might be reused for other tables in the future.

PhilippMa (changes requested):

src/Frontend/Components/AttributionCountPerSourcePerLicenseTable/AttributionCountPerSourcePerLicenseTable.tsx
```diff
@@ -96,16 +124,33 @@ export const AttributionCountPerSourcePerLicenseTable: React.FC<
...
+    if (
+      effectiveOrdering !== ordering &&
+      effectiveOrdering.orderedColumn === columnId
+    ) {
+      setOrdering({
+        ...effectiveOrdering,
+        orderDirection:
+          effectiveOrdering.orderDirection === 'asc' ? 'desc' : 'asc',
+      });
+    } else if (ordering.orderedColumn === columnId) {
```
PhilippMa, line 153: Can we not just go into the first case regardless of whether effectiveOrdering equals ordering?

src/Frontend/Components/AttributionCountPerSourcePerLicenseTable/__tests__/AttributionCountPerSourcePerLicenseTable.test.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+    expectHeaderTextsToEqual(screen, [
+      'Namesorted ascending', //correct, the sorted, ascending is for a11y
+      'Criticality',
+      'Classification',
+      'SourceA',
+      'SourceB',
+      'Total',
+    ]);
+  });
```
PhilippMa, line 71: Is there a way to get rid of the linter warnings that these tests have no assertions? Maybe it's better to move the expect statement out of the function back into the test cases themselves?

PhilippMa (changes requested):

src/ElectronBackend/main/createWindow.ts
```diff
@@ -9,13 +9,19 @@ import upath from 'upath';
 
 import { getIconPath } from './iconHelpers';
 
+const openDevTools = (mainWindow: BrowserWindow) => {
```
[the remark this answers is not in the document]
Hellgartner: No, thanks

src/Frontend/Components/AttributionCountPerSourcePerLicenseTable/AttributionCountPerSourcePerLicenseTable.tsx
Hellgartner: What I cannot do in the first case is using the function version of `setOrdering`. As we are toggling that ordering this is more safe for fast clicks, thus I wanted to keep it as much as possible.

PR #2868, opened 2025-03-25
fix: remove warn messages from recharts library

Hellgartner:
### Summary of changes

If not shown, do not even render recharts to avoid showing warning in the logs

When on the statistics table tab, recharts shows warnings in the logs
image

* Open the statistics tab
* Switch to the details table
* check js logs to ensure no entries are there

### DIFF
--- a/src/Frontend/Components/ProjectStatisticsPopup/ProjectStatisticsPopup.tsx
+++ b/src/Frontend/Components/ProjectStatisticsPopup/ProjectStatisticsPopup.tsx
@@ -111,7 +111,11 @@ export const ProjectStatisticsPopup: React.FC = () => {
             <MuiTab label={text.projectStatisticsPopup.tabs.overview} />
             <MuiTab label={text.projectStatisticsPopup.tabs.details} />
           </MuiTabs>
-          <TabPanel tabIndex={0} selectedTab={selectedTab}>
+          <TabPanel
+            tabIndex={0}
+            selectedTab={selectedTab}
+            drawNotSelectedTab={false}
+          >
             <ChartGrid>
               <ChartGridItem testId={'attributionBarChart'}>
                 <MuiTypography variant="subtitle1">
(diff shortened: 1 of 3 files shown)


src/Frontend/Components/ProjectStatisticsPopup/ProjectStatisticsPopup.tsx
```diff
@@ -213,13 +221,20 @@ export const ProjectStatisticsPopup: React.FC = () => {
 interface TabPanelProps extends React.PropsWithChildren {
   tabIndex: number;
   selectedTab: number;
+  drawNotSelectedTab: boolean;
```
mstykow, line 224: i find this new prop awkwardly named. in react, we usually speak of "rendering", not "drawing". also, by just looking at the name, i don't get what this does: what is a "not selected" tab? do you mean: "renderWhenInactive"? but then: why do we need to render any tab when it's not active, i.e., can this prop always be set to "false"?

mstykow (changes requested):

src/Frontend/Components/ProjectStatisticsPopup/ProjectStatisticsPopup.tsx
Hellgartner: I really do like the proposed name and adapted the code accordingly

> why do we need to render any tab when it's not active, i.e., can this prop always be set to "false"?

Rendering a tab when it is not visible (i.e. when it is hidden with `display: none`) can be required to keep the state in the component.
In the case at hand, I first tried not rendering both tabs by default. In this case, the `Licenses` tab "forgets" the selected sorting while you are on the overview tab.

Hellgartner: Agreed on solution: 
* The non visible tab is not rendered
* Keeping the state of the table sorting is deferred to follow up tast https://github.com/opossum-tool/OpossumUI/issues/2872

mstykow (approved):

Issue #2869, opened 2025-03-25
Update user guide


Hellgartner:
Update the user guide to explain the new options for classification

PR #2873, opened 2025-03-26
Feat/include classifications to documentation

Hellgartner:
### Summary of changes

Add classifications to the documentation

Update the documentation to keep up with the newest introduced feature
Closes https://github.com/opossum-tool/OpossumUI/issues/2869

Read the USERS_GUIDE.md


USER_GUIDE.md
```diff
@@ -64,24 +64,49 @@ to be recognized by the app.
 To open a `.opossum` file in the app, simply click the _Open File_ button on the left of the top bar (or on the entry in
 the _File_ menu with the same name).
 
+To quickly access recently opened Files, select the relevant file from the `Open Recent` submenu.
```
mstykow, line 67:
```suggestion
To quickly access recently opened files, select the relevant file from the `Open Recent` submenu.
```

USER_GUIDE.md
```diff
@@ -64,24 +64,49 @@ to be recognized by the app.
...
 To import a legacy opossum file (`.json`/`.json.gz`) or a file produced by a scanner (e.g. a ScanCode Json file), click
-the _Import File_ item in the _File_ menu and select the desired file format from the submenu. You will be presented
+the _Import_ item in the _File_ menu and select the desired file format from the submenu. You will be presented
 with a dialog window where you can select the file that you wish to import. Because OpossumUI converts this file into a
 new `.opossum` file as part of the import process, you will also be prompted to select where this new `.opossum` file
 should be saved. Clicking the _Import_ button converts the file into a new `.opossum` file and opens it in OpossumUI.
 Please note that edits made through OpossumUI are **ONLY** saved in this new `.opossum` file, **NOT** in the file you
 selected for import.
 
+### Merging additional information
```
mstykow, line 79: here and below: i think we're using title case throughout this document

USER_GUIDE.md
```diff
@@ -113,19 +138,20 @@ In the OpossumUI, a distinction between **signals** and **attributions** is made
...
-but no attribution.
+- The `Attributions` progress bar indicates how many resources have manually received an attribution (dark green), how many have an automatically **pre-selected** attribution (lighter green with gradient), and how many resources have a signal, but have not yet received an attribution (orange), with respect to the total number of files. Hovering on the bar shows a tooltip containing all four numbers. Clicking on the bar navigates to a resource that has a signal, but no attribution.
+- The `Criticalities` progress bar indicates how many files have a highly critical signal but no attribution (red), a critical signal but no attribution (orange) with respect to the total number of files not having an attribution. Hovering on the bar shows a tooltip containing all 4 numbers. Clicking on the bar navigates to a resource that has a critical signal, but no attribution.
```
mstykow, line 153: I'm not sure what you are referring to when you say "all 4 numbers". You never take about any number, especially not 4 of them, before this sentence. Also, the criticalities bar contains three colors. So why are you only explaining the case of highly critical and medium critical? Shouldn't the third state also be explained?

USER_GUIDE.md
```diff
@@ -113,19 +138,20 @@ In the OpossumUI, a distinction between **signals** and **attributions** is made
...
 
-Clicking the `Progress Bar Toggle` replaces the `Progress Bar` by the `Critical Signals Progress Bar`. The `Critical Signals Progress Bar` indicates how many files have a highly critical signal but no attribution (red), a critical signal but no attribution (orange) with respect to the total number of files not having an attribution. Hovering on the bar shows a tooltip containing all 4 numbers. Clicking on the bar navigates to a file that has a critical signal,
-but no attribution.
+- The `Attributions` progress bar indicates how many resources have manually received an attribution (dark green), how many have an automatically **pre-selected** attribution (lighter green with gradient), and how many resources have a signal, but have not yet received an attribution (orange), with respect to the total number of files. Hovering on the bar shows a tooltip containing all four numbers. Clicking on the bar navigates to a resource that has a signal, but no attribution.
```
mstykow, line 152: 1. let's avoid the words "manual" and "automatic". We can just say "attribution", "pre-selected attribution", and "no attribution".
2. again you are referring to "four numbers" after actually having spoken about three states. so it's not clear where "four" comes from and also not where "number" comes from.

mstykow (changes requested):

USER_GUIDE.md
Hellgartner: In both cases, the 4 is from having copy pasted the existing text without reading it properly enough. I just reformated the sentence into a list. Fixed.

mstykow (approved):

PR #2853, opened 2025-03-13
Feat/make classification and criticality optional

Hellgartner:
Closes https://github.com/opossum-tool/OpossumUI/issues/2849

### How can the changes be tested

* CI
* Use a file that has both classifications and criticality
   * Go to the view menu and toggle the corresponding features on and off and watch the corresponding features disappear
   * Default: both on
   * Edge case: Progress bar belongs to classifications and then switch off classifications. Same for criticality
   * Edge case: Selected sorting belongs to classifications and then switch off classifications. Same for criticality
   * Edge case: Both classifications and criticality are disabled --> Switch for progress bar to disappear

PR #2880, opened 2025-03-28
Fix circular dependencies

Hellgartner:
### Summary of changes

* Fix circular dependency check
* Fix detected circular dependencies

In https://github.com/opossum-tool/OpossumUI/pull/2853 a review by @PhilippMa detected a circular dependency which was not found by the tooling

run 
`yarn circular-import-check` 

### DIFF
--- a/src/ElectronBackend/main/menu/fileMenu.ts
+++ b/src/ElectronBackend/main/menu/fileMenu.ts
@@ -28,7 +28,7 @@ import {
   importFileListener,
   selectBaseURLListener,
 } from '../listeners';
-import { createMenu } from '../menu';
+import { UpdateMenu } from '../menu';
 import { ProcessingStatusUpdater } from '../ProcessingStatusUpdater';
 import { UserSettingsService } from '../user-settings-service';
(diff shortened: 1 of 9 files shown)


src/ElectronBackend/main/menu/fileMenu.ts
```diff
@@ -64,6 +64,7 @@ function getOpenFile(mainWindow: BrowserWindow): MenuItemConstructorOptions {
 
 async function getOpenRecent(
   mainWindow: BrowserWindow,
+  updateMenu: UpdateMenu,
```
PhilippMa, line 67: Personally I feel like just writing the full type, i.e. `updateMenu: () => Promise<void>`, would have better readability than repeating the same name for parameter name and type in `updateMenu: UpdateMenu`. Then you could also get rid of even the type import from menu.ts

PhilippMa (changes requested):
Great that you fixed the yarn command as well :+1:

PhilippMa (approved):

PR #2895, opened 2025-05-07
fix: disable path selection during file processing

alexzurbonsen:
### Summary of changes

Disbale input fields and early return in click handler if processing in progress.

Import and Merge dialog currently allow to select a different input and target path while the file is imported or merged. This interrupts the file processing.

Open OpossumUI, open import or merge dialog, select paths to import/merge, start the process and see that clicking on the input fields is no longer possible and the fields are disabled.

### DIFF
--- a/src/Frontend/Components/FilePathInput/FilePathInput.tsx
+++ b/src/Frontend/Components/FilePathInput/FilePathInput.tsx
@@ -25,6 +25,7 @@ interface FilePathInputProps {
   text: string;
   onClick: () => void;
   tooltipProps?: Partial<TooltipProps>;
+  disabled: boolean;
 }
 
 export const FilePathInput: React.FC<FilePathInputProps> = (props) => {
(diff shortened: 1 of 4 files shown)


mstykow:
> Do we require to test this functionality e2e? (My attempt failed and I couldn't easily iterate on it, because my local playwright still doesn't seem to recognize the opossum file binary. Need to give this another look if we want to e2e test this.)

I don't think not having an e2e test is a blocker for this non-critical functionality..

src/e2e-tests/page-objects/ImportDialog.ts
```diff
@@ -48,5 +48,11 @@ export class ImportDialog {
     showsError: async (): Promise<void> => {
       await expect(this.errorIcon).toBeVisible();
     },
+    inputFileSelectionIsDisabled: async (): Promise<void> => {
+      await expect(this.inputFileSelection).toBeDisabled();
+    },
+    opossumFileSelectionIsDisabled: async (): Promise<void> => {
+      await expect(this.opossumFileSelection).toBeDisabled();
+    },
```
Hellgartner, line 56: These two methods seem to be unused

Hellgartner (approved):

src/e2e-tests/page-objects/ImportDialog.ts
alexzurbonsen: Thanks, I attempted to add an e2e test once 😅

PR #2900, opened 2025-05-23
feat: input validation

mstykow:
### Summary of changes

This PR adds input validation for attribution details. This is achieved by giving dark orange feedback for invalid entries and light orange for incomplete entries. In addition, there is a tooltip on error fields explaining the problem.

image

We need a way to help users provide valid input values. Originally, we wanted to validate the license expression according to allowed SPDX identifier characters. However, this is not currently possible because the license expression field could be either an identifier or a full name which can contain any characters.

Enter invalid characters such as ";" into the package type field.


src/Frontend/Components/Autocomplete/Autocomplete.style.tsx
```diff
@@ -13,46 +13,56 @@ export const Container = styled('div')({
...
+        return OpossumColors.darkOrange;
+      case 'warning':
+        return OpossumColors.lightOrange;
+      default:
+        return OpossumColors.white;
+    }
+  })();
+  return {
+    '& .MuiInputLabel-root': {
+      backgroundColor: background || errorBackground,
```
Hellgartner, line 34: (Not really this change) Is is desired that the external background always overrides the error background. This would imply that any user of the background prop cannot use the error handling anymore.

src/Frontend/Components/Autocomplete/Autocomplete.tsx
```diff
@@ -50,7 +51,7 @@ type AutocompleteProps<
     ['aria-label']?: string;
     background?: string;
     endAdornment?: React.ReactNode | Array<React.ReactNode>;
-    error?: boolean;
+    error?: 'error' | 'warning';
```
Hellgartner, line 54: Is this the correct name here? An error which can have  a state warning seems a bit odd to me.

src/shared/text.ts
```diff
@@ -249,6 +249,8 @@ export const text = {
   generic: {
     unknown: 'unknown',
     noResults: 'No Results',
+    invalid: 'This expression contains invalid characters',
+    incomplete: 'This expression is incomplete',
```
Hellgartner, line 253: Is it really the expression which is incomplete? More like the object this expression belongs to. Maybe missing would even be a better wording here.
image

mstykow:
> Not sure whether this is desired it just struck my eye:
> 
> The change in background color also changes the background color of the label once you start entering input data image

i think it's fine

PR #2911, opened 2025-08-01
Preferred over origin ids not written

Hellgartner:
### Summary of changes
On attribution update, also check for changes in preferred status and update preferredOverOriginIds accordingly

Currently, preferred over origin ids are not properly written to the output part of the opossum file

* Take an oppossum file
* Mark an attribution as preferred and safe
* Unzip the opossum file and check that preferredOverOriginIds are present in resulting output file

### DIFF
--- a/src/ElectronBackend/main/createWindow.ts
+++ b/src/ElectronBackend/main/createWindow.ts
@@ -9,13 +9,19 @@ import upath from 'upath';
 
 import { getIconPath } from './iconHelpers';
 
+const openDevTools = (mainWindow: BrowserWindow) => {
+  const devtools = new BrowserWindow();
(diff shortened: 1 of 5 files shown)


abraemer:
Current state: In my test file, I had a structure like (several folders deep that I don't show):
```
folder
folder/subfolder
folder/subfolder/file1
folder/subfolder/file2
```
I selected folder, attached one of the signals from a child file and then marked it as preferred. This resulted in *no* `preferredOverOriginIDs` in the output file.
However, removing the preferred flag, selecting a child *file* and then setting the preferred flag again on the attribution (shown as "on parent") lead to the desired result, i.e. the correct `preferredOverOriginIDs` were written to the output file.

PR #2912, opened 2025-08-05
fix: remove trailing slashes for paths to files with childern

abraemer:
### Summary of changes

Before writing the `resourcesToAttributions`  to the opossum output file, remove the trailing slashes for all paths that correspond to `filesWithChildren`.

Currently, input and output formats are inconsistent: 
- In the input.json, we have paths in `resourcesToAttributions` with trailing slashes if and only if the resource is a folder. The list `filesWithChildren` contains the paths without trailing slashes accordingly.
- In the output.json, we write paths to files with children with a trailing slash as if they were a folder.

Open this file in OpossumUI and attach an attribution to `fileWithChildren.zip`, then inspect the `output.json` in the opossum file. Before this change it contains `"/fileWithChildren.zip/"` as path. After this change it contains `"/fileWithChildren.zip"`.

### DIFF
--- a/src/Frontend/util/__tests__/can-have-children.test.ts
+++ b/src/Frontend/util/__tests__/can-have-children.test.ts
@@ -2,9 +2,13 @@
 // SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
 //
 // SPDX-License-Identifier: Apache-2.0
-import { Resources } from '../../../shared/shared-types';
+import {
+  Resources,
+  ResourcesToAttributions,
+} from '../../../shared/shared-types';
 import {
   canResourceHaveChildren,
+  correctFilePathsInResourcesMapping,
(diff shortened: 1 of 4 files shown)


src/Frontend/util/__tests__/can-have-children.test.ts
```diff
@@ -35,3 +39,39 @@ describe('isIdOfResourceWithChildren', () => {
...
+    const filesWithChildren = new Set<string>();
+    expect(
+      correctFilePathsInResourcesMapping(
+        testResourcesToAttributions,
+        filesWithChildren,
+      ),
+    ).toEqual(testResourcesToAttributions);
+  });
+
+  it('returns false for a file id', () => {
```
Hellgartner, line 61: The test name does not match the content

Hellgartner (changes requested):

Hellgartner (approved):

PR #2914, opened 2025-08-11
fix: normalize trailing slashes for files with children

abraemer:
### Summary of changes

This PR adds more sanitization/normalization to the input:
- paths in `filesWithChildren` are normalized to end with a slash
- path from `resourceToAttributions` from the output file get the same sanitization as `resourcesToAttributions` in the input file

OpossumUI is quite fragile w.r.t. to paths in .opossum files. Especially confusing are the requirements of resourcesToAttributions (sanitized, so it does not matter if paths have trailing slashes or not) and filesWithChildren (requires a trailing slash although it is a file). After the last change in #2912 OpossumUI does not properly parse its own .opossum files.

Open a file that has filesWithChildren, add an attribution to the file with children and then reopen the file. Everything should work.
[fwc_slash_atr_noslash.zip](https://github.com/user-attachments/files/21714272/fwc_slash_atr_noslash.zip)

### DIFF
--- a/src/ElectronBackend/input/__tests__/importFromFile.test.ts
+++ b/src/ElectronBackend/input/__tests__/importFromFile.test.ts
@@ -274,7 +274,7 @@ describe('Test of loading function', () => {
         },
       },
       resourcesToAttributions: {
-        '/path/1': [testUuid],
+        '/folder/': [testUuid],
       },
       resolvedExternalAttributions: [],
     };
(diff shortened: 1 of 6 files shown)


src/ElectronBackend/input/__tests__/importFromFile.test.ts
abraemer, line 1: I needed to change the paths in these tests because they are now checked against the `resources` in the input file.

mstykow (approved):

PR #2943, opened 2025-11-05
feat: add url validation logic

abraemer:
### Summary of changes

This PR adds validation logic that checks whether the input can be parsed as an URL with some best effort adjustments. Note that this does not break opossum files that already contain bad data. The field will be highlighted but one can still convert a signal to an attribution. Only if the user edits some field, they will need to correct the bad URL in order to proceed.

Previously, the URL field did not have any validation logic.

- There are unit tests: `yarn test:unit input-validation.test.ts`
- Here is a test file (download -> rename to .opossum) containing two files with one signal each. One has a valid URL the other doesn't. Check out that the workflow works as intended (opossum file opens, signal can be converted to attribution, but upon edits, the URL needs to be corrected)
[test_url_validation.zip](https://github.com/user-attachments/files/23353428/test_url_validation.zip)

### DIFF
--- a/src/Frontend/util/input-validation.ts
+++ b/src/Frontend/util/input-validation.ts
@@ -90,37 +90,53 @@ export function isPackageAttributeInvalid(
       if (!url) {
         return false;
       }
-      return !isValidUrl(url);
+      return !isValidUpstreamAddress(url);
     }
     default:
       return false;
   }
 }
 
-function isValidUrl(urlString: string): boolean {
+function isValidUpstreamAddress(urlString: string): boolean {
   const trimmed = urlString.trim();
 
-  if (!trimmed || /[\s<>"]/.test(trimmed)) {
+  if (!trimmed) {
     return false;
   }
 
+  let url: URL;
   try {
-    const url = new URL(
-      trimmed.startsWith('http://') || trimmed.startsWith('https://')
-        ? trimmed
-        : `https://${trimmed}`,
-    );
-
-    if (!url.hostname || url.hostname.length === 0) {
-      return false;
-    }
-
-    if (!url.hostname.includes('.')) {
-      return false;
-    }
-
-    return true;
+    url = new URL(trimmed);
   } catch {
+    try {
+      url = new URL(`https://${trimmed}`);
+    } catch {
+      return false;
+    }
+  }
+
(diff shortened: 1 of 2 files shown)


src/Frontend/util/input-validation.ts
```diff
@@ -85,7 +85,42 @@ export function isPackageAttributeInvalid(
...
+
+    if (!url.hostname.includes('.')) {
+      return false;
+    }
+
+    return true;
+  } catch {
+    return false;
+  }
+}
```
mstykow: Me and my coding assistant are proposing the following changes:
1. rename to make it clear that this util is used for verifying an upstream address explicitly and not just any URL. it contains logic that is specific to our requirements for upstream addresses.
2. simplify some logic and remove some redundancy
3. cover more edge cases around IP addresses
4. add better documentation (in the form of comments) for why each check is done

```suggestion
function isValidUpstreamAddress(urlString: string): boolean {
  const trimmed = urlString.trim();

  if (!trimmed) {
    return false;
  }

  // Try parsing as-is first (handles URLs with protocols like http://, https://, ftp://, etc.)
  let url: URL;
  try {
    url = new URL(trimmed);
  } catch {
    // If parsing fails, try adding https:// prefix for URLs without protocol
    try {
      url = new URL(`https://${trimmed}`);
    } catch {
      // Both parsing attempts failed
      return false;
    }
  }

  // Only allow http and https protocols for upstream repository URLs
  if (url.protocol !== 'http:' && url.protocol !== 'https:') {
    return false;
  }

  if (!url.hostname) {
    return false;
  }

  // Reject IP addresses (IPv4 and IPv6) - upstream repos should use domain names
  // IPv4 addresses match pattern: digits separated by dots
  if (/^\d+\.\d+\.\d+\.\d+$/.test(url.hostname)) {
    return false;
  }
  // IPv6 addresses are wrapped in brackets in URLs
  if (url.hostname.startsWith('[') && url.hostname.endsWith(']')) {
    return false;
  }

  // Require a hostname with TLD (reject localhost and single-word hostnames)
  // This ensures we only accept publicly accessible upstream repositories
  if (!url.hostname.includes('.')) {
    return false;
  }

  return true;
}
```

PR #2946, opened 2025-11-06
feat: multi-edit auditing options

MarkusObendrauf:
### Summary of changes

Extend the multi-select options with a new menu that allows a user to toggle the auditing options (needs follow-up, needs review, exclude from notice) of all selected attributions.

image

Our users requested to be able to apply "Needs follow-up" and "Exclude from notice" to many attributions at once.

Open a file in OpossumUI and try out the feature.

### DIFF
--- a/src/Frontend/Components/AttributionPanels/AttributionsPanel/AttributionsPanel.tsx
+++ b/src/Frontend/Components/AttributionPanels/AttributionsPanel/AttributionsPanel.tsx
@@ -22,6 +22,7 @@ import { ConfirmButton } from './ConfirmButton/ConfirmButton';
 import { CreateButton } from './CreateButton/CreateButton';
 import { DeleteButton } from './DeleteButton/DeleteButton';
 import { LinkButton } from './LinkButton/LinkButton';
+import { MoreActionsButton } from './MoreActionsButton/MoreActionsButton';
 import { ReplaceButton } from './ReplaceButton/ReplaceButton';
 
 export function AttributionsPanel() {
@@ -73,6 +74,7 @@ export function AttributionsPanel() {
           <ConfirmButton {...props} />
           <ReplaceButton {...props} />
           <DeleteButton {...props} />
+          <MoreActionsButton {...props} />
         </>
       )}
       useFilteredData={useFilteredAttributions}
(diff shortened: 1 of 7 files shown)


src/Frontend/Components/AttributionPanels/AttributionsPanel/AttributionsPanel.tsx
```diff
@@ -73,6 +74,7 @@ export function AttributionsPanel() {
           <ConfirmButton {...props} />
           <ReplaceButton {...props} />
           <DeleteButton {...props} />
```
mstykow: i think it would be good to move these two buttons into the more-menu as well to reduce visual overwhelm.

mstykow (changes requested):
@MarkusObendrauf we have stopped using forks to contribute to opossum UI years ago. can you please re-open this PR directly in OpossumUI? this will make rebasing easier and also easier for me as a reviewer to check out your changes.

MarkusObendrauf:
> The replacement function no longer works after your refactoring: as soon as I enter replacement mode, it is no longer possible to select an actual replacement.
> 
> Now that I see what it looks like, I'm also unsure if it's really a good idea to move the replacement function into the "more" menu because it complicates entering and exiting the replacement mode. I'm now leaning toward having only the auditing options inside the "more" menu so basically undoing your last change. Let's talk if you disagree.

totally agree, will revert to an earlier revision

src/Frontend/Components/AttributionPanels/AttributionsPanel/AttributionsPanel.tsx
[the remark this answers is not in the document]
MarkusObendrauf: I've added another 20 pixels to packageListsWidth. I'm a little conscious of making this too much bigger.

src/Frontend/Components/AttributionPanels/AttributionsPanel/MoreActionsButton/MoreActionsButton.tsx
```diff
@@ -0,0 +1,140 @@
...
+import { useAppDispatch } from '../../../../state/hooks';
+import { useAttributionIdsForReplacement } from '../../../../state/variables/use-attribution-ids-for-replacement';
+import {
+  ExcludeFromNoticeIcon,
+  FollowUpIcon,
+  NeedsReviewIcon,
+} from '../../../Icons/Icons';
+import { PackagesPanelChildrenProps } from '../../PackagesPanel/PackagesPanel';
+
+type updatablePropertyType = 'needsReview' | 'followUp' | 'excludeFromNotice';
```
mstykow: 1. types usually start with a capital letter
2. this type is linked to `PackageInfo` so rather than defining it from scratch, it would be better to use `keyof` and `Extract` utility types.

PR #2964, opened 2026-01-05
fix: disable most menu items if popup is open

### LINKED ISSUE #2798: Disable native menu while popup is open
The native menu can currently be used even when a popup is open. This leads to weird situations, e.g. a file can be opened while the import dialog is open, and the import dialog stays open through the whole process. 

* Menu items that start a new workflow / open a new popup should be disabled while a popup is open
  * e.g. open, import etc. should be disabled
  * there may be some menu items that don't have to be disabled, such as e.g. the zoom in and zoom out menu items
* For this issue, it is sufficient to consider the popups that are handled through the GlobalPopup mechanism as well as the ProcessPopup
  * For any other popups (such as the ConfirmationDialog): investigate if there are any issues with allowing menu item interaction while these popups are open. If yes, create a separate issue to implement a similar solution for these other popups.
* Implementation suggestions/considerations:
  * Create an IPC channel so that the frontend can instruct the backend to disable the app menu
  * When reenabling the app menu, make sure to only enable menu items which were already enabled before the popup was opened (some menu items are disabled at application startup and only get enabled when a file is opened)
  * For popups handled through GlobalPopup, send IPC messages for disabling/enabling menu items when the openPopup/closePopup redux actions are created
  * For the ProcessPopup, send IPC messages when there are changes in the loading state which it uses as its open condition

An alternative would be to allow the user to close open popups when they click on an item in the app menu, but the solution described above is probably easier.

**Additional context**
\-

LMayTNG:
### Summary of changes

When a GlobalPopup or ProcessPopup is opened, disable most of the app menu icons.

closes #2798

### DIFF
--- a/src/ElectronBackend/main/menu/fileMenu.ts
+++ b/src/ElectronBackend/main/menu/fileMenu.ts
@@ -70,19 +70,26 @@ async function getOpenRecent(
     'recentlyOpenedPaths',
   );
 
+  const enabled =
+    !!recentlyOpenedPaths?.length && !getGlobalBackendState().frontendPopupOpen;
+
   return {
     icon: getIconBasedOnTheme('icons/open-white.png', 'icons/open-black.png'),
     label: text.menu.fileSubmenu.openRecent,
-    submenu: getOpenRecentSubmenu(mainWindow, recentlyOpenedPaths, updateMenu),
-    enabled:
-      !!recentlyOpenedPaths?.length &&
-      !getGlobalBackendState().frontendPopupOpen,
+    submenu: getOpenRecentSubmenu(
+      mainWindow,
+      recentlyOpenedPaths,
+      enabled,
+      updateMenu,
+    ),
+    enabled,
   };
 }
 
 function getOpenRecentSubmenu(
   mainWindow: BrowserWindow,
   recentlyOpenedPaths: Array<string> | null,
+  enabled: boolean,
   updateMenu: () => Promise<void>,
 ): MenuItemConstructorOptions['submenu'] {
   if (!recentlyOpenedPaths?.length) {
@@ -94,6 +101,7 @@ function getOpenRecentSubmenu(
       label: path.basename(recentPath, path.extname(recentPath)),
       click: ({ id }) => handleOpeningFile(mainWindow, id, updateMenu),
       id: recentPath,
+      enabled,
     })),
     { type: 'separator' },
     {
@@ -106,11 +114,13 @@ function getOpenRecentSubmenu(
         );
         await updateMenu();
       },
+      enabled,
     },
   ];
 }
 
 function getImportFile(mainWindow: BrowserWindow): MenuItemConstructorOptions {
+  const enabled = !getGlobalBackendState().frontendPopupOpen;
   return {
     icon: getIconBasedOnTheme(
       'icons/import-white.png',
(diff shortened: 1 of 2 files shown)


src/e2e-tests/page-objects/MenuBar.ts
```diff
@@ -38,103 +69,74 @@ export class MenuBar {
     return clickMenuItem(this.window.app, 'label', label);
   }
 
-  private async assertMenuItemDisabled(label: string) {
+  private async assertMenuItemEnabled(enabled: boolean, label: string) {
```
MarkusObendrauf: why this change? i think it's nicer to have two separate functions for `enabled` & `disabled`. same applies to the other assertion functions here

src/e2e-tests/page-objects/MenuBar.ts
```diff
@@ -38,103 +69,74 @@ export class MenuBar {
...
-      );
-
-      //need to call the asserts sequentially here, doing that in
-      //parallel via promise all somehow breaks the app object
-      for (const fileFormat of importFileFormats) {
-        await this.assertSubMenuItemDisabled(
-          text.menu.fileSubmenu.merge,
-          text.menu.fileSubmenu.mergeSubmenu(fileFormat),
-        );
+    initiallyDisableEntriesAreEnabled: async (
```
MarkusObendrauf: typo: `Disable` should be `Disabled`

src/shared/shared-types.ts
```diff
@@ -293,6 +293,7 @@ export interface ElectronAPI {
   on: (channel: AllowedFrontendChannels, listener: Listener) => () => void;
   getUserSettings: () => Promise<UserSettings>;
   updateUserSettings: (userSettings: Partial<UserSettings>) => Promise<void>;
+  setAppMenuDisabled: (enabled: boolean) => Promise<void>;
```
Hellgartner: The name of the argument should be disabled: boolean, correct?

src/shared/ipc-channels.ts
```diff
@@ -19,6 +19,7 @@ export enum IpcChannel {
   StopLoading = 'stop-loading',
   GetUserSettings = 'get-user-settings',
   UpdateUserSettings = 'update-user-settings',
+  SetAppMenuDisabled = 'set-app-menu-disabled',
```
Hellgartner: I am a bit unsure about this name: Because that is actually more like a "disable with some exceptions (e.g. quit is not disabled as well as the view/about menu. 
The alternative would be to just do a "frontendPopupOpen" and then just act on that state as part of the backends business logic. 
Yes I am aware that this would mean quite some changes as the name is used very widely.

src/ElectronBackend/main/menu/fileMenu.ts
```diff
@@ -332,6 +344,7 @@ function getExportSubMenu(
       getExportSpdxYaml(webContents),
       getExportSpdxJson(webContents),
     ],
+    enabled: !getGlobalBackendState().menuDisabled,
```
Hellgartner: This is handled inconsistently: In case there is no file loaded we allow opening the submenu but show all entries as disabled while in this case the whole submenu is disabled. I would like to unify that behavior.

Hellgartner (changes requested):

src/e2e-tests/page-objects/MenuBar.ts
LMayTNG: I now renamed the private functions to assert...EnabledState for more clarity, and split them into an `Enabled` and a `Disabled` function for all exported ones.

src/shared/ipc-channels.ts
LMayTNG: I agree, changed :+1:

src/ElectronBackend/main/main.ts
```diff
@@ -95,6 +96,13 @@ export async function main(): Promise<void> {
       (_, userSettings: Partial<UserSettings>) =>
         UserSettingsService.update(userSettings, { skipNotification: true }),
     );
+    ipcMain.handle(
+      IpcChannel.SetAppMenuDisabled,
+      async (_, disabled: boolean) => {
```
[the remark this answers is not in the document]
LMayTNG: Good idea, I extracted it :+1:

Hellgartner (approved):

MarkusObendrauf (approved):
lgtm with this new approach

src/ElectronBackend/main/menu/fileMenu.ts
```diff
@@ -332,6 +340,9 @@ function getExportSubMenu(
       getExportSpdxYaml(webContents),
       getExportSpdxJson(webContents),
     ],
+    enabled:
+      isFileLoaded(getGlobalBackendState()) &&
+      !getGlobalBackendState().frontendPopupOpen,
```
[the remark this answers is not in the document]
LMayTNG: Good point, I'll try that

PR #2975, opened 2026-01-14
Feat/autocomplete license expressions

mariusAsaTNG:
### Summary of changes

The autocomplete now gives you the SPDX identifiers when possible. You can also use 'AND', 'OR' Within the license field and get autocomplete recommendations. Signals and Attributes only get suggested at the beginning, since it is unlikely that you want to chain an attribute with another attribute.

Entering friendly license names leads to typos, increases review time, and distorts metrics. Switching to SPDX identifiers when possible leads to faster reviews and less errors. The better autocomplete should also help with this.

Open an opossum file with common licenses and try to fill out the license field. Once you have selected a license add an AND or OR afterwards and see the autocomplete

### DIFF
--- a/src/Frontend/Components/AttributionForm/LicenseSubPanel/LicenseSubPanelAutocomplete.tsx
+++ b/src/Frontend/Components/AttributionForm/LicenseSubPanel/LicenseSubPanelAutocomplete.tsx
@@ -43,14 +43,16 @@ export function LicenseSubPanelAutocomplete({
     if (input === undefined) {
       return ['', ''];
     }
-    return input.match(/(.*(?:AND \(*|OR \(*|^))(.*)$/i)?.slice(1) as [
-      string,
-      string,
-    ];
+    return input
+      .match(/(.*(?:(?: AND | OR | WITH |^)\(*))(.*)$/i)
+      ?.slice(1) as [string, string];
   }
 
   function capitalizeExpressions(input: string): string {
-    return input.replaceAll(/\bAND /gi, 'AND ').replaceAll(/\bOR /gi, 'OR ');
+    return input
+      .replaceAll(/ AND /gi, ' AND ')
+      .replaceAll(/ OR /gi, ' OR ')
+      .replaceAll(/ WITH /gi, ' WITH ');
   }
 
   type LicenseOption = {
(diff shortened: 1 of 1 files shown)


src/Frontend/Components/AttributionForm/LicenseSubPanel/LicenseSubPanelAutocomplete.tsx
```diff
@@ -0,0 +1,212 @@
...
+  const frequentLicensesNames = useAppSelector(getFrequentLicensesNameOrder);
+
+  const [{ attributions }] = useFilteredAttributions();
+  const [{ attributions: signals }] = useFilteredSignals();
+
+  function splitAtLastExpression(input: string | undefined): [string, string] {
+    if (input === undefined) {
+      return ['', ''];
+    }
+    return input.match(/(.*(?:AND |OR |^))(.*)$/i)?.slice(1) as [
```
MarkusObendrauf: There's also WITH:
https://spdx.github.io/spdx-spec/v2.3/SPDX-license-expressions/

MarkusObendrauf: There could be an opening bracket before the next part of the expression:
```
MIT AND (LGPL-2.1-or-later OR BSD-3-Clause)
```
it could be multiple brackets too like `((((` so add a `\(*` to the regex

src/Frontend/Components/AttributionForm/LicenseSubPanel/LicenseSubPanelAutocomplete.tsx
```diff
@@ -0,0 +1,212 @@
...
+      value={''}
+      options={licenseOptions}
+      title={text.attributionColumn.licenseExpression}
+      readOnly={!onEdit}
+      highlighting={
+        showHighlight && !packageInfo.licenseName ? 'warning' : undefined
+      }
+      inputValue={packageInfo.licenseName ?? ''}
+      getOptionLabel={(option) =>
+        typeof option === 'string' ? option : option.shortName
```
MarkusObendrauf: can option really be a string? when does that happen? why do we need to check the type of `option` everywhere?

src/Frontend/Components/AttributionForm/LicenseSubPanel/LicenseSubPanelAutocomplete.tsx
```diff
@@ -0,0 +1,212 @@
...
+    const hasExpressionBeforeLastWord = beforeLast !== '';
+    return options.filter((option) =>
+      option.replaceAll
+        ? option.shortName.toUpperCase().includes(inputValue.toUpperCase())
+        : hasExpressionBeforeLastWord || beforeLast === ''
+          ? `${option.shortName} ${option.fullName}`
+              .toUpperCase()
+              .includes(lastLicense.toUpperCase())
+          : false,
+    );
```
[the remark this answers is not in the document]
mariusAsaTNG: And I replaced the name from `replaceAll` to `replaceEntireSearch`

PR #2993, opened 2026-01-29
Create Backend API with SQlite

LMayTNG:
We want to move more of the business logic into the backend. The end goal would be a typical client-server pattern, so we don't have to send the whole (sometimes very large) opossum file to the frontend for analysis.

This PR starts the process
 - by creating and filling the database on file load
 - creating a backend api that is easily extensible by just adding a new command
 - moving the first functionality (filtering the resources) from the frontend to the API

### DIFF
--- a/src/e2e-tests/__tests__/project-statistics.test.ts
+++ b/src/e2e-tests/__tests__/project-statistics.test.ts
@@ -9,6 +9,9 @@ const [resourceName1] = faker.opossum.resourceNames({ count: 1 });
 const [attributionId1, packageInfo1] = faker.opossum.rawAttribution({
   criticality: RawCriticality[Criticality.Medium],
 });
+const [attributionId2, packageInfo2] = faker.opossum.rawAttribution({
+  criticality: RawCriticality[Criticality.Medium],
+});
 const [attributionId3, packageInfo3] = faker.opossum.rawAttribution({
   criticality: RawCriticality[Criticality.Medium],
 });
@@ -31,7 +34,7 @@ test.use({
     }),
     outputData: faker.opossum.outputData({
       manualAttributions: faker.opossum.rawAttributions({
-        [attributionId1]: packageInfo1,
+        [attributionId2]: packageInfo2,
       }),
     }),
   },


src/ElectronBackend/preload.ts
```diff
@@ -39,6 +39,12 @@ const electronAPI: ElectronAPI = {
     ipcRenderer.invoke(IpcChannel.UpdateUserSettings, userSettings),
   setFrontendPopupOpen: (open: boolean) =>
     ipcRenderer.invoke(IpcChannel.SetFrontendPopupOpen, open),
+  api: async (command, params) => {
+    console.time(`Calling API ${command}`);
```
mariusAsaTNG: I guess before merging we should remove this and other console logs

src/Frontend/Components/ResourceBrowser/ResourceBrowser.tsx
mariusAsaTNG: We could also move this function to the backend. It is also a resource search with much of the same functionality, except maybe one further where clause.

mariusAsaTNG (commented):
I would also expect some test changes.

src/ElectronBackend/preload.ts
```diff
@@ -39,6 +39,12 @@ const electronAPI: ElectronAPI = {
     ipcRenderer.invoke(IpcChannel.UpdateUserSettings, userSettings),
   setFrontendPopupOpen: (open: boolean) =>
     ipcRenderer.invoke(IpcChannel.SetFrontendPopupOpen, open),
+  api: async (command, params) => {
```
Hellgartner: I have to say I am not a fan of this interface:
It duplicates the main API. Why not just do a call per command. That would be way clearer.

src/Frontend/Components/AppContainer/AppContainer.tsx
```diff
@@ -16,7 +16,7 @@ import { Toaster } from '../Toaster/Toaster';
 dayjs.extend(localizedFormat);
 
 const store = createAppStore();
-const queryClient = new QueryClient({
+export const queryClient = new QueryClient({
```
Hellgartner: Wondering why this needs to be exported if it is provided via a dedicated context provider further down.

src/Frontend/Components/ResourceBrowser/ResourceBrowser.tsx
```diff
@@ -39,13 +37,11 @@ export function ResourceBrowser() {
...
-    () =>
-      resourceIds?.filter((path) =>
-        path.toLowerCase().includes(debouncedSearchAll),
-      ),
-    [resourceIds, debouncedSearchAll],
-  );
+  const allResourcesFiltered =
+    backend.searchResources.useQuery(
+      { searchString: debouncedSearchAll },
+      { placeholderData: keepPreviousData },
```
Hellgartner: What does that do? I do not find this as an argument to use query here.

src/ElectronBackend/api/commands.ts
```diff
@@ -0,0 +1,47 @@
...
+  async searchResources(props: { searchString: string }) {
+    const result = await getDb()
+      .selectFrom('resource')
+      .select([sql<string>`path || IF(can_have_children, '/', '')`.as('path')])
+      .where(sql<boolean>`instr(LOWER(path), LOWER(${props.searchString})) > 0`)
+      .execute();
+    return result.map((r) => r.path);
+  },
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+} satisfies Record<string, (param?: any) => unknown>;
```
Hellgartner: Again, all this type magic to not to use the proper interface?

src/ElectronBackend/api/commands.ts
```diff
@@ -0,0 +1,47 @@
...
+export type CommandReturn<C extends Command> = ReturnType<Commands[C]>;
+
+export function executeCommand<C extends Command>(
+  command: C,
+  params: CommandParams<C>,
+): CommandReturn<C> {
+  const fn = commands[command] as unknown as (
+    args: CommandParams<C>,
+  ) => CommandReturn<C>;
+  console.time(`Running ${command}`);
```
Hellgartner: Another console log

src/ElectronBackend/db/dbTypes.ts
```diff
@@ -0,0 +1,71 @@
...
+  priority: number;
+  is_relevant_for_preferred: SqliteBoolean;
+}
+
+export interface AttributionTable {
+  uuid: string;
+  /**
+   * All of the attribution data as JSON.
+   */
+  data: string;
```
Hellgartner: Doesn't sqllite have a dedicated datatype for JSON strings (best with query support)?

src/ElectronBackend/db/db.ts
```diff
@@ -0,0 +1,45 @@
+// SPDX-FileCopyrightText: Meta Platforms, Inc. and its affiliates
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+import BetterSqlite3 from 'better-sqlite3';
+import { Kysely, SqliteDialect } from 'kysely';
+
+import { Database } from './dbTypes';
+
```
Hellgartner: For me this whole file looks like you want to use a class with a constructor and two members.

src/ElectronBackend/preload.ts
mstykow: What do you mean by "main API"?

src/ElectronBackend/preload.ts
mariusAsaTNG: It does not, we will change the lint config

Hellgartner (approved):
Approved after discussing the critical points

src/Frontend/Components/AppContainer/AppContainer.tsx
LMayTNG: Not exported anymore :+1:

src/Frontend/util/backendClient.ts
```diff
@@ -0,0 +1,57 @@
...
+
+type BackendClient = {
+  [C in Command]: {
+    useQuery: (...args: Params<C>) => UseQueryReturn<C>;
+    invalidate: (params?: CommandParams<C>) => Promise<void>;
+    reset: (params?: CommandParams<C>) => Promise<void>;
+  };
+};
+
+export const backend = new Proxy({} as BackendClient, {
```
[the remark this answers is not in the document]
LMayTNG: Resolved in discussion ✔️

src/ElectronBackend/db/db.ts
LMayTNG: Resolved in discussion ✔️

Architecture.md
mstykow: what's the purpose of this doc? sounds like something we'd delete again after the migration is complete? is this here only for some agent to read?

package.json
```diff
@@ -217,6 +231,12 @@
   },
   "packageManager": "yarn@4.12.0",
   "dependenciesMeta": {
+    "better-sqlite3": {
+      "built": true
+    },
+    "better-sqlite3-electron": {
+      "built": true
+    },
```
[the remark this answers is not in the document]
LMayTNG: Both of them have a build script that needs to be executed on install.

Only `better-sqlite3-electron` then needs to be rebuilt with `electron-rebuild`, which is what we do in the postinstall script

mariusAsaTNG (approved):

PR #3007, opened 2026-02-04
Chore/vitest

mariusAsaTNG:
### Summary of changes

Move from jest to vitest, also from jsdom to happy-dom, since it is faster

Vitest is a more modern testing framework and works well with vite

run yarn test:unit

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -40,6 +40,7 @@ export default tseslint.config(
     plugins: {
       'check-file': eslintPluginCheckFile,
       vitest: vitest,
+      'jest-dom': fixupPluginRules(eslintConfigJestDom),
       '@tanstack/query': eslintPluginQuery,
       // @ts-ignore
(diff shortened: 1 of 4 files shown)


mariusAsaTNG:
Good point, we are missing some extra info here:
For two examples see here: [jest](https://github.com/opossum-tool/OpossumUI/actions/runs/21689658169/job/62545872398#step:11:1) vs [vitest](https://github.com/opossum-tool/OpossumUI/actions/runs/21709190982/job/62608075368?pr=3007#step:11:1)
- Vitest is actually slower than jest, we go from 37s -> 1m19s, which is ~2x slower
- 1m19s is still great and I do not think we should be worrying just yet
- Vitest is in active development and more future proof, while jest seems to be in maintenance mode
- The tests now show us how long they ran, so we could work on fixing some tests that are bad (e.g. AttributionForm takes ~8s)

PR #3014, opened 2026-02-09
Mirror data changes to the database

LMayTNG:
To keep the frontend and the db in sync, we have to mirror every data change from the frontend to the backend via the api.

### DIFF
--- a/src/ElectronBackend/api/mutations.ts
+++ b/src/ElectronBackend/api/mutations.ts
@@ -7,7 +7,11 @@ import { sql } from 'kysely';
 import { Attributions, PackageInfo } from '../../shared/shared-types';
 import { getDb } from '../db/db';
 import { QueryName, QueryParams } from './queries';
-import { removeRedundantAttributions } from './utils';
+import {
+  getAttributionOrThrow,
+  getResourceOrThrow,
+  removeRedundantAttributions,
+} from './utils';
 
 type QueryInvalidation<Q extends QueryName> = {
   queryName: Q;
@@ -25,10 +29,6 @@ type MutationFunction = (params?: any) => Promise<{
   invalidates?: Array<QueryInvalidationUnion>;
 }>;
 
-function removeTrailingSlash(path: string) {
-  return path.replace(/\/$/, '');
-}
-
 export const mutations = {
   invalidateGetAttributionData() {
     // to avoid typescript errors in backendClient, we need at least one mutation with no parameters, and an invalidation without parameters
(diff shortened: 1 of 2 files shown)


src/Frontend/Components/AppContainer/queryClient.ts
```diff
@@ -0,0 +1,9 @@
+// SPDX-FileCopyrightText: Meta Platforms, Inc. and its affiliates
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+import { QueryClient } from '@tanstack/react-query';
+
+export const queryClient = new QueryClient({
+  defaultOptions: { queries: { refetchOnWindowFocus: false, retry: false } },
+});
```
LMayTNG: We had to move the query client to a global constant so we can import it in the `mutation` function in the backend client for invalidation

src/Frontend/util/backendClient.ts
mariusAsaTNG: Double the in the comments

src/Frontend/util/backendClient.ts
```diff
@@ -102,6 +103,30 @@ export const backend = new Proxy({} as BackendClient, {
     const getQueryKey = (command: CommandName, params: unknown) =>
       ['backend', command, params] as const;
 
+    async function mutate(params: MutationParams<MutationName>) {
```
mariusAsaTNG: Perhaps calling it `run` might be better, since having useMutate.mutate and mutate is a bit confusing

mariusAsaTNG (changes requested):
I believe that we can improve the readability of the mutations.ts file with some refactorings

src/ElectronBackend/api/mutations.ts
```diff
@@ -16,39 +19,391 @@ type QueryInvalidationUnion = {
...
+          throw new Error("External attributions can't be replaced");
+        }
+
+        // Reassign resource links to the replacement attribution, skipping conflicts
+        // (conflicting links will be cascade deleted when the old attribution is removed)
+        await sql`
+        UPDATE OR IGNORE resource_to_attribution
+        SET attribution_uuid = ${params.attributionIdToReplaceWith}
+        WHERE attribution_uuid = ${params.attributionIdToReplace}
+      `.execute(trx);
```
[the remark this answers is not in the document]
LMayTNG: Yes, great find!

That's now fixed with the introduction of `removeRedundantAttributions`

PR #3018, opened 2026-02-12
Move Attribution/Signal count calculation from signals worker to backend

LMayTNG:
This improves the runtime (including communication) by ~30x

I recommend reviewing this commit-by-commit.
Also, the PR has some more complex logic, so feel free to approach me for explanations. And please tell me where the code is unclear, so I can add comments.

### DIFF
--- a/src/ElectronBackend/api/__tests__/queries.test.ts
+++ b/src/ElectronBackend/api/__tests__/queries.test.ts
@@ -3,6 +3,8 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 import { ParsedFileContent } from '../../../shared/shared-types';
+import { text } from '../../../shared/text';
+import { pathsToResources } from '../../../testing/global-test-helpers';
 import { initializeDb } from '../../db/initializeDb';
 import { queries } from '../queries';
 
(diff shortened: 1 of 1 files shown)


src/ElectronBackend/db/db.ts
```diff
@@ -20,12 +21,22 @@ let db: Kysely<DB> | undefined = undefined;
 
 function openDb() {
   // Empty filename for temporary (not in-memory) db: https://www.sqlite.org/inmemorydb.html#temp_db
-  rawDb = new BetterSqlite3('');
+  const filename = process.env.DEBUG ? 'dev.db' : '';
```
mariusAsaTNG: you should add it to the .env.sample

src/ElectronBackend/db/initializeDb.ts
```diff
@@ -362,15 +380,17 @@ async function initializeResourceToAttributionTable(
...
+    .column('attribution_uuid')
     .execute();
 
   await trx.schema
-    .createIndex('resource_to_attribution_attribution_uuid_idx')
+    .createIndex('resource_to_attribution_attribution_uuid_resource_id_idx')
     .on('resource_to_attribution')
     .column('attribution_uuid')
+    .column('resource_id')
     .execute();
```
mariusAsaTNG: Does it make sense to create an index with every column of a table in it? And also an index for the opposite order of that table?

mariusAsaTNG (commented):
Looks good, could not find too much. We should really discuss our testing strategy because you added no tests here and I am not sure if that is fine.

src/Frontend/Components/AttributionPanels/SignalsPanel/SignalsPanel.tsx
```diff
@@ -16,9 +19,21 @@ import { SignalsList } from './SignalsList/SignalsList';
...
+  const selectedResourceId = useAppSelector(getSelectedResourceId);
+
+  const [{ filters, search, selectedLicense }] = useFilteredSignals();
+
+  const counts = backend.filterCounts.useQuery({
+    external: true,
+    filters,
+    search,
+    license: selectedLicense,
+    resourcePathForRelationships: selectedResourceId,
```
mariusAsaTNG: you define the parameter showResolved but you should also use it here

src/ElectronBackend/db/db.ts
mariusAsaTNG: Somehow it does not work for me in the .env but maybe thats a me problem

PR #3023, opened 2026-02-17
feat/Move progressbar to backend

mariusAsaTNG:
### Summary of changes

Progressbar moved to a recursive sqlite query. This way we do not need to do it in the signals worker and are one step closer to removing the signals worker

### DIFF
--- a/src/Frontend/util/backendClient.ts
+++ b/src/Frontend/util/backendClient.ts
@@ -89,10 +89,8 @@ type BackendClient = {
  * This also invalidates all queries when flipped, otherwise old data could be shown.
  */
 let databaseInitialized = false;
-export function setDatabaseInitialized(props: {
-  databaseInitialized: boolean;
-}) {
-  databaseInitialized = props.databaseInitialized;
+export function setDatabaseInitialized(initialized: boolean) {
+  databaseInitialized = initialized;
   queryClient.clear();
   void queryClient.resetQueries();
 }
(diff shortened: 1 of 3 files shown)


src/Frontend/util/backendClient.ts
```diff
@@ -162,6 +175,19 @@ export const backend = new Proxy({} as BackendClient, {
...
+            if (typeof options === 'object' && 'enabled' in options) {
+              return Boolean(
+                typeof options.enabled === 'function'
+                  ? options.enabled()
+                  : options.enabled,
+              );
+            }
+            return true;
+          },
         }),
```
LMayTNG: We can improve the typing of options here. That made me notice that `enabled` takes a parameter we have to pass down

```
useQuery: (params?: QueryParams<QueryName>, options?: ClientQueryOptions<QueryName>) =>
useQuery({
queryKey: getQueryKey(command, params),
queryFn: () => {
return query(params);
},
...options,
          enabled: (query) => {
            if (!databaseInitialized) {
              return false;
            }
            if (options !== undefined && 'enabled' in options) {
              return Boolean(
                typeof options.enabled === 'function'
                  ? options.enabled(query)
                  : options.enabled,
              );
            }
            return true;
          },
}),
```

mariusAsaTNG: nice find, is fixed now!

PR #3025, opened 2026-02-20
Feat/move autocomplete to be

LMayTNG:
Move the autocomplete await from the signal worker's `useFilteredAttributions`/`useFilteredSignals`.

For the Autocompletes in the package info form, I added a new query `autocompleteOptions`.

For the Autocomplete in the filter button, I added a license list to the `filterCounts` query and renamed it to `filterProperties`.

Now the last uses of useFiltered... are the lists of Attributions/Signals in the middle panel.

It makes sense to review this commit by commit.

### DIFF
--- a/src/ElectronBackend/api/queries.ts
+++ b/src/ElectronBackend/api/queries.ts
@@ -31,6 +31,11 @@ type AutocompletableAttribute =
   | 'url'
   | 'licenseName';
 
+type AutocompleteOptions = Record<
+  'manual' | 'external',
+  Array<{ contained_uuid: string; value: string; count: number }>
+>;
+
 export type FilterProperties = FilterCounts & {
   total: number;
   licenses: Array<string>;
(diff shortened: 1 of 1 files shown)


src/ElectronBackend/api/queries.ts
```diff
@@ -44,7 +56,72 @@ export const queries = {
     return { result: result?.data ?? null };
   },
 
-  async filterCounts(props: {
+  async autoCompleteOptions({
+    attributeName,
+    onlyRelatedToResourcePath,
+  }: {
+    attributeName: AutocompletableAttribute;
+    onlyRelatedToResourcePath?: string;
```
mariusAsaTNG: Since we always have something selected it does not make sense to let this be undefined.

src/ElectronBackend/api/queries.ts
```diff
@@ -44,7 +56,72 @@ export const queries = {
...
+  }) {
```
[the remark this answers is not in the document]
mariusAsaTNG: If not PackageInfo I would still like to see a type definition with a name like AutoCompleteOption

src/ElectronBackend/api/queries.ts
mariusAsaTNG: well you could move the throws up and create a single block of query which is always nicer to read

LMayTNG: Yeah, you're right, YAGNI :grin:
Done :+1:

src/ElectronBackend/api/queries.ts
```diff
@@ -44,7 +61,71 @@ export const queries = {
...
+                  ? [eb('resource_id', '=', closestAncestorToResource)]
+                  : []),
+              ]),
+            ),
+        ),
+      );
+
+    query = query.groupBy(['value', 'is_external']);
+
+    const result = await query.execute();
```
mariusAsaTNG: Nit: why not just get the result right away?

PR #3030, opened 2026-02-25
Remove signals worker

LMayTNG:


src/ElectronBackend/api/queries.ts
```diff
@@ -130,6 +134,97 @@ export const queries = {
...
+    external: boolean;
+    filters: Array<Filter>;
+    resourcePathForRelationships: string;
+    license?: string;
+    search?: string;
+    showResolved?: boolean;
+    excludeUnrelated?: boolean;
+  }): Promise<{ result: Attributions }> {
+    console.log('Starting listAttributions');
+    console.time(`ListAttributions ${JSON.stringify(props)}`);
```
mariusAsaTNG: note, check that this is gone

PR #3032, opened 2026-02-27
feat: persist ordering of the statistics table across app restarts

hecadr:
### Summary of changes

the ordering of the statistics table is now persisted across app restarts

feature was requested by user

open the statistics table, change the ordering, close the statistics table (or the whole app), reopen the statistics table, observe that the ordering is the same as before

### DIFF
--- a/src/e2e-tests/__tests__/project-statistics.test.ts
+++ b/src/e2e-tests/__tests__/project-statistics.test.ts
@@ -2,6 +2,8 @@
 // SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
 //
 // SPDX-License-Identifier: Apache-2.0
+import { expect } from '@playwright/test';
+
 import { Criticality, RawCriticality } from '../../shared/shared-types';
 import { faker, test } from '../utils';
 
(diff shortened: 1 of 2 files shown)


src/shared/shared-types.ts
```diff
@@ -11,6 +11,7 @@ import {
   CommandParams,
   CommandReturn,
 } from '../ElectronBackend/api/commands';
+import { TableOrdering } from '../Frontend/Components/AttributionCountPerSourcePerLicenseTable/AttributionCountPerSourcePerLicenseTable.util';
```
hecadr: This looks out of place (compared to the other imports), but I think it's necessary here?

src/Frontend/Components/AttributionCountPerSourcePerLicenseTable/AttributionCountPerSourcePerLicenseTable.tsx
```diff
@@ -135,33 +135,37 @@ export const AttributionCountPerSourcePerLicenseTable: React.FC<
     ],
   );
 
-  const [ordering, setOrdering] = useState<TableOrdering>(defaultOrdering);
+  const [ordering, setOrdering] = useState<TableOrdering>(
+    userSettings.attributionTableSorting || defaultOrdering,
```
mariusAsaTNG: Since you aready define the default ordering in shared-constants you do not need the `||` operator, however, it does not hurt.

src/shared/shared-types.ts
hecadr: I fixed the cyclic imports by defining Order and TableOrdering in `shared-types.ts`, as suggested.

hecadr (commented):
I fixed the cyclic type imports.
It's still missing the e2e test.

mariusAsaTNG (approved):

PR #3063, opened 2026-03-17
refactor: align Electron runtime with native ESM

mstykow:
## Summary
- Migrate the Electron runtime path from `main`'s CommonJS-oriented setup to native ESM by emitting `.mjs` main/preload entrypoints, externalizing main-process runtime dependencies, and updating packaging metadata accordingly.
- Align TypeScript and import behavior with the new ESM setup by enabling `verbatimModuleSyntax`, migrating from `lodash` to `lodash-es`, fixing type-only imports that must be erased at runtime, replacing private package subpath imports, and standardizing Node-side JSON/path handling.
- Keep the repo’s dual native sqlite strategy intact under the new runtime model by separating plain-Node `better-sqlite3` usage from the Electron-only `better-sqlite3-electron` alias and rebuild path.
- Bundle and resolve packaged extra resources correctly under the new layout, including `bin/opossum-file-cli` and notices, so import/merge and packaged app behavior continue to work after the ESM migration.
- Modernize supporting tooling and docs so they match the ESM-first repo state: ESM lint-staged config, Babel config cleanup, lint rules that prefer `import type`, Knip cleanup, and contributor-guide updates.
- Harden the test harness around the migration by updating Vitest setup, stabilizing dialog/file-picker flows, and improving packaged E2E diagnostics.

## Benefits
- **Native ESM Electron entrypoints:** compared to `main`, the Electron app now ships `.mjs` main/preload outputs and packaging metadata that match the repo’s ESM direction.
- **Cleaner main-process bundle:** compared to `main`, Electron runtime dependencies stay externalized instead of being inlined into the ESM main bundle, which keeps the shipped main entry closer to native ESM semantics.
- **Safer runtime boundaries:** compared to `main`, true type-only imports are consistently erased, so renderer code no longer drags backend/native modules into the browser bundle through type imports.
- **Preserved native-module compatibility:** compared to `main`, the branch makes the Node-vs-Electron sqlite split explicit instead of relying on accidental ABI compatibility.
- **Correct packaged resource handling:** compared to `main`, packaged binaries and notices are resolved from the right runtime locations, which keeps import/merge flows working in release builds.
- **Better package-exports compatibility:** compared to `main`, the app no longer relies on private `dist/cjs/*` imports that break under modern package `exports` rules.
- **More stable verification paths:** compared to `main`, unit and packaged E2E flows are better aligned with the real runtime and collect better diagnostics when they fail.
- **More consistent tooling and source semantics:** compared to `main`, TypeScript, ESLint, JSON imports, path helpers, packaging scripts, Babel usage, and docs all reflect the same ESM-first setup.

## Verification
- `yarn copyright-lint-check`
- `yarn knip`
- `yarn lint-check`
- `yarn typecheck`
- `yarn test:unit`
- `yarn build`
- `yarn start` (startup probe)
- representative packaged E2E checks for import and merge dialog flows


src/ElectronBackend/main/listeners.ts
```diff
@@ -533,9 +530,7 @@ function setTitle(mainWindow: BrowserWindow, filePath: string): void {
 
   mainWindow.setTitle(
     getGlobalBackendState().projectTitle ||
-      decodeURIComponent(
-        upath.toUnix(filePath).split('/').pop() || defaultTitle,
-      ),
+      decodeURIComponent(path.basename(filePath) || defaultTitle),
```
mariusAsaTNG: I like the new way we handle file paths.

mariusAsaTNG (commented):
Initial look:
I did not check all files yet, however the migration seems to have little benefit and instead constrain us quite a bit. It makes imports weird and the whole ElectronInterop idea I do not quite understand.
Some changes I like, like the way we now handle paths, but this should in no way be in the same commit.

LMayTNG (changes requested):
I think this PR needs a lot of work before we can approve it.

The basic idea, migrating from CJS to ESM, is unmotivated. It solves no current issues. All of the benefits it claims to have are things that weren't broken before.
I know that the general direction of the ecosystem goes toward ESM, but I don't see a reason to force a migration before we have any pain with CJS.

 There a lot of drawbacks. The electron-interop adds a lot of complexity, and I'm not even sure it's needed. Electron added support for ESM in 2023. The PR also quietly adds the whole `node_modules` folder back into the bundle.

Last but not least, there are *a lot* of unrelated drive-by changes. While many of them seem like good changes, they don't belong in the same PR and certainly not in the same commit, especially not without any comments or motivation.

mstykow:
Superseded by #3142

PR #3065, opened 2026-03-18
fix: move computationally heavy calculations in backend to utility pr…

LMayTNG:
…ocess

The renderer process needs the IPC for rendering, so when we do blocking computations directly in the main thread, the UI freezes. This commit moves the calculation to a separate utility process to fix that.


src/ElectronBackend/dbProcess.ts
```diff
@@ -0,0 +1,167 @@
...
+          const { inputFileRaw: _, ...rest } = loadResult;
+          result = rest;
+        } else {
+          result = loadResult;
+        }
+        break;
+      }
+      case 'saveFile': {
+        const { id: _, type: __, ...params } = msg;
+        await saveFile(params as SaveFileParams, storedInputFileRaw);
```
LMayTNG: The `as` isn't needed. Also, why is the SaveFileParams type duplicated?

PR #3100, opened 2026-04-08
Bug: wrong ordering of resources that share a prefix

mariusAsaTNG:
See https://github.com/opossum-tool/OpossumUI/issues/3099 for more info

Now the ID works as a sort key and the sorting is defined in typescript

A test has been added to ensure the bug does not pop up again and getResourceTree > sorting tests cover that the old behaviour remains.


mariusAsaTNG:
9 files are touched but the only meaningful change is in initializeDb.ts

src/ElectronBackend/db/initializeDb.ts
```diff
@@ -278,6 +268,22 @@ async function initializeResourceTable(
...
+    bName: string,
+  ) {
+    if (aIsLeaf && !bIsLeaf) {
+      return 1;
+    }
+    if (!aIsLeaf && bIsLeaf) {
+      return -1;
+    }
+    // If both resources are files or both are directories, we sort them alphabetically
+    return aName.localeCompare(bName);
```
mstykow:
```suggestion
    // Use code-unit ordering so resource ids stay stable across locales and OSes.
    if (aName === bName) {
      return 0;
    }
    return aName < bName ? -1 : 1;
```

mstykow (commented):
It would be good to add some tests to ensure the new ordering.

For example, in `queries.test.ts`:
```
describe('getNextFileToReview', () => {
    it('returns the next attribution review target in deterministic resource order', async () => {
      await initializeDbWithTestData({
        resources: pathsToResources(['/a.ts', '/A.ts', '/b.ts']),
        externalAttributions: {
          attributions: {
            'ext-a': { id: 'ext-a', criticality: Criticality.None },
            'ext-A': { id: 'ext-A', criticality: Criticality.None },
            'ext-b': { id: 'ext-b', criticality: Criticality.None },
          },
          resourcesToAttributions: {
            '/a.ts': ['ext-a'],
            '/A.ts': ['ext-A'],
            '/b.ts': ['ext-b'],
          },
          attributionsToResources: {
            'ext-a': ['/a.ts'],
            'ext-A': ['/A.ts'],
            'ext-b': ['/b.ts'],
          },
        },
      });

      await expect(
        queries.getNextFileToReviewForAttribution({
          selectedResourcePath: '/A.ts',
        }),
      ).resolves.toEqual({ result: '/a.ts' });

      await expect(
        queries.getNextFileToReviewForAttribution({
          selectedResourcePath: '/b.ts',
        }),
      ).resolves.toEqual({ result: '/A.ts' });
    });

    it('returns the next criticality review target in deterministic resource order', async () => {
      await initializeDbWithTestData({
        resources: pathsToResources(['/a.ts', '/A.ts', '/b.ts']),
        externalAttributions: {
          attributions: {
            'ext-a': { id: 'ext-a', criticality: Criticality.High },
            'ext-A': { id: 'ext-A', criticality: Criticality.High },
            'ext-b': { id: 'ext-b', criticality: Criticality.High },
          },
          resourcesToAttributions: {
            '/a.ts': ['ext-a'],
            '/A.ts': ['ext-A'],
            '/b.ts': ['ext-b'],
          },
          attributionsToResources: {
            'ext-a': ['/a.ts'],
            'ext-A': ['/A.ts'],
            'ext-b': ['/b.ts'],
          },
        },
      });

      await expect(
        queries.getNextFileToReviewForCriticality({
          selectedResourcePath: '/A.ts',
        }),
      ).resolves.toEqual({ result: '/a.ts' });
    });

    it('returns the next classification review target in deterministic resource order', async () => {
      await initializeDbWithTestData({
        resources: pathsToResources(['/a.ts', '/A.ts', '/b.ts']),
        externalAttributions: {
          attributions: {
            'ext-a': {
              id: 'ext-a',
              criticality: Criticality.None,
              classification: 1,
            },
            'ext-A': {
              id: 'ext-A',
              criticality: Criticality.None,
              classification: 1,
            },
            'ext-b': {
              id: 'ext-b',
              criticality: Criticality.None,
              classification: 1,
            },
          },
          resourcesToAttributions: {
            '/a.ts': ['ext-a'],
            '/A.ts': ['ext-A'],
            '/b.ts': ['ext-b'],
          },
          attributionsToResources: {
            'ext-a': ['/a.ts'],
            'ext-A': ['/A.ts'],
            'ext-b': ['/b.ts'],
          },
        },
      });

      await expect(
        queries.getNextFileToReviewForClassification({
          selectedResourcePath: '/A.ts',
          classifications: {
            1: { description: 'Permissive', color: '#00ff00' },
          },
        }),
      ).resolves.toEqual({ result: '/a.ts' });
    });
  });
```

mstykow (changes requested):

src/ElectronBackend/db/initializeDb.ts
mariusAsaTNG: I would not do this, since localeCompare is the more natural way to alphabetically sort resources. With localCompare 'a' would appear before 'B' but with `aName < bName ? -1 : 1;` it would appear after 'B'.
We could add `toLowerCase` and other things but then we would be rebuilding our own localeCompare.

mariusAsaTNG:
> it('returns the next criticality review target in deterministic resource order', async () => {
>       await initializeDbWithTestData({
>         resources: pathsToResources(['/a.ts', '/A.ts', '/b.ts']),
>         externalAttributions: {
>           attributions: {
>             'ext-a': { id: 'ext-a', criticality: Criticality.High },
>             'ext-A': { id: 'ext-A', criticality: Criticality.High },
>             'ext-b': { id: 'ext-b', criticality: Criticality.High },
>           },
>           resourcesToAttributions: {
>             '/a.ts': ['ext-a'],
>             '/A.ts': ['ext-A'],
>             '/b.ts': ['ext-b'],
>           },
>           attributionsToResources: {
>             'ext-a': ['/a.ts'],
>             'ext-A': ['/A.ts'],
>             'ext-b': ['/b.ts'],
>           },
>         },
>       });
> 
>       await expect(
>         queries.getNextFileToReviewForCriticality({
>           selectedResourcePath: '/A.ts',
>         }),
>       ).resolves.toEqual({ result: '/a.ts' });
>     });

Thanks, I added the second test. there are other tests in the getResourceTree test file, so I do not think we need to add all 3 tests. Especially since they test the progress bar more than the sorting.

src/ElectronBackend/db/initializeDb.ts
mstykow: If we used plain `aName.localeCompare(bName)`, JS would use the host’s default locale/collation rules. That means the same input could be inserted in a different order on different machines.

Example:

- Children under one folder: A.ts, a.ts, b.ts
- Machine 1 might order them as a.ts, A.ts, b.ts
- Machine 2 might order them as A.ts, a.ts, b.ts

That matters because insertion order becomes resource.id, and this branch now uses resource.id for:
- resource tree ordering
- “next file to review” navigation
- export ordering

So without a fixed collator, two users could load the same file and get different “next file” behavior or export order. The fix keeps alphabetical sorting, but pins the rules so the result is stable everywhere.

However, your point is valid and I have another suggestion:
```javascript
const resourceNameCollator = new Intl.Collator('en', {
    sensitivity: 'variant',
    caseFirst: 'lower',
  });

function sortChildren(
    aIsLeaf: boolean,
    aName: string,
    bIsLeaf: boolean,
    bName: string,
  ) {
    if (aIsLeaf && !bIsLeaf) {
      return 1;
    }
    if (!aIsLeaf && bIsLeaf) {
      return -1;
    }
    // Keep alphabetical ordering stable across environments.
    const result = resourceNameCollator.compare(aName, bName);
    if (result !== 0) {
      return result;
    }
    return aName < bName ? -1 : aName > bName ? 1 : 0;
}
```

mstykow (changes requested):

src/ElectronBackend/db/initializeDb.ts
mariusAsaTNG: With the same logic could we not just use `aName.localeCompare(bName, "en")`?

mariusAsaTNG: However I am fine with your option, I will implement it

mstykow (approved):

PR #3106, opened 2026-04-13
Fix/speed up loading db

mariusAsaTNG:
### Summary of changes

Speed up the loading time of the DB. The changes are to the progressBar, the insertion of resources, the insertion of resource_to_attribution values, and some DB settings.

We want the db loading time to be smaller than the sending of data to the FE. This way we can state that our startup time has gone down.

All the preexisting tests ensure that nothing has broken

### DIFF
--- a/src/ElectronBackend/api/__tests__/utils.test.ts
+++ b/src/ElectronBackend/api/__tests__/utils.test.ts
@@ -267,7 +267,7 @@ describe('removeRedundantAttributions', () => {
       });
 
     await expectDbContent(props.expectedResourcesToAttributions);
-    await expectCAAConsistency();
+    await expectCaaConsistency();
   });
 });
 
@@ -342,7 +342,7 @@ async function expectDbContent(
   }
 }
 
-async function expectCAAConsistency() {
+async function expectCaaConsistency() {
   const resources = await getDb()
     .selectFrom('resource')
     .innerJoin(
@@ -370,7 +370,7 @@ async function expectCAAConsistency() {
     ).map((r) => r.resource_id),
   );
 
-  const cAAManualByResourceId = new Map(resources.map((r) => [r.id, r.manual]));
+  const caaManualByResourceId = new Map(resources.map((r) => [r.id, r.manual]));
 
   for (const resource of resources) {
     if (resourcesWithManualAttributions.has(resource.id)) {
@@ -383,7 +383,7 @@ async function expectCAAConsistency() {
       resource.parent_id !== null
     ) {
       const parentManual =
-        cAAManualByResourceId.get(resource.parent_id) ?? null;
+        caaManualByResourceId.get(resource.parent_id) ?? null;
       expect(
         resource.manual,
         `Closest ancestors with manual attributions for '${resource.path}' should match its parent's`,
--- a/src/ElectronBackend/api/__tests__/utils.test.ts
+++ b/src/ElectronBackend/api/__tests__/utils.test.ts
@@ -267,7 +267,7 @@ describe('removeRedundantAttributions', () => {
       });
 
     await expectDbContent(props.expectedResourcesToAttributions);
-    await expectCwaConsistency();
+    await expectCAAConsistency();
   });
 });
 
@@ -342,16 +342,20 @@ async function expectDbContent(
   }
 }
 
-async function expectCwaConsistency() {
+async function expectCAAConsistency() {
   const resources = await getDb()
     .selectFrom('resource')
-    .innerJoin('cwa', 'cwa.resource_id', 'resource.id')
+    .innerJoin(
+      'closest_attributed_ancestors',
+      'closest_attributed_ancestors.resource_id',
+      'resource.id',
+    )
     .select([
       'resource.id',
       'resource.path',
       'resource.parent_id',
       'resource.is_attribution_breakpoint',
-      'cwa.manual',
+      'closest_attributed_ancestors.manual',
     ])
     .execute();
 
@@ -366,28 +370,28 @@ async function expectCwaConsistency() {
     ).map((r) => r.resource_id),
   );
 
-  const cwaManualByResourceId = new Map(resources.map((r) => [r.id, r.manual]));
+  const cAAManualByResourceId = new Map(resources.map((r) => [r.id, r.manual]));
 
   for (const resource of resources) {
     if (resourcesWithManualAttributions.has(resource.id)) {
       expect(
         resource.manual,
-        `CWA manual for '${resource.path}' should point to itself`,
+        `Closest ancestors with manual attributions for '${resource.path}' should point to itself`,
       ).toBe(resource.id);
     } else if (
       resource.is_attribution_breakpoint === 0 &&
       resource.parent_id !== null
     ) {
       const parentManual =
-        cwaManualByResourceId.get(resource.parent_id) ?? null;
+        cAAManualByResourceId.get(resource.parent_id) ?? null;
       expect(
         resource.manual,
-        `CWA manual for '${resource.path}' should match its parent's CWA`,
+        `Closest ancestors with manual attributions for '${resource.path}' should match its parent's`,
       ).toBe(parentManual);
     } else {
       expect(
         resource.manual,
-        `CWA manual for '${resource.path}' should be null`,
+        `Closest_attributed_ancestors manual for '${resource.path}' should be null`,
       ).toBeNull();
     }
   }
--- a/src/ElectronBackend/db/initializeDb.ts
+++ b/src/ElectronBackend/db/initializeDb.ts
@@ -86,7 +86,6 @@ export async function initializeDb(inputFile: ParsedFileContent) {
 
       await initializeProgressBarTable(trx);
     });
-  getRawDb().exec('VACUUM');
 }
 
 async function initializeProgressBarTable(trx: Transaction<DB>) {
@@ -271,15 +270,15 @@ async function initializeResourceTable(
   });
 
   function sortChildren(
-    aIsLeaf: boolean,
+    aIsFile: boolean,
     aName: string,
-    bIsLeaf: boolean,
+    bIsFile: boolean,
     bName: string,
   ) {
-    if (aIsLeaf && !bIsLeaf) {
+    if (aIsFile && !bIsFile) {
       return 1;
     }
-    if (!aIsLeaf && bIsLeaf) {
+    if (!aIsFile && bIsFile) {
       return -1;
     }
     // If both resources are files or both are directories, we sort them alphabetically
@@ -312,12 +311,12 @@ async function initializeResourceTable(
         ([childName, childChildren]) => ({
           name: childName,
           children: childChildren,
-          isLeaf:
+          isFile:
             childChildren === 1 ||
             trimmedFilesWithChildren.has(`${currentPath}/${childName}`),
         }),
       );
-      entries.sort((a, b) => sortChildren(a.isLeaf, a.name, b.isLeaf, b.name));
+      entries.sort((a, b) => sortChildren(a.isFile, a.name, b.isFile, b.name));
       for (const { name, children } of entries) {
         lastDescendantId = recursivelyCollectResource(
           name,
@@ -342,11 +341,11 @@ async function initializeResourceTable(
     return lastDescendantId;
   }
 
-  // The root resource has "" as name and path
-  console.time('collectResources');
   const resourcesToInsert: Array<ResourceRow> = [];
+  // The root resource has '' as name and path
   recursivelyCollectResource('', resources, null, null, resourcesToInsert);
   insertMany(resourcesToInsert);
+
   await trx.schema
     .createIndex('resource_parent_id_covering_idx')
     .on('resource')
--- a/src/ElectronBackend/db/initializeDb.ts
+++ b/src/ElectronBackend/db/initializeDb.ts
@@ -46,7 +46,6 @@ export const comments: Record<string, Record<string, string>> = {
 
 export async function initializeDb(inputFile: ParsedFileContent) {
   resetDb();
-
   await getDb()
     .transaction()
     .execute(async (trx) => {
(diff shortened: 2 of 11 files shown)


mariusAsaTNG:
Some numbers:
The fb file took 23.7s to initialize the DB and now it takes 15.4s, so a reduction of 8.3s in startup time.
The main gains look as follows:
- initializeResourceTable: 9.1s -> 4.7s
  - done by removing the update max_descendant_id and only doing 1 insert instead
- initializeResourceToAttributionTable: 5.2s -> 4.6s
  - setting is_external in typescript and not getting the value each time in SQL
- Progress bar: 7.2s -> 5.2s
  - combined 3 tree walks into 1

also changing the `page_size` and `cache_size` values helped a lot.
Calling VACUUM after initializaion: 1.14GB -> 1.04GB

mariusAsaTNG:
Some changes that are suboptimal:

- We remove the foreign key constraint on `max_descendant_id`, so that we can set it before the child exists
- We need to store all resources first and only then start writing them into the DB. This creates a memory overhead
- Bulk instering resourceToAttribution values is "ugly" code and hard to maintain, however it does offer a speedup
- Nothing to note on the Progress Bar. That is simply an improvement

src/ElectronBackend/db/initializeDb.ts
```diff
@@ -86,6 +86,7 @@ export async function initializeDb(inputFile: ParsedFileContent) {
 
       await initializeProgressBarTable(trx);
     });
+  getRawDb().exec('VACUUM');
```
MarkusObendrauf: Had to look up what this means, but doing it right after initialization seems questionable to me.
https://sqlite.org/lang_vacuum.html
The only reason to vacuum that sounds like it could make sense here is that you initialized in a way that created a fragmented table - is that what this is about?
Did you measure a performance improvement with this?

src/ElectronBackend/db/initializeDb.ts
```diff
@@ -95,62 +96,48 @@ async function initializeProgressBarTable(trx: Transaction<DB>) {
...
-    .addColumn('external', 'integer', (col) => col.references('resource.id'))
     .addColumn('breakpoint', 'integer', (col) =>
       col.notNull().references('resource.id'),
     )
+    .addColumn('manual', 'integer', (col) => col.references('resource.id'))
+    .addColumn('external', 'integer', (col) => col.references('resource.id'))
     .execute();
 
   await sql`
   -- closest with ancestor table
```
MarkusObendrauf: not a fan of this name, could you rename to `closest_attributed_ancestors`?

src/ElectronBackend/db/initializeDb.ts
```diff
@@ -259,14 +244,27 @@ async function initializeResourceTable(
   const rawDb = getRawDb();
   const insertStmt = rawDb.prepare(`
     INSERT INTO resource
-      (id, path, name, parent_id, is_attribution_breakpoint, is_file, can_have_children, base_url)
+      (id, path, name, parent_id, is_attribution_breakpoint, is_file, can_have_children, base_url, max_descendant_id)
```
MarkusObendrauf, line 245: I don't see why we need max_descendant_id at all - you should be able to query for all children of a folder using a query like `WHERE path LIKE "/path/to/resource%"`. If you add an index on path, then this should be fast.

src/ElectronBackend/db/initializeDb.ts
```diff
@@ -293,66 +291,63 @@ async function initializeResourceTable(
...
-        last_inserted_id = recursivelyInsertResource(
-          childName,
-          childChildren,
+      const entries = Object.entries(children).map(
+        ([childName, childChildren]) => ({
+          name: childName,
+          children: childChildren,
+          isLeaf:
+            childChildren === 1 ||
+            trimmedFilesWithChildren.has(`${currentPath}/${childName}`),
```
MarkusObendrauf, line 316: this condition looks wrong, if the child has children, then it's not a leaf

src/ElectronBackend/db/initializeDb.ts
```diff
@@ -293,66 +291,63 @@ async function initializeResourceTable(
...
+      is_attribution_breakpoint: Number(isAttributionBreakpoint),
+      is_file: Number(isFile),
+      can_have_children: Number(!isLeaf),
+      base_url: trimmedBaseUrlsForSources[currentPath],
+      max_descendant_id: lastDescendantId,
+    };
+    return lastDescendantId;
   }
 
   // The root resource has "" as name and path
```
MarkusObendrauf: this comment is misplaced

src/ElectronBackend/db/initializeDb.ts
MarkusObendrauf, line 239: why is this 1-indexed? now the indexes in result don't match the `result` array and there's this awkward `- 1` in the code. We're creating all IDs by hand, so can't we just start at 1?

MarkusObendrauf (changes requested):
Nice improvements. I've got a few questions especially as this is my first time seeing this code.

src/ElectronBackend/db/initializeDb.ts
mariusAsaTNG: The VACUUM command rebuilds the database file, repacking it into a minimal amount of disk space.
I felt like after initialization is the perfect time to run this, since the DB will not change much after that.
A smaller file with VACUUM is 52MB and without is 54MB and a larger file is 1.04GB with and 1.14GB without.

mariusAsaTNG: Since the reduction is quite small and it does take 100ms for the small file and 3s(had not measured this before) for the large file, I am fine to remove it. What do you think?

src/ElectronBackend/db/initializeDb.ts
mariusAsaTNG: fair point, I like closest_attributed_ancestors 👍

src/ElectronBackend/db/initializeDb.ts
mariusAsaTNG: using the path is more complex in many situations. For example `WHERE path LIKE "/path/to/resource%"` would also match folders that are in `/path/to/resource-util` and Like is case insensitive, so it would match `/path/to/Resource` as well. Furthermore indexing often breaks when you use Like. It is much easier if you know that your descendants are between your `id` and your `max_descendant_id`.

src/ElectronBackend/db/initializeDb.ts
mariusAsaTNG: good catch, it should be called `isFile`

src/ElectronBackend/db/initializeDb.ts
mariusAsaTNG: my bad

src/ElectronBackend/db/initializeDb.ts
mariusAsaTNG: Good question. Since we write the id into the DB and sql is usually 1-indexed, we decided to also have a 1-index here.

mariusAsaTNG: I feel like changing it should be done in another PR, if we want to do so.

src/ElectronBackend/db/initializeDb.ts
MarkusObendrauf: Yeah, definitely remove it

src/ElectronBackend/db/initializeDb.ts
MarkusObendrauf: Apparently case-sensitive-like exists but is deprecated: https://sqlite.org/pragma.html#pragma_case_sensitive_like

You should still be able to use string ordering somewhat like this (where x'FF' is the highest possible byte):
```
WHERE path > 'path/to/folder/' AND path <  'path/to/folder/' || x'FF';
```
but it's starting to feel a bit hacky (implicitly depends on string encoding). I still like the idea though because `max_descendant_id` is redundant, and I get the feeling that calculating it has an impact on load times. Would you have to use any recursion at all if it weren't for this field?

src/ElectronBackend/db/initializeDb.ts
MarkusObendrauf: Sure. I don't think it's a big deal either way, just surprised me.

MarkusObendrauf (approved):

MarkusObendrauf:
lgtm, i still am not convinced by `max_descendant_id` but that can be another pr

src/ElectronBackend/db/initializeDb.ts
LMayTNG: To work on paths, `WHERE path GLOB 'path/to/folder/*'` would be the best option, because `GLOB` is always case sensitive and uses the indexes well.

However, `max_descendant_id` still has better performance. I tried around a lot with that when writing the resource tree query, and introducing `max_descendant_id` was basically the breakthrough that brought the runtime down from seconds to tens of milliseconds.

1. Numbers are a bit faster than strings in general, because they are much smaller and fit better in caches etc.
2. Rows in SQLite are identified and accessed by a [Rowid](https://www.sqlite.org/rowidtable.html), which in our table is equal to `id`. So when querying by path, SQLite needs to query the path index to get the rowid and then query the table to get the data. With `max_descendant_id`, we can directly query the table. That is a free 2x speedup for simple queries.
   - That could be mitigated by using a [WITHOUT ROWID](https://sqlite.org/withoutrowid.html) table and using the path as primary key, but we don't really want long strings as primary keys.
3. Often, we aren't really interested in the resources, but their attributions. Using `max_descendant_id`, we can directly query `resource_to_attribution.resource_id` instead of having to join with `resource` for the path.
   - That could be mitigated by further denormalizing `resource_to_attribution` and including the resource path, but that seems messy.
4. The query planner seems to behave better with integer ids, maybe because it expects the string lookups to be slower. But that's just a feeling, I can't really back that up.


I also like the semantics of treating the path as data that we don't want to work on, and identifying resources only by their id. The path should basically only be a display property. But that's probably a matter of taste.

Thanks for bringing that question up! It's always good to validate architectural decisions like this.

src/ElectronBackend/api/__tests__/utils.test.ts
```diff
@@ -342,16 +342,20 @@ async function expectDbContent(
   }
 }
 
-async function expectCwaConsistency() {
+async function expectCAAConsistency() {
```
LMayTNG: In camel case, I'd always treat abbreviations as words, i.e. `expectCaaConsistency`. Especially `cAAManualByResourceId` down below is really confusing and would be better as `caaManualByResourceId`.

See e.g. the [Google Styleguide](https://google.github.io/styleguide/jsguide.html#naming-camel-case-defined)

src/ElectronBackend/db/initializeDb.ts
```diff
@@ -95,62 +95,48 @@ async function initializeProgressBarTable(trx: Transaction<DB>) {
...
+  INSERT INTO cwa WITH RECURSIVE 
+  has_manual_attribution AS MATERIALIZED (
+    SELECT DISTINCT resource_id
+    FROM resource_to_attribution
+    WHERE attribution_is_external = 0
+  ),
+  has_unresolved_external_attribution AS MATERIALIZED (
+    SELECT DISTINCT resource_id
+    FROM resource_to_attribution
+    WHERE attribution_uuid IN (SELECT uuid FROM attribution WHERE is_external = 1 AND is_resolved = 0)
```
LMayTNG, line 115: We could use `attribution_is_external` here too.

```
WHERE attribution_is_external = 1 AND attribution_uuid IN (SELECT uuid FROM attribution WHERE is_resolved = 0)
```

src/ElectronBackend/db/initializeDb.ts
```diff
@@ -95,62 +95,48 @@ async function initializeProgressBarTable(trx: Transaction<DB>) {
...
-      closest_ancestor_with_breakpoint.ancestor
-    from resource
-    left join closest_ancestor_with_manual_attributions on resource.id = closest_ancestor_with_manual_attributions.resource_id
-    left join closest_ancestor_with_external_attributions on resource.id = closest_ancestor_with_external_attributions.resource_id
-    left join closest_ancestor_with_breakpoint on resource.id = closest_ancestor_with_breakpoint.resource_id
-    `.execute(trx);
+    FROM resource as child
+    JOIN cwa as parent ON child.parent_id = parent.resource_id
+  )
+  SELECT resource_id, is_file, breakpoint, manual, external FROM cwa
```
LMayTNG: Big improvement 👍

src/ElectronBackend/db/initializeDb.ts
LMayTNG: Yup, this was just chosen for consistency with auto-numerated ids

src/ElectronBackend/db/initializeDb.ts
```diff
@@ -540,14 +540,31 @@ async function initializeResourceToAttributionTable(
...
-  const insertStmt = rawDb.prepare(`
-    INSERT OR IGNORE INTO resource_to_attribution
+  function insertRows(
+    rows: {
+      resource_id: number;
+      attribution_uuid: string;
+    }[],
+    attribution_is_external: 0 | 1,
+  ) {
+    const singleValuesSql = attribution_is_external ? '(?, ?, 1)' : '(?, ?, 0)';
```
LMayTNG: Easier:

```
`(?, ?, ${attribution_is_external})`
```

LMayTNG (commented):
Great changes! Only a few minor nitpicks

src/ElectronBackend/db/initializeDb.ts
mariusAsaTNG: true

src/ElectronBackend/db/initializeDb.ts
mariusAsaTNG: When I add it it somehow makes the query slower, but I am not sure why. The best way to speed up this subquery would be to add is_resolved to the resource_to_attribution table

src/ElectronBackend/api/__tests__/utils.test.ts
mariusAsaTNG: good point, it looks better and is more understandable this way

src/ElectronBackend/db/initializeDb.ts
mariusAsaTNG: but this is a bit out of scope for this PR imo

LMayTNG (approved):

PR #3110, opened 2026-04-15
fix: add the right os for the tests

mariusAsaTNG:
### Summary of changes

Fix os version, so that we do not break mac intel.
Also added a step that runs mac intel in the E2E tests

### DIFF
--- a/.github/workflows/build-and-e2e-test.yml
+++ b/.github/workflows/build-and-e2e-test.yml
@@ -29,6 +29,8 @@ jobs:
             E2E: xvfb-run --auto-servernum --server-args="-screen 0 1280x960x24" -- yarn test:e2e:ci
           - os: macos-latest
             E2E: yarn test:e2e:ci
+          - os: macos-15-intel
+            E2E: yarn test:e2e:ci
           - os: windows-latest
             E2E: yarn test:e2e:ci
     env:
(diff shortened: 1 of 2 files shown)


mariusAsaTNG:
To fix mac-intel we will need to merge the PR and release

.github/workflows/build-and-e2e-test.yml
```diff
@@ -29,6 +29,8 @@ jobs:
             E2E: xvfb-run --auto-servernum --server-args="-screen 0 1280x960x24" -- yarn test:e2e:ci
           - os: macos-latest
             E2E: yarn test:e2e:ci
+          - os: macos-15-intel
+            E2E: yarn test:e2e:ci
```
mstykow: I think a smoke test for macos-15-intel would suffice and save us from doubling the CI check duration.

mstykow (changes requested):

.github/workflows/build-and-e2e-test.yml
mariusAsaTNG: good point, we now only run the `opens Opossum file and shows project as recently opened`, since this test is closest thing to a `hello world` I could find.

mstykow (approved):

PR #3119, opened 2026-04-21
feat: move classifications to database

LMayTNG:


src/Frontend/Components/ProgressBar/ProgressBar.tsx
```diff
@@ -77,7 +75,7 @@ export const ProgressBar: React.FC<ProgressBarProps> = (props) => {
       { enabled: props.selectedProgressBar === 'criticality' },
     );
 
-  const classifications = useAppSelector(getClassifications);
+  const classifications = useClassifications();
   const classificationProgressBarData =
     backend.getClassificationProgressBarData.useQuery(
```
mariusAsaTNG: in the progressBar, the classification only get called to be sent to the BE, would it not make sense to get them in the ProgressBar query instead of the FE

src/Frontend/Components/ResourceBrowser/ResourcesTree/ResourcesTreeNode/ResourcesTreeNode.tsx
```diff
@@ -23,7 +22,7 @@ import {
 import { type TreeNode } from '../../../VirtualizedTree/VirtualizedTreeNode/VirtualizedTreeNode';
 
 export function ResourcesTreeNode({ resource }: TreeNode) {
-  const classification_mapping = useAppSelector(getClassifications);
+  const classification_mapping = useClassifications();
```
mariusAsaTNG: the Classifications are only used in ClassificationIcon, would it make sense to use useClassifications inside the icon instead?

src/ElectronBackend/db/initializeDb.ts
```diff
@@ -203,6 +209,27 @@ async function initializeExternalAttributionSourceTable(
   }
 }
 
+async function initializeClassificationTable(
+  trx: Transaction<DB>,
+  classificationConfig: RawClassificationsConfig,
+) {
+  await trx.schema
+    .createTable('classification')
+    .addColumn('number', 'integer', (col) => col.primaryKey().notNull())
```
mariusAsaTNG: I do not like calling the column number, `classification` would make more sense, however the table is called that, so how about `classification_value`

src/Frontend/state/actions/resource-actions/types.ts
```diff
@@ -3,90 +3,41 @@
...
-export const ACTION_REPLACE_ATTRIBUTION_WITH_MATCHING =
-  'ACTION_REPLACE_ATTRIBUTION_WITH_MATCHING';
-export const ACTION_LINK_TO_ATTRIBUTION = 'ACTION_LINK_TO_ATTRIBUTION';
-export const ACTION_UNLINK_RESOURCE_FROM_ATTRIBUTION =
-  'ACTION_UNLINK_RESOURCE_FROM_ATTRIBUTION';
-export const ACTION_REMOVE_RESOLVED_EXTERNAL_ATTRIBUTIONS =
-  'ACTION_REMOVE_RESOLVED_EXTERNAL_ATTRIBUTIONS';
 export const ACTION_SET_PROJECT_METADATA = 'ACTION_SET_PROJECT_METADATA';
-export const ACTION_SET_BASE_URLS_FOR_SOURCES =
-  'ACTION_SET_BASE_URLS_FOR_SOURCES';
```
mariusAsaTNG, line 48: Nice find! Not sure how we missed all this

src/ElectronBackend/db/initializeDb.ts
LMayTNG: I agree. I think we can just call them `classification`, I have no big problem with `classification.classification`

src/Frontend/Components/ProgressBar/ProgressBar.tsx
```diff
@@ -77,12 +77,9 @@ export const ProgressBar: React.FC<ProgressBarProps> = (props) => {
...
   const classifications = useClassifications();
   const classificationProgressBarData =
-    backend.getClassificationProgressBarData.useQuery(
-      { classifications },
-      {
-        enabled: props.selectedProgressBar === 'classification',
-      },
-    );
+    backend.getClassificationProgressBarData.useQuery(undefined, {
+      enabled: props.selectedProgressBar === 'classification',
```
mariusAsaTNG, line 80: why not use the skipToken

src/Frontend/types/types.ts
```diff
@@ -24,16 +24,7 @@ export type SelectedProgressBar =
...
-  description: string;
-  resourceCount: number;
-  color: string;
-}
-
-export type ClassificationStatistics = Record<
-  number,
-  ClassificationStatisticsEntry
->;
+export type ClassificationStatistics = Record<number, number>;
```
mariusAsaTNG, line 27: nice change, the BE should not care about the color of the ClassificationStatistics

src/ElectronBackend/api/progressBarQueries.ts
mariusAsaTNG: For this function we could do the same with a `SELECT classification FROM classification ORDER BY classification DESC`

src/Frontend/Components/ProgressBar/ProgressBar.tsx
mariusAsaTNG: because skip token only works if the function takes props

src/ElectronBackend/api/progressBarQueries.ts
mariusAsaTNG: now it only takes selected resource as a prop

mariusAsaTNG (approved):

PR #3120, opened 2026-04-22
fix: jump to section buttons not working after selecting a signal

mstykow:
## Problem

The jump to next/previous section buttons in the signals panel had no effect once a signal was selected. They worked correctly before any selection was made.

## Root Cause

`styled()` was called **inside the render functions** (`renderJumpUp` / `renderJumpDown`), creating a new React component type on every render. React treated each render's `Icon` as a completely different component, unmounting the old DOM element and mounting a new one. This destroyed the button the user was clicking mid-interaction, silently swallowing the click event.

The MUI Tooltip `anchorEl` warning confirmed this — the anchor element was being removed from the document during re-renders triggered by selection state changes.

## Fix

Replaced inline `styled()` calls with the raw MUI icon components (`KeyboardArrowUpIcon`, etc.) and the `sx` prop for dynamic styles. The component type is now stable across renders, so the DOM element persists and click handlers fire correctly.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -204,6 +204,15 @@ export default tseslint.config(
         { avoidEscape: true, allowTemplateLiterals: false },
       ],
       yoda: 'error',
+      'no-restricted-syntax': [
+        'error',
+        {
+          selector:
+            ':function VariableDeclarator[init.callee.callee.name="styled"]',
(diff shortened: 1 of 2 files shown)


mariusAsaTNG (approved):
Nice find and fix

PR #3129, opened 2026-04-24
Fix/resolve many

mariusAsaTNG:
### Summary of changes

fix the bug, when you select >15000 attributions to resolve it does not work.
Also improve the indexes for the progress bar query

### DIFF
--- a/src/ElectronBackend/api/mutations.ts
+++ b/src/ElectronBackend/api/mutations.ts
@@ -23,6 +23,7 @@ import {
   replaceAttributions,
   unlinkAttributions,
   updateAttribution,
+  withBatching,
 } from './utils';
 
 type QueryInvalidation<Q extends QueryName> = {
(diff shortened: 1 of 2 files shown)


src/ElectronBackend/api/mutations.ts
```diff
@@ -495,8 +489,43 @@ export const mutations = {
...
+  async resolveAttributions(params: { attributionUuids: Array<string> }) {
+    return resolveAttributionsWithBatching(params.attributionUuids, true);
+  },
+
+  async unresolveAttributions(params: { attributionUuids: Array<string> }) {
+    return resolveAttributionsWithBatching(params.attributionUuids, false);
+  },
 } satisfies Record<string, MutationFunction>;
 
+async function resolveAttributionsWithBatching(
```
LMayTNG: Why introduce a new function for this instead of just adding batching to the existing `setAttributionsResolvedStatus` function?

mariusAsaTNG: because of the invalidations, we would have to collect all invalidations returned from the `setAttributionsResolvedStatus` function and deduplicate them, so this was the easier option

LMayTNG (approved):

PR #3136, opened 2026-04-24
fix: don't hide cursor in empty text box

LMayTNG:

### DIFF
--- a/src/Frontend/Components/TextBox/TextBox.tsx
+++ b/src/Frontend/Components/TextBox/TextBox.tsx
@@ -22,7 +22,7 @@ const classes = {
       backgroundColor: OpossumColors.white,
       borderRadius: '0px',
     },
-    '& label': {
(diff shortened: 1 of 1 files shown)


src/Frontend/Components/TextBox/TextBox.tsx
```diff
@@ -22,7 +22,7 @@ const classes = {
       backgroundColor: OpossumColors.white,
       borderRadius: '0px',
     },
-    '& label': {
+    '& label[data-shrink=true]': {
```
LMayTNG, line 25: Before, the label background hid the cursor. Now we only add the background when the label is moved to the top

mariusAsaTNG: nice fix

mariusAsaTNG (approved):

PR #3140, opened 2026-05-04
fix(electron): confirm before opening external files

mstykow:
## Summary
- before this change, double-clicking a second `.opossum` file from Finder while OpossumUI was already open could do nothing because the external open path did not reliably reach the app's existing open-file flow
- that external path also bypassed the existing unsaved-changes confirmation, so there was no clear user choice between keeping the current file and replacing it with the newly clicked one
- after this change, external `.opossum` opens are routed through the same discard-or-cancel flow as in-app opens, so users can explicitly keep their current work or discard local edits and continue opening the new file
- add regression coverage for startup/runtime macOS open-file handling and the path-aware unsaved-popup flow

## Testing
1. Launch OpossumUI and open an `.opossum` file.
2. Make a local change without saving, for example by editing attribution/package details.
3. While OpossumUI is still open, double-click a different `.opossum` file in Finder.
4. Verify that the existing unsaved-changes dialog appears instead of the second file being ignored.
5. Click **Cancel** and verify that the currently open file stays in place and the unsaved local change is still there.
6. Repeat steps 2-4, then click **Discard and Proceed**.
7. Verify that the newly clicked `.opossum` file opens and the previous unsaved local change is discarded.
8. Optional regression check: with no unsaved changes, double-click a different `.opossum` file and verify it opens immediately.

### DIFF
--- a/src/ElectronBackend/main/openFileRequests.ts
+++ b/src/ElectronBackend/main/openFileRequests.ts
@@ -1,0 +1,40 @@
+// SPDX-FileCopyrightText: Meta Platforms, Inc. and its affiliates
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+
+const pendingMacOsOpenFilePaths: Array<string> = [];
+
+let runtimeMacOsOpenFileHandler: ((filePath: string) => void) | null = null;
+
+export function queueMacOsOpenFile(filePath: string): void {
+  if (runtimeMacOsOpenFileHandler) {
+    runtimeMacOsOpenFileHandler(filePath);
+    return;
+  }
+
+  pendingMacOsOpenFilePaths.push(filePath);
+}
+
+export function consumeStartupMacOsOpenFilePath(): string | null {
+  return pendingMacOsOpenFilePaths.shift() ?? null;
+}
+
+export function clearPendingMacOsOpenFilePaths(): void {
+  pendingMacOsOpenFilePaths.length = 0;
+}
+
+export function setRuntimeMacOsOpenFileHandler(
+  handler: (filePath: string) => void,
+): void {
+  runtimeMacOsOpenFileHandler = handler;
+
+  for (const filePath of pendingMacOsOpenFilePaths.splice(0)) {
+    runtimeMacOsOpenFileHandler(filePath);
+  }
+}
+
+export function resetMacOsOpenFileHandlingForTests(): void {
+  clearPendingMacOsOpenFilePaths();
+  runtimeMacOsOpenFileHandler = null;
+}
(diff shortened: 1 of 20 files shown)


src/ElectronBackend/main/openFileRequests.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  for (const filePath of pendingMacOsOpenFilePaths.splice(0)) {
+    runtimeMacOsOpenFileHandler(filePath);
+  }
+}
+
+export function resetMacOsOpenFileHandlingForTests(): void {
+  clearPendingMacOsOpenFilePaths();
+  runtimeMacOsOpenFileHandler = null;
+}
```
mariusAsaTNG, line 40: Why is this all named MacOs, the double click also works on windows and on linux with certain app stores

src/Frontend/util/use-ipc-renderer.ts
```diff
@@ -87,5 +93,5 @@ export function useIpcRenderer<T extends Listener>(
       removeListener();
     };
     // eslint-disable-next-line @eslint-react/exhaustive-deps
-  }, dependencies);
+  }, [...dependencies]);
```
mariusAsaTNG: why this change

src/Frontend/state/actions/view-actions/types.ts
```diff
@@ -19,6 +19,10 @@ export const ACTION_SET_IMPORT_FILE_REQUEST = 'ACTION_SET_IMPORT_FILE_REQUEST';
 export const ACTION_SET_MERGE_REQUEST = 'ACTION_SET_MERGE_REQUEST';
 export const ACTION_SET_EXPORT_FILE_REQUEST = 'ACTION_SET_EXPORT_FILE_REQUEST';
 
+export interface OpenFileRequest {
+  filePath?: string;
+}
```
mariusAsaTNG: You only ever use the type OpenFileRequest together with null. It is also not that clear what the difference is between `open({})` and `open(null)`perhaps you could get rid of one of these options

mariusAsaTNG (commented):
Looks good, the whole `openFileRequests.ts` file is somewhat confusing but it is well tested and a somewhat small change

src/Frontend/state/actions/view-actions/types.ts
mstykow: Agreed. I replaced the optional object shape with an explicit union so the states are now `null` = no pending open, `{ kind: "dialog" }` = continue with the regular open dialog, and `{ kind: "path", filePath }` = continue opening a specific external file.

src/Frontend/util/use-ipc-renderer.ts
mstykow: Agreed — that dependency-array spread was incidental and not part of the feature. I reverted it so `useIpcRenderer` is back to its previous caller-controlled dependency behavior.

src/ElectronBackend/main/openFileRequests.ts
mstykow: I kept the `MacOs` naming here because this queue only bridges Electron's `app.on('open-file')` event path, which is the macOS-specific hook in this code. Windows and Linux external opens still come through different startup/runtime paths, so renaming this helper to sound cross-platform would make the scope less precise.

mariusAsaTNG (approved):

src/ElectronBackend/main/openFileRequests.ts
mariusAsaTNG: understood thanks for the explanation

PR #3142, opened 2026-05-07
refactor: split runtime prep into focused reviewable slices

mstykow:
## Summary
- The first three prep commits from this branch have already been fast-forwarded directly onto `main`.
- This PR now covers only the remaining discussion-heavy work: packaged converter/resource handling, E2E stabilization, and tooling/docs/CI alignment.
- Review focus should be on whether these three changes belong together, are scoped correctly, and represent the right follow-up after the groundwork already landed on `main`.

### `fix(electron): resolve packaged converter resources reliably`
**What it does**
- Packages `bin/opossum-file-cli` as an explicit extra resource.
- Routes notice and converter-resource resolution through shared helper paths.
- Makes converter execution failures distinguishable from invalid input failures.
- Adds focused path/converter tests.

**Why this needs review**
- It changes how packaged builds locate and execute conversion resources.
- It affects release-build behavior and error reporting, so it is worth validating the packaging assumptions and failure semantics carefully.

### `test(e2e): stabilize dialog and menu interactions`
**What it does**
- Adds stable file-picker test hooks.
- Replaces brittle `electron-playwright-helpers` sync dialog stubbing with a local helper that returns the real sync API shapes.
- Hardens dialog/menu/page-object interactions under Playwright, including the no-focused-window dialog fallback and more reliable picker click targets.
- Updates the affected E2E tests to assert selected paths before continuing.

**Why this needs review**
- It changes test helpers and interaction patterns in a fairly broad way.
- The goal is better determinism, but the abstractions and test ergonomics are still worth discussing before merge.

### `chore(tooling): align docs and CI with current workflow`
**What it does**
- Renames `.lintstagedrc.js` to `.lintstagedrc.mjs` and updates ESLint ignores accordingly.
- Removes the obsolete Babel config and the unused Babel preset/plugin dependencies that supported it.
- Introduces `ship:e2e:*` scripts for CI and switches the E2E workflow to use them.
- Updates `CONTRIBUTING.md` and `knip.json` so the docs and tooling metadata match the current workflow.

**Why this needs review**
- It changes contributor workflow documentation and CI behavior in the same slice.
- The end state likely makes sense, but it is exactly the kind of tooling/process update where reviewer alignment is useful.

## Context
The already-landed groundwork on `main` is:
- `feat(eslint): enhance TypeScript linting rules and improve import type usage`
- `refactor(deps): replace lodash with lodash-es for improved tree-shaking`
- `refactor(electron): update module system and streamline imports`

## Verification
- targeted backend/unit verification for the packaged converter/resource path changes
- targeted Playwright verification for import, merge, open-file, and menu interaction flows
- CI workflow verification for the updated packaged E2E path


package.json
```diff
@@ -211,6 +210,7 @@
       "main": "build/ElectronBackend/app.js"
     },
     "extraResources": [
+      "./bin/opossum-file-cli*",
```
LMayTNG, line 209: Don't we need to include `bin/opossum-file-cli.exe` here for Windows?

src/ElectronBackend/opossum-file/FileConverter.ts
```diff
@@ -9,18 +9,73 @@ import { join } from 'path';
...
+    const packagedCandidates = executableNames.map((executableName) =>
+      join(getPathOfExtraResource('bin'), executableName),
+    );
+
+    if (app?.isPackaged) {
+      return (
+        packagedCandidates.find((candidate) => fs.existsSync(candidate)) ??
+        packagedCandidates[0]
+      );
+    }
```
LMayTNG: Why are we unsure whether it's `.exe` or not?

src/ElectronBackend/opossum-file/FileConverter.ts
```diff
@@ -9,18 +9,73 @@ import { join } from 'path';
...
+      ]),
+    );
+
+    const candidates = candidateDirectories.flatMap((directory) =>
+      executableNames.map((executableName) => join(directory, executableName)),
+    );
+
+    return (
+      candidates.find((candidate) => fs.existsSync(candidate)) ?? candidates[0]
+    );
```
LMayTNG: Here too: Why are there so many possible places to look? Was this broken before or our changes make it more unstable?

src/e2e-tests/__tests__/linking-attributions.test.ts
```diff
@@ -121,7 +121,10 @@ test('links multiple attributions on resource at once', async ({
   resourcesTree,
   attributionsPanel,
   attributionDetails,
+  projectStatisticsPopup,
 }) => {
+  await projectStatisticsPopup.closeIfVisible();
+
```
LMayTNG: Why would the popup be open here? Why in no other test?

src/e2e-tests/page-objects/ImportDialog.ts
```diff
@@ -45,6 +57,12 @@ export class ImportDialog {
     titleIsHidden: async (): Promise<void> => {
       await expect(this.title).toBeHidden({ timeout: 30000 });
     },
+    inputFilePathIsVisible: async (filePath: string): Promise<void> => {
+      await expect(this.inputFileSelectionValue).toContainText(filePath);
+    },
+    opossumFilePathIsVisible: async (filePath: string): Promise<void> => {
+      await expect(this.opossumFileSelectionValue).toContainText(filePath);
+    },
```
LMayTNG: The `Visible` in the name confuses me. Why not just `inputFilePathIs(...)`? We don't assert visibility.

src/e2e-tests/page-objects/MenuBar.ts
```diff
@@ -27,7 +27,6 @@ const initiallyDisabledMenuItems = [
 
 const popupDisabledMenuItems = [
   text.menu.fileSubmenu.open,
-  text.menu.fileSubmenu.openRecent,
```
LMayTNG, line 30: Why?

src/e2e-tests/utils/fixtures.ts
```diff
@@ -163,7 +167,7 @@ export const test = base.extend<{
       info.fixme();
     });
   },
-  modKey: async ({}, use) => {
+  modKey: async ({ data: _data }, use) => {
```
LMayTNG: Why?

src/e2e-tests/playwright.config.ts
```diff
@@ -11,6 +11,24 @@ dotenv.config({ path: path.resolve(process.cwd(), '.env') });
...
+    return workers as `${number}%`;
+  }
+
+  const numericWorkers = Number.parseInt(workers, 10);
+
+  return Number.isInteger(numericWorkers) && numericWorkers > 0
+    ? numericWorkers
+    : 1;
+}
+
```
LMayTNG, line 31: Why? Wasn't it working before?

src/Frontend/Components/FilePathInput/FilePathInput.tsx
```diff
@@ -26,23 +28,50 @@ interface FilePathInputProps {
...
+        startIcon={<AttachFileIcon sx={baseIcon} />}
+        cursor={'pointer'}
+        showTooltip={true}
+        tooltipProps={props.tooltipProps}
+        // using a custom input component allows us to disable a lot of TextField
+        // behavior (e.g. horizontal text scrolling) that we don't want here
+        inputComponent={CustomInput}
+        disabled={props.disabled}
+      />
+    </MuiBox>
```
LMayTNG: Why? 

This also introduced a bug: We now open two File Picker dialogs when you click the input.

package.json
```diff
@@ -147,11 +143,19 @@
...
     "ship:auto:linux": "yarn ship-linux",
     "ship:auto:win32": "yarn ship-win",
+    "ship:e2e:auto": "cross-env NODE_OPTIONS=--no-deprecation run-script-os",
+    "ship:e2e:auto:darwin": "yarn node tools/runScriptArch.mjs ship:e2e:mac",
+    "ship:e2e:auto:linux": "yarn ship:e2e:linux",
+    "ship:e2e:auto:win32": "yarn ship:e2e:win32",
+    "ship:e2e:linux": "yarn install-deps:linux && yarn build && NODE_OPTIONS=--no-deprecation electron-builder --linux --x64 --publish never --dir",
+    "ship:e2e:mac:x64": "yarn install-deps:darwin:x64 && yarn build && NODE_OPTIONS=--no-deprecation electron-builder --mac --x64 --publish never --dir",
+    "ship:e2e:mac:arm64": "yarn install-deps:darwin:arm64 && yarn build && NODE_OPTIONS=--no-deprecation electron-builder --mac --arm64 --publish never --dir",
+    "ship:e2e:win32": "yarn install-deps:win32 && yarn build && NODE_OPTIONS=--no-deprecation electron-builder --win --x64 --publish never --dir",
```
LMayTNG: What does this do? Why is it called `e2e`? What do we need the non-`e2e` scripts for?

LMayTNG (commented):
I've got a few questions, but all-in-all it looks like solid improvements across a few areas

package.json
mstykow: The glob already covers it: ./bin/opossum-file-cli* matches both the suffixless binary and opossum-file-cli.exe, so no extra Windows-specific entry is needed here.

src/ElectronBackend/opossum-file/FileConverter.ts
mstykow: Agreed. The suffixless fallback was compensating for a mismatch between our older runtime lookup (always opossum-file-cli) and the new Windows download path (opossum-file-cli.exe). I tightened this to be explicit now: Windows resolves .exe, other platforms resolve the suffixless binary.

src/ElectronBackend/opossum-file/FileConverter.ts
mstykow: This was previously implicit rather than newly broken: packaged builds and dev/e2e runs resolve the app root differently. I simplified the lookup so we now only distinguish the two layouts we actually support: packaged resources vs dev/e2e repo paths.

src/e2e-tests/__tests__/linking-attributions.test.ts
mstykow: It should not be open here. I removed that workaround instead of keeping a blind closeIfVisible().

src/e2e-tests/page-objects/ImportDialog.ts
mstykow: Agreed. I renamed these to inputFilePathIs(...) and opossumFilePathIs(...), and did the same for the merge-dialog equivalent so the assertion name matches what it actually checks.

src/e2e-tests/page-objects/MenuBar.ts
mstykow: I kept this change. Open Recent is data-dependent now that the tests reset user settings on startup; with no recent entries it is disabled regardless of popup state, so it no longer belongs in the fixed popupDisabled expectation set.

src/e2e-tests/utils/fixtures.ts
mstykow: That fixture dependency was just noise. I removed it and kept modKey independent again.

src/e2e-tests/playwright.config.ts
mstykow: This keeps WORKERS aligned with what Playwright actually accepts here (a number or a percentage string) instead of passing arbitrary env strings through. So 1 still behaves the same, but values like 50% and invalid input now resolve deterministically.

src/Frontend/Components/FilePathInput/FilePathInput.tsx
mstykow: Agreed. The wrapper click handler plus the inner TextBox click handler could trigger the picker twice. I removed the wrapper click handling so only the actual input activation path fires.

package.json
mstykow: These were introduced to separate build an unpacked app directory for Playwright from the normal distributable ship:* flows. The e2e scripts use electron-builder --dir because the tests only need something runnable, while the non-e2e ship scripts still produce the actual release artifacts. This also results in a speedup of the e2e flows.

package.json
LMayTNG: You're right, I'm blind :smile:

src/Frontend/Components/FilePathInput/FilePathInput.tsx
LMayTNG: I still don't understand why we need a wrapper at all.

mstykow: The wrapper was originally introduced during the E2E stabilization work to provide a stable outer locator for the clickable trigger, separate from the inner locator used to assert the rendered path value. That worked, but I agree it was not the cleanest shape.

I followed up by moving those root-level semantics into TextBox itself instead: the stable root test hook, button role, tabIndex, and keyboard activation now live on the TextBox root, and FilePathInput is back to being a thin adapter without its own extra wrapper.

package.json
LMayTNG: I think it did make sense to do a full build in the CI, otherwise the first time we do that is in a release.
Also, running the E2E-Tests on the complete build also sounds good.

And the speedup on my machine is only about 7s, I don't think that's worth it.

src/e2e-tests/__tests__/linking-attributions.test.ts
LMayTNG: Now `ProjectStatisticsPopup.closeIfVisible` is unused, right?

src/Frontend/Components/FilePathInput/FilePathInput.tsx
```diff
@@ -26,23 +27,46 @@ interface FilePathInputProps {
...
+        }
+
+        if (event.key === 'Enter' || event.key === ' ') {
+          event.preventDefault();
+          onActivate();
+        }
+      }}
+      rootRole={'button'}
+      rootTabIndex={props.disabled ? -1 : 0}
+      inputDataTestId={props.testId ? `${props.testId}-input` : undefined}
```
LMayTNG: This is worse than before. The interface is really sloppy. Why does the textbox expose that many attributes of its root element? Why would anybody set `rootAriaDisabled` different to `disabled`? Why does the root element have such an active role anyway? We pass in a custom input element that we define in this file, why doesn't that have e.g. the role, tabIndex and keydown handler?

Also, I don't think that anybody will actually use the keyboard input. I don't really see why we would increase our complexity for that.

src/e2e-tests/__tests__/linking-attributions.test.ts
mstykow: Yes. It is unused now, and I removed it.

package.json
mstykow: I think that is a fair point, and I reverted the workflow back to the full `ship:auto` path.

The original motivation for the `ship:e2e:*` split was not primarily performance. The idea was to decouple “build something runnable for Playwright” from “build the actual release artifacts”, because the E2E job only needs an unpacked app directory while `ship:*` also carries release-specific packaging steps. The speedup was only a side effect.

That said, I agree that the existing setup has an important benefit: PR/main CI also exercises the real packaging path before release. Since that tradeoff is mostly a project preference and the older setup already gave us that coverage, I reverted to the previous approach.

src/Frontend/Components/FilePathInput/FilePathInput.tsx
mstykow: Agreed. The previous follow-up overcorrected and made `TextBox` too leaky.

I simplified it again: the extra wrapper is gone, all of the root-level role/ARIA/keyboard props are gone, and `TextBox` only keeps a narrow `rootDataTestId` hook so the E2E tests still have a stable click target. The rendered value continues to use `inputDataTestId` for assertions.

So the shape is back to: thin `FilePathInput` adapter, narrow `TextBox` API, stable selectors, and no extra interaction semantics in the general component.

LMayTNG (approved):
Thanks for the adjustments, looks good 👍

PR #3156, opened 2026-06-10
fix: write created output through the canonical save path

braemeradrian-meta:
## Problem

When OpossumUI creates an output file on first load (no existing output), resources that are **files-with-children** (e.g. an expanded `__metadata.zip`) were written into `resourcesToAttributions` with a **trailing slash** — e.g. `/path/__metadata.zip/` instead of `/path/__metadata.zip`. This is inconsistent with the path written on a normal save, and downstream tooling can reject it.

## Root cause

There were two independent output serializers:

- **First-load creation** (`createJsonOutputFile`) copied resource keys from the sanitized input, where `sanitizeResourcesToAttributions` → `canResourceHaveChildren` appends a trailing slash to any node with children — including files-with-children (treating them like folders).
- **Save** (`getSaveFileArgs`) keys off the DB `is_file` column, which marks files-with-children as files, so it correctly emits **no** trailing slash.

The two paths disagreed, and first-load output used the wrong one.

## Fix

Build the created output in memory only, initialize the DB, then persist it through `saveFile` — the exact same `is_file`-aware serializer every later save uses. First-load output is now identical to saved output by construction, so the two paths can no longer drift. Removes the now-redundant `createOutputInOpossumFile` and `parseOrCreateOutputJsonFile` helpers.

## Testing

- Added a regression test: loading an input with a files-with-children resource that has a pre-selected attribution now writes its path **without** a trailing slash.
- Verified end-to-end on a real ~13 MB `.opossum`: all 83 files-with-children attributions are now written without a trailing slash (was 83 with), regular folders still keep their slash (18,682 unchanged), and no attributions were dropped.
- `tsc --noEmit` clean; backend test suite passes (the only failures are pre-existing environmental ones in `opossum-file.test.ts` that shell out to a bundled binary).

### DIFF
--- a/src/ElectronBackend/input/loadFile.ts
+++ b/src/ElectronBackend/input/loadFile.ts
@@ -11,7 +11,7 @@ import type {
   ParsedFileContent,
   ResourcesToAttributions,
 } from '../../shared/shared-types';
-import { writeFile, writeOpossumFile } from '../../shared/write-file';
+import { saveFile } from '../api/saveFile';
 import { initializeDb } from '../db/initializeDb';
 import type {
   FileNotFoundError,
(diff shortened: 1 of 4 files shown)


src/ElectronBackend/input/loadFile.ts
```diff
@@ -156,28 +156,27 @@ export async function loadFile(
...
-        resourcesToExternalAttributions,
-        parsedInputData.metadata.projectId,
-        inputFileRaw,
-      );
+    const outputJsonPath = isOpossumFileFormat(filePath)
+      ? undefined
+      : getFilePathWithAppendix(filePath, '_attributions.json');
+
+    if (outputJsonPath !== undefined && fs.existsSync(outputJsonPath)) {
+      parsedOutputData = parseOutputJsonFile(outputJsonPath);
```
mariusAsaTNG, line 170: It feels like the if and else conditions switched but I think I just dont understand the `_attributions.json` path fully.

mariusAsaTNG (approved):
Looks good, only point is that I would have prefered a branch

src/ElectronBackend/input/loadFile.ts
braemeradrian-meta: The branches are switched because previously the creation logic was doubled between the opossum file and _attribution.json code paths (seen in the deleted parseOrCreateOutputJsonFile below). Previous flow:
```
isOpossumFile?
-> YES: Create output data and write as .opossum
-> NO: Check if _attributions.json exists
    -> YES: Load that
    -> NO: Create output data and write as _attribution.json
```

Now the saving separated to later. Here we have:
```
does _attribution.json exist?
 -> YES: load that
 -> NO: Create output data

<later when needing to save>
isOpossumFile?
 -> YES: write .opossum
 -> NO: write _attribution.json
```

PR #3157, opened 2026-06-10
refactor: remove redundant resource-path sanitization on load

braemeradrian-meta:
~~Stacked on #3156 — please review after that merges (this branch is based on it, so until then the diff also shows PR1's commits; I'll rebase + un-draft once PR1 lands).~~

## What

Removes `sanitizeResourcesToAttributions` and its helpers (`getAllResourcePaths`, `getListOfResourcePaths`), and deletes the `canResourceHaveChildren` helper that only they (in production) used.

## Why it's safe

The sanitizer did two things to the input attribution maps: (1) normalize each key to the canonical trailing-slash form, and (2) drop keys with no matching resource. Both are now fully handled by the database layer:

- `initializeDb` resolves attribution keys via `removeTrailingSlash`, so the slash form of the key is irrelevant.
- It returns no row for unknown resource paths, so dangling keys are dropped anyway.

The sanitized maps are only ever fed into the DB (they are not returned to the renderer), so the normalization was dead weight.

## Tests

Adds a characterization test in `loadFile.test.ts` asserting the load→DB linkage is correct for trailing-slash variants and that unknown paths are dropped — it passed both before and after removing the sanitizer, demonstrating equivalence.


braemeradrian-meta:
force pushed to rebase - branch now contains only the intended commit

abraemer (approved):

PR #3160, opened 2026-06-15
fix: allow opossum to open files larger than the maximum V8 string length

LMayTNG:

### DIFF
--- a/src/ElectronBackend/input/opossumArchive.ts
+++ b/src/ElectronBackend/input/opossumArchive.ts
@@ -1,0 +1,98 @@
+// SPDX-FileCopyrightText: Meta Platforms, Inc. and its affiliates
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+import * as fflate from 'fflate';
+import fs from 'fs';
+
+import {
+  INPUT_FILE_NAME,
+  OUTPUT_FILE_NAME,
+} from '../../shared/write-file-utils';
+
+// First two bytes of the local-file-header signature ('P', 'K').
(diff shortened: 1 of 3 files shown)


src/ElectronBackend/input/opossumArchive.ts
```diff
@@ -0,0 +1,97 @@
...
+  unzipper.onfile = (file) => {
+    if (file.name === INPUT_FILE_NAME) {
+      inputSeen = true;
+      file.ondata = (err, chunk) => {
+        if (err) {
+          entryError = err;
+          return;
+        }
+        inputChunks.push(chunk);
+      };
```
abraemer: This logic is duplicated between the 2 files. I think it would be nice to extract it into a function.
Also you could just throw the err as it will be caught by the try catch around the reading loop resulting in the same error message

LMayTNG: Good idea, done :+1:

src/ElectronBackend/input/opossumArchive.ts
```diff
@@ -0,0 +1,83 @@
...
+      inputChunks = collect(file);
+    } else if (file.name === OUTPUT_FILE_NAME) {
+      outputChunks = collect(file);
+    }
+  };
+
+  // fflate's streaming `Unzip` silently ignores bytes that aren't a valid
+  // local-file-header signature, so we sniff the zip magic ourselves to
+  // detect non-zip input.
+  let isZip: boolean | null = null;
```
abraemer: I think it would be better to extract the isZip check into its own function and run that as first thing here. The price is negligible (open the file, read 2 bytes) but it makes the loop much more readable imo

abraemer (approved):

PR #3164, opened 2026-06-17
refactor(zip): replace fflate with adm-zip for opossum file handling

abraemer:
## Summary

Replace fflate with adm-zip as the zip library for `.opossum` file reading and writing.

### Problem
The full decompressed `input.json` bytes (`storedInputFileRaw`) were held in memory for the entire session so they could be written back verbatim on each save. On every save, the entire zip was recreated from scratch via `fflate.zip()`, recompressing both `input.json` and `output.json`.

### Solution
adm-zip's `updateFile()` allows modifying individual zip entries in-place. Unchanged entries (like `input.json`) pass through with their original compressed bytes intact — no decompression or recompression needed.

Key changes:
- **opossumArchive.ts**: Use `AdmZip` instead of fflate streaming unzip; return the `AdmZip` instance for later reuse on save
- **write-file.ts**: Use `AdmZip.updateFile()` for saves — only `output.json` is recompressed; `input.json` compressed bytes pass through unchanged; `writeOpossumFile` is now synchronous
- **parseFile.ts**: Replace `fflate.strFromU8()` with `Buffer.toString()`
- **loadFile.ts / dbProcess.ts**: Replace `storedInputFileRaw` (`Uint8Array`) with `storedOpossumZip` (`AdmZip` instance)
- **types.ts**: Replace `inputFileRaw` with `opossumZip` in `ParsedOpossumInputAndOutput`
- **LegacyFileConverter.ts / fixtures.ts**: Remove `await` from now-synchronous `writeOpossumFile` calls
- Remove `fflate` dependency, add `adm-zip` + `@types/adm-zip`

### Compression level note
adm-zip uses Node.js `zlib` with the default compression level (6), whereas fflate was configured at level 5. The practical difference is negligible (~0.1% size difference).

### What stays
`stream-json` + `bytesAsStream()` + `parseJsonStream()` remain — they are still needed for parsing large input JSON to avoid V8's ~512 MB string-length cap.

### Checklist
- [x] All 856 unit tests pass
- [x] Typecheck clean
- [x] Lint clean
- [x] Format clean
- [x] No circular imports
- [x] Knip (unused code) clean


mariusAsaTNG:
awesome idea, I was trying to do this but never found the right package. Did not know about `adm-zip`

mariusAsaTNG:
Some numbers would have been nice. On the facebook file with some manual testing I can see that the zipping time goes from ~4s -> ~1s

src/shared/write-file.ts
```diff
@@ -32,49 +32,46 @@ export async function writeFile({
...
+    zip.writeZip(path);
+  } else {
+    zip = new AdmZip();
+    if (input) {
+      zip.addFile(INPUT_FILE_NAME, toBuffer(input));
+    }
+    if (output) {
+      zip.addFile(OUTPUT_FILE_NAME, toBuffer(output));
+    }
+    zip.writeZip(path);
```
mariusAsaTNG: Nit: you can move this before the return, since in all branches we want to write the Zip

src/shared/write-file.ts
```diff
@@ -32,49 +32,46 @@ export async function writeFile({
...
-          reject(err);
-        } else {
-          await fs.promises.writeFile(path, data);
+  zip?: AdmZip;
+}): string {
+  if (zip) {
+    if (output) {
+      zip.updateFile(OUTPUT_FILE_NAME, toBuffer(output));
+    }
+    zip.writeZip(path);
```
mariusAsaTNG: Instead of writeZip I would use `await zip.writeZipPromise(path)`, since this is non blocking. Perhaps there is a reason why you decided to go the sync route?
Leaving it async would minimize the blast radius of this PR

src/ElectronBackend/api/__tests__/saveFile.test.ts
```diff
@@ -76,7 +77,7 @@ describe('saveFile', () => {
   });
 
   it('writes to opossumFilePath as .opossum format', async () => {
-    const inputFileRaw = new Uint8Array([1, 2, 3]);
+    const opossumZip = new AdmZip();
```
mariusAsaTNG, line 80: Probably no reason, but why did the Uint8Array contain [1, 2, 3]?

src/shared/write-file.ts
abraemer: Oh good catch 👍

src/ElectronBackend/api/__tests__/saveFile.test.ts
abraemer: I have no idea 😅

abraemer:
> Some numbers would have been nice. On the facebook file with some manual testing I can see that the zipping time goes from ~4s -> ~1s

Thanks for timing it. I did this on the train and didn't have any sufficiently large opossum file at hand

abraemer:
rebased and addressed your comments

src/e2e-tests/utils/fixtures.ts
```diff
@@ -266,17 +266,15 @@ async function createTestFiles({
...
-  ]);
+  await writeFile({
+    path: json,
+    content: inputData,
+  });
+  await writeOpossumFile({
+    input: inputData,
+    path: opossum,
+    output: outputData,
+  });
```
mariusAsaTNG, line 277: Nit: could be await Promise.all again. But it really does not matter.

mariusAsaTNG (approved):

package.json
```diff
@@ -67,6 +67,7 @@
     "@testing-library/react": "16.3.2",
     "@testing-library/user-event": "14.6.1",
     "@trivago/prettier-plugin-sort-imports": "6.0.2",
+    "@types/adm-zip": "^0",
```
mariusAsaTNG, line 70: wait is this correct?

PR #3167, opened 2026-06-17
refactor(load): unify legacy and opossum file loading paths

abraemer:
## Summary

After the fflate-to-adm-zip migration (#3164), the loading code still branched on file format throughout the entire stack. Every layer — from listeners through the DB process to `loadFile.ts` — had `isOpossumFileFormat` checks that selected different parsing paths, different save targets, and different state fields depending on whether the input was a `.opossum` file or a legacy JSON input.

This PR eliminates that dual-path complexity by ensuring every load produces the same uniform result — a `LoadedArchive` containing an `AdmZip` instance, parsed input, and parsed output — regardless of whether the source was a `.opossum` file or a legacy JSON input. It also simplifies the save path so that `loadFile` passes the output data it already has in hand directly to `writeOpossumFile`, avoiding a pointless round-trip through the DB.

**Rebased onto main (includes #3164 adm-zip migration).**

## Problem

Before this PR, the import flow for legacy files was:

1. `importFileConvertAndLoadListener` → `convertToOpossum()` writes a `.opossum` file to disk
2. `initializeGlobalBackendState(opossumFilePath, true)` → opens it as `.opossum`
3. `loadFile()` → re-reads the file it just wrote, branches on `isOpossumFileFormat()`
4. Downstream code branches on format for save paths, state fields, etc.

This was wasteful (write + immediate re-read) and needlessly complex: the listener already *knew* the format but threw that information away, forcing `loadFile` to re-derive it from the file extension.

Meanwhile, the "direct legacy load" path in `loadFile.ts` (the `else` branch of `isOpossumFileFormat`) was unreachable from the UI and broken — saving would throw `Cannot save: no input file loaded` because `storedOpossumZip` was `undefined`.

Additionally, `saveFile` covertly depended on global DB state via `getSaveFileArgs()`, even in the initial load path where the output data was already in hand — forcing a pointless DB round-trip.

### Unified loading

Two format-specific loader functions produce a uniform `LoadedArchive` type:

- **`loadOpossumFile(path)`** — opens `AdmZip`, extracts and parses `input.json`/`output.json`
- **`loadLegacyFile(path)`** — reads raw JSON (`.json`/`.json.gz`), parses it, reads sidecar `_attributions.json` if present, then constructs a new in-memory `AdmZip` with both entries packed in

From that point on, all downstream code is format-agnostic: always an `AdmZip`, always an `opossumFilePath` for saves.

The listener layer explicitly picks the right loader (it already knows the format), so `loadFile()` no longer needs to infer it.

### Explicit save paths

`loadFile` now adds `output.json` to the in-memory zip when it creates output from scratch (the null-output case), then passes `resolvedOutputData` directly to `writeOpossumFile` — no round-trip through the DB. The user-initiated save path in `dbProcess.ts` explicitly calls `buildOpossumOutputFile` (which reads from the DB) then `writeOpossumFile`. The global-state dependency is visible at each call site, not hidden inside a wrapper function.

### Core loading
- **`parseFile.ts`**: Replace `parseOpossumFile` / `parseInputJsonFile` / `parseOutputJsonFile` with `loadOpossumFile` / `loadLegacyFile` returning uniform `LoadedArchive`
- **`loadFile.ts`**: Remove all `isOpossumFileFormat` branching; accepts `LoadedArchive` + `opossumFilePath`. When creating output from scratch, adds `output.json` to the zip and writes via `writeOpossumFile` directly (no DB round-trip)

### IPC layer
- **`dbProcess.ts`**: Split `LoadFileMessage` into `LoadOpossumFileMessage` (format: `"opossum"`) and `LoadLegacyFileMessage` (format: `"legacy"`); routes to correct loader, passes uniform result to `loadFile`. Save handler calls `buildOpossumOutputFile` + `writeOpossumFile` explicitly
- **`dbProcessClient.ts`**: Split `loadFile()` into `loadOpossumFile()` and `loadLegacyFile()`; `saveFile()` takes explicit `(projectId, opossumFilePath)` instead of `SaveFileParams`
- **`importFromFile.ts`**: Split into `loadOpossumFileFromPath()` and `loadLegacyFileFromPath()`

### Listeners
- **`listeners.ts`**: `importFileConvertAndLoadListener` no longer calls `convertToOpossum()` (disk write + re-read) for legacy files; calls `loadLegacyFileFromPath()` directly. ScanCode/OWASP still use `convertToOpossum`. `saveFileListener` simplified — no more `attributionFilePath`/`inputFileChecksum`. `initializeGlobalBackendState` simplified — no `isOpossumFormat` boolean.

### Save path
- **Deleted `saveFile.ts`** — `SaveFileParams` and `persistOutputFile` removed; callers use `writeOpossumFile` directly
- **`buildOpossumOutputFile.ts`** (renamed from `getSaveFileArgs.ts`): exports `buildOpossumOutputFile` (assembles `OpossumOutputFile` from DB state) and `getPreferredOverOriginIds` (for its own test); `querySaveFileArgs` is private
- **`writeOpossumFile`** in `write-file.ts`: always uses `zip.updateFile(OUTPUT_FILE_NAME, ...)` — the entry is guaranteed to exist because `loadFile` adds it when creating output

### Types & state
- **`types.ts`**: Remove `resourceFilePath`/`attributionFilePath` from `GlobalBackendState`; remove `ParsedOpossumInputAndOutput` (replaced by `LoadedArchive`)
- **`getLoadedFile.ts`**: Remove `getLoadedFileType()` (dead code); simplify `getLoadedFilePath()` and `isFileLoaded()`

### Deleted files
- `enums.ts` (only contained `LoadedFileFormat` enum)
- `isOpossumFileFormat.ts`
- `opossumArchive.ts` (already inlined in previous commit)
- `saveFile.ts` (replaced by direct `writeOpossumFile` calls)

## Test plan

- [x] `parseFile.test.ts` — 12 tests pass (updated to test `loadOpossumFile`/`loadLegacyFile`; added tests for sidecar output reading, `.json.gz`, error handling on sidecar)
- [x] `loadFile.test.ts` — updated to use new two-step flow (load archive → `loadFile`); tests that `output.json` is written correctly for legacy input
- [x] `buildOpossumOutputFile.test.ts` — simplified (only tests `buildOpossumOutputFile`, not the deleted wrapper)
- [x] `getLoadedFile.test.ts` — simplified (removed `LoadedFileFormat.Json` tests)
- [x] `globalBackendState.test.ts` — updated (removed `resourceFilePath`/`attributionFilePath`)
- [x] `listeners.test.ts` / `get-save-file-listener.test.ts` / `importFromFile.test.ts` / `errorHandling.test.ts` — all pass with updated mocks
- [x] `opossum-file.test.ts` — updated to use `loadOpossumFile` instead of `parseOpossumFile`
- [x] TypeScript type check passes (`tsc --noEmit`)
- [x] Knip (unused export check) passes
- [ ] Manual test: open a `.opossum` file
- [ ] Manual test: import a legacy `.json` file via the import dialog

## Rebase notes

- Commits from #3164 were dropped (already merged). 5 commits remain.
- DCO sign-off added to all commits.
- Floating promise in `dbProcess.ts` `saveFile` case fixed (added `await`).


src/ElectronBackend/api/saveFile.ts
```diff
@@ -26,7 +24,6 @@ export async function saveFile(
     metadata: {
       projectId: params.projectId,
       fileCreationDate: String(Date.now()),
-      inputFileMD5Checksum: params.inputFileChecksum,
```
abraemer, line 29: Why was this removed? Doesn't that change the output format?

src/ElectronBackend/main/listeners.ts
```diff
@@ -189,7 +188,7 @@ export const importFileConvertAndLoadListener =
   async (
     _: Electron.IpcMainInvokeEvent,
     resourceFilePath: string,
-    fileType: FileType,
+    _fileType: FileType,
```
abraemer: Do we need to keep _fileType? Or is that always opossum now because the most recent files and everything always points to the opossum file that was worked on?

src/ElectronBackend/api/saveFile.ts
```diff
@@ -4,16 +4,14 @@
 // SPDX-License-Identifier: Apache-2.0
 import type AdmZip from 'adm-zip';
 
-import { writeFile, writeOpossumFile } from '../../shared/write-file';
+import { writeOpossumFile } from '../../shared/write-file';
 import { serializeAttributions } from '../input/parseInputData';
 import type { OpossumOutputFile } from '../types/types';
 import { getSaveFileArgs } from './getSaveFileArgs';
 
 export interface SaveFileParams {
```
abraemer, line 12: This function does 2 things:
1) create the output data
2) write the file
Perhaps it would make sense to split it. Then we could test the logic of the output file without mocking.

abraemer:
Closing in favor of two smaller PRs that split this work:
1. #TBD — refactor(load): unify legacy and opossum file loading paths
2. #TBD — refactor: remove SaveFileParams, inline save path, delete saveFile.ts

PR #3168, opened 2026-06-19
fix: remove copyright & license info when first party is true

LMayTNG:
First-party attributions can't have a copyright or license information. But previously, when you entered them and switched to first party, they did have it in the database, we exported it to the opossum file and it caused an error when importing it.


LMayTNG:
After talking to Dominikus, it probably makes sense to keep the data in the database in case the click on first party was accidental, and just removing it on export. I'll do that

MarkusObendrauf (approved):

PR #3177, opened 2026-07-09
Compare any attribution or signal

### LINKED ISSUE #2563: Create general Diff functionality to compare attributions and signals as chosen by the user
It can be helpful to compare attributions to signals to figure out what changes were made.
Currently, that is possible to a certain extent by using the newly introduced "compare to original" function which opens a popup in which the selected attribution is compared to its original signal. This is only possible if there is an original signal.  

Extend the functionality of the Diff Popup to compare an attribution to a source signal or attribution chosen by the user.
Instead of opening a popup, this could be integrated into the audit view which should dynamically change if comparison / diff view is activated.

Further changes to the current diff popup functionality that should be considered:
- make the attribution side editable
- show the revert arrows on the side of the signal or attribution that the current attribution is compared to (like in the diff view of IDEs)
- no need for redo errors anymore and an internal buffer state of the popup anymore, as all revert actions would modify the `temporaryDisplayPackageInfo` of the attribution directly (user should use the revert button from audit view to undo changes instead)
- include auditing options in the comparison as well. This should be a natural consequence of showing the comparison in audit view directly. As the current attribution remains editable, the auditing options do as well and need no separate revert functionality

philipp-karcher:
### Summary of changes

- "compare with" button that allows you to select any other attribution or signal to compare with
- improved replacement selection alongside this: opacity, disabled unrelated buttons, more obvious cancel button, highlighting attributions to be replaced
- diff view is not interactive yet

#2563


LMayTNG:
Things I noticed while testing:
- We still have the arrows in the fields in the diff view, but no option to save the changes
- When Editing an attribution and clicking compare, we get the unsaved changes modal when selecting a second attribution. I'd expect to see it in the beginning when clicking the compare button, or just disable the button when we have unsaved changes.

philipp-karcher:
> Things I noticed while testing:
> 
>     * We still have the arrows in the fields in the diff view, but no option to save the changes
> 
>     * When Editing an attribution and clicking compare, we get the unsaved changes modal when selecting a second attribution. I'd expect to see it in the beginning when clicking the compare button, or just disable the button when we have unsaved changes.

Made the diff view fully readonly (except for the comparison to the original source) and disabled the compare button when there are unsaved changes. Making the diff view editable should be done along with some improvements in a separate PR.

src/Frontend/state/variables/use-compare-selection.ts
```diff
@@ -0,0 +1,19 @@
...
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+import { useVariable } from './use-variable';
+
+export const COMPARE_SELECTION_SOURCE = 'compare-selection-source';
+
+export interface CompareSelectionSource {
+  id: string;
+  label: string;
```
LMayTNG: I think the label is unused, right? I'd remove and just use the id.

src/Frontend/state/variables/use-compare-selection.ts
```diff
@@ -0,0 +1,19 @@
...
+}
+
+export function useCompareSelectionSource() {
```
LMayTNG, line 12: We could also do the query for the source in here, I think that would be cleaner encapsulation. So this wouldn't just return the id, but the whole attribution.

But that's more of a preference, I'm fine with keeping it this way if you like it better.

src/Frontend/state/variables/use-picker-mode.ts
```diff
@@ -0,0 +1,45 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+import { useAttributionIdsForReplacement } from './use-attribution-ids-for-replacement';
+import {
+  type CompareSelectionSource,
+  useCompareSelectionSource,
+} from './use-compare-selection';
+
+export type PickerMode =
+  | { mode: 'inactive'; isActive: false }
```
LMayTNG, line 10: The `isActive` is redundant. Why not just use `mode !== 'inactive'`?

src/Frontend/state/variables/use-picker-mode.ts
```diff
@@ -0,0 +1,45 @@
...
+      isActive: true;
+      attributionIdsForReplacement: Array<string>;
+    }
+  | {
+      mode: 'compare';
+      isActive: true;
+      compareSelectionSource: CompareSelectionSource;
+    };
+
+export function usePickerMode(): PickerMode {
```
LMayTNG, line 22: The `usePickerMode` hook is a good solution 👍

src/shared/text.ts
```diff
@@ -72,7 +72,13 @@ export const text = {
...
     commonEcosystems: 'Common Ecosystems',
     commonLicenses: 'Common Licenses',
-    compareToOriginal: 'Compare to original signal',
+    compareConfirm: 'Compare',
+    compareToOriginal: 'Compare with original',
+    compareWith: 'Compare with…',
+    comparedLicenseInformation: 'Compared License Information',
+    comparedPackageCoordinates: 'Compared Component Coordinates',
+    comparingWith: (name: string) => `Comparing with ${name}`,
+    replacing: (attributions: string) => `Replacing ${attributions}`,
```
LMayTNG: I think these two aren't used anymore

src/Frontend/Components/AttributionForm/AuditingOptions/AuditingOptions.tsx
```diff
@@ -32,9 +32,7 @@ export const AuditingOptions: React.FC<Props> = ({
 }) => {
   const options = useAuditingOptions({ packageInfo, isEditable });
   const [anchorEl, setAnchorEl] = useState<HTMLElement>();
-  const unselectedOptions = options.filter(
-    ({ selected, interactive }) => !selected && interactive,
-  );
```
LMayTNG, line 37: This leads to an empty dropdown when there are non-interactive unselected options:

Image

src/Frontend/Components/AttributionPanels/AttributionsPanel/ReplaceButton/ReplaceButton.tsx
```diff
@@ -21,24 +20,13 @@ export const ReplaceButton: React.FC<PackagesPanelChildrenProps> = ({
...
   const label = attributionIdsForReplacement.length
     ? text.packageLists.cancelReplace
     : text.packageLists.replace;
 
   const mutationsPending = useIsMutating() > 0;
 
-  const selectedAttributionIsExternal = useSelectedAttributionIsExternal();
-
-  useEffect(() => {
-    if (attributionIdsForReplacement.length && selectedAttributionIsExternal) {
```
LMayTNG, line 33: This introduces a weird bug: When the selected attribution is external, and I select a manual one via checkbox, I can click "use as replacement". This tries to use the external attribution as replacement for the manual one, which throws an error in the backend.

I think the best solution is to just disable the Replace button when the selected attribution is external.

src/Frontend/Components/AttributionForm/AttributionForm.tsx
```diff
@@ -59,22 +66,32 @@ export function AttributionForm({
...
       <MuiDivider variant={'middle'}>
         <MuiTypography>
           {isDiff
-            ? variant === 'diff-original'
-              ? text.attributionColumn.originalPackageCoordinates
-              : text.attributionColumn.currentPackageCoordinates
+            ? (sectionLabels?.packageCoordinates ??
+              (variant === 'diff-original'
+                ? text.attributionColumn.originalPackageCoordinates
+                : text.attributionColumn.currentPackageCoordinates))
```
LMayTNG: I think this is a bit chaotic. What do you think about a parameter like `sectionPrefix` which can be empty, some string like `Selected`, `Compared`, `Original`, ... and just prepending it to the sections?

LMayTNG (changes requested):
I've looked at most of the non-test files now. I've got a few comments, but it looks really good!

On the UI side: In the diff view, currently the selected component is right and the compared is left. That seems unintuitive to me, do we want to swap that?

philipp-karcher:
> I've looked at most of the non-test files now. I've got a few comments, but it looks really good!
> 
> On the UI side: In the diff view, currently the selected component is right and the compared is left. That seems unintuitive to me, do we want to swap that?

I don't have a super strong opinion on it, but to me the way it is seems more intuitive. If you click compare to original, the attribution you're on is right. Putting it left when you click the other compare seems confusing to me.

src/Frontend/state/variables/use-picker-mode.ts
philipp-karcher: > The `isActive` is redundant. Why not just use `mode !== 'inactive'`?

I was back and forth on this as well, but we're using `isActive` in quite a lot of places and it's a lot easier to read than `mode !== 'inactive'`

src/Frontend/state/variables/use-compare-selection.ts
philipp-karcher: Good idea, I like that!

src/Frontend/state/variables/use-picker-mode.ts
LMayTNG: Okay, I'm fine with that 👍

LMayTNG:
> I don't have a super strong opinion on it, but to me the way it is seems more intuitive. If you click compare to original, the attribution you're on is right. Putting it left when you click the other compare seems confusing to me.

I don't have a really strong opinion either, but a quick survey said that the expectation would be more that left is the "first" attribution, which is the original attribution when comparing to the original and the first selected attribution when comparing to another

src/e2e-tests/page-objects/AttributionDetails.ts
```diff
@@ -111,6 +126,28 @@ export class AttributionDetails {
     compareButtonIsVisible: async (): Promise<void> => {
       await expect(this.compareButton).toBeVisible();
     },
+    compareWithButtonIsVisible: async (): Promise<void> => {
+      await expect(this.compareWithButton).toBeVisible();
+    },
+    compareWithButtonIsHidden: async (): Promise<void> => {
+      await expect(this.compareWithButton).toBeHidden();
+    },
+    comparingWithTextIsVisible: async (_name: string): Promise<void> => {
```
LMayTNG: Why pass a name in here and not use it?

philipp-karcher: excellent question

LMayTNG (approved):
Thanks for the adjustments.

This is great work, especially for your first PR here. Thanks!

PR #3188, opened 2026-07-13
Add license based filtering to file tree

### LINKED ISSUE #3178: Add license based filtering to file tree
In OpossumUI, there is currently no easy solution to check all files which do have a given license

I want to be able to filter the license tree to all files which contain a certain license.
For the first version, an file contains a license if it has an attribution with this license
Possible follow up: Extend to signals after  feedback
* Where to add this filtering functionality (UI)
   * Add filter button right to search menu 
   * Filter button works like for attributions (including select license dropdown)
   * Filter button allows to clear the filter
   * Make the filter button change color to signify active filtering

philipp-karcher:
### Summary of changes

- License filter for resource tree
- Moved clear filter button into dropdown

#3178
I tried multiple approaches with the clear filter button. Either it was too small to comfortably click (imho the existing button was already too small), or it needed too much space and caused layout issues. I opted to move it into the dropdown, and introduced a new visual indicator for active filters.

### DIFF
--- a/src/Frontend/Components/FilterButton/FilterButton.tsx
+++ b/src/Frontend/Components/FilterButton/FilterButton.tsx
@@ -6,10 +6,10 @@ import Filter3Icon from '@mui/icons-material/Filter3';
 import FilterAltIcon from '@mui/icons-material/FilterAlt';
 import SentimentDissatisfiedIcon from '@mui/icons-material/SentimentDissatisfied';
 import SentimentSatisfied from '@mui/icons-material/SentimentSatisfied';
-import type { SxProps } from '@mui/material';
-import MuiBadge from '@mui/material/Badge';
+import MuiBadge, { type BadgeProps } from '@mui/material/Badge';
+import MuiIconButton from '@mui/material/IconButton';
 import MuiTooltip from '@mui/material/Tooltip';
-import { type CSSProperties, useCallback, useMemo, useState } from 'react';
+import { useCallback, useMemo, useState } from 'react';
 
 import { text } from '../../../shared/text';
 import type { Filter } from '../../shared-constants';
@@ -35,7 +35,7 @@ import {
   type SelectMenuOption,
   type SelectMenuProps,
 } from '../SelectMenu/SelectMenu';
-import { ClearMenuIcon, IconButton } from './FilterButton.style';
+import { ClearMenuIcon } from './FilterButton.style';
 import { LicenseAutocomplete } from './LicenseAutocomplete/LicenseAutocomplete';
 
 const FILTER_ICONS: Record<Filter, React.ReactElement<unknown>> = {
@@ -62,6 +62,13 @@ const FILTER_ICONS: Record<Filter, React.ReactElement<unknown>> = {
   [text.filters.modifiedPreferred]: <ModifiedPreferredIcon noTooltip />,
 };
 
+export interface FilterButtonTriggerProps {
+  content: React.ReactElement;
+  disabled: boolean;
+  isSomeFilterActive: boolean;
+  onClick: React.MouseEventHandler<HTMLElement>;
+}
+
 interface Props extends Pick<
   SelectMenuProps,
   'anchorArrow' | 'anchorPosition'
@@ -70,10 +77,9 @@ interface Props extends Pick<
   availableFilters: Array<Filter>;
   disabled?: boolean;
   emptyAttributions?: boolean;
-  activeIconSx?: SxProps;
-  activeBadgeStyle?: CSSProperties;
-  iconSx?: SxProps;
+  badgeColor?: BadgeProps['color'];
   mode: FilterPropsMode;
+  renderTrigger?: (props: FilterButtonTriggerProps) => React.ReactNode;
 }
 
 export const FilterButton: React.FC<Props> = ({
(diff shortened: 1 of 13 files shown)


src/Frontend/Components/FilterButton/FilterButton.tsx
```diff
@@ -143,31 +170,35 @@ export const FilterButton: React.FC<Props> = ({
         aria-label={'filter button'}
         onClick={(event) => setAnchorEl(event.currentTarget)}
         disabled={disabled || (emptyAttributions && !isSomeFilterActive)}
-        isClearHovered={isClearHovered}
         size={'small'}
-        color={isSomeFilterActive ? 'primary' : undefined}
+        color={iconSx ? undefined : isSomeFilterActive ? 'primary' : undefined}
```
abraemer: this chained tertiary is a bit hard to read

abraemer (approved):
1 minor comment. Manual test works. LGTM!

src/Frontend/Components/FilterButton/FilterButton.tsx
philipp-karcher: During reworking this I noticed the styling was slightly off anyway compared to the other buttons, and I ended up refactoring it, to separate the styling of the button from the actual dropdown

src/e2e-tests/page-objects/ResourcesTree.ts
```diff
@@ -45,4 +57,14 @@ export class ResourcesTree {
       await this.node.getByText(resourceName, { exact: true }).click();
     }
   }
+
+  async closeMenu(): Promise<void> {
+    await this.window.keyboard.press('Escape');
+  }
```
LMayTNG, line 63: I think escape doesn't close the menu, at least for me it doesn't. That seems like a bug. But it's not something you introduced, so I'd be fine with keeping this PR as is and maybe opening a bug ticket

src/Frontend/Components/ResourceBrowser/ResourceBrowser.tsx
```diff
@@ -46,12 +71,33 @@ export function ResourceBrowser() {
...
+      resourceTreeAttributionsWithSelectedLicense.data,
+    ],
+  );
   const resourceTreeAll = backend.getResourceTree.useQuery(
     {
       expandedNodes: expandedIdsAll,
+      onAttributionUuids: resourceTreeAttributionUuids,
       search: debouncedSearchAll,
       selectedResourcePath: selectedResourceId,
     },
```
LMayTNG, line 81: I think that we should do the filtering in the resource tree query instead of getting the attributions here and then passing them down. But I noticed that you already do that in your [next PR](https://github.com/opossum-tool/OpossumUI/pull/3191/changes), so I'll ignore it here.

src/Frontend/Components/FilterButton/FilterButton.tsx
```diff
@@ -127,50 +147,48 @@ export const FilterButton: React.FC<Props> = ({
...
+          isSomeFilterActive
+            ? {
+                selected: false,
+                id: 'clear-filters',
+                label: text.packageLists.clearFilters,
+                icon: <ClearMenuIcon />,
+                onAdd: () => clearFilters(),
+              }
+            : [],
+        ),
```
LMayTNG, line 155: I like this solution, it's an improvement over the previous X button 👍

src/Frontend/Components/FilterButton/FilterButton.tsx
```diff
@@ -69,7 +77,9 @@ interface Props extends Pick<
   availableFilters: Array<Filter>;
   disabled?: boolean;
   emptyAttributions?: boolean;
+  badgeColor?: BadgeProps['color'];
   mode: FilterPropsMode;
+  renderTrigger?: (props: FilterButtonTriggerProps) => React.ReactNode;
```
LMayTNG: This feels like a weird interface to me. You can pass in a trigger component that renders the button around the content, but the content is rendered here? That's a lot of complexity for just changing the color and padding of the button.

It feels like this could be accomplished easier with just a `triggerStyle` prop

LMayTNG (changes requested):
I'm still reviewing, but I figured I should probably already send the first comments so you don't merge with just Adrian's approval

src/Frontend/Components/ResourceBrowser/ResourceBrowser.tsx
```diff
@@ -46,12 +71,33 @@ export function ResourceBrowser() {
...
   // All resources
+  const [resourceTreeFilters] = useResourceTreeFilters();
+  const { selectedLicense: resourceTreeSelectedLicense } = resourceTreeFilters;
   const [searchAll, setSearchAll] = useVariable(ALL_RESOURCES_SEARCH, '');
   const debouncedSearchAll = useDebouncedInput(searchAll);
   const expandedIdsAll = useAppSelector(getExpandedIds);
+  const resourceTreeAttributionsWithSelectedLicense =
+    backend.listAttributions.useQuery(
+      {
+        external: false,
```
LMayTNG, line 60: I think filtering only on manual attributions is a sensible choice, but it's probably something we should mention to Indira and check if she has any objections

src/e2e-tests/page-objects/ResourcesTree.ts
philipp-karcher: The keyboard navigation was a bit iffy in general in this dropdown, I fixed the following:
- cleaner escape from nested dropdown
- nested autocomplete blocking keyboard navigation
- persistent highlighting when closing dropdown with escape

src/Frontend/Components/ResourceBrowser/ResourceBrowser.tsx
philipp-karcher: Yes, that's also what it says in the issue description

src/e2e-tests/page-objects/ResourcesTree.ts
LMayTNG: I'm on the fence about this. On the one hand it improves usability, on the other it now adds some complex dependencies to enable a feature that probably nobody will use.

Passing down an inputRef from FilterButton through LicenseAutocomplete to Autocomplete, only so FilterButton can expose a new focusContent option that can then be called on enter is a lot of new moving parts. And `disableRestoreFocus` is a pretty cryptic option.

I'd be in favor of moving that commit out of this PR so we can merge this today, and then thinking about the keyboard-friendliness afterwards

LMayTNG (approved):
Thanks! Looks great

src/e2e-tests/page-objects/ResourcesTree.ts
philipp-karcher: Yeah, that sounds reasonable :+1:

PR #3192, opened 2026-07-14
chore: correct graphviz version

philipp-karcher:
The new version was introduced by 151a710db7bdfe330256b8f34f3f8a9d60ffe075


abraemer (approved):

abraemer:
Just checking: This PR only contains a regenerated version of the database diagram. Is that what you intended?

philipp-karcher:
> Just checking: This PR only contains a regenerated version of the database diagram. Is that what you intended?

yeah, my agent always silently reverted this change as unrelated to the current work, so I thought yeah they have a point :laughing:

PR #3203, opened 2026-07-16
Fine granular filtering of incomplete attributions

### LINKED ISSUE #3180: Improve filtering of incomplete attributions
Improve the missing attribution filter to be specific on what is missing (copyright, license, version, ...)
Current version:

Image

Make the new possibilities accessible via a sub-menu

philipp-karcher:
### Summary of changes

- Fine-granular filters for incomplete attributions
- Highlight selected filters
- Disable license dropdown if empty

#3180 

Another approach I though about was to make the dropdowns multiselect, so that you can freely combine missing attributes. I decided against it, because the UX is just too hard to make work. We probably would want it to work as an OR (if you want to turn licenses into multiselect as well at least), but how do you visually communicate that? Especially as the top-level multiselect works as an AND. And then how do you show which are active without opening the dropdown?

That is why I settled on this simpler approach with an additional "Any" option.

I chose to add highlighting for selected filters, because especially when choosing "Any" the existing visual language for it is confusing: for the simple filters it's a checkmark, for the dropdowns it's the clear button


src/Frontend/Components/FilterButton/LicenseAutocomplete/LicenseAutocomplete.tsx
```diff
@@ -25,6 +25,8 @@ export const LicenseAutocomplete = ({
   return (
     <ValueFilterAutocomplete
       ariaLabel={'license names'}
+      disabled={licenses.length === 0 && !selectedLicense}
+      faded={licenses.length === 0}
```
mariusAsaTNG: I would have the same condition here `{licenses.length === 0 && !selectedLicense}`, since you can still disable a selected license and it does not seem that way with faded

mariusAsaTNG: Perhaps unify the two and have faded enabled anytime it is disabled.

mariusAsaTNG:
I do not think this PR added the bug, however, if you select a manual attribution and filter that attribution out, you are left without a 3rd column in the UI. See image
Screenshot 2026-07-23 at 07 57 01

src/Frontend/Components/FilterButton/ValueFilterAutocomplete/ValueFilterAutocomplete.tsx
```diff
@@ -0,0 +1,59 @@
...
+  setSelectedValue,
+  startAdornment,
+}) => {
+  return (
+    <Autocomplete<string, false, false, false>
+      sx={{ height: '38px' }}
+      background={'transparent'}
+      variant={'filled'}
+      placeholder={placeholder}
+      options={options}
```
mariusAsaTNG: Maybe worth discussing, however, I do not like that after selecting a value you can still select the others. I would change this to `options={selectedValue ? [selectedValue] : options}`, since this is also how the license filter seems to work. Then you would have to deselect it before selecting something else.

mariusAsaTNG: Otherwise you run into the problems you described: is this a multiselect, is this an AND, or OR filter and so on.

mariusAsaTNG (commented):
not fully done with the review

src/Frontend/Components/FilterButton/ValueFilterAutocomplete/ValueFilterAutocomplete.tsx
philipp-karcher: I'm unsure about this. I think at least after selecting something, it's visually pretty clear that the filters aren't combined. And I dislike that it basically just adds another click the user has to do if they want to change the filter.

But I do agree that it's not great that it works differently from the license filter. 

Actually, I think the license filter is behaving weirdly here, if you click into it, you can still type into the autocomplete but won't get any suggestions. What would you say about adjusting the license filter to work more like the new ones?

src/Frontend/Components/FilterButton/LicenseAutocomplete/LicenseAutocomplete.tsx
philipp-karcher: Yeah, good point, we can just remove faded then I think

philipp-karcher:
> I do not think this PR added the bug, however, if you select a manual attribution and filter that attribution out, you are left without a 3rd column in the UI. See image Screenshot 2026-07-23 at 07 57 01

I'm not sure what you mean. When I try this out, we are actually missing the attribution details view (if that is what you mean by 3rd column), but that's not depicted in the image :sweat_smile: 

The most consistent thing would probably be to show the new attribution form, since that is what we do when clicking on a folder in the resource tree. Though it seems kind of counterintuitive, because you probably want to look at attributions after filtering, not create new ones.

So the other option would be to just pre-select the first attribution in the list. Though if we do that, we should probably do it this way too for selecting a folder in the resource tree.

mariusAsaTNG:
> > I do not think this PR added the bug, however, if you select a manual attribution and filter that attribution out, you are left without a 3rd column in the UI. See image Screenshot 2026-07-23 at 07 57 01
> 
> I'm not sure what you mean. When I try this out, we are actually missing the attribution details view (if that is what you mean by 3rd column), but that's not depicted in the image 😅
> 
> The most consistent thing would probably be to show the new attribution form, since that is what we do when clicking on a folder in the resource tree. Though it seems kind of counterintuitive, because you probably want to look at attributions after filtering, not create new ones.
> 
> So the other option would be to just pre-select the first attribution in the list. Though if we do that, we should probably do it this way too for selecting a folder in the resource tree.

Sorry yeah I'm always a bit weary uploading images of opossum files here, since this is the open web. Exactly the attribution details view is missing. I am fine with either a new attribution or selecting the first one. Probably preselecting is better. For the resource selection it makes sense to create a new one, so we can keep it that way.

We also need to think about what happens to unsaved changes

src/Frontend/Components/FilterButton/ValueFilterAutocomplete/ValueFilterAutocomplete.tsx
mariusAsaTNG: The thing with the extra click is true but i would also not expect a box with only an x to be selectable. Changing the license filter might be an extra level of complexity. 
It is up to you, we could just unify the filters in this pr and then think about how we can improve them in a future one.

philipp-karcher: Yeah you're right, let's keep it simple in this PR. There are some other issues with the UX of the autocomplete, especially with keyboard navigation, I think it makes sense to do all these improvements together. I will create a ticket for that.

mariusAsaTNG:
Also I could not find anything else in the review, other than the 2 open points.

src/Frontend/Components/FilterButton/ValueFilterAutocomplete/ValueFilterAutocomplete.tsx
philipp-karcher: #3218

mariusAsaTNG (approved):
Would you fix the bug with the attribution details in another PR?

philipp-karcher:
> Would you fix the bug with the attribution details in another PR?

I'm fine with doing it in this PR, cleanly handling the unsaved changes just gave me a bit of trouble :sweat_smile:

mariusAsaTNG:
Looked over the new changes and it looks great. A question I had: Wouldn't it make sense to open the filters to the right, because if you are looking for an attribution and the filter is blocking the entire attribution view, it is not great.

philipp-karcher:
> Looked over the new changes and it looks great. A question I had: Wouldn't it make sense to open the filters to the right, because if you are looking for an attribution and the filter is blocking the entire attribution view, it is not great.

I don't have a strong opinion on this. I see your point that it looks awkward. I guess the advantage now is that you can see the details and that it opens like the other dropdowns. But that seems unrelated to this PR anyway. Feel free to create a task for it.

PR #3204, opened 2026-07-17
Use signals for license filter on resource tree

philipp-karcher:
### Summary of changes
- Base resource license filter on signals instead of attributions
- small refactoring to make this setting explicit

Feedback from Indira

### DIFF
--- a/src/ElectronBackend/api/__tests__/resourceTree.test.ts
+++ b/src/ElectronBackend/api/__tests__/resourceTree.test.ts
@@ -506,7 +506,7 @@ describe('getResourceTree', () => {
   });
 
   describe('license filtering', () => {
-    it('only shows resources linked to matching signals once', async () => {
+    it('only shows resources linked to matching external attributions once', async () => {
       const firstMitUuid = 'first-mit-uuid';
       const secondMitUuid = 'second-mit-uuid';
       const apacheUuid = 'apache-uuid';
--- a/src/ElectronBackend/api/__tests__/resourceTree.test.ts
+++ b/src/ElectronBackend/api/__tests__/resourceTree.test.ts
@@ -506,7 +506,7 @@ describe('getResourceTree', () => {
   });
 
   describe('license filtering', () => {
-    it('only shows resources linked to matching manual attributions once', async () => {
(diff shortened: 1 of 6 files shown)


src/e2e-tests/__tests__/interacting-with-resources.test.ts
LMayTNG: The link button promotes this from an external attribution to external **and** manual attribution, so this test succeeds no matter whether we filter on external or manual.

Maybe we could improve it by:
- Not clicking on the link button
- Adding a manual-only attribution with the same license and checking that it isn't included

src/ElectronBackend/api/__tests__/resourceTree.test.ts
```diff
@@ -506,7 +506,7 @@ describe('getResourceTree', () => {
   });
 
   describe('license filtering', () => {
-    it('only shows resources linked to matching manual attributions once', async () => {
+    it('only shows resources linked to matching signals once', async () => {
```
LMayTNG: I'd call them `external attributions` here because we do that in most of the codebase, but to be honest it's sadly not consistent.

src/Frontend/Components/ResourceBrowser/ResourceBrowser.tsx
```diff
@@ -30,6 +30,7 @@ import { ResourcesTree } from './ResourcesTree/ResourcesTree';
 
 const ALL_RESOURCES_SEARCH = 'all-resources-search';
 const LINKED_RESOURCES_SEARCH = 'linked-resources-search';
+const isLicenseFilterBasedOnExternalAttributions = true;
```
LMayTNG: I'd write that in UPPER_SNAKE_CASE to denote that it's configuration

LMayTNG (approved):
I've got a few small nitpicks, but it looks great 👍

src/e2e-tests/__tests__/interacting-with-resources.test.ts
philipp-karcher: I also renamed the variables, because it was really hard to tell what was going on in the tests

PR #3208, opened 2026-07-20
feat: show subtree of filter matches and add highlights

LMayTNG:


philipp-karcher:
2 things that are not included here:

- autoexpand for single children in filtered results
- search highlighting in linked resources

They don't need to be in this PR, but we should have tickets for them at least.

philipp-karcher (approved):

LMayTNG:
> 2 things that are not included here:
>
>    autoexpand for single children in filtered results
>    search highlighting in linked resources
>
>They don't need to be in this PR, but we should have tickets for them at least.

Added issues #3210 and #3211

PR #3209, opened 2026-07-20
fix: don't show parent tab for signals

### LINKED ISSUE #3205: Signals on parents
**Describe the bug**
Take a resource without a signal. If its parent has a signal but no attribution, the child resource will not show its "On Parents" tab in the signals list, until you add an attribution to the parent

LMayTNG:
Closes #3205

### DIFF
--- a/src/ElectronBackend/api/listAttributions.ts
+++ b/src/ElectronBackend/api/listAttributions.ts
@@ -39,12 +39,14 @@ export async function listAttributions(props: {
         ? await getResourceOrThrow(trx, props.resourcePathForRelationships)
         : undefined;
 
-      const closestAncestor = resourceForRelationships
-        ? await getClosestAncestorWithManualAttributionsBelowBreakpoint(
-            trx,
-            resourceForRelationships.id,
-          )
-        : undefined;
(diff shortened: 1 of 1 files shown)


src/ElectronBackend/api/listAttributions.ts
```diff
@@ -39,12 +39,14 @@ export async function listAttributions(props: {
...
-          )
-        : undefined;
+      const closestAncestor = props.external
+        ? undefined // External resources don't have inference, so showing the ancestor attributions would be confusing
+        : resourceForRelationships
+          ? await getClosestAncestorWithManualAttributionsBelowBreakpoint(
+              trx,
+              resourceForRelationships.id,
+            )
+          : undefined;
```
philipp-karcher, line 49: Not a big fan of the chained ternary, especially as two paths lead to the same result

philipp-karcher (approved):

src/ElectronBackend/api/listAttributions.ts
LMayTNG: Good point, that's cleaned up now

PR #3213, opened 2026-07-21
feat: split opossum files

### LINKED ISSUE #3194: Split opossum files into locked / editable
Multiple person could work on the same project simultaneously by splitting it up

Splitting simply copies the opossum file, marking only the selected files as editable in each copy

**Additional context**
See https://github.com/opossum-tool/opossum-file/pull/252 for a similar implementation idea
Open questions:
- Use opossum-file or implement directly in opossumUI?
  - Decision: use opossumUI 
- determine the right format to save which files are locked: output file? a new third file?

philipp-karcher:
### Summary of changes

- logic for splitting opossum files
- new optional file in .opossum zip, containing the split metadata
- split metadata is stored in db, to validate further splits

#3194 
This includes a very crude UI in a separate commit, so splitting can be tested manually. We can either hide it behind a feature flag, leave it as is, or remove it before merging


src/ElectronBackend/api/splitOpossumFile.ts
```diff
@@ -0,0 +1,65 @@
...
+} from '../split/split-opossum-file';
+import { saveFile, type SaveFileParams } from './saveFile';
+
+export interface SplitOpossumFileParams extends SaveFileParams {
+  opossumFilePath: string;
+  selectedFolderPaths: Array<string>;
+  selectedPartitionPath: string;
+}
+
+export async function splitOpossumFile(
```
LMayTNG, line 35: This function (and the parameters) probably deserve a comment

src/ElectronBackend/api/splitOpossumFile.ts
```diff
@@ -0,0 +1,65 @@
...
+  {
+    selectedFolderPaths,
+    selectedPartitionPath,
```
LMayTNG: I found the naming a bit confusing, because it suggests that the folder paths and the partition path are the same kind of path, when actually the folder paths are in the context of the resource tree and the partition path is the path for the output file.

Maybe change `selectedPartitionPath` to something like `partitionOutputPath`?

src/ElectronBackend/api/splitOpossumFile.ts
```diff
@@ -0,0 +1,65 @@
...
+  const selectedResourcePathsInDatabase = new Set(
+    selectedResources.map((resource) => resource.path),
+  );
+  for (const selectedFolderPath of normalizedSelectedFolderPaths) {
+    if (!selectedResourcePathsInDatabase.has(selectedFolderPath)) {
+      throw new Error(
+        `Selected resource '${selectedFolderPath}' does not exist`,
+      );
+    }
+  }
```
LMayTNG: This seems like it should also be done in `validateSelectedFolderPaths`

src/ElectronBackend/db/initializeDb.ts
```diff
@@ -705,3 +718,47 @@ async function initializeMetadataTable(
       .execute();
   }
 }
+
+async function initializeSplitInfoTable(
```
LMayTNG: The name hides that this also creates the `readonly_rule` table. I'd just split this into two functions, but I'd also be fine with renaming it

src/ElectronBackend/db/initializeDb.ts
```diff
@@ -705,3 +718,47 @@ async function initializeMetadataTable(
...
+    .createTable('split_info')
+    .addColumn('singleton', 'integer', (col) =>
+      col
+        .primaryKey()
+        .notNull()
+        .check(sql`singleton = 1`),
+    )
+    .addColumn('split_id', 'text', (col) => col.notNull())
+    .addColumn('input_sha256', 'text', (col) => col.notNull())
+    .execute();
```
LMayTNG: I don't really get the motivation for this table. What problems does it help against?

For the hash:
If we want to prevent accidentally merging split files from different projects, we can just check the projectId from the metadata. And if we really want to check that the input files are equal, we could just compare them when merging instead of saving the hashes before.

For the id:
If someone wants to merge two files from the same project where the paths allow that, I see no reason to prevent that. And if we want to prevent that, one id isn't enough, we'd need a whole list of ids.

src/ElectronBackend/db/split-info.ts
```diff
@@ -0,0 +1,85 @@
...
+          singleton: SPLIT_INFO_SINGLETON,
+          split_id: splitInfo.splitId,
+          input_sha256: splitInfo.inputSha256,
+        })
+        .execute();
+      await insertReadonlyRules(trx, splitInfo.readonlyRules);
+    });
+}
+
+export async function setReadonlyRules(
```
LMayTNG: Maybe `replaceReadonlyRules` to make it clearer?

src/ElectronBackend/split/split-opossum-file.ts
```diff
@@ -0,0 +1,287 @@
...
+  selectedPartitionPath: string;
+}
+
+interface SplitOpossumFileResult {
+  splitId: string;
+  selectedFolderPaths: Array<string>;
+  selectedPartitionPath: string;
+}
+
+export interface SplitOpossumArchiveArgs extends SplitOpossumArchivePaths {
```
LMayTNG: Nit: I feel like instead of extending it, composition could be cleaner. So just another entry `paths: SplitOpossumArchivePaths`

LMayTNG: Another nit: This should probably move up to below `SplitOpossumArchivePaths`

src/ElectronBackend/split/split-opossum-file.ts
```diff
@@ -0,0 +1,287 @@
...
+  readonlyRules: Array<ReadonlyRule>,
+): Array<string> {
+  if (selectedFolderPaths.length === 0) {
+    throw new SplitOpossumFileError(
+      'invalid-selection',
+      'Select at least one writable folder',
+    );
+  }
+
+  const normalizedPaths = [...selectedFolderPaths].sort();
```
LMayTNG: So normalizing them is just sorting them? Why do we return that? Why do other places need a sorted list?

src/ElectronBackend/split/split-opossum-file.ts
```diff
@@ -0,0 +1,287 @@
...
+    }
+    if (getReadonlyState(selectedPath, rulesByPath)) {
+      throw new SplitOpossumFileError(
+        'invalid-selection',
+        `'${selectedPath}' is readonly`,
+      );
+    }
+    if (
+      index > 0 &&
+      (selectedPath === normalizedPaths[index - 1] ||
```
LMayTNG: You can remove the equality check because the `isEqualOrDescendant` function already check for that`

src/ElectronBackend/split/split-opossum-file.ts
```diff
@@ -0,0 +1,287 @@
...
+        isEqualOrDescendant(selectedPath, normalizedPaths[index - 1]))
```
LMayTNG: This approach doesn't catch everything, because there are characters that are allowed in filenames that sort before `/`:

```
> ["/foo", "/foo/bar", "/foo.txt"].toSorted()

[ "/foo", "/foo.txt", "/foo/bar" ]
```

I think it would work if we sorted them after splitting at `/`:

```
> ["/foo", "/foo/bar", "/foo.txt"].toSorted((a, b) => a.split("/") > b.split("/"))

[ "/foo", "/foo/bar", "/foo.txt" ]
```

src/ElectronBackend/split/split-opossum-file.ts
```diff
@@ -0,0 +1,287 @@
...
+      readonlyRules: createReadonlyRules(
+        currentReadonlyRules,
+        selectedPaths,
+        'selected',
+      ),
+    },
+  };
+}
+
+function createReadonlyRules(
```
LMayTNG: I don't really understand this function. I think it needs some comments explaining how it works.

src/ElectronBackend/split/split-opossum-file.ts
```diff
@@ -0,0 +1,287 @@
...
+    currentReadonlyRules.map((rule) => [rule.path, rule.readonly]),
+  );
+  const rulesByPath = new Map<string, boolean>();
+  const boundaryPaths = new Set([
+    '/',
+    ...currentReadonlyRules.map((rule) => rule.path),
+    ...selectedPaths,
+  ]);
+
+  for (const resourcePath of [...boundaryPaths].sort()) {
```
LMayTNG: Does this have the same sorting problem as above? Or does it not matter because we only need the parent to be somewhere before the child?

src/shared/write-file.ts
```diff
@@ -59,10 +66,28 @@ export async function writeOpossumFile({
...
 }
 
+function updateSplitInfo(
+  zip: AdmZip,
+  splitInfo: SplitInfo | null | undefined,
+): void {
+  if (splitInfo === undefined) {
+    return;
+  }
+  if (splitInfo === null) {
```
LMayTNG: I don't like that `null` and `undefined` mean different things here. Do we actually need the "Leave splitInfo untouched" functionality or could we just make it mandatory to set it?

LMayTNG (changes requested):

src/ElectronBackend/db/initializeDb.ts
philipp-karcher: No longer relevant

src/shared/write-file.ts
philipp-karcher: This is an array now, so instead of null and undefined we have empty array and undefined, that should be clear enough

src/ElectronBackend/split/split-opossum-file.ts
philipp-karcher: I don't think it had that problem, but no longer relevant

src/ElectronBackend/split/split-opossum-file.ts
philipp-karcher: Yeah, simplified it

src/ElectronBackend/split/split-opossum-file.ts
philipp-karcher: True, we no longer rely on sorting at all, performance doesn't matter here, so I went for something simpler

src/ElectronBackend/split/split-opossum-file.ts
philipp-karcher: Yeah, no need for that

LMayTNG (approved):
Looks great 👍

PR #3216, opened 2026-07-22
test: add tests for saving output files

philipp-karcher:


src/e2e-tests/__tests__/saving-opossum-files.test.ts
```diff
@@ -0,0 +1,140 @@
...
+        }),
+      }),
+    },
+    packageInfo: existingPackageInfo,
+  },
+];
+
+for (const { data, packageInfo, title } of saveScenarios) {
+  test.describe(title, () => {
+    test.use({ data, openFromCLI: false });
```
mariusAsaTNG: It could make sense to also have this test for openFromCLI: true, but then we would have to open it the 2nd time from CLI as well, which is hard to do I think

mariusAsaTNG (approved):
Really nice tests, these would've caught our bugs

src/e2e-tests/__tests__/saving-opossum-files.test.ts
philipp-karcher: Good point! Re-opening from CLI is maybe a bit too much, but I think it still provides value to open it just the first time from CLI

mariusAsaTNG (approved):

PR #3220, opened 2026-07-23
refactor(load): unify legacy and opossum file loading paths

abraemer:
## Summary

After the fflate-to-adm-zip migration (#3164), the loading code still branched on file format throughout the entire stack. Every layer — from listeners through the DB process to `loadFile.ts` — had `isOpossumFileFormat` checks that selected different parsing paths, different save targets, and different state fields depending on whether the input was a `.opossum` file or a legacy JSON input.

This PR eliminates that dual-path complexity by ensuring every load produces the same uniform result — a `LoadedArchive` containing an `AdmZip` instance, parsed input, and parsed output — regardless of whether the source was a `.opossum` file or a legacy JSON input.

**Rebased onto current main. Splits #3167 (part 1 of 2).**

## Problem

Before this PR, the import flow for legacy files was:

1. `importFileConvertAndLoadListener` → `convertToOpossum()` writes a `.opossum` file to disk
2. `initializeGlobalBackendState(opossumFilePath, true)` → opens it as `.opossum`
3. `loadFile()` → re-reads the file it just wrote, branches on `isOpossumFileFormat()`
4. Downstream code branches on format for save paths, state fields, etc.

This was wasteful (write + immediate re-read) and needlessly complex: the listener already *knew* the format but threw that information away, forcing `loadFile` to re-derive it from the file extension.

Meanwhile, the "direct legacy load" path in `loadFile.ts` (the `else` branch of `isOpossumFileFormat`) was unreachable from the UI and broken — saving would throw `Cannot save: no input file loaded` because `storedOpossumZip` was `undefined`.

### Unified loading

A single loader function produces a uniform `LoadedArchive` type:

- **`loadOpossumFile(path)`** — opens `AdmZip`, extracts and parses `input.json`/`output.json`

Legacy JSON files go through the existing `convertToOpossum` (via `LegacyFileConverter`) and are then loaded via `loadOpossumFile` — the same path as ScanCode/OWASP. No separate `loadLegacyFile` parser exists; `LegacyFileConverter` is the single implementation of legacy JSON parsing.

From that point on, all downstream code is format-agnostic: always an `AdmZip`, always an `opossumFilePath` for saves.

### Save path (preliminary)

`loadFile` now adds `output.json` to the in-memory zip when it creates output from scratch (the null-output case), then saves via `saveFile`. The save path is further cleaned up in the follow-up PR.

### Core loading
- **`parseFile.ts`**: Replace `parseOpossumFile` / `parseInputJsonFile` / `parseOutputJsonFile` with `loadOpossumFile` returning uniform `LoadedArchive`
- **`loadFile.ts`**: Remove all `isOpossumFileFormat` branching; accepts `LoadedArchive` + `opossumFilePath`. When creating output from scratch, adds `output.json` to the zip and writes via `saveFile`

### IPC layer
- **`dbProcess.ts`**: Simplified `LoadFileMessage` (no discriminated union — single format); inlines `loadOpossumFile` call directly, preserving the `onProgress` message
- **`dbProcessClient.ts`**: Single `loadOpossumFile()` method
- **`importFromFile.ts`**: Single `loadOpossumFileFromPath()` function

### Listeners
- **`listeners.ts`**: `importFileConvertAndLoadListener` routes all file types (including legacy) through `convertToOpossum` + `openOpossumFile`. `saveFileListener` simplified — no more `attributionFilePath`/`inputFileChecksum`. `initializeGlobalBackendState` simplified — no `isOpossumFormat` boolean.

### Types & state
- **`types.ts`**: Remove `resourceFilePath`/`attributionFilePath` from `GlobalBackendState`; remove `ParsedOpossumInputAndOutput` (replaced by `LoadedArchive`)
- **`getLoadedFile.ts`**: Remove `getLoadedFileType()` (dead code); simplify `getLoadedFilePath()` and `isFileLoaded()`

### Deleted files
- `enums.ts` (only contained `LoadedFileFormat` enum)
- `isOpossumFileFormat.ts`

## Test plan

- [x] `parseFile.test.ts` — `loadOpossumFile` tests pass
- [x] `loadFile.test.ts` — tests use `convertToOpossum` + `loadOpossumFile` as fixture (matches production path)
- [x] `getLoadedFile.test.ts` — simplified (removed `LoadedFileFormat.Json` tests)
- [x] `globalBackendState.test.ts` — updated (removed `resourceFilePath`/`attributionFilePath`)
- [x] `listeners.test.ts` / `get-save-file-listener.test.ts` / `importFromFile.test.ts` / `errorHandling.test.ts` — all pass with updated mocks
- [x] `opossum-file.test.ts` — updated to use `loadOpossumFile` instead of `parseOpossumFile`
- [x] TypeScript type check passes (`yarn typecheck`)
- [x] Knip (unused export check) passes (`yarn knip`)
- [x] Circular import check passes
- [x] Manual test: open a `.opossum` file
- [x] Manual test: import a legacy `.json` file via the import dialog


src/ElectronBackend/input/importFromFile.ts
```diff
@@ -33,37 +33,41 @@ async function handleParsingError(
...
+}
+
+async function handleLoadResult(
+  mainWindow: BrowserWindow,
+  result: {
+    ok: boolean;
+    error?: ParsingError;
+    projectTitle?: string;
+    projectId?: string;
+  },
```
philipp-karcher: Why not type this as `LoadFileIpcResult`, then you don't need the `result.error!` below

src/ElectronBackend/input/importFromFile.ts
```diff
@@ -33,37 +33,41 @@ async function handleParsingError(
...
+export async function loadOpossumFileFromPath(
   mainWindow: BrowserWindow,
-  filePath: string,
+  opossumFilePath: string,
 ): Promise<void> {
-  mainWindow.webContents.send(AllowedFrontendChannels.ResetLoadedFile, {
-    resetState: true,
+  const globalState = getGlobalBackendState();
+  const result = await getMainDbClient().loadOpossumFile(opossumFilePath, {
+    inputFileChecksum: globalState.inputFileChecksum,
```
philipp-karcher: We're not passing a `LoadFileProgressCallback`, so we don't have any progress updates anymore. We should also have an e2e test that catches future regressions like this.

philipp-karcher (changes requested):

PR #3225, opened 2026-07-24
fix: prevent delayed unsaved popup after resource navigation

philipp-karcher:
The tree previously triggered a second unsaved-changes check after its asynchronous attribution lookup completed. This could open a delayed popup after the original resource-navigation popup had been cancelled, causing unrelated E2E interactions to time out. Example: https://github.com/opossum-tool/OpossumUI/actions/runs/30076489731

Now autoselecting is owned centrally instead of being scattered across multiple places.

### DIFF
--- a/src/Frontend/Components/AttributionPanels/PackagesPanel/PackagesPanel.tsx
+++ b/src/Frontend/Components/AttributionPanels/PackagesPanel/PackagesPanel.tsx
@@ -11,7 +11,7 @@ import {
   intersection,
   isEqual,
 } from 'lodash-es';
-import { useCallback, useEffect, useMemo, useState } from 'react';
+import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 
 import type { Attributions, Relation } from '../../../../shared/shared-types';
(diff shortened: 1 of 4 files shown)


src/Frontend/Components/AttributionPanels/PackagesPanel/PackagesPanel.tsx
```diff
@@ -110,6 +114,51 @@ export const PackagesPanel = ({
...
+      lastResourceIdWithAutoSelectionRef.current !== selectedResourceId
+    ) {
+      lastResourceIdWithAutoSelectionRef.current = selectedResourceId;
+      const closestAttribution =
+        Object.values(attributions).find(
+          ({ relation }) => relation === 'resource',
+        ) ??
+        Object.values(attributions).find(
+          ({ relation }) => relation === 'parents',
+        );
```
LMayTNG: We could probably use `groupedIds` from below here, right? So we don't have to search through the resources twice

LMayTNG (approved):

PR #3226, opened 2026-07-24
feat: UI for splitting opossum files

### LINKED ISSUE #3195: UI for splitting opossum files
We need something in the UI to trigger #3194, e.g. context menu on a folder in the resource tree

philipp-karcher:
### Summary of changes

* New dialog for splitting opossum files
* Can be triggered via context menu on resource tree or file menu
* Includes a resource tree where you can select multiple folders

#3195


src/ElectronBackend/main/menu/fileMenu.ts
```diff
@@ -169,6 +170,24 @@ function getSaveFile(webContents: WebContents): MenuItemConstructorOptions {
...
 }
 
+function getCreateCollaborativePartition(
+  webContents: WebContents,
+): MenuItemConstructorOptions {
+  const globalBackendState = getGlobalBackendState();
+  return {
+    icon: getIconBasedOnTheme(
+      'icons/follow-up-white.png',
+      'icons/follow-up-black.png',
```
LMayTNG: It might be confusing to use the same icon as for the follow-up export.

We could think about using a different icon like [this](https://fonts.google.com/icons?selected=Material+Symbols+Outlined:folder_shared:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=collaboration&icon.size=24&icon.color=%23e3e3e3), [this](https://fonts.google.com/icons?selected=Material+Symbols+Outlined:handshake:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=collaboration&icon.size=24&icon.color=%23e3e3e3), [this](https://fonts.google.com/icons?selected=Material+Symbols+Outlined:export_notes:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=share&icon.size=24&icon.color=%23e3e3e3) or [this](https://fonts.google.com/icons?selected=Material+Symbols+Outlined:arrow_split:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=split&icon.size=24&icon.color=%23e3e3e3)

src/Frontend/Components/ResourceBrowser/ResourcesTree/ResourcesTree.tsx
```diff
@@ -70,6 +84,9 @@ export const ResourcesTree = ({ resources, sx }: Props) => {
   const handleContextMenu = useCallback(
     (event: MouseEvent<HTMLElement>, resource: ResourceTreeNodeData) => {
       event.preventDefault();
+      void window.electronAPI
+        .isOpossumFileLoaded()
+        .then(setIsOpossumFileLoaded);
```
LMayTNG: This whole flow is unneeded, because we only load opossum files, so any loaded file has the opossum format.

Looks like the option to open JSON files was removed in February 2025 (in favor of the import flow that converts them to Opossum files), I didn't know that there still were some remnants there. We should probably remove them in a follow-up PR.

So in essence, we can remove this, the guard on the context menu and the new IPC channel.

src/ElectronBackend/main/menu/fileMenu.ts
```diff
@@ -169,6 +170,24 @@ function getSaveFile(webContents: WebContents): MenuItemConstructorOptions {
...
+    ),
+    label: text.menu.fileSubmenu.createCollaborativePartition,
+    click: () =>
+      webContents.send(AllowedFrontendChannels.ShowSplitDialogWithUnsavedCheck),
+    enabled:
+      getLoadedFileType(globalBackendState) === LoadedFileFormat.Opossum &&
```
LMayTNG: This can just be replaced with `isFileLoaded`, see my other comment.

src/ElectronBackend/main/setUpIpcHandling.ts
```diff
@@ -46,6 +49,14 @@ export function setupIpcHandling(
...
   );
   ipcMain.handle(IpcChannel.SaveFile, saveFileListener(window));
+  ipcMain.handle(
+    IpcChannel.SelectSplitDestination,
+    selectSplitDestinationListener(window),
+  );
+  ipcMain.handle(IpcChannel.IsOpossumFileLoaded, () => {
+    const globalBackendState = getGlobalBackendState();
+    return getLoadedFileType(globalBackendState) === LoadedFileFormat.Opossum;
+  });
```
LMayTNG: Can be removed, see my other comment.

src/ElectronBackend/split/split-opossum-file.ts
```diff
@@ -79,24 +79,27 @@ export async function splitOpossumArchive({
...
       'invalid-destination',
-      'Destination directory does not exist',
+      'Destination file must use the .opossum extension',
     );
   }
-  if (fs.existsSync(resolvedPartitionOutputPath)) {
+  if (!fs.existsSync(path.dirname(resolvedPartitionOutputPath))) {
     throw new SplitOpossumFileError(
       'invalid-destination',
-      'Destination file already exists',
```
LMayTNG, line 99: Don't we want to guard against the file already existing?

src/ElectronBackend/split/split-opossum-file.ts
```diff
@@ -119,6 +122,22 @@ async function validateResourcesExist(paths: string[]) {
...
+    const firstStats = fs.statSync(firstPath);
+    const secondStats = fs.statSync(secondPath);
+    return (
+      firstStats.dev === secondStats.dev && firstStats.ino === secondStats.ino
+    );
+  } catch {
+    return false;
+  }
+}
+
```
LMayTNG: This is overkill. I'd just check for path equality.

src/Frontend/Components/BackendCommunication/BackendCommunication.tsx
```diff
@@ -124,6 +126,11 @@ export const BackendCommunication: React.FC = () => {
     (_, fileFormat) => dispatch(showMergeDialogOrOpenUnsavedPopup(fileFormat)),
     [dispatch],
   );
+  useIpcRenderer<ShowSplitDialogWithUnsavedCheckListener>(
+    AllowedFrontendChannels.ShowSplitDialogWithUnsavedCheck,
```
LMayTNG: I'd rename this to just `ShowSplitDialog`, whether to do an unsaved check or not is the decision of the frontend.

LMayTNG (changes requested):

src/shared/text.ts
```diff
@@ -237,6 +238,7 @@ export const text = {
   resourceBrowser: {
     allResources: (selectedResources: number, totalCount: number) =>
       `Resources (${new Intl.NumberFormat().format(selectedResources)} / ${new Intl.NumberFormat().format(totalCount)})`,
+    splitHere: 'Create collaborative partition',
```
LMayTNG: It seems like you call the feature "collaborative partition" in the UI, but "split" in the code. I feel like inconsistencies like that are normal (see "external attribution" vs "signal"), but they normally arise from renaming things in the UI. I wouldn't introduce that in the first PR.

I also like the name 'split' a bit better. I think it communicates the idea in a simpler way.

src/ElectronBackend/split/split-opossum-file.ts
philipp-karcher: The idea was to avoid differences in case sensitivity between platforms, but yeah path equality is probably enough here because we use a file picker.

src/ElectronBackend/split/split-opossum-file.ts
philipp-karcher: The file picker on my OS at least already prompted me to confirm if I want to overwrite before I could select an already existing file (and I would have to look into whether this decision is somehow accessible to us). Is that not consistent for different platforms?

src/Frontend/types/types.ts
```diff
@@ -34,11 +34,20 @@ export interface ResourceCriticalityCounts {
...
+  | PopupType.UpdateAppPopup;
+
+export type PopupInfo =
+  | { popup: PopupWithoutPayload }
+  | { popup: PopupType.NotSavedPopup; attributionId?: string }
+  | {
+      popup: PopupType.ImportDialog | PopupType.MergeDialog;
+      fileFormat: FileFormatInfo;
+    }
+  | { popup: PopupType.SplitDialog; resourcePath?: string };
```
LMayTNG, line 50: Clean 👍

src/Frontend/Components/SplitDialog/MultiResourcePicker.tsx
```diff
@@ -0,0 +1,198 @@
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
```
LMayTNG, line 2: For all of the files: We should also include the meta copyright

src/ElectronBackend/split/split-opossum-file.ts
LMayTNG: Do we ever want to allow overriding the files? It's probably never what the user wants, right?

But I do see your point that if the dialog asks the user to confirm overriding but then we don't allow it, that's weird. It would be great if we could just disable that in the dialog, but I don't see an option for that.

philipp-karcher: >  It's probably never what the user wants, right?

I don't think that's a decision we should make for the user, if they select that file, that's where we should write. It's probably not common, but idk maybe they have an old partition that was created by mistake lying around that they want to overwrite

LMayTNG: Okay, sounds sensible :+1:

src/ElectronBackend/main/menu/fileMenu.ts
```diff
@@ -169,6 +166,19 @@ function getSaveFile(webContents: WebContents): MenuItemConstructorOptions {
   };
 }
 
+function getCreateCollaborativePartition(
```
LMayTNG: The old naming still survived in a few places

src/Frontend/Components/SplitDialog/MultiResourcePicker.tsx
LMayTNG, line 2: This one file is still missing the Meta copyright

src/ElectronBackend/main/menu/fileMenu.ts
philipp-karcher: should be fixed now

LMayTNG (approved):

PR #3229, opened 2026-07-27
feat: loading of partially locked opossum files

### LINKED ISSUE #3196: Loading of partially locked opossum files
Files from #3194 should not be editable in full.

As a first step, we do not need to adjust the UI, we simply only show resources/attributions that would be editable. An attribution is editable iff it is attached to at least one editable resource.
We can calculate editable flags once on load, and adjust all queries to hide readonly resources/attributions (watch the performance)
When modifying an attribution attached to both editable and readonly resources, the attribution should be cloned/split into a readonly and editable part.

philipp-karcher:
### Summary of changes
* hide readonly resources and attributions
  * readonly ancestors of editable resources still need to be shown
* disable creation and linking of attributions to readonly resources
* clone attributions linked to readonly and editable resources before modifying / deleting

#3196


src/ElectronBackend/db/initializeDb.ts
LMayTNG: Incidental: This function should probably be called `initializeClosestAttributedAncestorsTable`

src/ElectronBackend/db/initializeDb.ts
```diff
@@ -187,6 +209,18 @@ async function initializeProgressBarTable(trx: Transaction<DB>) {
...
+    .createIndex('closest_attributed_ancestors_editable_file_idx')
+    .on('closest_attributed_ancestors')
+    .columns([
+      'resource_is_readonly',
+      'is_file',
+      'manual_is_readonly',
+      'external_is_readonly',
+      'resource_id',
+    ])
+    .execute();
```
LMayTNG, line 228: This index is surprising to me. So this comment is a note to myself to come back here after I read the queries 😄

src/ElectronBackend/db/initializeDb.ts
```diff
@@ -392,6 +452,18 @@ async function initializeResourceTable(
...
+    .createIndex('resource_is_readonly_id_idx')
+    .on('resource')
+    .columns(['is_readonly', 'id'])
+    .execute();
+
+  await sql`
+    CREATE INDEX resource_has_editable_descendant_id_idx
+    ON resource (id)
+    WHERE has_editable_descendant = 1
+  `.execute(trx);
```
LMayTNG: Can be done with kysely, you just have to include `has_editable_descendant` too (which seems a bit more idiomatic to me anyway)

```
await trx.schema
    .createIndex('resource_has_editable_descendant_id_idx')
    .on('resource')
    .columns(['id', 'has_editable_descendant'])
    .where('has_editable_descendant', '=', 1)
    .execute();
```

src/ElectronBackend/db/initializeDb.ts
```diff
@@ -436,6 +508,9 @@ async function initializeAttributionTable(
     .addColumn('data', 'text', (col) => col.notNull())
     .addColumn('is_external', 'integer', (col) => col.notNull())
     .addColumn('is_resolved', 'integer', (col) => col.notNull().defaultTo(0));
+  schema = schema.addColumn('resource_access', 'integer', (col) =>
```
LMayTNG: Why not add that to the `.addColumn` chain?

src/ElectronBackend/db/initializeDb.ts
```diff
@@ -436,6 +508,9 @@ async function initializeAttributionTable(
...
+    col.notNull().defaultTo(AttributionResourceAccess.Unlinked),
```
LMayTNG: I think we should use a check constraint here to make sure this can only take allowed values.

src/ElectronBackend/db/initializeDb.ts
```diff
@@ -653,6 +728,37 @@ async function initializeResourceToAttributionTable(
...
+      ),
+      ${AttributionResourceAccess.Unlinked}
+    )
+  `.execute(trx);
+
+  await sql`
+    CREATE INDEX attribution_resource_access_audit_idx
+    ON attribution (resource_access, is_external, is_resolved, uuid)
+  `.execute(trx);
+}
```
LMayTNG: Can be done with kysely:

```
await trx.schema
      .createIndex('attribution_resource_access_audit_idx')
      .on('attribution')
      .columns(['resource_access', 'is_external', 'is_resolved', 'uuid'])
      .execute();
```

src/ElectronBackend/db/generated/databaseTypes.ts
```diff
@@ -24,6 +24,10 @@ export interface Attribution {
   data: string;
   is_external: number;
   is_resolved: Generated<number>;
+  /**
+   * Whether this attribution has no resources, only readonly resources, only writable resources, or both.
+   */
+  resource_access: Generated<number>;
```
LMayTNG, line 30: It would be nice if we could make this an enum type, but that would probably be too complex to be worth it

src/ElectronBackend/api/utils.ts
LMayTNG, line 42: Do we need to watch out that we don't remove attributions from a readonly resource here?

src/ElectronBackend/api/queries.ts
LMayTNG: I just noticed that this query is unused, we can just remove it

src/ElectronBackend/db/initializeDb.ts
LMayTNG: If that is for the `getManualFilesQuery` query, we should probably use `is_file`, `resource_is_readonly`, `manual`, `manual_is_readonly`, `resource_id` as columns, right?

LMayTNG (commented):
Currently, the behavior after splitting a file is a bit weird. You can still edit the selected resources, but you can't set a new location or split the file.

Also, the tree collapses after selecting a resource to split. That's not a big problem, but it might be easy to improve

src/ElectronBackend/db/initializeDb.ts
philipp-karcher: Doesn't this just make the index bigger than it needs to be? The [docs](https://kysely-org.github.io/kysely-apidoc/classes/CreateIndexBuilder.html#examples-7) recommend using `sql.ref` in this scenario:

```
await trx.schema
    .createIndex('resource_has_editable_descendant_id_idx')
    .on('resource')
    .column('id')
    .where(sql.ref('has_editable_descendant'), '=', 1)
    .execute();
```

src/ElectronBackend/api/utils.ts
```diff
@@ -605,18 +605,71 @@ async function updateAttributionResourceAccess(
...
+
+    oldUuidsToNewUuids[attribution.uuid] = newUuid;
+  }
+
+  await updateAttributionResourceAccess(trx, [
+    ...mixedAttributions.map((attribution) => attribution.uuid),
+    ...Object.values(oldUuidsToNewUuids),
+  ]);
+
+  return oldUuidsToNewUuids;
```
LMayTNG, line 702: Wouldn't it make sense to also include non-mixed attribution uuids here and just point them to themselves? Then we wouldn't have to do `oldUuidsToNewUuids[selectedAttribution.id] ?? selectedAttribution.id` everywhere

src/ElectronBackend/db/generated/databaseTypes.ts
philipp-karcher: Yeah I don't think there is a clean way to easily achieve this

src/Frontend/Components/AttributionDetails/ButtonRow/ButtonRow.tsx
```diff
@@ -266,6 +270,7 @@ export function ButtonRow({ packageInfo, isEditable }: Props) {
   function renderLinkButton() {
     if (
       isSelectedResourceBreakpoint ||
+      isSelectedResourceReadonly ||
```
LMayTNG, line 273: I didn't think about the fact that we now have visible readonly ressources and have to handle stuff like this. That means that the MVP is even closer to the VP than we thought. But I guess there's no way around that

src/ElectronBackend/db/initializeDb.ts
LMayTNG: Sounds good :+1:

src/ElectronBackend/db/initializeDb.ts
```diff
@@ -121,6 +121,48 @@ export async function initializeDb(inputFile: ParsedFileContent) {
     });
 }
 
+export async function refreshReadonlyDataAfterSplit() {
```
LMayTNG: This function probably doesn't belong in `initializeDb`, right?

src/ElectronBackend/api/utils.ts
philipp-karcher: I also added a check that raises an error if something slips through

src/ElectronBackend/db/initializeDb.ts
LMayTNG: Also, why the "afterSplit" in the name? `refreshReadonlyData` seems enough

src/ElectronBackend/db/initializeDb.ts
```diff
@@ -121,6 +121,48 @@ export async function initializeDb(inputFile: ParsedFileContent) {
...
+  await getDb()
+    .transaction()
+    .execute(async (trx) => {
+      await sql`
+        WITH RECURSIVE resource_readonly(id, is_readonly) AS (
```
LMayTNG: It would be cooler to do this in kysely for typesafety 😁 
But I'm also fine with this, as this is a pretty complex query

src/ElectronBackend/db/initializeDb.ts
```diff
@@ -121,6 +121,48 @@ export async function initializeDb(inputFile: ParsedFileContent) {
...
+        UPDATE resource
+        SET is_readonly = (
+          SELECT is_readonly
+          FROM resource_readonly
+          WHERE resource_readonly.id = resource.id
+        )
+      `.execute(trx);
+
+      await sql`
+        UPDATE resource AS parent
```
LMayTNG: But this can definitely be done in kysely

src/Frontend/Components/AttributionDetails/ButtonRow/ButtonRow.tsx
philipp-karcher: Yeah, I only noticed this during implementation as well.

src/ElectronBackend/db/initializeDb.ts
philipp-karcher: Migrated it

src/ElectronBackend/db/initializeDb.ts
philipp-karcher: Agreed on both

src/ElectronBackend/api/utils.ts
```diff
@@ -605,7 +605,14 @@ async function updateAttributionResourceAccess(
...
       ),
-      ${AttributionResourceAccess.Unlinked}
+      (
+        SELECT CASE
+          WHEN is_readonly = 1 THEN ${AttributionResourceAccess.Readonly}
+          ELSE ${AttributionResourceAccess.Writable}
+        END
+        FROM resource
+        WHERE path = ''
+      )
```
LMayTNG: We don't expect this to happen, so this is just defensive coding, right? I'd add that as comment

src/ElectronBackend/db/initializeDb.ts
```diff
@@ -39,7 +39,7 @@ export const comments: Record<string, Record<string, string>> = {
       "External attributions (UI: 'signals') and manual attributions (UI: 'attributions')",
     data: 'All of the attribution as JSON',
     resource_access:
-      'Whether this attribution has no resources, only readonly resources, only writable resources, or both.',
+      'Whether this attribution has only readonly resources, only writable resources, or both. Attributions without resources use the root resource access.',
```
LMayTNG: It's something we don't expect to happen, so I wouldn't add it to the comment

src/e2e-tests/page-objects/PackageCard.ts
```diff
@@ -61,9 +61,15 @@ export class PackageCard {
...
     },
+    isEditable: async (packageInfo: RawPackageInfo): Promise<void> => {
+      await expect(this.node(packageInfo)).toBeVisible();
+    },
     isHidden: async (packageInfo: RawPackageInfo): Promise<void> => {
       await expect(this.node(packageInfo)).toBeHidden();
     },
+    isReadonly: async (packageInfo: RawPackageInfo): Promise<void> => {
+      await expect(this.node(packageInfo)).toBeHidden();
+    },
```
LMayTNG, line 72: Good idea 👍

src/ElectronBackend/db/refresh-readonly-data.ts
```diff
@@ -14,37 +12,71 @@ export async function refreshReadonlyData() {
...
-        )
-      `.execute(trx);
+        .updateTable('resource')
+        .set((eb) => ({
+          is_readonly: eb
+            .selectFrom('resource_readonly')
+            .select('is_readonly')
+            .whereRef('resource_readonly.id', '=', 'resource.id'),
+        }))
+        .execute();
```
LMayTNG, line 53: Image

LMayTNG:
Great job :+1: I've now looked at all the code and at your adjustments after my comments, so there's only the two last nitpicks and then we can merge this.

src/ElectronBackend/api/utils.ts
philipp-karcher: Done, also migrated this to kysely

LMayTNG (approved):

PR #3235, opened 2026-07-29
fix: preserve nested readonly rules in partitions

philipp-karcher:
Fixed a bug in the splitting logic, where the new partition dropped pre-existing readonly rules nested under one of the selected folders


LMayTNG (approved):
Good point, I missed that too

PR #3238, opened 2026-07-30
test: immediately fail on unstubbed dialog call

philipp-karcher:
Previously, this resulted in a teardown timeout because the app could not be closed


LMayTNG (approved):
Nice 👍

Issue #3198, opened 2026-07-15
UI for merging partially locked opossum files


philipp-karcher:
**Is your feature request related to a problem? Please describe.**
We need UI to trigger #3197

**Describe the solution you'd like**
There needs to be a button to merge files back together. Potential errors and warnings emitted by the merging logic need to be displayed

PR #3241, opened 2026-07-31
feat: merge partially locked opossum files

### LINKED ISSUE #3197: Merge partially locked opossum files
To become useful, there needs to be an opposite operation of #3194

* On merge, merge output.json-s
* Un-locked files wins
    * Attributions touched in both files are duplicated (check: this should not never occur, we clone them already on edit)
    * Assert no out of order merges
    * Check if the unlocked files overlap -- warn and default to import wins

philipp-karcher:
### Summary of changes

* backend functionality for merging split files back together
* checks for conflicts in purely readonly path
  * provides a boolean flag to ignore this
  * intended workflow: run into conflicts -> warn in UI -> user can say proceed anyway -> retry with flag set

#3197
No UI yet, that will be done in #3198 (inlcuding things like progress updates)

### DIFF
--- a/src/ElectronBackend/api/mergeOpossumFiles.ts
+++ b/src/ElectronBackend/api/mergeOpossumFiles.ts
@@ -8,6 +8,7 @@ import { mergeOpossumArchives } from '../split/merge-opossum-files';
 import { saveFile, type SaveFileParams } from './saveFile';
 
 export interface MergeOpossumFilesParams {
+  ignoreReadonlyResourceOutputConflicts: boolean;
   saveFileParams: Omit<SaveFileParams, 'opossumFilePath'> & {
     opossumFilePath: string;
   };
@@ -15,23 +16,34 @@ export interface MergeOpossumFilesParams {
 }
 
 export interface MergeOpossumFilesFromPathsParams {
+  ignoreReadonlyResourceOutputConflicts: boolean;
   inputPaths: Array<string>;
   outputPath: string;
 }
 
 export async function mergeOpossumFilesFromPaths({
+  ignoreReadonlyResourceOutputConflicts,
   inputPaths,
   outputPath,
 }: MergeOpossumFilesFromPathsParams): Promise<void> {
-  await mergeOpossumArchives({ inputPaths, outputPath });
+  await mergeOpossumArchives({
+    ignoreReadonlyResourceOutputConflicts,
+    inputPaths,
+    outputPath,
+  });
 }
 
 export async function mergeOpossumFiles(
-  { saveFileParams, partitionPaths }: MergeOpossumFilesParams,
+  {
+    ignoreReadonlyResourceOutputConflicts,
+    saveFileParams,
+    partitionPaths,
+  }: MergeOpossumFilesParams,
   opossumZip: AdmZip,
 ): Promise<void> {
   await saveFile(saveFileParams, opossumZip);
   await mergeOpossumArchives({
+    ignoreReadonlyResourceOutputConflicts,
     inputPaths: [saveFileParams.opossumFilePath, ...partitionPaths],
     outputPath: saveFileParams.opossumFilePath,
   });
--- a/src/ElectronBackend/api/mergeOpossumFiles.ts
+++ b/src/ElectronBackend/api/mergeOpossumFiles.ts
@@ -1,0 +1,38 @@
+// SPDX-FileCopyrightText: Meta Platforms, Inc. and its affiliates
+// SPDX-FileCopyrightText: TNG Technology Consulting GmbH <https://www.tngtech.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+import type AdmZip from 'adm-zip';
+
+import { mergeOpossumArchives } from '../split/merge-opossum-files';
+import { saveFile, type SaveFileParams } from './saveFile';
+
+export interface MergeOpossumFilesParams {
+  saveFileParams: Omit<SaveFileParams, 'opossumFilePath'> & {
+    opossumFilePath: string;
+  };
+  partitionPaths: Array<string>;
+}
+
+export interface MergeOpossumFilesFromPathsParams {
+  inputPaths: Array<string>;
+  outputPath: string;
+}
+
+export async function mergeOpossumFilesFromPaths({
+  inputPaths,
+  outputPath,
+}: MergeOpossumFilesFromPathsParams): Promise<void> {
+  await mergeOpossumArchives({ inputPaths, outputPath });
+}
+
+export async function mergeOpossumFiles(
+  { saveFileParams, partitionPaths }: MergeOpossumFilesParams,
+  opossumZip: AdmZip,
+): Promise<void> {
+  await saveFile(saveFileParams, opossumZip);
+  await mergeOpossumArchives({
+    inputPaths: [saveFileParams.opossumFilePath, ...partitionPaths],
+    outputPath: saveFileParams.opossumFilePath,
+  });
+}
--- a/src/ElectronBackend/api/mergeOpossumFiles.ts
+++ b/src/ElectronBackend/api/mergeOpossumFiles.ts
@@ -9,9 +9,7 @@ import { saveFile, type SaveFileParams } from './saveFile';
 
 export interface MergeOpossumFilesParams {
   ignoreReadonlyResourceOutputConflicts: boolean;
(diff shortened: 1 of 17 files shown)


src/ElectronBackend/api/mergeOpossumFiles.ts
```diff
@@ -0,0 +1,50 @@
...
+import type AdmZip from 'adm-zip';
+
+import { mergeOpossumArchives } from '../split/merge-opossum-files';
+import { saveFile, type SaveFileParams } from './saveFile';
+
+export interface MergeOpossumFilesParams {
+  ignoreReadonlyResourceOutputConflicts: boolean;
+  saveFileParams: Omit<SaveFileParams, 'opossumFilePath'> & {
+    opossumFilePath: string;
+  };
```
LMayTNG: `opossumFilePath` is a string in `SaveFileParams` too, so this Omit & join doesn't make sense:


```suggestion
  saveFileParams: SaveFileParams;
```

src/ElectronBackend/main/listeners.ts
```diff
@@ -106,6 +107,41 @@ export const splitCurrentOpossumFileListener =
...
+    });
+  };
+
+export async function mergeOpossumFilesFromPathsListener(
+  _: Electron.IpcMainInvokeEvent,
+  inputPaths: Array<string>,
+  outputPath: string,
+  ignoreReadonlyResourceOutputConflicts: boolean,
+): Promise<void> {
+  await mergeOpossumFilesFromPaths({
```
LMayTNG: Don't we want to do this in the DB process to avoid blocking the main thread?

src/ElectronBackend/split/merge-opossum-files.ts
```diff
@@ -0,0 +1,299 @@
...
+export interface MergeOpossumArchivesArgs {
+  ignoreReadonlyResourceOutputConflicts?: boolean;
+  inputPaths: Array<string>;
+  outputPath: string;
+}
+
+interface MergeArchive {
+  output: ParsedOpossumOutputFile;
+  readonlyRules: Array<ReadonlyRule>;
+  readonlyRulesByPath: Map<string, boolean>;
```
LMayTNG: That seems redundant

src/ElectronBackend/split/merge-opossum-files.ts
```diff
@@ -0,0 +1,299 @@
...
+  OPOSSUM_FILE_EXTENSION,
+  OUTPUT_FILE_NAME,
+  SPLIT_INFO_FILE_NAME,
+} from '../../shared/write-file-utils';
+import { parseOutputJsonContent, parseReadonlyRules } from '../input/parseFile';
+import type { ParsedOpossumOutputFile } from '../types/types';
+import {
+  getReadonlyRuleMap,
+  getReadonlyState,
+  mergeReadonlyRules as mergePureReadonlyRules,
```
LMayTNG, line 23: That seems unnnecessarily confusing to me. I think I'd just do the try-catch directly in `mergeOposumArchives`, then this file doesn't need a separate `mergeReadonlyRules` function.

LMayTNG: Or, better solution: Why is there a special `MergeReadonlyRulesError`? Why doesn't `mergeReadonlyRules` just throw a `MergeOpossumFilesError` and then you don't need to catch anything

src/ElectronBackend/split/readonly-rules.ts
```diff
@@ -0,0 +1,167 @@
...
+  }
+  return editableArchiveCount === 0;
+}
+
+function comparePaths(firstPath: string, secondPath: string): number {
+  const depthDifference = getPathDepth(firstPath) - getPathDepth(secondPath);
+  return depthDifference === 0
+    ? firstPath.localeCompare(secondPath)
+    : depthDifference;
+}
```
LMayTNG: Is there a specific reason that you sort the paths breadth-first?

src/ElectronBackend/dbProcess/dbProcess.ts
```diff
@@ -140,6 +151,37 @@ async function executeDbProcessMessage(
...
+          ignoreReadonlyResourceOutputConflicts,
+          saveFileParams,
+          partitionPaths,
+        },
+        storedOpossumZip,
+      );
+      const loadResult = await loadFile(
+        saveFileParams.opossumFilePath,
+        onProgress,
+      );
```
LMayTNG, line 187: Saving the file, doing the merging on the files and then loading it again is pretty expensive. You prioritized the path where you don't want to open the file at all, you just want to do a merge on the paths. That doesn't really seem like the most common usecase to me (but I could be wrong).

LMayTNG (commented):
From a dev perspective, it would imo have been cleaner to have a loaded file and then take a file to merge and add that to the database. That seems more in line with how the db process and the backend are supposed to work.

And working with an arbitrarily long list of files seems more complicated to me than having a "host file" and an "imported file" (potentially repeating that)

src/ElectronBackend/split/merge-opossum-files.ts
```diff
@@ -0,0 +1,299 @@
...
+  const unlinkedAttributions =
+    rootEditableArchive?.output.manualAttributions ??
+    archives[0].output.manualAttributions;
+  for (const [attributionUuid, attribution] of Object.entries(
+    unlinkedAttributions,
+  )) {
+    if (!referencedAttributionUuids.has(attributionUuid)) {
+      manualAttributions[attributionUuid] = attribution;
+    }
+  }
```
LMayTNG, line 207: I think a comment here saying that we don't expect this to happen would be good. Also, calling it `unlinkedAttributions` is a bit confusing if it contains all attributions.

src/ElectronBackend/dbProcess/dbProcess.ts
philipp-karcher: Yes, this has the benefit of making the scenario faster where we don't open the file at all, but it's not the main reason I chose this approach.

My assumption was that merging is something you do rarely (less than once a day, and even then only the one person who does the merging), where we prioritize correctness over performance (as long as it's in the same ballpark). 

The alternative would be to perform the merge operation on the database. The data model in the database is more complex than the json files, because of all the performance optimizations, and the logic would then live partly in sql instead of typescript. That's a lot more opportunities for bugs to sneak in.

src/ElectronBackend/split/merge-opossum-files.ts
philipp-karcher: Agreed, but I would rather fix this in the UI PR, as I had to touch some of the error handling there

src/ElectronBackend/split/readonly-rules.ts
philipp-karcher: Not really, tbh I'm not even entirely convinced that we should sort them at all here. But my agent was very insistent that having some form predictable order in the returned array was good, which I can see being useful I guess, so I kept it in. So this is just about having _a_ reasonable order, not specifically breadth-first

LMayTNG: I'm think that makes sense, but then I'd just call `.sort()`, so we don't have to introduce complexity like the `getPathDepth` function.

src/ElectronBackend/dbProcess/dbProcess.ts
LMayTNG: I feel like I'd still prefer the DB version, but I'm not sure enough of that to push for it :smile:
Your solution is good too, let's keep it :+1:

src/ElectronBackend/split/readonly-rules.ts
philipp-karcher: I think a bare `.sort()` just sorts the entries by their `toString()`, which is always `'[object Object]'`.  We should either not sort at all or do something simple like 

```
.sort((first, second) => first.path.localeCompare(second.path));
```

LMayTNG: Ah, good point. I'd be fine with both

LMayTNG (approved):

LMayTNG:
- When splitting, you shouldn't be able to change the selection in the resource tree. And maybe we could make the loading state a bit more obvious?

philipp-karcher:
> * When splitting, you shouldn't be able to change the selection in the resource tree. And maybe we could make the loading state a bit more obvious?

Agreed, I'm working on a follow up PR for this

PR #3245, opened 2026-08-03
feat: UI for merging partially locked opossum files

### LINKED ISSUE #3198: UI for merging partially locked opossum files

philipp-karcher:
### Summary of changes

* New dialog for merging split files back together
  * includes a progress log like for loading
* combined previous merge and import file menu entries
  * the decision whether to create a new project or to merge into the existing one now lives inside the dialog

#3198


LMayTNG:
Looks great! I like how it works with the current file, and the refactoring in the imports.

Some things I noticed while testing:
- The Merge dialog should directly tell you if there are no readonly paths, so merging doesn't make sense
- The merge icon should probably be horizontal like the split icon
- Should we offer an option to delete the merged-in split files?

philipp-karcher:
> Looks great! I like how it works with the current file, and the refactoring in the imports.
> 
> Some things I noticed while testing:
> 
>     * The Merge dialog should directly tell you if there are no readonly paths, so merging doesn't make sense
> 
>     * The merge icon should probably be horizontal like the split icon
> 
>     * Should we offer an option to delete the merged-in split files?

Agreed on the first two. I also like the delete option, but let's do it in a separate PR: #3248

src/shared/text.ts
```diff
@@ -2,9 +2,13 @@
...
 // SPDX-License-Identifier: Apache-2.0
-import { Criticality, type FileFormatInfo } from './shared-types';
+import {
+  Criticality,
+  type FileFilter,
+  type FileFormatInfo,
+} from './shared-types';
 
-function menuLabelForFileFormat(fileFormat: FileFormatInfo): string {
+function menuLabelForFileFormat(fileFormat: FileFilter): string {
```
LMayTNG, line 11: Should we unify the naming?

```
function menuLabelForFileFilter(fileFilter: FileFilter): string {
```

src/ElectronBackend/main/listeners.ts
LMayTNG, line 1: Good refactoring 👍

LMayTNG (approved):
Looks great! Only one really small nitpick

LMayTNG:
I noticed that in the split file, the count in the attribution list shows the wrong number. But I don't think that's a problem. We can just keep that as is now and when we improve the UI for locked resources, we can think about how we want to display that

Image

PR #3252, opened 2026-08-04
feat: improve loading indicator in split dialog

philipp-karcher:
### Summary of changes

* disable resource tree selection when creating split
* make loading indicator more obvious

### DIFF
--- a/src/Frontend/Components/SplitDialog/SplitDialog.tsx
+++ b/src/Frontend/Components/SplitDialog/SplitDialog.tsx
@@ -3,6 +3,7 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 import MuiAlert from '@mui/material/Alert';
+import MuiLinearProgress from '@mui/material/LinearProgress';
 import MuiTypography from '@mui/material/Typography';
 import { useEffect, useState } from 'react';
 
@@ -98,7 +99,6 @@ export const SplitDialog: React.FC<SplitDialogProps> = ({
           !destinationPath ||
           selectedResourcePaths.length === 0 ||
           splitInProgress,
-        loading: splitInProgress,
       }}
       rightButtonConfig={{
         onClick: onClose,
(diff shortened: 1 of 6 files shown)


src/Frontend/Components/SplitDialog/SplitDialog.tsx
```diff
@@ -127,12 +128,16 @@ export const SplitDialog: React.FC<SplitDialogProps> = ({
           testId={'split-destination-path'}
           disabled={splitInProgress || selectedResourcePaths.length === 0}
         />
-        {splitSucceeded ? (
+        {splitInProgress ? (
+          <div style={{ marginTop: '20px' }}>
+            <MuiTypography>{text.splitDialog.inProgress}</MuiTypography>
+            <MuiLinearProgress sx={{ marginTop: '8px' }} />
+          </div>
```
Hellgartner, line 148: I am not much of a fan of pixel perfect styling. Do we have some common margins/distances defined somewhere which we could use at least?

src/Frontend/Components/SplitDialog/SplitDialog.tsx
```diff
@@ -127,12 +128,16 @@ export const SplitDialog: React.FC<SplitDialogProps> = ({
           testId={'split-destination-path'}
           disabled={splitInProgress || selectedResourcePaths.length === 0}
         />
-        {splitSucceeded ? (
+        {splitInProgress ? (
```
Hellgartner: This whole construct again leads to jumps of the ui during each of the three state. Is that fine or should we do something about it (e.g. reserve space from the beginning)

Hellgartner (approved):

src/Frontend/Components/SplitDialog/SplitDialog.tsx
philipp-karcher: I agree that that would be good, but I don't think we have anything like that, all our components I have seen do their styling like this. 

If you want, I can create a ticket to introduce something like `theme.spacing()`

src/Frontend/Components/SplitDialog/SplitDialog.tsx
philipp-karcher: I tried reserving space, but that looked awkward. I added a `MuiCollapse` to make this look smoother at least now

PR #3253, opened 2026-08-04
test: remove retries from e2e tests

### LINKED ISSUE #3234: Remove retries from e2e tests
In our pipeline, we retry failed e2e tests and do not even fail the job for flaky tests. That leads to flaky tests not being fixed and even more sneaking in.

Remove the retry or fail on flaky tests. Run e2e tests a few times, fix any failures.

philipp-karcher:
### Summary of changes

* test: remove retries from e2e tests in CI
* test: stability improvements for e2e tests
  * use stable ids for all menu entries
  * wait for loading to finish
* feat: disable editing while attributions are loading
* fix: prevent stale dirty flag in attribution
* fix: prevent losing attribution edits on navigation

#3234


src/e2e-tests/page-objects/MenuBar.ts
```diff
@@ -51,109 +64,60 @@ export class MenuBar {
...
-      return this.findByLabel(text.menu.fileSubmenu.openRecent);
-    },
-  };
-
-  private clickMenuItem(label: string) {
-    return clickMenuItem(this.window.app, 'label', label);
-  }
-
-  private async assertMenuItemEnabledState(label: string, enabled: boolean) {
+  private async clickEnabledMenuItem(menuItemId: string): Promise<void> {
```
Hellgartner, line 70: Idea: we could improve typing here by not using any string but by restricting to the keys of the menuItemIds

Same for all other points where the menuItemId is a parameter. Probably worth defining a type.

src/Frontend/Components/AttributionDetails/AttributionDetails.tsx
```diff
@@ -115,15 +124,23 @@ export function AttributionDetails() {
...
-      <ButtonRow
-        isEditable={isEditable}
-        packageInfo={temporaryDisplayPackageInfo}
-      />
+      {!isSelectedAttributionLoading && (
+        <ButtonRow
+          isEditable={isEditable}
+          packageInfo={temporaryDisplayPackageInfo}
+        />
+      )}
```
Hellgartner, line 143: I am not sure I do like this too much. It causes a lot of screen flickering and visual noise. I would have preferred to just disable the buttons during loading. 

The same is true for the add auditing option (might not be this PR). This causes a lot of screen flicker.

Image

src/e2e-tests/utils/fixtures.ts
```diff
@@ -147,6 +148,11 @@ export const test = base.extend<{
     await window.waitForLoadState('domcontentloaded', {
       timeout: LOAD_TIMEOUT,
     });
+    if (opossumFilePath && openFromCLI) {
+      await expect(window.getByTestId('resources-tree')).toBeVisible({
+        timeout: LOAD_TIMEOUT,
+      });
+    }
```
Hellgartner, line 159: Do my knowledge, doing an expect in a fixture is bad practice. I also asked TAIA about it and apparently this would lead to wrong or confusing test failure reports.

src/e2e-tests/page-objects/MenuBar.ts
```diff
@@ -165,116 +129,80 @@ export class MenuBar {
...
+    if (filePath) {
+      await stubOpenDialogSync(this.window.app, [filePath]);
+    }
+    await this.clickEnabledMenuItem(menuItemIds.openFile);
   }
 
-  async createSplit(): Promise<void> {
-    await this.clickMenuItem(text.menu.fileSubmenu.split);
+  async openFileAndWaitForLoad(filePath: string): Promise<void> {
+    const resourcesTree = this.window.getByTestId('resources-tree');
```
Hellgartner: I do not like this too much. Can't we get that as the other things via the page object?

Hellgartner (changes requested):

src/e2e-tests/page-objects/MenuBar.ts
philipp-karcher: I thought about this, but I'm not sure it's worth it, because for the recent open files we do still need to use arbitrary strings as ids

src/e2e-tests/page-objects/MenuBar.ts
philipp-karcher: Yeah, you're right. We do still need to do some not so pretty stuff on the element handle, but we use the page object for most of it now

src/e2e-tests/utils/fixtures.ts
philipp-karcher: Yeah, we can use waitFor on the locator instead. The playwright docs warn on some of the locator methods that they can be flaky, which is why I didn't think about using them, but apparently waitFor is fine

src/Frontend/Components/AttributionDetails/AttributionDetails.tsx
philipp-karcher: Fully agreed on the add auditing option, will create a ticket for that.

I totally see what you mean when working on a small file with fast load times. Have you tried this with a large file? To me it feels much smoother now. Especially when switching between attributions, before it would completely clear the form, and then fill it with the loaded attribution.

But I'm no UX expert, so I don't really have a strong opinion here.

Hellgartner: Chatted on this. Fine as it is

Hellgartner (approved):

PR #3265, opened 2026-08-10
feat: show locked files as readonly

### LINKED ISSUE #3199: Show locked files as readonly
As a simplification, #3196 simply hides all locked files. Instead we should show them as readonly

- locked resources/attributions should appear greyed out and be readonly
- Disable extract functionality on locked files

philipp-karcher:
### Summary of changes

- Show readonly resources and attributions
- Readonly attributions are NOT shown in the unrelated tab
  - Reasoning: this would clutter the tab, and you can't really act on them anyway
- Readonly attributions ARE shown in the on children/on parent tab
  - Reasoning: These are still relevant to the current resource. Other than on root, this is an edge case anyway if we assume that splits usually happen top-level

#3199

Create a split and click around


philipp-karcher:
One decision that must be made is what we display when selecting an editable resource. What I ended up doing:

- Readonly attributions are NOT shown in the unrelated tab
  - Reasoning: this would clutter the tab, and you can't really act on them anyway 
- Readonly attributions ARE shown in the on children/on parent tab
  - Reasoning: These are still relevant to the current resource. Other than on root, this is an edge case anyway if we assume that splits usually happen top-level

Ideas I experimented with but did not like the UX for:

- separate readonly tab
- something similar to the show/hide resolve button for signals

hecadr:
> * Readonly attributions are NOT shown in the unrelated tab
>   
>   * Reasoning: this would clutter the tab, and you can't really act on them anyway
> * Readonly attributions ARE shown in the on children/on parent tab
>   
>   * Reasoning: These are still relevant to the current resource. Other than on root, this is an edge case anyway if we assume that splits usually happen top-level

I think these are reasonable decisions.

hecadr (approved):

PR #3268, opened 2026-08-11
feat: force unlock

### LINKED ISSUE #3200: Force unlock in admin mode
If somehow other parts of a split opossum file get lost or are not accessible, there needs to be a way to recover the whole file

Add an option to force unlock the file in admin mode

philipp-karcher:
### Summary of changes

Add a file menu entry for removing all readonly rules from the current file

#3200


hecadr (commented):
This will unlock all previously locked parts. My naive expectation was that it should be possible to select only some locked files which should be unlocked.

How would we handle a review that was split into three parts and one of them got lost? Would it then be possible to merge the two remaining opossum files first and then unlock the missing part afterwards?

philipp-karcher:
> This will unlock all previously locked parts. My naive expectation was that it should be possible to select only some locked files which should be unlocked.
> 
> How would we handle a review that was split into three parts and one of them got lost? Would it then be possible to merge the two remaining opossum files first and then unlock the missing part afterwards?

Yes, that would be possible. But you have a point, it's still awkward, because you have to wait for the other person to give you their work so you can merge it first before the lost part becomes unlocked. Or you would need to unlock everything and then re-lock some parts.

But I don't think this should really be part of any normal workflow anyway, it's more of a break glass in case of emergency thing.  In the "normal" case where the person you sent the file to becomes sick or something, you probably still have the original file you sent them somewhere, which can be merged back.

I would suggest to keep it like this for now. If it turns out that this is really needed, I would probably prefer to just offer a "merge anyway" button on not matching readonly rules.

hecadr (approved):

hecadr:
Ok, since this is intended as a last resort, I think the "unlock all" feature is sufficient.

PR #3271, opened 2026-08-11
test: make wait for saving changes more robust

philipp-karcher:
### Summary of changes

Make wait for saving to finish more robust

Before, it was possible for the assertions to finish before the save had actually completed (all of them were true before saving had started). See https://github.com/opossum-tool/OpossumUI/actions/runs/31494815870


hecadr (approved):
LGTM

PR #3273, opened 2026-08-12
fix: keep auditing option trigger stable while loading

### LINKED ISSUE #3256: Reduce flicker from auditing options
When switching between attributions, the auditing options add the top changing causes the screen to flicker. Especially "Add Auditing Option" always disappears and re-appears

AnkitParekh007:
## Summary

- keep the Add Auditing Option chip mounted while a newly selected attribution is loading
- disable the chip during the existing read-only loading state
- add regression coverage for the stable disabled control

## Root cause

AttributionDetails intentionally makes the form read-only while attribution data is loading. AuditingOptions derived whether to render its trigger from editability, so the transient read-only state removed the chip and then added it back when loading completed.

## User impact

Switching between editable attributions no longer makes the Add Auditing Option control disappear and reappear. The control remains visible but disabled until editing is safe again.

## Validation

- focused AttributionDetails regression test
- TypeScript checks for frontend and Electron backend
- ESLint on all changed files
- Prettier check on all changed files
- `git diff --check`

The broader AttributionDetails test file was also run: 30 tests passed, while two unrelated pre-existing tests exceeded their 5-second local timeout. The focused regression test passes independently.

Closes #3256

## Disclosure

This contribution was prepared with AI assistance under human-directed scope and was source-verified and tested before submission.

### DIFF
--- a/src/Frontend/Components/AttributionDetails/AttributionDetails.tsx
+++ b/src/Frontend/Components/AttributionDetails/AttributionDetails.tsx
@@ -142,6 +142,7 @@ export function AttributionDetails() {
         packageInfo={temporaryDisplayPackageInfo}
         onEdit={isEditable ? confirmEditWasPreferred : undefined}
         dimmed={pickerMode.isActive}
+        keepAddButtonVisibleWhenDisabled={isSelectedAttributionLoading}
       />
       {!isSelectedAttributionLoading && (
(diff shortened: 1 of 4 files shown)


philipp-karcher:
This only partially addresses the problem, and even then, just keeping the button visible is not great either

PR #3277, opened 2026-08-13
fix: disable buttons during force unlock

philipp-karcher:
### Summary of changes
* disable buttons while force unlock is in progress
* show a loading indicator


mariusAsaTNG (approved):
looks good

PR #3278, opened 2026-08-13
feat: make split here target initially visible

philipp-karcher:
### Summary of changes
Auto expand and scroll to the split-here target in the split dialog

When the target is nested or in a long list of files, it's awkward to have to search for it

### DIFF
--- a/src/Frontend/Components/SplitDialog/MultiResourcePicker.tsx
+++ b/src/Frontend/Components/SplitDialog/MultiResourcePicker.tsx
@@ -15,6 +15,7 @@ import type { ResourceTreeNodeData } from '../../../ElectronBackend/api/resource
 import { text } from '../../../shared/text';
 import { ROOT_PATH } from '../../shared-constants';
 import { readonlyStyle } from '../../shared-styles';
+import { getParents } from '../../state/helpers/get-parents';
 import { backend } from '../../util/backendClient';
 import { Checkbox } from '../Checkbox/Checkbox';
 import { ReadonlyIcon } from '../Icons/Icons';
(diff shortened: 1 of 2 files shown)


src/Frontend/Components/SplitDialog/MultiResourcePicker.tsx
```diff
@@ -152,6 +168,11 @@ function renderResource(
   return (
     <ResourceRow
       key={resource.id}
+      ref={
+        path === initiallySelectedPath
+          ? scrollToInitiallySelectedResource
+          : undefined
+      }
```
LMayTNG, line 172: That's neat, I didn't know about ref callbacks

src/Frontend/Components/SplitDialog/MultiResourcePicker.tsx
```diff
@@ -45,9 +46,21 @@ export function MultiResourcePicker({
...
+  const [expandedPaths, setExpandedPaths] = useState<Array<string>>(() => [
+    ...new Set([
+      ROOT_PATH,
+      ...initialSelectedPaths.flatMap((selectedPath) =>
+        getParents(selectedPath),
+      ),
+    ]),
   ]);
+  const initiallySelectedPath = initialSelectedPaths[0];
+  const scrollToInitiallySelectedResource = useCallback(
```
LMayTNG: I think I'd just call this "scrollToResource", as the function doesn't know whether it's scrolling initially or what a selected resource is

src/Frontend/Components/SplitDialog/MultiResourcePicker.tsx
```diff
@@ -45,9 +46,21 @@ export function MultiResourcePicker({
...
+    (element: HTMLDivElement | null) => {
+      element?.scrollIntoView({ block: 'center' });
```
LMayTNG: We could add `behavior: 'smooth'` here, which looks cool but adds visual noise. I'll leave that to your judgement 😄

LMayTNG (approved):
Nice improvement 👍

src/Frontend/Components/SplitDialog/MultiResourcePicker.tsx
philipp-karcher: I like it!

src/Frontend/Components/SplitDialog/MultiResourcePicker.tsx
philipp-karcher: Me neither before my agent used it :laughing:

PR #3283, opened 2026-08-19
test(performance): add performance tests

### LINKED ISSUE #3281: Automated performance tests
Performance on last files is always an issue. Automated performance tests will help us catch regressions and enable further improvements

philipp-karcher:
### Summary of changes

First draft for automated performance tests on generated files

#3281 

### Things to discuss / possible follow ups
* Do we want to integrate this into our CI and if yes in what way?
  * it's hard to reliably separate small improvements/regressions from noise
* Do we need an easy way to run single tasks in the tests instead of the whole suite?
  * this is difficult, because the biggest bottleneck for the large file is the initial load
  * we could attempt to bypass this somehow by directly injecting it into the database, but then we leave e2e territory
* Do we need more in-depth profiling/tracing? 
  * probably helpful for more in-depth analysis, but in the first draft I did not want to clutter the production code to much


src/performance-tests/playwright/mutation-performance-workflows.ts
```diff
@@ -0,0 +1,175 @@
...
+        signalsPanel.assert.loadingIndicatorIsHidden(),
+        attributionDetails.assert.loadingIndicatorIsHidden(),
+      ]);
+      await signalsPanel.assert.linkButtonIsDisabled();
+    },
+  });
+
+  await runScenario({
+    id: 'edit-and-save-attribution',
+    title: 'edit and save an attribution',
```
mariusAsaTNG, line 62: For me this failed once and then never again, so it was probably a fluke. Just wanted to note it.

src/performance-tests/synthetic-file/profiles.ts
```diff
@@ -0,0 +1,75 @@
...
+  large: LARGE_PROFILE,
+} as const satisfies Record<SyntheticFileProfileName, SyntheticFileProfile>;
+
+export function getSyntheticFileProfile(name: string): SyntheticFileProfile {
+  if (
+    !SYNTHETIC_FILE_PROFILE_NAMES.includes(name as SyntheticFileProfileName)
+  ) {
+    throw new Error(`Unsupported synthetic file profile: ${name}`);
+  }
+  return SYNTHETIC_FILE_PROFILES[name as SyntheticFileProfileName];
```
mariusAsaTNG: Why not change the type from string to `small | large`? Then you would not need this error

src/performance-tests/synthetic-file/writer.ts
```diff
@@ -0,0 +1,246 @@
...
+  }
+  return attributions;
+}
+
+function createInput(
+  profile: SyntheticFileProfile,
+  model: Fixture,
+): ParsedOpossumInputFile {
+  const frequentLicenses: Array<RawFrequentLicense> =
+    SYNTHETIC_LICENSE_NAMES.map((name) => ({
```
mariusAsaTNG, line 132: I do not have a full overview of the synthetic data but do you use faker for some of these? So far it seems like most data is taken from pre defined arrays

mariusAsaTNG (approved):
I did not look over every LoC but it works and the direction is correct.

src/performance-tests/synthetic-file/profiles.ts
philipp-karcher: Because we call it with arbitrary strings from env variables

src/performance-tests/synthetic-file/writer.ts
philipp-karcher: No, we're not using faker here atm. The pre-defined arrays are more deterministic, and I don't see much benefit in more "realistic" names.

Issue #3101, opened 2026-04-08
long load times for large files


mariusAsaTNG:
**Describe the bug**
When opening a large file with >1mil. resources the clicks on the resource tree become unreasonably slow. ~2 sec. Even expanding the files without clicking on them takes too long. 
Perhaps this is because of limited ram

**To Reproduce**
Open a large opossum file e.g. facebook-for-android and click around

**Expected behavior**
we want loading times in the sub ~500ms range

PR #3286, opened 2026-08-27
perf: pagination for attributions

philipp-karcher:
#### Paginate attribution lists

- Paginate the audit lists, report table, and query-wide confirmation previews.
- Represent multi-selection as either explicit UUIDs or an “all matching” selection with exclusions. Symbolic selections are resolved in the backend when an action executes.
- Navigation from the report view locates the target attribution and loads the result-set prefix through the page containing it. This can be expensive for targets deep in the result set; a bounded or bidirectional solution is deferred because it requires broader list-state changes.
- Opening an attribution from the report view resets the manual audit filters and search while preserving the selected sort order. Otherwise, an active filter could exclude the target and prevent navigation.

#### Improve attribution cache handling after mutations

- Previously, all affected active queries were refetched and awaited together, allowing secondary aggregate queries to delay visible updates.
- Now, queries required for immediate UI consistency—attribution details, loaded attribution pages, and resolved-signal state—are refreshed first. Inactive queries are marked stale, while secondary active queries refresh afterward in the background.

#### Performance gains

Measured with `yarn performance:measure:large` averaged across 3 runs each

| Metric | Baseline | New | Improvement |
|---|---:|---:|---:|
| link-signal | 5245 ms | 898 ms | 83% |
| resolve-high-fanout-signal | 8396 ms | 4143 ms | 51% |
| navigate-next-review-resource.attributions | 3141 ms | 275 ms | 91% |
| edit-and-save-attribution | 4046 ms | 1210 ms | 70% |
| edit-high-fanout-attribution | 10930 ms | 8538 ms | 22% |
| navigate-next-review-resource.classifications | 2147 ms | 263 ms | 88% |
| search-dense-signal-list | 2364 ms | 716 ms | 70% |
| open-project-statistics | 2023 ms | 917 ms | 55% |
| bulk-link-signals | 8036 ms | 7058 ms | 12% |
| expand-and-select-resource | 2019 ms | 1046 ms | 48% |

#3101


mariusAsaTNG:
1. It would be nice if the scrollbar went all the way down and did not jump
2. You get a Failed to invalidate error when deleting an attribution and selecting the same attribution or selecting another one quickly
3. Selecting an attribution, going to report view and clicking the edit button on the selected attribution does not jump to it.

mariusAsaTNG:
Minimal reproduction:
Scroll down in the attributions select an attribution. Scroll back up and select the first attribution. You will jump back down

philipp-karcher:
> Minimal reproduction: Scroll down in the attributions select an attribution. Scroll back up and select the first attribution. You will jump back down

Fixed. That bug was pre-existing, but made more easily reproducible by artifically increasing the virtuoso viewport for pre-loading new pages

philipp-karcher:
> 1. It would be nice if the scrollbar went all the way down and did not jump
> 
>     2. You get a Failed to invalidate error when deleting an attribution and selecting the same attribution or selecting another one quickly
> 
>     3. Selecting an attribution, going to report view and clicking the edit button on the selected attribution does not jump to it.

1. was a bit more work than anticipated, because scrolling all the way down means loading many pages at once, but all 3 are fixed now

src/Frontend/util/invalidate-mutation-queries.ts
```diff
@@ -0,0 +1,75 @@
...
+
+export async function invalidateMutationQueries({
+  queryClient,
+  invalidations,
+}: {
+  queryClient: QueryClient;
+  invalidations: Array<MutationInvalidation>;
+}): Promise<void> {
+  await invalidateAll({ queryClient, invalidations, awaitRefetch: true });
+  void invalidateAll({ queryClient, invalidations, awaitRefetch: false });
```
mariusAsaTNG: Would it make sense to have the void queries before the await ones? Since then they can start while we are awaiting the other ones?

mariusAsaTNG: Nit I would prefer the filtering to happen before invalidateAll, or rename the function to invalidateQueriesFilteredByRefetch or something like that

src/Frontend/Components/ProgressBar/ProgressBar.tsx
```diff
@@ -82,7 +102,7 @@ export const ProgressBar: React.FC<ProgressBarProps> = (props) => {
   const getNextClassificationResource =
     backend.getNextFileToReviewForClassification.useQuery(
       { selectedResourcePath },
-      { enabled: props.selectedProgressBar === 'classification' },
+      { enabled: false },
```
mariusAsaTNG, line 105: You disable the progress bar resource fetching until someone clicks it? I like that

src/ElectronBackend/api/progressBarUtils.ts
```diff
@@ -200,6 +201,9 @@ export async function removeManualOrExternalCaaFromResources(
     resourceIds?: Array<number> | OperandExpression<number>;
   },
 ) {
+  const attributionUuidSelection = attributionUuids
+    ? sql<string>`(select value from json_each(${JSON.stringify(attributionUuids)}))`
```
mariusAsaTNG, line 205: I do not really understand, why (select value from json_each(${JSON.stringify(attributionUuids)})) is better than just passing attributionUuids

src/ElectronBackend/api/utils.ts
```diff
@@ -584,19 +589,25 @@ export async function linkAttributions(
...
+            attribution_uuid: attributionUuid,
+            attribution_is_external: 0,
+          })),
+        )
+        .$if(options?.ignoreExisting ?? false, (eb) =>
+          eb.onConflict((oc) => oc.doNothing()),
+        )
+        .execute();
+    },
+    { batchSize: 10000 },
```
mariusAsaTNG: How did you arrive at 10'000 for a batch size?

mariusAsaTNG:
As mentioned in person, the pagination seems somewhat broken right now. When you click on a resource it fetches the attributions 4 times, instead of 2 times(might be okay), but then when you scroll it seems to fetch the entire attributions list instead of just the next 200.

mariusAsaTNG:
>     1. was a bit more work than anticipated, because scrolling all the way down means loading many pages at once, but all 3 are fixed now

It looks really nice now. You cannot feel the pagination anymore and it just feels smoother with no downsides.

src/Frontend/util/invalidate-mutation-queries.ts
philipp-karcher: Having the void queries before the awaited ones would defeat the purpose of this mechanism, no? The idea is to start them after the awaited ones, so they don't block them, they all use the same resource (the DB)

src/ElectronBackend/api/utils.ts
philipp-karcher: The parameter limit of 30.000 divided by 3 parameters. But actually, one of them was always constant, so we can get away with 15.000. Also added a comment explaining this

src/ElectronBackend/api/progressBarUtils.ts
philipp-karcher: It's always one parameter (the json string) instead of however many attribution ids we have, and there is a strict limit on the amount of parameters in sqlite

philipp-karcher:
> As mentioned in person, the pagination seems somewhat broken right now. When you click on a resource it fetches the attributions 4 times, instead of 2 times(might be okay), but then when you scroll it seems to fetch the entire attributions list instead of just the next 200.

The 4 requests instead of 2 is because we optimistically fetch the on resource tab before we know the counts. If that turns out to be empty, we then fetch the non-empty tab we want. We can instead just move the decision for the "right" first tab to the backend.

The scrolling thing is because I implemented this too late in the evening yesterday, will fix :melting_face:

src/Frontend/util/invalidate-mutation-queries.ts
mariusAsaTNG: Oh yeah i forgot that bettersqlite3 is not async in that regard. I though that starting everything at the same time would be better, but in this case not.

src/ElectronBackend/api/utils.ts
mariusAsaTNG: Makes sense 👍

philipp-karcher:
> > As mentioned in person, the pagination seems somewhat broken right now. When you click on a resource it fetches the attributions 4 times, instead of 2 times(might be okay), but then when you scroll it seems to fetch the entire attributions list instead of just the next 200.
> 
> The 4 requests instead of 2 is because we optimistically fetch the on resource tab before we know the counts. If that turns out to be empty, we then fetch the non-empty tab we want. We can instead just move the decision for the "right" first tab to the backend.
> 
> The scrolling thing is because I implemented this too late in the evening yesterday, will fix 🫠

Scrolling should be fixed now, but moving this into the backend introduced too much complexity and actually made things slower (might be optimizable, but idk), so I would just keep the optimistic approach

mariusAsaTNG:
For me the flow of selecting a manual attribution, going to the report page and clicking on that attribution to go back is broken again.

mariusAsaTNG:
Did you not have to add any indexes whatsoever for this?

philipp-karcher:
> Did you not have to add any indexes whatsoever for this?

Good catch, I ran the performance tests and they looked fine, but yes I think we should add an index

philipp-karcher:
> For me the flow of selecting a manual attribution, going to the report page and clicking on that attribution to go back is broken again.

I had an e2e test for this, but it didn't cover the case of switching the tab away from the standard "On Resource". Fixed now

mariusAsaTNG (approved):
Looks great! You added a bunch of tests, and with all the manual testing you get a go ahead from me.

PR #3290, opened 2026-09-01
ci: run performance tests

### LINKED ISSUE #3285: Run performance tests in CI
We have performance tests now, we should run them in CI

There probably is too much noise to compare two runs exactly. They should be averaged across multiple (3?) runs.

Running nightly on main instead of in every PR might be enough.

It might make sense to tinker with the size of the generated files (the current "large" may be too big)

philipp-karcher:
### Summary of changes

* run performance tests nightly
* compare results to previous run and previous release

#3285

There is no hard gate yet to make the action fail on regressions, we will first have to assess how big the natural variance is

### DIFF
--- a/.github/workflows/performance.yml
+++ b/.github/workflows/performance.yml
@@ -15,6 +15,7 @@ jobs:
     timeout-minutes: 60
     env:
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      PERFORMANCE_RUN_ID: ${{ github.run_id }}-${{ github.run_attempt }}
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
 
--- a/.github/workflows/performance.yml
+++ b/.github/workflows/performance.yml
@@ -7,6 +7,8 @@ name: Performance Test
 
 on:
   workflow_dispatch:
+  release:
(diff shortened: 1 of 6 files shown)


.github/workflows/performance.yml
```diff
@@ -0,0 +1,149 @@
...
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    permissions:
+      actions: read
+      contents: read
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      PERFORMANCE_RUN_ID: ${{ github.run_id }}-${{ github.run_attempt }}
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
```
hecadr, line 26: Are you sure we have to specify the commit hash here? I think version tags should work, too, and would be sufficient.

The same applies to the other steps.

hecadr (approved):

.github/workflows/performance.yml
philipp-karcher: We explicitly configured `helpers:pinGitHubActionDigests` in our `renovate.json`, so this is the standard of our repo. 

You're right that the risk with official actions is probably negligible, but renovate will take care of it, so it doesn't hurt to have it.

PR #3295, opened 2026-09-07
fix: ensure unit tests work regardless of underlying OS

KramerKlara:
### Summary of changes

- New: src/testing/mock-posix-path.ts — exports posixPathModule ({ ...path.posix, default: path.posix }) via vi.importActual (which is what prevents mock recursion).
- FileConverter.test.ts — posix mock + per-test platform pinning via a small stubProcessPlatform helper.
- getPath.test.ts, get-save-file-listener.test.ts — posix mock only.

Caveat
- The guarantee is file-scoped, not suite-wide: any future test mixing POSIX string fixtures with native path needs the same 3-line vi.mock('path', ...) incantation.

Unit tests up until now were dependent on a linux OS due to filepath structures.
With this change, the tests should be OS-agnostic

`yarn test:unit`


src/ElectronBackend/main/__tests__/get-save-file-listener.test.ts
```diff
@@ -45,6 +45,11 @@ vi.mock('../dialogs', () => ({
   selectSaveFile: vi.fn(),
 }));
 
+vi.mock('path', async () => {
+  const { posixPathModule } = await import('../../../testing/mock-posix-path');
+  return posixPathModule;
+});
```
philipp-karcher: What's the point of this helper over just doing `vi.mock('path', () => import('node:path/posix'))`?

src/ElectronBackend/main/__tests__/getPath.test.ts
philipp-karcher: I think the core problem is that we have hardcoded paths like this in our test code instead of using `path.join`. If we consistently do that we probably don't even need to mock anything

philipp-karcher (changes requested):

src/ElectronBackend/opossum-file/__tests__/FileConverter.test.ts
```diff
@@ -51,16 +56,20 @@ class TestExternalFileConverter extends ExternalFileConverter {
   }
 }
 
+function stubProcessPlatform(platform: NodeJS.Platform): void {
```
philipp-karcher, line 54: good refactor 👍

src/ElectronBackend/main/__tests__/get-save-file-listener.test.ts
KramerKlara: Your version does make more sense, with one small change:
`() => import('node:path/posix')` relies on Vitest unwrapping the promise and on the namespace's synthetic default matching the CJS object. The explicit form 
`vi.mock('path', async () => {
  const posix = await import('node:path/posix')
  return { ...posix, default: posix }
})` 
guarantees both named and default imports resolve to the real posix object, and reads clearly. Safer + self-documenting. I will implement this now.

src/ElectronBackend/main/__tests__/getPath.test.ts
KramerKlara: Reduce Electron mock dependency in getPath tests

Production Code (src/ElectronBackend/main/getPath.ts)
- Added optional basePath?: string parameter to getBasePathOfAssets() and getBasePathOfIcons() functions
- When basePath is provided: uses path.join(basePath, 'public', 'assets/icons') directly
- When basePath is omitted: falls back to original Electron-dependent logic (app?.isPackaged checks, getAppPath(), getDevAppRoot()) for backward compatibility

Test Code (src/ElectronBackend/main/__tests__/getPath.test.ts)
- Removed all Electron mock setup (vi.hoisted, vi.mock('electron'), beforeEach with mock return values)
- Tests now call functions directly with base path arguments: getBasePathOfAssets('/repo'), getBasePathOfIcons('/repo'), etc.
- All 3 tests pass without needing to mock electron.app.getAppPath()

philipp-karcher: Ah I think there was a misunderstanding, by not mocking anything I meant the posix path mocking you added, not all mocks in this test, I don't think we should do this :sweat_smile:

KramerKlara: Oh-oh, sorry about that - how do you think we can avoid mocking the posix paths only?

philipp-karcher: My understanding was that the tests fail on windows because the production code uses windows paths on your machine, but some of our test assertions hardcode posix separators. If we just use path.join in the assertions, the separators should agree again.

KramerKlara: Changes:
- src/ElectronBackend/main/getPath.ts — reverted to the pre-refactor API (no basePath param, original getPathOfExtraResource)
- getPath.test.ts — Electron mocks and all 4 original tests restored (incl. getPathOfExtraResource coverage); no path mock; multi-segment mock values now built with path.join (the endsWith Windows fix)
- FileConverter.test.ts — path mock removed; dev-fallback getAppPath mock now path.join('/repo', 'build', 'ElectronBackend')
- get-save-file-listener.test.ts — path mock removed; path import added; defaultPath assertion now path.join('/my', 'file-source.opossum')

philipp-karcher (approved):
If the tests pass on windows now, it's ready to merge. But we should probably squash the 5 commits into one for a cleaner history, I don't think keeping them separate adds any value here

PR #3296, opened 2026-09-08
feat: add search highlighting for the Linked Resources Tree

KramerKlara:
Added search highlighting functionality for the linked resources tree. When attribution UUIDs are provided and a search term is active, the tree now:

- Filters linked resources to show only those matching the search term (case-insensitive, matching both path and name)
- Highlights matching nodes with a light blue background
- Highlights folder nodes when their name matches the search term

### Summary of changes

- src/ElectronBackend/api/resourceTree.ts: Added highlight_matches SQL filter and updated matchesFilters logic to use search term when onAttributionUuids are present
- src/Frontend/Components/ResourceBrowser/LinkedResourcesTree/LinkedResourcesTreeNode/LinkedResourcesTreeNode.tsx: Added light blue background styling when matchesFilters is true
- src/Frontend/Components/ResourceBrowser/LinkedResourcesTree/LinkedResourcesTreeNode/__tests__/LinkedResourcesTreeNode.test.tsx: Added highlighting tests
- src/Frontend/Components/ResourceBrowser/LinkedResourcesTree/__tests__/LinkedResourcesTree.test.tsx: Added search/filter highlight tests
- src/ElectronBackend/api/__tests__/resourceTree.test.ts: Added search with attribution filter test suite

Further cosmetic changes:

- Remove the "wave" background picture from src/Frontend/Components/App/App.style.ts
- Make disabled buttons on bottom right more easily visible (src/Frontend/Components/AttributionDetails/ButtonRow/ButtonRow.style.ts & src/Frontend/Components/AttributionDetails/ButtonRow/ButtonRow.tsx)

Beforehand, the search field in the Main Resource Tree highlighted matching paths and files. The Linked Resources Tree now does the same.

### DIFF
--- a/src/ElectronBackend/api/resourceTree.ts
+++ b/src/ElectronBackend/api/resourceTree.ts
@@ -62,6 +62,10 @@ export function getResourceTree({
 
       const hasActiveFilters = Boolean(search || hasActiveNonSearchFilters);
 
+      const searchLastPart = removeTrailingSlash(search ?? '')
+        .split('/')
+        .at(-1);
+
       /*
        * FILTERED_RESOURCE_TEMP_TABLE contains the resources included by the active filters.
        * Without active filters, it is a view on `resource` with no runtime overhead.
@@ -296,6 +300,18 @@ export function getResourceTree({
                 ),
             )
             .as('is_expandable'),
+        )
+        .select((eb) =>
+          eb
+            .and([
+              eb(
+                'shown_resources.path',
+                'like',
+                `%${removeTrailingSlash(search ?? '')}%`,
+              ),
+              eb('shown_resources.name', 'like', `%${searchLastPart}%`),
+            ])
+            .as('highlight_matches'),
         );
 
       query = query.orderBy('id');
@@ -332,7 +348,15 @@ export function getResourceTree({
         criticality: node.max_criticality_on_unresolved_external_attribution,
         classification:
           node.max_classification_on_unresolved_external_attribution,
-        matchesFilters: Boolean(node.matches_filters),
+        /*
+         * For attribution-filtered queries (linked resources tree), the
+         * highlight comes from the search term only, mirroring the main
+         * tree's search-only behavior (path and name match). Without a
+         * search, no node is highlighted.
+         */
+        matchesFilters: onAttributionUuids
+          ? Boolean(search && node.highlight_matches)
+          : Boolean(node.matches_filters),
       }));
 
       await dropTempTable();
--- a/src/ElectronBackend/api/resourceTree.ts
+++ b/src/ElectronBackend/api/resourceTree.ts
@@ -215,13 +215,14 @@ export function getResourceTree({
                   }
 
                   // Search only: Only highlight where the file matches, not the entire subtree
-                  const lastSearchPart = removeTrailingSlash(search!)
-                    .split('/')
-                    .at(-1);
                   return eb
                     .and([
-                      eb('r.path', 'like', `%${removeTrailingSlash(search!)}%`),
-                      eb('r.name', 'like', `%${lastSearchPart}%`),
+                      eb(
+                        'r.path',
+                        'like',
+                        `%${removeTrailingSlash(search ?? '')}%`,
+                      ),
+                      eb('r.name', 'like', `%${searchLastPart}%`),
                     ])
                     .as('matches_filters');
                 })
--- a/src/Frontend/Components/App/App.style.ts
+++ b/src/Frontend/Components/App/App.style.ts
@@ -29,9 +29,6 @@ export const ViewContainer = styled(MuiBox)({
   height: '100vh',
   flexDirection: 'column',
   background: OpossumColors.lightGrey,
-  // backgroundImage: 'url("icons/wave.svg")',
-  // backgroundPosition: 'bottom',
-  // backgroundRepeat: 'no-repeat',
 });
 
 export const theme = createTheme({
--- a/src/Frontend/Components/App/App.style.ts
+++ b/src/Frontend/Components/App/App.style.ts
@@ -29,9 +29,9 @@ export const ViewContainer = styled(MuiBox)({
   height: '100vh',
   flexDirection: 'column',
   background: OpossumColors.lightGrey,
-  backgroundImage: 'url("icons/wave.svg")',
-  backgroundPosition: 'bottom',
-  backgroundRepeat: 'no-repeat',
+  // backgroundImage: 'url("icons/wave.svg")',
+  // backgroundPosition: 'bottom',
+  // backgroundRepeat: 'no-repeat',
 });
 
 export const theme = createTheme({
--- a/src/Frontend/Components/ResourceBrowser/LinkedResourcesTree/LinkedResourcesTreeNode/LinkedResourcesTreeNode.tsx
+++ b/src/Frontend/Components/ResourceBrowser/LinkedResourcesTree/LinkedResourcesTreeNode/LinkedResourcesTreeNode.tsx
@@ -6,11 +6,7 @@ import MuiBox from '@mui/material/Box';
 import MuiTypography from '@mui/material/Typography';
 
 import { text } from '../../../../../shared/text';
-import {
-  OpossumColors,
-  readonlyStyle,
-  treeItemClasses,
-} from '../../../../shared-styles';
+import { readonlyStyle, treeItemClasses } from '../../../../shared-styles';
 import {
   BreakpointIcon,
   DirectoryIcon,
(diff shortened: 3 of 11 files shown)


src/Frontend/Components/App/App.style.ts
```diff
@@ -29,9 +29,9 @@ export const ViewContainer = styled(MuiBox)({
   height: '100vh',
   flexDirection: 'column',
   background: OpossumColors.lightGrey,
-  backgroundImage: 'url("icons/wave.svg")',
-  backgroundPosition: 'bottom',
-  backgroundRepeat: 'no-repeat',
+  // backgroundImage: 'url("icons/wave.svg")',
+  // backgroundPosition: 'bottom',
+  // backgroundRepeat: 'no-repeat',
```
philipp-karcher: We should remove this properly, including the .svg file

src/Frontend/Components/ResourceBrowser/LinkedResourcesTree/__tests__/LinkedResourcesTree.test.tsx
```diff
@@ -1,15 +1,17 @@
+/* eslint-disable testing-library/no-node-access */
```
philipp-karcher: I don't love disabling this, especially on the entire file

src/Frontend/Components/ResourceBrowser/LinkedResourcesTree/LinkedResourcesTreeNode/__tests__/LinkedResourcesTreeNode.test.tsx
```diff
@@ -1,10 +1,12 @@
+/* eslint-disable testing-library/no-node-access */
```
philipp-karcher: See above

src/ElectronBackend/api/resourceTree.ts
philipp-karcher, line 227: Your implementation sort of duplicates the existing search logic here

src/Frontend/Components/ResourceBrowser/LinkedResourcesTree/LinkedResourcesTreeNode/LinkedResourcesTreeNode.tsx
```diff
@@ -27,6 +31,12 @@ export const LinkedResourcesTreeNode: React.FC<TreeNode> = ({ resource }) => {
       sx={{
         ...treeItemClasses.labelRoot,
         ...(resource.isReadonly ? readonlyStyle : {}),
+        ...(resource.matchesFilters
+          ? {
+              backgroundColor: OpossumColors.lightBlue,
+              borderRadius: '3px',
+            }
+          : {}),
```
philipp-karcher: We could move this to the other `treeItemClasses` in `shared-styles.ts` so that the linked tree and normal tree stay consistent

philipp-karcher (changes requested):
We should also add an e2e test for the highlighting (or just expand the existing `shows only linked resources matching search` test)

src/Frontend/Components/App/App.style.ts
KramerKlara: Done

src/Frontend/Components/ResourceBrowser/LinkedResourcesTree/__tests__/LinkedResourcesTree.test.tsx
KramerKlara: Done

src/Frontend/Components/ResourceBrowser/LinkedResourcesTree/LinkedResourcesTreeNode/__tests__/LinkedResourcesTreeNode.test.tsx
KramerKlara: Also done :)

src/Frontend/Components/ResourceBrowser/LinkedResourcesTree/__tests__/LinkedResourcesTree.test.tsx
philipp-karcher: Mhh, but now the tests don't actually assert what we want them to, they just check whether the resource is there, not whether it's highlighted. E.g. `highlights a node matching the filters` and `does not highlight a node not matching the filters` should be opposites, but they now have the exact same assertion

src/ElectronBackend/api/resourceTree.ts
KramerKlara: In my latest commit I have deduplicated this as far as possible - at least the search logic itself

src/Frontend/Components/ResourceBrowser/LinkedResourcesTree/__tests__/LinkedResourcesTree.test.tsx
KramerKlara: You're right, I see that now!

Here's some updates:
- LinkedResourcesTreeNode.tsx — added data-testid={linked-resources-tree-node-${resource.id}} to the root MuiBox 
- LinkedResourcesTreeNode.test.tsx — the two highlight tests now assert toHaveStyle({ backgroundColor: OpossumColors.lightBlue }) vs .not.toHaveStyle(...) via getByTestId
- LinkedResourcesTree.test.tsx —the two search tests now additionally assert folder1/folder2/resource_1 highlight states

src/Frontend/Components/ResourceBrowser/LinkedResourcesTree/LinkedResourcesTreeNode/LinkedResourcesTreeNode.tsx
KramerKlara: Done - the highlight styling is now reusable

src/ElectronBackend/api/resourceTree.ts
philipp-karcher: Mhh the actual search query is still duplicated, I think this should be refactored into a common helper. 

There might even be a way to not have to call this in two places at all, but then we would probably need to actually inspect how that affects the resulting query plan and performance, so I'm fine with keeping the existing structure.

KramerKlara: Done. Extracted the duplicated search-match expression into getSearchMatchExpression (src/ElectronBackend/api/resourceTree.ts:479)

src/e2e-tests/__tests__/interacting-with-linked-resources.test.ts
```diff
@@ -77,16 +77,22 @@ test('shows only linked resources matching search', async ({
...
+  await linkedResourcesTree.assert.resourceIsNotHighlighted(resourceName4);
   await linkedResourcesTree.assert.resourceIsVisible(resourceName1);
   await linkedResourcesTree.assert.resourceIsVisible(resourceName4);
 
   await linkedResourcesTree.goto(resourceName1);
   await window.keyboard.press(`${modKey}+F`);
   await window.keyboard.type(resourceName4);
+  await linkedResourcesTree.assert.resourceIsHighlighted(resourceName4);
+  await linkedResourcesTree.assert.resourceIsNotHighlighted(resourceName1);
   await linkedResourcesTree.assert.resourceIsHidden(resourceName1);
```
philipp-karcher, line 93: If we assert that the resource is hidden, we don't need to also assert that it's not highlighted. Same above. It would probably also make more sense to check that the resource is visible before checking the highlighting instead of after

src/e2e-tests/page-objects/LinkedResourcesTree.ts
```diff
@@ -34,6 +39,28 @@ export class LinkedResourcesTree {
...
+      const node = this.node
+        .getByText(resourceName, { exact: true })
+        .locator('..');
+      await expect
+        .poll(async () => {
+          if ((await node.count()) === 0) {
+            return true;
+          }
+          return node.evaluate(
+            (el) => getComputedStyle(el).borderRadius !== '3px',
```
philipp-karcher: Mhh I think I would rather just have the check for the colour, explicitly checking for 3px is too brittle for my taste (either way, checking for "any other value than 3px" here is not so great)

philipp-karcher (changes requested):

src/e2e-tests/page-objects/LinkedResourcesTree.ts
philipp-karcher: Ah, I meant to just remove checking the border radius in both the new methods and just check the colour

KramerKlara: Ah yes, that makes sense :)
Done

philipp-karcher (approved):

PR #3298, opened 2026-09-09
fix: always re-fetch state in file menu click handler

philipp-karcher:
### Summary of changes

Always re-fetch state in file menu click handler

This could lead to stale backend state being used in the click handler, see e.g. https://github.com/opossum-tool/OpossumUI/actions/runs/34283946400
We already do this for the other click handlers.


KramerKlara:
Good fix. The PR correctly identifies that file menu click handlers could use stale backend state, and adds the re-fetch pattern already used by other handlers. The test confirms the behavior works correctly. This eliminates bugs from stale state.

KramerKlara (approved):

PR #3299, opened 2026-09-09
fix: destroy the db to reset it to fix EBUSY errors on e2e tests

KramerKlara:
### Summary of changes

- src/ElectronBackend/db/db.ts — `resetDb()` is now async. It awaits `db.destroy()` on any existing connection before opening a new one, properly releasing the file handle.
- src/ElectronBackend/db/initializeDb.ts — `initializeDb()` now awaits `resetDb()`.

Problem: On Windows, `resetDb()` opened a new SQLite connection without closing the old one. The lingering connection kept a file handle on the .opossum file, causing EBUSY errors when e2e tests reset/reopened the database.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,5 +55,3 @@ yarn-error.log*
 
 dev.db
 dev.db-journal
-local-reuse.js
-.local-pre-commit/pre-commit
--- a/.gitignore
+++ b/.gitignore
@@ -55,3 +55,5 @@ yarn-error.log*
 
 dev.db
 dev.db-journal
+local-reuse.js
+.local-pre-commit/pre-commit
--- a/src/ElectronBackend/db/db.ts
+++ b/src/ElectronBackend/db/db.ts
@@ -42,9 +42,7 @@ function openDb() {
 }
 
 export async function resetDb() {
-  if (db) {
-    await db.destroy()
-  }
+  await db?.destroy()
   db = openDb();
 }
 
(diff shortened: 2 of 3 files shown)


src/ElectronBackend/db/db.ts
```diff
@@ -41,7 +41,10 @@ function openDb() {
   });
 }
 
-export function resetDb() {
+export async function resetDb() {
+  if (db) {
+    await db.destroy()
+  }
```
philipp-karcher: Nit:

```suggestion
  await db?.destroy()
```

philipp-karcher (approved):
I noticed that you're commits sometimes don't pass formatting. We have pre-commit hooks that should take care of that automatically, do those not work on Windows?

.gitignore
```diff
@@ -55,3 +55,5 @@ yarn-error.log*
 
 dev.db
 dev.db-journal
+local-reuse.js
+.local-pre-commit/pre-commit
```
philipp-karcher: Did you add this for the pre-commit hooks on windows? If so that should probably be a separate PR, and we should aim for a solution that is checked into git. Maybe we can take a look at that together tomorrow.

philipp-karcher (changes requested):

.gitignore
KramerKlara: Yes I did... How would you like to approach this?

KramerKlara: Removed

philipp-karcher (approved):
Let's also squash the commits again, as the actual change is quite small. Other than that it's ready to merge

PR #3303, opened 2026-09-10
ci: sync graphviz version in diagram

philipp-karcher:
### Summary of changes

* re-generate the database diagram when graphviz is updated

The database diagram contains the graphviz version and we run generation with every typecheck. So after graphviz was updated, running typecheck would give you changes unrelated to your current work.

The alternative would be to do this via `postUpgradeTasks`, but you somehow need to put the commands you want to execute on an allowlist, and I have no idea how/if we can configure this (in mend?), and this approach is already established with `sync-node-version`


KramerKlara (approved):
Clean PR: The workflow file follows project conventions, the changes are focused on diagram regeneration, and infinite loops are prevented.

PR #3305, opened 2026-09-11
fix: consistently prevent variable overflow

philipp-karcher:
### Summary of changes

* Consistently use batching and json_each
* Calculate batch sizes from the number of parameters

Sqlite has a hard limit for how many parameters are supported and we had some leftover gaps that could run into that limit


src/ElectronBackend/api/utils.ts
```diff
@@ -33,6 +33,14 @@ import { removeManualOrExternalCaaFromResources } from './progressBarUtils';
 export type ResourceRelationship =
   'same' | 'ancestor' | 'descendant' | 'unrelated';
 
+export function jsonArraySelection<T extends string | number>(
```
KramerKlara: This function seems to be duplicated in multiple places:
- src/ElectronBackend/api/attributions/attribution-list-query-utils.ts:66-70
- src/ElectronBackend/api/progressBarUtils.ts:204-206 (inline)
- src/ElectronBackend/api/progressBarUtils.ts:318 (inline)
Can this be turned into a generic helper and moved down a layer?

src/ElectronBackend/api/mutations.ts
```diff
@@ -633,7 +633,7 @@ async function setAttributionsResolvedStatus(
             .execute();
         },
         // The main problem is updating the caa table, which can only take 15_000 attributionUuids
```
KramerKlara: This comment seems outdated as it no longer matches the code - could be misleading

src/ElectronBackend/api/utils.ts
```diff
@@ -499,15 +511,24 @@ export function toCanonicalLicenseName(
...
-export async function withBatching<P, R>(
+export async function withSqlBatching<P, R>(
   input: Array<P>,
   f: (arg: Array<P>) => Promise<R>,
-  props?: { batchSize: number },
+  props?: { parametersPerItem: number },
 ): Promise<Array<R>> {
-  const batchSize = props?.batchSize ?? DEFAULT_BATCH_SIZE;
+  const batchSize = Math.floor(
+    SQL_PARAMETER_LIMIT / (props?.parametersPerItem ?? 1),
```
KramerKlara, line 515: This does not properly guard against the case where parametersPerItem==0: 
30000 / 0 yields Infinity, and Math.floor(Infinity) is still Infinity. The ?? 1 fallback only catches undefined/null — not 0, since 0 ?? 1 evaluates to 0.

src/ElectronBackend/api/utils.ts
```diff
@@ -499,15 +511,24 @@ export function toCanonicalLicenseName(
...
+}
 
+export async function withBatching<P, R>(
+  input: Array<P>,
+  f: (arg: Array<P>) => Promise<R>,
+  batchSize: number,
+): Promise<Array<R>> {
   const results: Array<R> = [];
 
   const numBatches = Math.ceil(input.length / batchSize);
```
KramerKlara, line 531: Follow-up from the above comment:
With batchSize = Infinity flowing in from the above division, 40000 / Infinity is 0, so Math.ceil(0) = 0 and the loop never executes. the callback is never called and [] is returned.

philipp-karcher: This is a bit pathological, but sure, can't hurt to be defensive

src/ElectronBackend/api/utils.ts
philipp-karcher: See below, guarding in one place should be enough

KramerKlara (approved):

PR #3308, opened 2026-09-14
feat: editable diff view

### LINKED ISSUE #3202: Improved diff view
#2563 expanded the diff view to any two attributions, but made it readonly

- Both sides are editable (if the attributions themselves are editable, i.e. not readonly or signals)
- there are left/right arrows to takeover changes like in an IDE diff view
  - how to deal with multiple fields in one row?

philipp-karcher:
### Summary of changes

* Make the diff view fully editable
* IDE style left/right arrows
* include flags and confidence chips in diff view

#3202

Compare two attributions/signals and play around


src/Frontend/Components/DiffPopup/ComparisonFieldEditor.tsx
```diff
@@ -0,0 +1,450 @@
...
+  const onFirstPartyChange = useEventCallback(
+    (value: boolean) =>
+      void onEdit(() => onChange(side, { firstParty: value })),
+  );
+  const onLicenseToggle = useEventCallback(() => onToggleLicenseText());
+  const undo = useMemo(
+    () =>
+      isDirty ? (
+        <UndoFieldButton
+          key={'undo'}
```
KramerKlara, line 115: I think this key doesn't do anything because the element isn't rendered as part of a list - is this a leftover?

src/Frontend/Components/ConfirmSavePopup/ConfirmSavePopup.tsx
```diff
@@ -2,52 +2,69 @@
...
   });
   const modifyOrMatchOnlyOnOneResource =
     backend.modifyOrMatchOnlyOnOneResource.useMutation({
-      onBeforeInvalidation: handleFocusedAttributionOutcome,
+      onBeforeInvalidation: handleBeforeInvalidation,
     });
   const isSaving =
     updateOrMatch.isPending || modifyOrMatchOnlyOnOneResource.isPending;
+  const isMutating = useIsMutating() > 0;
+  const isBusy = isSaving || isMutating;
```
KramerKlara: With this definition, buttons are disabled during any app-wide mutation, not just this popup's. Is this conservative approach intentional?

src/Frontend/Components/DiffPopup/ComparisonFieldEditor.tsx
philipp-karcher: It is part of a list if there are other action buttons, like in upstream address. If you remove the key, you will get a warning when modifying that field:

> Each child in a list should have a unique "key" prop.
>
> Check the render method of `div`. It was passed a child from ComparisonFieldEditor. See https://react.devlink/warning-keys for more information.

src/Frontend/Components/ConfirmSavePopup/ConfirmSavePopup.tsx
philipp-karcher: Good point. Unfortunately this pattern is used in some other places already, but I fully agree that it's no good. I will improve this usage and create a ticket to clean this up in general

src/Frontend/Components/DiffPopup/ComparisonFieldEditor.tsx
KramerKlara: Ah I see, thank you. This makes sense then, I just didn't realise it would be part of a list in some cases :)

src/Frontend/Components/ConfirmSavePopup/ConfirmSavePopup.tsx
philipp-karcher: Done, and created https://github.com/opossum-tool/OpossumUI/issues/3315

KramerKlara (approved):

PR #3310, opened 2026-09-15
chore: use renovate best practices

philipp-karcher:
### Summary of changes

Use renovate best practices. The most important change is that it includes [`:maintainLockFilesWeekly`](https://docs.renovatebot.com/presets-default/#maintainlockfilesweekly), which ensures that our transitive dependencies are up to date.

https://docs.renovatebot.com/upgrade-best-practices


KramerKlara (approved):
Looks good;
It turns on lockFileMaintenance, which Renovate runs weekly (early Monday mornings). It opens a separate, scheduled PR that refreshes lock files only:
- bumps existing dependencies to the latest in-range minor/patch versions in the lock file
- removes stale/phantom lock-file entries

PR #3311, opened 2026-09-16
chore: resolve yarn warnings

philipp-karcher:
### Summary of changes

* explicitly declare rolldown as a dev dependency
   * rolldown already comes with vite transitively, but @rolldown/plugin-babel expects it to be declared explicitly
* remove superfluous extensions

```
❯ yarn install
➤ YN0000: · Yarn 4.18.0
➤ YN0000: ┌ Resolution step
➤ YN0000: └ Completed
➤ YN0000: ┌ Post-resolution validation
➤ YN0060: │ better-sqlite3 is listed by your project with version 13.0.3 (p29a30f), which doesn't satisfy what kysely-codegen requests (>=7.6.2 <13.0.0).
➤ YN0002: │ opossum-ui@workspace:. doesn't provide rolldown (p972dc1), requested by @rolldown/plugin-babel.
➤ YN0086: │ Some peer dependencies are incorrectly met by your project; run yarn explain peer-requirements <hash> for details, where <hash> is the six-letter p-prefixed code.
➤ YN0069: │ @tanstack/eslint-plugin-query ➤ peerDependencies ➤ typescript: This rule seems redundant when applied on the original package; the extension may have been applied upstream.
➤ YN0068: │ eslint-plugin-jest ➤ peerDependencies ➤ typescript: No matching package in the dependency tree; you may not need this rule anymore.
```

better-sqlite3 13.0.0 is just newer than the latest release of kysely-codegen, so it is not explicitly declared as supported yet. Kysely-codegen is just a dev-dependency and it seems to work, so I think it's fine to leave it as is.


.yarnrc.yml
```diff
@@ -6,12 +6,6 @@ enableGlobalCache: false
 
 packageExtensions:
   # ESLint plugins need to forward typescript to @typescript-eslint/utils
-  '@tanstack/eslint-plugin-query@*':
```
KramerKlara, line 9: Just for my understanding; Why do we not need these anymore?

philipp-karcher: tanstack/eslint-plugin-query added its own typescript dependency in v5.91.4 and we simply don't have eslint-plugin-jest as a dependency anymore. In both cases we just forgot to remove them here.

KramerKlara (approved):

PR #3316, opened 2026-09-16
ci: trigger required checks after automated commits

philipp-karcher:
### Summary of changes

Explicitly dispatch the build/E2E, code-quality and license checks after node version synchronization or database diagram regeneration create a commit

Commits created with GITHUB_TOKEN [do not trigger push workflows](https://docs.github.com/en/actions/concepts/security/github_token#when-github_token-triggers-workflow-runs). This leaves required checks pending on the PR, preventing merges.

The cleanest alternative would be postUpgradeTasks in renovate, but those can only be allowlisted in the paid mend plan.

Another alternative would be to create a github app for these actions, but that seems like to much hassle.


KramerKlara (approved):
Ensures automated commits properly re-trigger required checks instead of leaving them pending. Clean, minimal changes. Approved.

PR #3324, opened 2026-09-17
feat: auto expand on filtered resource tree

### LINKED ISSUE #3210: Auto-expand on filtered resource tree - Persist filtered resources in DB
Currently, when calculating the resource tree with filters applied, we send the filter as parameter with the `resourceTree` query. In the query, we build a `FILTERED_RESOURCE_TEMP_TABLE` based on the filters that we then use to calculate the shown resources.

The problem is that that table is not reusable. Every time we rerun the query without changing the filters (e.g. when expanding resources), we have to calculate it again, which can take half a second on large files.

We also have a feature in `getNodePathsToExpand` that auto-expands single resources, so when you expand a resource that has a single child, that child is expanded as well. Currently that doesn't take the filters into account, so you may have a single child in the filtered view that doesn't get auto-expanded. We'd like to change that, but the half-second latency would make that painful.

The proposal would be to convert the filters from parameters of `resourceTree` to something that is persisted in the backend. We'd have a mutation `setFilters` that calculates the filtered resource tree and persists it, and then invalidates the `resourceTree` query. That way we don't have to refilter the resources on every request, only every time the filters are changed.

philipp-karcher:
### Summary of changes
* auto-expand if a resource if it only has one child in the filtered tree
  * previously, auto-expansion only worked if the resource had one child in the total tree
* persist the previously temporary filtered resource tree table
  * without this, auto-expansion is not feasible performance-wise 
* keep a cache of filtered tables
  * this allows you to switch between the resource tree with filters and the linked resources tree without losing the cached table

#3210


src/ElectronBackend/api/resourceTree.ts
```diff
@@ -2,366 +2,278 @@
...
+    query = query.orderBy('id');
+
+    const treeNodes = (await query.execute()).map((node) => ({
+      id: node.path + (node.can_have_children ? '/' : ''), // For compatibility with legacy code
+      labelText: node.name || '/',
+      level: node.level,
+      isExpandable: Boolean(node.is_expandable),
+      isExpanded:
+        expandedNodes === 'expandAll' ||
+        expandedNodes.includes(node.path + (node.can_have_children ? '/' : '')),
```
KramerKlara: Performance: Line 234 and 240 both compute the same path, and expandedNodes.includes() searches the array element-by-element for every node. Can this be optimised?

src/ElectronBackend/api/resource-tree-cache.ts
```diff
@@ -0,0 +1,114 @@
...
+          .execute();
+        cache.entries.delete(oldestKey);
+      }
+      const cacheId = cache.nextCacheId;
+      cache.nextCacheId += 1;
+      await populateFilteredResources(trx, filters, cacheId);
+      cache.entries.set(key, cacheId);
+      return query(trx, cacheId);
+    });
+  } catch (error) {
```
KramerKlara, line 108: Do we need to drop the whole cache on any error, or would it be enough just to drop the failed key to be less wasteful?

src/ElectronBackend/db/initializeDb.ts
```diff
@@ -185,6 +187,19 @@ export async function initializeDb(inputFile: ParsedFileContent) {
     });
 }
 
+async function initializeFilteredResourcesTable(trx: Transaction<DB>) {
+  await trx.schema
+    .createTable('filtered_resources')
```
KramerKlara, line 192: Table filtered_resources does not have a PK or UNIQUE constraint enforced - would make sense on [cache_id, id]

src/ElectronBackend/api/resource-tree-cache.ts
philipp-karcher: This error handling is purely defensive, if we reach this we are in an unexpected state where something has gone quite wrong, so I think being conservative is the right approach here

src/ElectronBackend/db/initializeDb.ts
philipp-karcher: Good catch

src/ElectronBackend/api/resourceTree.ts
philipp-karcher: I think this was a bit messy in general (with one correctness issue), so I cleaned it up

KramerKlara: I see, that looks good now :)

KramerKlara (approved):
The auto-expansion on filtered trees works nicely, and persisting the filtered resources with the multi-entry cache is a clean solution to the cache-miss problem on folder expansion. Looks good to me; approved.

Issue #3315, opened 2026-09-16
Reconsider use of global mutation guards


philipp-karcher:
**Is your feature request related to a problem? Please describe.**
We use global mutation guards `useIsMutating() > 0` in a lot of places. That works, but at least in some places it may be cleaner to have more targeted guards instead

**Describe the solution you'd like**
Look at all usages of these global guards and decide whether they should be more targeted

PR #3326, opened 2026-09-19
fix: disable the progress bar when mutations are pending

### LINKED ISSUE #3284: Progress bar can navigate to stale resource during loading
The progress bar stays interactive while refetching, causing navigation to the wrong resource.

Possible solution: Disable interactions while the progress bar is loading

mariusAsaTNG:
### Summary of changes

Disable clicking of the progressbar when mutations are happening and lower the opacity to 0.5

See #3284 

start opossumui with a larger file and trigger any mutation. To be discussed if this is even needed. I am fine with dropping this PR but then we should also close #3284


mariusAsaTNG:
Possible improvements:
1. Technically you do not need to freeze it at every mutation. Only on ones that touch the progress bar directly or indirectly. However, this then becomes harder to maintain.
2. Look into tests. The tests did not break which surprises me. I guess we had no tests where the progress bar was clicked right after a mutation. Or playwright waits long enough for this not to be a problem

philipp-karcher:
> Possible improvements:
> 
>     1. Technically you do not need to freeze it at every mutation. Only on ones that touch the progress bar directly or indirectly. However, this then becomes harder to maintain.
> 
>     2. Look into tests. The tests did not break which surprises me. I guess we had no tests where the progress bar was clicked right after a mutation. Or playwright waits long enough for this not to be a problem

1. I think we can leave it as is for now, and then make a general decision how we want to handle this in #3315

2. This doesn't surprise me, basically all things you do with playwright have waits/readiness checks baked into them (which is why failures are always a TimeoutError)

philipp-karcher:
Switching between progress bars (if they had already been loaded before, but need to be re-fetched) looks pretty rough because of the delay:

1. Switch to any other progress bar
2. Make a mutation (you can wait for it to finish)
3. Switch back to to the other progress bar
4. It shows the outdated progress bar as ready
5. Then after the delay it becomes inactive
6. Then it shows the updated progress bar

philipp-karcher:
I would also be fine with just dropping this, the CI flake hasn't re-occured, and if it does we can then decide what to do about it

PR #3328, opened 2026-09-21
fix: prevent scrollbar from jumping to selected resource when expanding

### LINKED ISSUE #3312: Resource tree scollbar jumps on expand
**Describe the bug**
When having a file selected in the resource tree, expanding another resource above it will make the scrollbar jump back to the selected resource if it was out of view.

**To Reproduce**

1. Scroll down in the resource tree
2. Select a resource
3. Scroll up so that it is out of view
4. Expand another resource
5. Observe how it jumps back down to the selected resource

**Expected behavior**
You can see the expanded resource after expanding

philipp-karcher:
### Summary of changes

Prevent scrollbar from jumping to selected resource when expanding

#3312


src/e2e-tests/__tests__/resource-tree-scroll-preservation.test.ts
```diff
@@ -0,0 +1,55 @@
...
+      },
+    }),
+  },
+});
+
+test('preserves the selected resource viewport when expanding an upper folder', async ({
+  resourcesTree,
+}) => {
+  await resourcesTree.scrollToBottom();
+  await resourcesTree.clickResource('99-selected');
```
KramerKlara, line 41: It might be worth adding a small wait after this to prevent e2e test flakiness;
in the test, clicking 99-selected schedules a deferred (~1 ms) scroll-to-center; on a slow machine that deferred scroll could land after line 44's scrollToTop, breaking the viewport assertions

philipp-karcher: I'm not a big fan of hardcoded waits. Instead I removed the defer, because I don't see the point in it

KramerKlara (approved):
PR makes sense and is to the point. Approved.

PR #3330, opened 2026-09-21
fix: compare complete attribution sets before removing redundant links

philipp-karcher:
### Summary of changes

Require exact equality between a resource's manual attributions and its closest attributed ancestor using EXCEPT in both directions.

The previous scalar comparison could treat partially matching sets as equal, removing explicit assignments and changing the resource's effective attributions.


KramerKlara (approved):
Sensible fix and well-tested. Approved.

PR #3334, opened 2026-09-22
test: add unit and e2e tests for export logic

### LINKED ISSUE #2795: Add tests for export logic
There should be unit tests for the export logic in export-actions.ts

KramerKlara:
### Summary of changes

Adds missing unit and e2e test coverage for the export feature.
No production code changes.

New e2e tests (src/e2e-tests/__tests__/exporting-opossum-files.test.ts)
- Exports all five formats (follow-up CSV, compact BOM, detailed BOM, SPDX YAML/JSON) via the File menu and verifies each file is written next to the loaded .opossum file with the expected naming and recognizable content (also checks follow-up packages are excluded from the compact BOM).
- Exports after discarding unsaved changes: export → unsaved-changes popup → discard → file is produced.
- Support changes: utils/shell.ts stubs/records shell.showItemInFolder (prevents file-explorer popups during tests, asserts the revealed path), and MenuBar gains the four previously missing export menu-item methods.

New/extended unit tests
- popup-actions.test.ts: direct tests for exportFileOrOpenUnsavedPopup, showImportDialogOrOpenUnsavedPopup, showMergeOpossumFilesDialogOrOpenUnsavedPopup (clean-state vs. unsaved-changes branches) and for createSplit (success invalidates queries; errors pass through without invalidation).
- exportCommands.test.ts: SPDX license-text fallbacks (own text vs. frequent_license table vs. empty string), detailed-BOM filter (follow-up/first-party excluded), follow-up path resolution via closest_attributed_ancestors in resource-ID order, and empty-result exports.

[#2795](https://github.com/orgs/opossum-tool/projects/1/views/1?pane=issue&itemId=97963159&issue=opossum-tool%7COpossumUI%7C2795)


src/e2e-tests/__tests__/exporting-opossum-files.test.ts
```diff
@@ -0,0 +1,162 @@
...
+    menuBar,
+    notSavedPopup,
+    resourcesTree,
+  }) => {
+    await stubShowItemInFolder(window.app);
+
+    await resourcesTree.goto(resourceName1);
+    await attributionDetails.attributionForm.comment.fill(
+      faker.lorem.sentences(),
+    );
```
philipp-karcher: This test sort of duplicates `warns user of unsaved changes if user attempts to export data before saving`. It additionally wants to test that we don't include the discarded changes in the export, but not properly: we only change a comment, which isn't even included in the followup csv anyway

src/ElectronBackend/api/__tests__/exportCommands.test.ts
```diff
@@ -232,4 +333,104 @@ describe('export tests', () => {
...
+        }),
+        uuid3: expect.objectContaining({
+          packageName: 'pkg-with-unknown-license',
+          licenseText: '',
+        }),
+      },
+    });
+  });
+
+  it('exports an empty CSV when no attributions match the follow-up filter', async () => {
```
philipp-karcher: `when no attributions match the follow-up filter` is a bit misleading, because the database is completely empty

src/ElectronBackend/api/__tests__/exportCommands.test.ts
```diff
@@ -70,6 +70,51 @@ describe('export tests', () => {
     });
   });
 
+  it('resolves follow-up paths via closest attributed ancestors ordered by resource id', async () => {
```
philipp-karcher: specifying `ordered by resource id` is a bit weird.  And we don't even really test that we use the _closest_ attributed ancestor, because there is only on attributed ancestor in our test setup

philipp-karcher (changes requested):

src/e2e-tests/utils/shell.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export async function stubShowItemInFolder(
+  app: ElectronApplication,
+): Promise<void> {
+  await app.evaluate(({ shell }, stateKey) => {
+    const state: ShellStubState = { showItemInFolderCalls: [] };
+    Reflect.set(shell, stateKey, state);
+    Reflect.set(shell, 'showItemInFolder', (filePath: string) => {
+      state.showItemInFolderCalls.push(String(filePath));
+      return true;
```
philipp-karcher: `filePath` already is a string and `showItemInFolder` returns void
```suggestion
      state.showItemInFolderCalls.push(filePath);
```

philipp-karcher (changes requested):

src/ElectronBackend/api/__tests__/exportCommands.test.ts
KramerKlara: Changed to `exports an empty follow-up CSV when there are no attributions`

src/e2e-tests/__tests__/exporting-opossum-files.test.ts
KramerKlara: Done: Made the test more concretely check that exports do not include unsaved changes after they were discarded; Now uses the copyright field rather than the comment field

src/ElectronBackend/api/__tests__/exportCommands.test.ts
KramerKlara: Renamed to `exports follow-up attributions with the resources of their closest attributed ancestors`;
Setup now has two followUp attributions with files and subfolders;
Now actually tests the closest-ancestor behaviour

philipp-karcher (approved):

PR #3335, opened 2026-09-22
test: stabilize resource search

philipp-karcher:
### Summary of changes

* add data-applied-search attribute to resource browser header
* wait for search to complete in tests

Failed performance test: https://github.com/opossum-tool/OpossumUI/actions/runs/35700064307/job/106655691745


src/e2e-tests/page-objects/ResourcesTree.ts
```diff
@@ -150,29 +150,55 @@ export class ResourcesTree {
...
+        parentResource
+          .getByLabel(`collapse ${parentResourcePath}`)
+          .or(parentResource.getByLabel(`collapse ${normalizedPath}`)),
+      ).toBeVisible();
+      await this.waitForReady();
     }
     await expect(this.getResourceByPath(resourcePath)).toBeVisible();
   }
 
-  async expandResourceAtPath(resourcePath: string): Promise<void> {
```
KramerKlara, line 157: If instead of this old `expandResourceAtPath` we just use `expandResource`, I think we lose the `.or()` fallback for no-trailing-slash paths. Also, `expandResource` skips `waitForReady` after expanding. Is this intentional?

src/e2e-tests/page-objects/ResourcesTree.ts
```diff
@@ -150,29 +150,55 @@ export class ResourcesTree {
...
 
+  async selectResourceAtPath(resourcePath: string): Promise<void> {
+    await this.withRevealedResource(resourcePath, (resource) =>
+      resource.click(),
+    );
+  }
+
+  async search(value: string): Promise<void> {
+    await this.searchField.fill(value);
+    await this.waitForReady();
```
KramerKlara: I might be wrong, but;;
Doesn't `waitForReady()` just look for `aria-busy="false"`, but "false" is also the state **before** React has processed the `fill()`? So if thefirst poll lands in that gap, we declare "ready" without the search ever having started because we are silently skipping a click against a stale DOM.
We might need a two-phase wait here to first expect `aria-busy="true"` and after that `aria-busy="false"`

src/e2e-tests/page-objects/ResourcesTree.ts
```diff
@@ -150,29 +150,55 @@ export class ResourcesTree {
...
+    } finally {
+      await this.clearSearch();
+    }
   }
 
   async selectRevealedResource(resourcePath: string): Promise<void> {
     await this.getResourceByPath(resourcePath).click();
   }
 
+  async selectResourceAtPath(resourcePath: string): Promise<void> {
```
KramerKlara: I cannot find a caller for this anywhere - is this dead code?
If so, the same applies to `withRevealedResource`

src/e2e-tests/page-objects/ResourcesTree.ts
philipp-karcher: You're right that this is possibly flaky, but a two-phase wait will have a similar problem, we might be too slow and miss the "false". I will have to think of something else

src/e2e-tests/page-objects/ResourcesTree.ts
philipp-karcher: no longer relevant

src/e2e-tests/page-objects/ResourcesTree.ts
philipp-karcher: Should be more stable now

KramerKlara: The change lets the test check that the search was actually applied, so it can never declare success too early. Should be more reliable now!

KramerKlara (approved):

